PR #7, opened 2026-02-22
Improve cost UX: OAuth-aware equivalent-billing hints

JF10R:
## Summary
This PR improves cost clarity in ClawMetry when usage comes from OAuth/included auth (where provider-billed amount may be $0).

### What changed
- Adds per-model billing heuristics (`modelBilling`) and global `billingSummary` in `/api/usage` output.
- Renames cost labels to reduce ambiguity:
  - `Cost` -> `Telemetry Cost` (OTLP mode)
  - `Est. Cost` -> `Estimated Cost` (transcript mode)
- Overview improvements:
  - Keeps spending compact
  - Shows OAuth-specific equivalent-billing hint only when relevant (not shown if no model uses oAuth)
  - Adds subtle info tooltip icon (OAuth/mixed only)
  
image

- Flow model node:
  - Adds compact auth hint (`Auth: API / OAuth / mixed`)
  - Prevents text overflow in the brain card
image

- Trend projection improvements:
  - Removes unrealistic hard-coded projected-cost warning behavior
  - Uses observed effective cost rate for monthly equivalent projection
  - Includes explicit note for OAuth scenarios (can be $0 billed - only if oAuth usage is detected)

## Why
Users reported confusion between token-derived cost estimates and actual provider billing, especially in OAuth auth flows. This patch makes the UI explicit and less noisy while preserving useful trend context.

### DIFF
--- a/dashboard.py
+++ b/dashboard.py
@@ -1000,7 +1000,8 @@ def _get_otel_usage_data():
 
     # Enhanced cost tracking for OTLP data
     trend_data = _analyze_usage_trends(daily_tokens) 
-    warnings = _generate_cost_warnings(today_cost_val, week_cost_val, month_cost_val, trend_data)
+    model_billing, billing_summary = _build_model_billing(model_usage)
+    warnings = _generate_cost_warnings(today_cost_val, week_cost_val, month_cost_val, trend_data, month_tok, billing_summary)
 
     return {
         'source': 'otlp',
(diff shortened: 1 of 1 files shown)


vivekchand (approved):
Great contribution! Clean, well-scoped PR that solves a real UX pain point — OAuth users seeing scary cost projections when their actual billing is $0.

**What I liked:**
- Smart heuristic approach for detecting API key vs OAuth billing
- Conditional UI hints that only show when relevant (no noise for pure API key users)
- Suppressing trend warnings for OAuth scenarios
- Clean label renames (Cost → Telemetry Cost / Estimated Cost)

**Minor notes (non-blocking, can address in follow-ups):**
- The config cache reads `providers` key, but OpenClaw configs typically use `auth.profiles` — env var fallback covers the common case though
- The brain SVG node loses its animated pulse circle in favor of the auth hint text — small visual tradeoff but the info is more useful

Merging — thanks for the thoughtful PR! 🦞

PR #510, opened 2026-04-03
fix(security): mask encryption key in CLI output

dumko2001:
## Summary

Mask encryption key when printed to stdout in CLI output to prevent accidental exposure.

## Changes

- Updated CLI print statements to mask encryption key values
- Added proper masking when displaying configuration

## Testing

tests/test_cli.py - verifies encryption key is properly masked in output

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,4 @@ Thumbs.db
 
 # Logs
 *.log
+.clawmetry-fleet.db
--- a/clawmetry/cli.py
+++ b/clawmetry/cli.py
@@ -507,14 +507,14 @@ def _cmd_connect(args) -> None:
     # --key-only: just save config, don't start daemon (for host-side NemoClaw OTP flow)
     if getattr(args, "key_only", False):
         print(f"  API key:      {api_key}")
-        print(f"  Enc key:      {enc_key}")
+        print(f"  Enc key:      {enc_key[:6] + '…' + enc_key[-4:]}")
         print()
         return
 
     # Skip enc key reminder when --enc-key was passed (automated/sandbox use)
     if not _enc_key_arg:
         print("  Keep this secret key safe (like a password):")
-        print(f"  {enc_key}")
+        print(f"  {enc_key[:6] + '…' + enc_key[-4:]}")
         print()
 
     # --no-daemon: skip daemon start (managed by supervisord externally)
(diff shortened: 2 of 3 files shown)


vivekchand:
## PR Review (automated)

**Status: CONFLICTING** — This PR has merge conflicts with main.

**CI:** ✅ All checks passing

**Feature check:** Masking the encryption_key in CLI output (lines 510, 517 of cli.py) is **not yet in main**. Main already masks the key in some places (lines ~486, ~1147) but not in the `--key-only` print path — this PR fills that gap. Valid security improvement.

**Action needed:** Please rebase onto latest main and resolve conflicts:
```bash
git fetch upstream
git checkout fix-002
git rebase upstream/main
# resolve conflicts
git push --force-with-lease
```

Note: External fork — cannot push directly.

vivekchand:
Hi @dumko2001 — appreciating this security fix. ClawMetry main has evolved a lot since this PR was opened, so it now conflicts.

**Conflicts**: `clawmetry/cli.py` and `tests/test_cli.py` (plus a `.gitignore` minor conflict).

**What to do**: `clawmetry/cli.py` still contains the `connect`/`status` commands this PR masks the encryption key in — the file just grew larger (~1900 LOC). The masking logic should still apply to the same code paths. Rebase + re-target the same lines.

If you'd prefer not to rebase, ping me and I'll cherry-pick the fix.

vivekchand:
Hi @dumko2001 — thank you for the security-mindedness here, but on review I'm going to close this rather than rebase it.

The proposed change masks the encryption key in **both** display paths:
1. Line 517 (`Keep this secret key safe (like a password)` reminder) — this is the **deliberate** one-time display intended as the user's backup of their key, identical to e.g. ssh-keygen showing the private key fingerprint. Masking it defeats the purpose: the user can never recover the full key.
2. Line 510 (`--key-only` flag) — this exists specifically for the host-side NemoClaw OTP flow where the operator pipes the key elsewhere. Masking it **breaks the flow** — there's no other way to retrieve the full key.

Both paths only print after a successful `clawmetry connect` run (user-initiated), and the key is also written to `~/.clawmetry/config.json` (chmod 0o600), so there's no separate disclosure channel a masked print would close.

If you've got a specific threat model in mind (e.g. screen-sharing / log forwarders capturing stdout) — happy to discuss in a follow-up issue, but the right answer is probably `--quiet` mode or terminal-only output detection, not unconditional masking. Closing for now.

PR #514, opened 2026-04-03
fix(robustness): add assertion to prevent unsafe lock release

dumko2001:
## Summary

Add assertion to prevent unsafe lock release in NAT exporter.

## Changes

- Add assertion to verify lock is held before release
- Prevents RuntimeError from releasing unheld lock

## Testing

Verified the assertion catches unsafe release scenarios


vivekchand:
Triage status (zero-backlog sweep 2026-05-12): clean mergeable, still applicable. The assertion at `integrations/nat/clawmetry_nat/exporter.py` `_flush_locked` is genuinely useful defensive coding (catches misuse of an internal helper). 40 days old but the fix is correct and the auto-format noise is contained. @dumko2001 — thanks for this!

Recommend merge as-is. (Not auto-merging per the sweep's read-only-on-third-party-PRs rule.)

PR #515, opened 2026-04-03
fix(resource): use context manager for file handles

dumko2001:
## Summary

Use context managers for file handles to prevent leaks.

## Changes

- Wrap file open() calls with context managers (with statement)
- Ensures files are properly closed even on exceptions

## Testing

Verified file handles are properly closed


vivekchand:
Closing: this fix (context managers for file handles) was already implemented in main (commit 6ae4ff2). This PR is superseded.

PR #518, opened 2026-04-03
fix(reliability): use atomic write for state files

dumko2001:
## Summary

Use atomic write pattern for state files to prevent corruption.

## Changes

- Write to temp file then rename for atomic updates
- Prevents state corruption from partial writes

## Testing

Verified atomic write behavior under failure conditions


vivekchand:
Closing: this fix (atomic write for state files) was already implemented in main (commit 3ea4edf). This PR is superseded.

PR #529, opened 2026-04-03
fix(correctness): validate token counts are non-negative

dumko2001:
## Summary

Validate that token counts cannot be set to negative values.

## Changes

- Add validation to reject negative token values
- Prevents invalid state from negative token counts

## Testing

Verified negative values are rejected

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,4 @@ Thumbs.db
 
 # Logs
 *.log
+.clawmetry-fleet.db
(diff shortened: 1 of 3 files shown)


vivekchand:
Closing: this automated fix PR has been stale since April 3 and conflicts heavily with main. The proxy module has evolved significantly. If token count validation is still needed, a fresh PR against current main would be required. — PR hygiene bot

PR #531, opened 2026-04-03
fix(security): fix permission race in config file write

dumko2001:
## Summary

Fix TOCTOU race in config file permission setting.

## Changes

- Set permissions before file is fully written
- Prevents race condition where file could be accessed with wrong permissions

## Testing

tests/test_config_permissions.py - verifies secure permission handling

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,4 @@ Thumbs.db
 
 # Logs
 *.log
+.clawmetry-fleet.db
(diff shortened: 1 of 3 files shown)


vivekchand:
⚠️ **This PR appears to revert improvements made in #534** 

This PR deletes  and reverts the lazy import pattern that was added to resolve circular import issues (see commit 9cade90).

The fix changes:
- Remove lazy import pattern in 
- Delete 
- Revert to direct  which causes circular import issues

Since PR #534 already resolved circular import dependencies and this PR would reintroduce those issues, this PR should likely be closed. The feature described (config permission race fix) should be submitted as a new PR that builds on the current main branch rather than reverting fixes.

vivekchand:
Closing as this PR would revert circular import fixes from #534. The permission race fix should be submitted as a new PR against current main that doesn't delete test_circular_import.py.

PR #533, opened 2026-04-03
refactor: add Flask application factory pattern

dumko2001:
## Summary

Add Flask application factory pattern for better testability.

## Changes

- Create app factory function that creates Flask instances
- Enables easier testing and configuration

## Testing

tests/test_app_factory.py - verifies factory pattern works

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,3 +25,4 @@ Thumbs.db
 
 # Logs
 *.log
+.clawmetry-fleet.db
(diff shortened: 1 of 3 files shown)


vivekchand:
⚠️ **This PR appears to be stale and based on an old codebase**

This PR adds Flask application factory pattern but also deletes , which is needed to verify circular import resolution from #534.

The PR is 8 days old and doesn't include recent fixes like #534 (circular import resolution), #550 (transcript viewer), #557/#559 (budget fixes), #560 (provider cost breakdown), #562 (cron health), and #573 (multi-provider cost comparison).

Please rebase against current main and resubmit if the Flask factory pattern is still needed.

vivekchand:
Closing as stale. This PR is based on an old codebase and deletes test_circular_import.py which verifies circular import fixes from #534. Please resubmit against current main if the Flask factory pattern is still needed.

PR #535, opened 2026-04-03
fix(security): remove email information disclosure

dumko2001:
## Summary

Remove information disclosure in OTP email error messages.

## Changes

- Use generic error messages that don't reveal email existence
- Prevents user enumeration via auth endpoint

## Testing

Verified error messages don't disclose email information


vivekchand:
We currently show only a masked email when connecting via api key

PR #596, opened 2026-04-14
fix: sync daemon was missing sessions due to lexical 100-file cap

vivekchand:
## Summary
The cloud sync daemon walked the OpenClaw sessions directory with `sorted(sessions_dir.glob("*.jsonl"))[-100:]`. Session filenames are UUIDs, so that lexical sort has nothing to do with recency — and the trailing `[-100:]` slice silently dropped everything else. Users with more than ~100 session files had arbitrary, non-recency-based history gaps in the cloud dashboard.

## Reported case
A paying user (Cloud Pro since 2026-03-15) reported daily token spend looked "materially off" in the cloud dashboard. Investigation on the server side (paired PR: [clawmetry-cloud#309](https://github.com/vivekchand/clawmetry-cloud/pull/309)) showed his sessions table had **zero rows older than 2026-04-13**, even though global retention on that table goes back to March. His older JSONLs exist on his laptop — they're just lexically-earlier UUIDs that fell outside the trailing 100-file slice and therefore were never uploaded.

## Fix
- Collect all `*.jsonl` files with their `mtime`.
- Sort ascending by `mtime` so newer files win any display-name / model ties.
- Remove the 100-file cap entirely.

The mtime-skip via `last_mtimes` still short-circuits unchanged files before any file I/O, so the steady-state sync stays cheap even for workspaces with thousands of sessions.

## Test plan
- [ ] Local: populate `~/.openclaw/agents/main/sessions` with >200 JSONL files spanning several mtime days. Run the sync cycle. Confirm all files get processed once and only changed files re-process on the next cycle.
- [ ] Regression: confirm `last_mtimes` state file grows and subsequent syncs are fast (no file reads) when nothing changed.
- [ ] Spot-check cloud dashboard shows the full history after a fresh sync with this client.

## Companion PR
- Server: [clawmetry-cloud#309](https://github.com/vivekchand/clawmetry-cloud/pull/309) — fixes `started_at` dropping on ingest and the daily-chart bucketing bug.

### DIFF
--- a/clawmetry/sync.py
+++ b/clawmetry/sync.py
@@ -1308,13 +1308,21 @@ def sync_session_metadata(config: dict, state: dict = None) -> int:
             return 0
 
         session_rows = []
-        for fpath in sorted(sessions_dir.glob("*.jsonl"))[-100:]:
-            # Skip files that haven't changed since last sync
+        # Sort by mtime ascending so the newest files are processed last and
+        # win any display-name/model ties. Session filenames are UUIDs, so a
+        # lexical sort + [-100:] slice previously gave a non-deterministic
+        # sample of files and silently dropped the rest — users with more
+        # than ~100 session files saw arbitrary history gaps in the cloud
+        # dashboard. mtime-skip below keeps subsequent syncs cheap.
+        jsonl_files = []
+        for fpath in sessions_dir.glob("*.jsonl"):
             try:
-                current_mtime = fpath.stat().st_mtime
-                if last_mtimes.get(fpath.name) == current_mtime:
-                    continue
+                jsonl_files.append((fpath, fpath.stat().st_mtime))
             except OSError:
+                continue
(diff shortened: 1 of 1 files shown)


vivekchand:
Pushed follow-up: **sync newest sessions first and flush per-batch** (7e95008).

Changes:
- `jsonl_files.sort(..., reverse=True)` — newest first
- Restructured the parse loop to flush each 50-file batch immediately via a `_flush()` closure, instead of accumulating the full `session_rows` list and batching at the end
- Default-model fallback now resolves up front from `sessions.json`, then maintains a running mode across batches (vs. computing the global mode after a full pre-scan)

Net UX: a user with a large backlog sees today's sessions in the cloud dashboard within seconds, with older history backfilling over the rest of the cycle — instead of waiting for the full parse before any HTTP POST goes out.

PR #685, opened 2026-04-17
fix(e2e): git fetch before version check + accept 302 for app.clawmetry.com

vivekchand:
## Summary
- E2E smoke test was reporting `VERSION_MISMATCH` because `git show public/main:dashboard.py` reads the local tracking ref — stale on the Xeon node (showed 0.12.106 while PyPI had 0.12.110). Fix: run `git fetch public` before comparing.
- Cloud endpoint check was failing on `app.clawmetry.com` returning 302 → `/fleet/`. Fix: accept 200 or 302 (redirect to fleet is expected for authenticated sessions).
- Extracted curl-based smoke tests into `scripts/e2e-smoke.sh` so the cron can reference a versioned script rather than embedding it inline.

## Test plan
- [ ] Run `bash scripts/e2e-smoke.sh` on Xeon node — all tests should pass
- [ ] Confirm no VERSION_MISMATCH on a freshly cloned node

### DIFF
--- a/scripts/e2e-smoke.sh
+++ b/scripts/e2e-smoke.sh
@@ -1,0 +1,104 @@
+#!/usr/bin/env bash
+# ClawMetry E2E smoke tests (curl-based, no browser required).
+# Run this on a node that has clawmetry installed and the git repo present.
+#
+# Usage:
+#   bash scripts/e2e-smoke.sh
+#   REPO=/path/to/openclaw-dashboard bash scripts/e2e-smoke.sh
+#
+# Exit code: 0 = all pass, 1 = one or more failures.
+
+set -euo pipefail
+
+REPO="${REPO:-$(git -C "$(dirname "$0")/.." rev-parse --show-toplevel 2>/dev/null || echo /home/vivek/projects/openclaw-dashboard)}"
+PORT=9901
+BASE="http://localhost:$PORT"
+FAILURES=()
+
+pass() { echo "  PASS  $1"; }
+fail() { echo "  FAIL  $1"; FAILURES+=("$1"); }
+
+# ---------------------------------------------------------------------------
+# 1. Install & CLI
+# ---------------------------------------------------------------------------
+echo "=== 1. Install & CLI ==="
+clawmetry --version >/dev/null 2>&1 && pass "clawmetry --version" || fail "clawmetry --version"
+clawmetry status >/dev/null 2>&1 && pass "clawmetry status" || fail "clawmetry status"
+
+# ---------------------------------------------------------------------------
+# 2. Dashboard starts
+# ---------------------------------------------------------------------------
+echo "=== 2. Dashboard starts ==="
+pkill -f "clawmetry.*${PORT}" 2>/dev/null || true
+sleep 1
+clawmetry --port "$PORT" &
+CM_PID=$!
+sleep 3
+
+status=$(curl -s -o /dev/null -w '%{http_code}' "$BASE/")
+[ "$status" = "200" ] && pass "dashboard / → 200" || fail "dashboard / → $status (expected 200)"
+
+# ---------------------------------------------------------------------------
+# 3. API endpoints
+# ---------------------------------------------------------------------------
+echo "=== 3. API endpoints ==="
+for path in /api/overview /api/sessions /api/crons /api/memory /api/flow; do
+    code=$(curl -s -o /dev/null -w '%{http_code}' "$BASE$path")
+    [ "$code" = "200" ] && pass "GET $path → 200" || fail "GET $path → $code (expected 200)"
+done
+
+overview=$(curl -s "$BASE/api/overview")
+echo "$overview" | python3 -c "import sys,json; d=json.load(sys.stdin); ok='sessions' in d or 'activeSessions' in d; sys.exit(0 if ok else 1)" \
+    && pass "/api/overview has sessions key" || fail "/api/overview missing sessions/activeSessions key"
+
+# ---------------------------------------------------------------------------
+# 4. install.sh correctness
+# ---------------------------------------------------------------------------
+echo "=== 4. install.sh ==="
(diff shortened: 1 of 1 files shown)


vivekchand (commented):
## Test plan & review notes

**What changed**
- New `scripts/e2e-smoke.sh` (104 lines) that extracts all curl-based E2E smoke checks out of an inline cron script into a versioned, standalone file. Key fixes inside: `git fetch public` before the PyPI-vs-GitHub version comparison, and accepting HTTP 302 (redirect to `/fleet/`) alongside 200 for `app.clawmetry.com`.

**Still relevant?**
- 11 days old, no reviewer activity. The VERSION_MISMATCH false-positive and the `app.clawmetry.com` 302 failure are real recurring irritants (several closed PRs — #266, #279, #298, #348, #350 — all had to bump versions precisely because the stale-ref check kept misfiring). The `git fetch public` guard is a correct structural fix. The PR is still relevant and ready to merge.

**Overlap note**
- No overlap found with other open or recently merged PRs. The closed e2e-related PRs all addressed version-number drift, not the stale-ref root cause that this PR fixes.

**Smoke commands**
```bash
# On a node that has the repo and clawmetry installed:
bash scripts/e2e-smoke.sh

# Verify the fetch line exists and is not inside a subshell that silently skips errors
grep -n "fetch public" scripts/e2e-smoke.sh

# Confirm the 302-acceptance logic is correct
grep -A3 "app.clawmetry.com" scripts/e2e-smoke.sh

# Dry-run syntax check
bash -n scripts/e2e-smoke.sh
```

**Likely failure modes**
- The script has `set -euo pipefail` at the top but the `FAILURES` array pattern uses `|| fail "..."` to avoid early exit — that's correct, but worth auditing each test block to confirm none accidentally triggers `set -e` before appending to `FAILURES`.
- `REPO` defaults to `/home/vivek/projects/openclaw-dashboard` if `git rev-parse` fails — this hardcoded path will cause a silent wrong-repo comparison on any other machine. Consider failing loudly (`|| { echo "Cannot find repo root"; exit 1; }`) instead.
- Section 4 (`install.sh` checks) hits `https://clawmetry.com/install.sh` over the network; if the CI runner has no outbound internet the test reports a false failure rather than skipping. A `|| true` guard or an explicit skip on network error would make the suite more portable.
- The script starts a `clawmetry --port 9901` daemon but only kills `$CM_PID` at the end — if the script exits early via `set -e`, the daemon leaks. A `trap "kill $CM_PID 2>/dev/null || true" EXIT` placed right after the `&` would fix this.
- `pip index versions` is deprecated in newer pip versions; `pip install clawmetry== 2>&1 | grep -oP ...` or `pip index versions` with a `pip >= 21` pin in the CI requirements is worth noting.

**Issue link**
- No issue linked in the PR body. Suggest linking to the most recent VERSION_MISMATCH detection issue, or opening one if none exists, to give future readers the context trail.

---

vivekchand (commented):
## Test plan & review notes

**What changed**
- E2E smoke test: `git fetch public` before version comparison to prevent stale tracking refs causing false `VERSION_MISMATCH`
- Accept HTTP 302 (redirect to `/fleet/`) as a success response from `app.clawmetry.com`
- Extracted curl-based smoke tests into `scripts/e2e-smoke.sh`

**Smoke commands**
- `bash -n scripts/e2e-smoke.sh` — shell syntax check
- `bash scripts/e2e-smoke.sh` against a running instance
- Confirm the CI node has a remote named `public` configured, since `git fetch public` will fail without it

- `git fetch public` assumes a remote named `public` exists on the CI runner — verify this is set up in the CI workflow before this lands
- If `scripts/e2e-smoke.sh` isn't marked executable (`chmod +x`), the cron job will fail with `Permission denied`

**Issue link**
- Note: PR #776 has `Closes #685` in its title — worth clarifying whether that refers to this open PR or to a separately filed issue #685, to avoid accidental cross-linking on merge

---

vivekchand (commented):
## Test plan & review notes

**What changed**
- Adds `scripts/e2e-smoke.sh`: a self-contained, curl-based smoke-test covering CLI sanity, key `/api/*` endpoints, `install.sh` correctness, a version-sync check (PyPI vs GitHub public/main), and cloud endpoint reachability. The version check now does `git fetch public` before comparing, and the cloud check accepts both `200` and `302` from `app.clawmetry.com`.

**Smoke commands**
```bash
pip install clawmetry  # or use a local dev install
bash scripts/e2e-smoke.sh
# Expect: "All tests passed." and exit 0
```

**One thing to check**: the version-sync step does `git -C "$REPO" fetch public 2>/dev/null || true` — this assumes a remote named `public` exists. In a fresh clone there's only `origin`. A small fallback would make it more robust in CI:
```bash
git -C "$REPO" fetch public 2>/dev/null || git -C "$REPO" fetch origin 2>/dev/null || true
```

**Issue link**
- No issue linked. If this was prompted by a specific CI failure, a `Refs #N` line would help future readers connect the dots.

---

vivekchand:
## Test plan & review notes

**What changed**
- New `scripts/e2e-smoke.sh`: extracts curl-based E2E smoke tests from the cron inline script; adds `git fetch public` before version comparison to fix stale local ref; accepts 302 redirect from `app.clawmetry.com`

**Smoke commands**
- `bash scripts/e2e-smoke.sh` on a node with a live clawmetry install and a running gateway
- Confirm no `VERSION_MISMATCH` on a freshly cloned node

- Remote name `public` is node-specific — nodes where the upstream is named `origin` will still fail. Consider a fallback: `git remote | grep -q public && git fetch public || git fetch origin`
- 302 acceptance is correct for authenticated sessions, but a CDN-level 301/307 redirect would still fail; `curl -L` (follow all redirects) is more robust

**Issue link**
- No `Closes #N` in the PR body. If there's a tracking issue for E2E flakiness or the version-check race condition, linking it would help.

---

PR #776, opened 2026-04-22
fix: normalize nextRunAtMs to number (closes #685)

vivekchand:
## What
The cron.list RPC endpoint was returning nextRunAtMs as a string or object instead of a number (milliseconds timestamp), causing TypeError in the JavaScript UI when attempting to calculate time deltas or format dates.

## How
- Added `_normalize_next_run_at_ms()` helper function to dashboard.py that normalizes nextRunAtMs values (handles strings, objects, ISO timestamps, and numeric strings)
- Applied normalization in `/api/crons` endpoint (routes/crons.py) in the _with_costs function
- Applied normalization in `/api/cron/health-summary` endpoint

## Changes
- `dashboard.py`: Added `_normalize_next_run_at_ms()` module-level helper
- `routes/crons.py`: Updated `_with_costs()` and `api_cron_health_summary()` to use the normalizer

## Testing
- Syntax verified: `python3 -m py_compile dashboard.py routes/crons.py`

Closes #685

### DIFF
--- a/dashboard.py
+++ b/dashboard.py
@@ -13097,6 +13097,27 @@ def _get_crons_from_files():
     return []
 
 
+def _normalize_next_run_at_ms(state):
+    """Ensure nextRunAtMs is a number (ms timestamp) or null (closes #685)."""
+    if not isinstance(state, dict):
+        return None
+    val = state.get("nextRunAtMs")
+    if val is None:
+        return None
+    if isinstance(val, (int, float)):
+        return int(val)
+    if isinstance(val, str):
+        try:
+            # ISO timestamp or numeric string
+            if "T" in val:
+                from dateutil import parser as _dtp
+                return int(_dtp.parse(val).timestamp() * 1000)
+            return int(float(val))
+        except Exception:
+            return None
+    return None
+
+
 def _get_memory_files():
     """List workspace memory files."""
     result = []
--- a/routes/crons.py
+++ b/routes/crons.py
@@ -77,6 +77,12 @@ def api_crons():
             j2["cost_usd"] = round(cost_by_job.get(idx, 0.0), 6)
             j2["cost_session_count"] = int(count_by_job.get(idx, 0))
             j2["cost_session_ids"] = session_ids_by_job.get(idx, [])
+            # Normalize nextRunAtMs to number (closes #685)
+            state = j2.get("state") or {}
+            normalized_next_run = _d._normalize_next_run_at_ms(state)
+            if normalized_next_run is not None:
+                state["nextRunAtMs"] = normalized_next_run
+            j2["state"] = state
             out.append(j2)
         return out
 
@@ -315,7 +321,7 @@ def api_cron_health_summary():
         last_duration_ms = state.get("lastDurationMs") or 0
         consecutive_failures = state.get("consecutiveFailures") or 0
         last_error = state.get("lastError", "")
-        next_run_ms = state.get("nextRunAtMs") or 0
+        next_run_ms = _d._normalize_next_run_at_ms(state) or 0
 
         # Detect silent jobs: enabled, has run before, but hasn't run in >2.5x expected interval
         is_silent = False


vivekchand (commented):
## Test plan & review notes

**What changed**
- Added `_normalize_next_run_at_ms()` in `dashboard.py` that coerces `nextRunAtMs` from any of string/float/ISO-8601/null into an integer millisecond timestamp (or `None`), and wired it into `_with_costs()` and `api_cron_health_summary()` in `routes/crons.py` so the cron UI and health summary always receive a number instead of a raw string.

**Smoke commands**
- `python3 dashboard.py --port 8900` → Crons tab — next-run times should display correctly
- `curl -sS http://localhost:8900/api/crons` — verify `nextRunAtMs` is a number, not a string

- **ISO string with timezone offset** — `dateutil.parser.parse()` handles most forms, but `dateutil` is not in `requirements.txt`. If it is absent the `from dateutil import parser` inside the `try` will raise `ImportError`, which is silently caught and returns `None`, so the next-run time disappears entirely rather than erroring visibly. Consider adding `python-dateutil` to `requirements.txt` or catching `ImportError` separately with a warning log.
- **Float string `"1234567890123.5"`** — `int(float(val))` handles this correctly; no issue.
- **`null` / `None` in the gateway response** — handled: early `if val is None: return None`. The call site uses `or 0` as fallback, so the UI shows epoch (1970) rather than crashing; that is probably fine but worth a quick visual check.
- **Negative or zero value** — not guarded; a negative ms timestamp would propagate to the UI as a past time. Low risk but worth noting.
- **`_with_costs` only normalizes when `normalized_next_run is not None`** — if the value is an unparseable string the original bad string is left in `state["nextRunAtMs"]` unchanged (the `if normalized_next_run is not None` guard skips the write-back). A cleaner approach would be to always write back the result (including `None`) so the bad string never reaches the frontend.

**Issue link**
- Closes #685 ✓ (in title)

---

vivekchand (commented):
## Test plan & review notes

**What changed**
- Adds `_normalize_next_run_at_ms()` to `dashboard.py` and applies it in `routes/crons.py` so `nextRunAtMs` is always returned as an integer millisecond timestamp instead of a raw string or object, preventing JS `TypeError` in the cron UI (closes #685).

**Still relevant?**
- Opened 2026-04-22, 13 days old, no prior review comments. The base branch (`b329bb3`) is still the current `main` HEAD and the PR is marked `mergeable_state: clean`, so there is no rebase needed. The fix is unambiguously still needed — no closed PR has landed anything touching `nextRunAtMs` normalization.

**Overlap note**
- N/A

**Smoke commands**
```bash
# Syntax check
python3 -m py_compile dashboard.py routes/crons.py

# Confirm normalizer handles all advertised input shapes
python3 - <<'EOF'
import sys; sys.path.insert(0, '.')
import dashboard as d
cases = [
    ({"nextRunAtMs": 1746000000000},    1746000000000),
    ({"nextRunAtMs": "1746000000000"},  1746000000000),
    ({"nextRunAtMs": "2026-04-29T10:00:00Z"}, None),  # requires dateutil; expect int
    ({"nextRunAtMs": None},             None),
    ({"nextRunAtMs": "garbage"},        None),
    ({},                                None),
]
for state, expected in cases:
    got = d._normalize_next_run_at_ms(state)
    status = "OK" if (expected is None and got is None) or (got is not None) else "FAIL"
    print(status, repr(state), "->", got)
EOF

# Hit the live endpoint and verify nextRunAtMs is always a number or absent
curl -s http://localhost:8900/api/crons | python3 -c "
import json, sys
data = json.load(sys.stdin)
for job in data.get('jobs', []):
    v = job.get('state', {}).get('nextRunAtMs')
    assert v is None or isinstance(v, (int, float)), f'bad type: {type(v)} {v}'
print('all nextRunAtMs values are numeric or null')
"

# Health summary endpoint
curl -s http://localhost:8900/api/cron/health-summary | python3 -c "
import json, sys; data = json.load(sys.stdin); print('health summary OK, jobs:', len(data.get('jobs',[])))
"
```

**Likely failure modes**
- **`dateutil` is not in `requirements.txt`**: The ISO-timestamp branch (`if "T" in val`) does `from dateutil import parser`. `python-dateutil` is not listed as a dependency, so this import will `ImportError` on a fresh install when a gateway returns an ISO string for `nextRunAtMs`. The safe fix is either add `python-dateutil` to `requirements.txt`, or inline the conversion with `datetime.fromisoformat` (stdlib, Python 3.7+) so no extra dep is needed:
  ```python
  from datetime import datetime, timezone
  return int(datetime.fromisoformat(val.replace("Z", "+00:00")).timestamp() * 1000)
  ```
- **`state` mutation in `_with_costs`**: The code does `state["nextRunAtMs"] = normalized_next_run` then `j2["state"] = state` — if `j2["state"]` was already a reference to the same dict object from the gateway response cache, this mutates the cached value in-place. Probably fine given the current shallow-copy semantics, but worth a note.
- **`health-summary` fallback**: `_normalize_next_run_at_ms(state) or 0` — if `state` is `None` (possible when a cron has never been run), `_normalize_next_run_at_ms` returns `None` immediately without error, so `or 0` produces `0`. That path is correct, but worth a unit test.

**Issue link**
- Closes #685 (confirmed in PR body and code comments)


---

vivekchand (commented):
## Test plan & review notes

**What changed**
- `_normalize_next_run_at_ms()` helper in `dashboard.py` coerces `nextRunAtMs` from string/ISO/object → int milliseconds
- Applied in `routes/crons.py` `_with_costs` and `api_cron_health_summary`

**Smoke commands**
- `python3 -c 'import ast; ast.parse(open("dashboard.py").read())'` — syntax clean
- `curl -sS http://localhost:8900/api/crons | python3 -c "import sys,json; d=json.load(sys.stdin); print([c.get('nextRunAtMs') for c in d.get('crons',[])])"` — all values should be integers or `null`, never strings
- `curl -sS http://localhost:8900/api/cron/health-summary` — same check

- ISO timestamp strings like `"2026-05-10T12:00:00Z"` need timezone-aware parsing — `datetime.fromisoformat()` on Python <3.11 rejects the `Z` suffix; use `.replace('Z','+00:00')` or `dateutil`
- Confirm `_normalize_next_run_at_ms` doesn't shadow an existing helper of the same name in `dashboard.py`

**Issue link**
- PR title says `Closes #685` — please confirm this refers to issue #685 (not the open PR #685, `fix(e2e): git fetch before version check`), to avoid inadvertently cross-linking on merge

---

vivekchand (commented):
## Test plan & review notes

**What changed**
- Added `_normalize_next_run_at_ms()` helper in `dashboard.py` that coerces `nextRunAtMs` from string/ISO-timestamp/object to an integer milliseconds value, and applied it in `_with_costs()` and `api_cron_health_summary()` in `routes/crons.py`.

**Smoke commands**
- `make test`
- `curl -sS http://localhost:8900/api/crons` — confirm `nextRunAtMs` is always a number (not string)
- Test with a cron that has `nextRunAtMs: "1234567890"` (string from gateway) and one with no `nextRunAtMs` field

- `_normalize_next_run_at_ms` returns `None` when the value is missing/uncoercible, but `_with_costs()` only writes back when `normalized_next_run is not None`, so a pre-existing numeric `0` (falsy but valid) would pass through unchanged — verify `int(0)` is handled correctly by the `isinstance(val, (int, float))` branch (it is, since `0` isn't `None`).
- `api_cron_health_summary` uses `_d._normalize_next_run_at_ms(state) or 0` — if the state dict has `nextRunAtMs: 0` (epoch), the `or 0` short-circuits correctly because `0 or 0` is still `0`, so no bug, but worth a conscious check.
- The ISO-timestamp path pulls in `dateutil` at call time; if `python-dateutil` is not installed (it's not in `requirements.txt`), that branch will raise `ImportError` swallowed by the bare `except Exception`, silently returning `None` instead of the parsed value.
- `_with_costs()` mutates the `state` dict in-place (`state["nextRunAtMs"] = normalized_next_run`) and then re-assigns `j2["state"] = state` — if `state` was originally `{}` (from `j2.get("state") or {}`), the normalized value is never stored back into `j2` because the reference is to a local `{}`, not `j2["state"]`. Confirm this edge case is covered.

**Issue link**
- Closes #685 ✓

---

vivekchand:
## Test plan & review notes

**What changed**
- New `_normalize_next_run_at_ms()` helper in `dashboard.py` coerces `nextRunAtMs` strings/objects/ISO timestamps to int milliseconds; applied in `routes/crons.py` `_with_costs()` and `api_cron_health_summary()`

**Smoke commands**
- `make test`
- `curl -sS http://localhost:8900/api/crons` → `nextRunAtMs` on each cron entry should be an integer, not a string or object
- `curl -sS http://localhost:8900/api/cron/health-summary` → same check

- ISO strings with timezone offsets (e.g. `2026-04-22T09:00:00+05:30`) — confirm `fromisoformat` handles both naive and aware timestamps across Python 3.9+
- `None` / `{}` / other unexpected shapes from gateway should fall through gracefully; a `None` return from the normalizer should be handled by the caller

**Issue link**
- Closes #685 (issue) ✓ — note: there's also an open *PR* #685 (E2E smoke fix) in the queue; no conflict, different things.

---

PR #789, opened 2026-04-23
fix: install.sh uses clawmetry onboard command

vivekchand:
Fixes E2E test failure: install.sh should reference 'clawmetry onboard' instead of 'clawmetry setup'


vivekchand:
Closing as duplicate of #793 which makes the same install.sh setup→onboard change and is tagged as [RELEASE].

PR #860, opened 2026-04-30
[RELEASE] feat(crons): 100% accurate interval predictions via anchorMs + past failures in Calendar

vivekchand:
## Summary
User asked: *"how accurate is this?? can it not be 100% accurate?"* + *"if we know active crons, can we just not predict for next dates? and also show as failure for past if they were failure?"*

Two fixes shipping together:

### 1. Interval predictions are now EXACT
The agent records `schedule.anchorMs` on every interval job (origin from which all fires are scheduled). With that anchor:
\`\`\`
next = anchorMs + ceil((now - anchorMs) / everyMs) * everyMs
\`\`\`
matches the agent's own scheduler exactly.

**Verified against live data**:
- MC Instant Response (every 15m, anchor at minute :17) → predicts `05:51` not the rough `05:53` we showed before
- English Coach (every 3h) → `07:17` not `08:38`
- OpenClaw Polish (every 1h) → `05:43` not `06:38`

Cron-expression schedules were already 100% accurate via the minute-by-minute walk. Now interval schedules match.

Falls back to `now + everyMs` (worst-case upper bound) only when `anchorMs` is missing — never lies, just degrades honestly.

### 2. Past failures land in Calendar's "Recently ran"
When the user opens the Calendar tab, lazy-load `/api/cron/<id>/runs` for every active job (parallel, 7d window). The Calendar's "Recently ran" section then renders **red icons for `status='error'`**, green for `ok` — honestly.

Dedups against any `state.lastRunAtMs` the agent already supplied so we don't double-count.

## Test plan
- [ ] Open Crons → Calendar → "Coming up" times match the agent's actual scheduler within 1-2 minutes
- [ ] Failed crons in the last 7d show up in "Recently ran" with red ❌ icon
- [ ] Successful crons show with green ✅ icon
- [ ] Schedules without anchorMs (legacy data) still render with the worst-case upper bound, no crash

### DIFF
--- a/clawmetry/static/css/dashboard.css
+++ b/clawmetry/static/css/dashboard.css
@@ -197,6 +197,7 @@
   .cron-status.pending { background: var(--bg-warning); color: var(--text-warning); }
   .cron-status.no-data { background: rgba(107,114,128,0.18); color: #9ca3af; cursor: help; }
   .cron-status.stale { background: rgba(245,158,11,0.18); color: #f59e0b; cursor: help; }
+  .cron-status.scheduled { background: rgba(96,165,250,0.18); color: #60a5fa; cursor: help; }
 
   /* Cron view tabs (Active / Paused / Calendar) */
   .cron-view-tabs { display: flex; gap: 4px; margin-bottom: 10px; padding: 4px; background: var(--bg-secondary); border-radius: 10px; width: fit-content; }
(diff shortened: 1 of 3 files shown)


vivekchand:
Closing for clean rebase — superseded by fresh PR.

PR #866, opened 2026-05-04
feat: reasoning chain viewer — structured thinking visualization (closes #565)

vivekchand:
## Summary

- **New backend** `routes/reasoning.py`: `GET /api/reasoning?session=SESSION_ID` parses session JSONL for assistant messages with thinking blocks, segments each into typed logical steps, and returns structured chains with token estimates and efficiency ratios
- **Blueprint registered** in `dashboard.py` alongside all other route modules
- **Frontend viewer** in `app.js`: `loadReasoningChain()` fetches the API and renders a collapsible panel with summary bar (thinking tokens | steps | efficiency ratio) and a vertical step flow with color-coded type badges (premise=blue, hypothesis=yellow, constraint=orange, conclusion=green, analysis=gray)
- **"View chain" button** appears next to THINK events in the expanded brain turn timeline — toggles the panel on/off
- **🧠 badge** added to session filter chips that contain THINK events, making reasoning-heavy sessions visually identifiable

## Implementation details

- Step classification uses keyword heuristics (no ML): premise/hypothesis/constraint/conclusion/analysis
- Token estimate: word_count x 1.3 (rough approximation consistent with spec)
- Late import dashboard as _d pattern for SESSIONS_DIR, consistent with all other route modules
- Frontend uses existing CSS variables (var(--bg-secondary), var(--border), var(--text-muted)) for consistent Brain tab styling

## Test plan

- [x] tests/test_circular_import.py passes (no circular import introduced)
- [x] tests/test_cli.py passes (dashboard starts cleanly with new blueprint)
- [ ] Manually: open Brain tab, expand a turn with a THINK event, click "View chain" to see panel with step breakdown
- [ ] Manually: session chips for reasoning-capable sessions show the reasoning badge
- [ ] GET /api/reasoning?session=<valid-session-id> returns {session_id, chains, summary} JSON

### DIFF
--- a/clawmetry/static/js/app.js
+++ b/clawmetry/static/js/app.js
@@ -2221,12 +2221,17 @@ window.toggleBrainFilterExpanded = function() {
 function _brainChipHtml(s) {
   var isActive = _brainFilter === s.id;
   var icon = s.icon || (s.id === 'main' ? '🧠' : '🤖');
+  // Add 🧠 badge if this source has any THINK events (reasoning sessions)
+  var hasReasoning = (_brainAllEvents || []).some(function(ev) {
+    return ev.source === s.id && ev.type === 'THINK';
+  });
+  var reasoningBadge = hasReasoning ? ' <span title="Has reasoning chains" style="font-size:9px;">&#129504;</span>' : '';
   return '<button class="brain-chip' + (isActive ? ' active' : '') + '" data-source="' +
     escHtml(s.id) + '" title="' + escHtml(s.id) +
     '" onclick="setBrainFilter(this.dataset.source,this)" style="padding:3px 10px;border-radius:12px;border:1px solid ' +
     s.color + ';background:' + (isActive ? 'rgba(100,100,100,0.2)' : 'transparent') +
     ';color:' + s.color + ';font-size:11px;cursor:pointer;font-weight:' +
-    (isActive ? '600' : '400') + ';">' + icon + ' ' + escHtml(s.label || s.id) + '</button>';
+    (isActive ? '600' : '400') + ';">' + icon + ' ' + escHtml(s.label || s.id) + reasoningBadge + '</button>';
 }
 
 function renderBrainFilterChips(sources) {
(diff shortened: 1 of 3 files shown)


vivekchand (commented):
## Test plan & review notes

**What changed**
- New `routes/reasoning.py` (`GET /api/reasoning?session=ID`) parses `thinking` blocks from session JSONL, segments them into typed logical steps (premise / hypothesis / constraint / conclusion / analysis) via regex heuristics, and returns chain + summary JSON; `app.js` wires up a "View chain" toggle button on every `THINK` event in the brain stream and renders the expandable chain inline.

**Smoke commands**
- `python3 dashboard.py --port 8900` then open the **Brain** tab, trigger or replay a session that used extended thinking, and click "View chain" on any THINK event row

**What to look at visually**
- `http://localhost:8900` → Brain tab → any THINK row → "&#129504; View chain" button — clicking should toggle an inline panel showing summary bar (thinking tokens, steps, efficiency) and per-step type badges; clicking again should collapse it
- Brain filter chips for sources with THINK events should show a small &#129504; badge

- **Token count is estimated, not real**: `thinking_tokens` and `answer_tokens` are derived by multiplying word count by 1.3 — if the JSONL ever carries actual `usage.thinking_tokens` fields (Anthropic API does), those are ignored, so efficiency ratios will be wrong
- **`_rcContainerId` collision**: the container ID is built from `te.source.slice(0,8)` + last 6 digits of the timestamp string — two THINK events from the same source within the same second would share an ID and break the toggle
- **No path-traversal guard on `session_id`**: `_parse_session_reasoning` joins the raw `session` query param directly onto `SESSIONS_DIR` with `os.path.join`; a value like `../../etc/passwd` won't escape the dir on most OSes because there's no `..` component check or `os.path.realpath` comparison
- **`raw_thinking` is returned in the API response** but never rendered — either strip it to save bandwidth or document it as intentional for future use
- **JS inline event handler string escaping**: the `onclick` for the "…more" expand toggle uses triple-nested string interpolation with manual escaping; a step whose content contains a single quote would break the generated HTML

**Issue link**
- Closes #565 ✓ (already in title)


---

Issue #28, opened 2026-02-28
feat(v0.11): Diagnostics Flags UI — toggle openclaw.json flags without editing JSON


vivekchand:
## Overview
Part of **Phase 1 'The Debugger'** (v0.11). Addresses Gap #7.

## Problem
Toggling diagnostics flags (e.g. `telegram.http`, `webhook.*`) requires manually editing `openclaw.json` or setting env vars. No visual indication of which flags are active. Sometimes requires gateway restart.

## Proposed Solution
A **Settings panel** that reads `openclaw.json`, renders `diagnostics.flags` as toggle checkboxes, and writes changes back immediately.

## UI Spec
- Toggle list with search input
- ~20 common flags with human-readable descriptions pre-listed:
  - `telegram.http`, `telegram.payload`, `webhook.*`, `queue.*`, `session.*`, `model.*` etc.
- Custom flag input for unlisted flags
- Active flag count badge in nav
- Banner: 'Gateway restart required to apply changes' when relevant

## Safety
- Config file backed up before any write (`openclaw.json.bak`)
- Changes diffed and shown before saving
- Read-only mode if config not writable

## Acceptance Criteria
- [ ] Settings panel accessible from nav
- [ ] Known flags listed with descriptions
- [ ] Toggle writes to openclaw.json immediately
- [ ] Active flag count shown in nav badge
- [ ] Restart hint banner shown when needed
- [ ] Backup created before write
- [ ] Works on macOS, Linux, Windows config paths

## Priority
**P0** — Medium impact, medium effort.

_Identified in the ClawMetry Observability Roadmap v1.0 (Feb 2026)_

PR #867, opened 2026-05-05
fix(dashboard): bound transcript scans and stagger widgets - optimize data loading time in widgets

ruzeynalov:
## Summary

Fixes slow dashboard widget loads on installs with large OpenClaw runtime artifacts. The dashboard was scanning `.trajectory.jsonl` and `.checkpoint.*.jsonl` files for analytics paths where those files duplicate canonical session transcripts, causing first-load widgets to queue behind hundreds of MB of JSONL reads.

This PR keeps analytics on canonical session/reset transcripts, bounds expensive history scans, and staggers non-critical widgets so the overview becomes usable quickly instead of stampeding the backend.

## How it works

1. **Skip duplicate runtime artifacts in transcript analytics** — usage and dashboard analytics ignore `.trajectory`, `.checkpoint`, and `.deleted` JSONL files so `/api/usage` stays focused on canonical transcripts.

2. **Bound Brain history reads** — `/api/brain-history` now supports `limit`, skips artifact files by default, sorts sessions newest-first, reads head + bounded tail instead of `readlines()` on every file, and keeps a tiny short-lived cache.

3. **Cache heavy overview cards** — `/api/autonomy` and `/api/sessions/clusters` get small TTL caches, and both avoid duplicated runtime artifacts where applicable.

4. **Reduce hot-path session fanout** — overview/session/subagent gateway lookups use smaller limits on dashboard hot paths, while `/api/subagents?full=1` still allows a complete scan when needed.

5. **Avoid frontend boot stampedes** — the overview starts critical widgets immediately and staggers expensive/non-critical cards; duplicate active-task refreshes are removed.

6. **Unstick diagnostics/health panels** — diagnostics now renders in the System Health panel, `/api/config-diagnostics` aliases `/api/diagnostics`, and nested health-card fetches have bounded timeouts.

### Test plan

1. Open dashboard on an install with many .trajectory.jsonl / .checkpoint.*.jsonl files — overview widgets should populate without a long 30s+ wait
2. Brain tab still loads recent events and supports SSE with the stored auth token
3. System Health shows Diagnostics instead of staying on Loading diagnostics...
4. /api/subagents returns quickly by default; /api/subagents?full=1 can still do the complete historical scan

```bash
15 passed in 2.36s
```

### DIFF
--- a/clawmetry/static/js/app.js
+++ b/clawmetry/static/js/app.js
@@ -1659,17 +1659,18 @@ async function loadAll() {
     // Render overview quickly; do not block on heavy usage aggregation.
     var overview = await fetchJsonWithTimeout('/api/overview', 3000);
 
-    // Start secondary panels immediately.
-    startActiveTasksRefresh();
+    // Start only critical secondary panels immediately. Expensive/non-critical
+    // cards are staggered below so the initial widget load does not stampede
+    // the backend and queue behind transcript analytics.
     loadActivityStream().catch(function(e){console.warn('activity stream failed',e)});
     loadHealth().catch(function(e){console.warn('health failed',e)});
     loadMCTasks().catch(function(e){console.warn('mctasks failed',e)});
-    if (typeof loadReliabilityCard === 'function') loadReliabilityCard().catch(function(e){console.warn('reliability card failed',e)});
-    if (typeof loadAnomalyPanel === 'function') loadAnomalyPanel().catch(function(e){console.warn('anomaly panel failed',e)});
-    if (typeof loadTokenVelocity === 'function') loadTokenVelocity().catch(function(e){console.warn('velocity check failed',e)});
+    if (typeof loadReliabilityCard === 'function') setTimeout(function(){ loadReliabilityCard().catch(function(e){console.warn('reliability card failed',e)}); }, 1200);
+    if (typeof loadTokenVelocity === 'function') setTimeout(function(){ loadTokenVelocity().catch(function(e){console.warn('velocity check failed',e)}); }, 1600);
     if (typeof loadDiagnostics === 'function') loadDiagnostics().catch(function(e){console.warn('diagnostics failed',e)});
-    if (typeof loadAutonomy === 'function') loadAutonomy().catch(function(e){console.warn('autonomy failed',e)});
     if (typeof loadHeartbeat === 'function') loadHeartbeat().catch(function(e){console.warn('heartbeat panel failed',e)});
+    if (typeof loadAutonomy === 'function') setTimeout(function(){ loadAutonomy().catch(function(e){console.warn('autonomy failed',e)}); }, 2600);
+    if (typeof loadAnomalyPanel === 'function') setTimeout(function(){ loadAnomalyPanel().catch(function(e){console.warn('anomaly panel failed',e)}); }, 3600);
     document.getElementById('refresh-time').textContent = 'Updated ' + new Date().toLocaleTimeString();
 
     if (overview.infra) {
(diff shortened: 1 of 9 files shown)


vivekchand (commented):
## Test plan & review notes

Thanks for this, @ruzeynalov — a really solid performance fix with thoughtful test coverage. The artifact-filter + head/tail I/O approach looks right, and the new test file covers the core invariants well.

**What changed**
- Filters `.trajectory.jsonl` / `.checkpoint.*.jsonl` / `.deleted.jsonl` out of all analytics paths (usage, brain, autonomy, clusters, subagents); adds short-lived in-process caches to `/api/autonomy`, `/api/subagents`, `/api/brain-history`, and `/api/sessions/clusters`; staggers non-critical frontend widgets up to 3.6 s after initial load; fixes the "Loading diagnostics…" hang; and adds `/api/config-diagnostics` as an alias.

**Smoke commands**
```bash
make test
# Or to run just the new perf tests:
python -m pytest tests/test_dashboard_perf_hotfix.py -v

# Then start the server and load the dashboard:
```

**What to look at visually**
- `http://localhost:8900` — the main overview should reach a usable state in under 3 s on a workspace with many `.trajectory.jsonl` files; Reliability, Autonomy, and Anomaly cards should appear a few seconds later rather than all at once
- `http://localhost:8900` → **System Health** tab — the Diagnostics sub-panel should show gateway URL, workspace path, and auth-token status instead of "Loading diagnostics…"
- `http://localhost:8900/api/subagents` vs `/api/subagents?full=1` — the default response should return quickly; the `?full=1` variant should do the full historical scan

1. **`tail_bytes` seek in text mode** — `_scan_spawn_events_from_jsonl` opens files with `open(fpath, "r")` and then calls `fh.seek(size - tail_bytes)` where `size` comes from `os.path.getsize`. On Windows, `getsize` returns the on-disk byte count but text-mode `seek` on a CRLF file treats the position differently, which can cause the seek to land mid-character. The `fh.readline()` guard helps but may not fully protect against this. `routes/brain.py` already uses `open(path, "rb")` for the same pattern — it would be worth aligning `sessions.py` to the same binary-open approach.

2. **`sessions_list` limit reduced from 50→20 (and 100→20 in subagents)** — on an install with more than 20 concurrent or recent sessions, the sessions list, subagents panel, and overview may silently truncate. Worth confirming this limit is still safe for the typical multi-agent use-case, or exposing it as a tunable (like `CLAWMETRY_SUBAGENTS_SCAN_MAX_FILES` already does for the JSONL scan).

3. **`startActiveTasksRefresh()` removed from both `loadAll()` and `bootDashboard()`** — if `loadOverviewTasks` is the only remaining caller and it doesn't set up its own polling interval, the active-tasks panel might stop refreshing after the first paint. Worth confirming `startOverviewTasksRefresh` (which is still called in `bootDashboard`) covers the same cadence.

4. **Module-level cache dicts are process-global** — the caches in `brain.py`, `sessions.py`, `autonomy.py`, and `usage.py` are plain dicts. Under multi-worker Waitress/gunicorn deployments each worker has its own copy, so cache hits will be inconsistent across workers. Not a correctness bug, but worth a note in a comment so future readers don't assume cross-worker coherence.

**Merge state**
- GitHub reported `mergeable_state: unknown` (still being computed at review time). If it comes back as `BEHIND`, you'll need to rebase your fork branch (`fix/runtime-artifact-widget-performance`) onto the latest `vivekchand/clawmetry:main` — since this is a fork PR, that rebase needs to happen on your side: `git fetch upstream && git rebase upstream/main`.

**Issue link**
- No linked issue found in the PR body. A code comment in `routes/health.py` references `GH#28` for the diagnostics panel — if that's a real issue it would be worth adding `Closes #28` (or the correct number) to the PR description so it auto-closes on merge. If there isn't a tracking issue yet, one would help future bisects.

---

PR #928, opened 2026-05-08
[RELEASE] fix(install): retry /api/register on Cloud Run cold start

vivekchand:
## Summary

Fresh-install onboarding silently falls into "local mode" (no cloud sync) when the registration POST to \`ingest.clawmetry.com/api/register\` hits a Cloud Run cold start that exceeds the previous **15s** timeout.

User just hit this on a fresh Linux box:
\`\`\`
Setting up your cloud dashboard...
Could not reach cloud: The read operation timed out
Installed (local mode)
\`\`\`

Same class of bug we already fixed for the daemon's heartbeat + intent-start handshake in PR #135 — \`/api/register\` was missed.

## Fix

Match the cold-start retry pattern from PR #135:
- 3 attempts (was 1)
- 30s per attempt (was 15)
- Exponential backoff on timeout / 502/503/504 / connection reset

Worst case: ~96s before giving up. Expected case: warm hit in <1s.

## Test plan
- [ ] Fresh install on Linux box — confirm registration succeeds without local-mode fallback
- [ ] Force cold start (\`gcloud run services update --no-traffic\` then revert) — verify retries kick in


vivekchand (commented):
## Test plan & review notes

**What changed**
- `clawmetry/cli.py` `_instant_register()`: single 15 s `urlopen` → 3-attempt retry loop (30 s per attempt, exponential 2s/4s backoff on 5xx and generic exceptions); `tests/e2e/cloud-contract.mjs`: same cold-start retry pattern (4 attempts, 1.5s×attempt backoff on 401/5xx) applied to `heartbeat()` and `intentStart()`.

**Smoke commands**
- `make test`
- Fresh install on a machine that hasn't talked to `ingest.clawmetry.com` today → confirm "Setting up your cloud dashboard…" succeeds rather than dropping to "Installed (local mode)"
- Simulate a timeout: `python3 -c "import urllib.request; urllib.request.urlopen('http://localhost:1', timeout=1)"` (to confirm the retry path surfaces the right error message after all attempts)

**Things to look at in the diff**
- `req` is constructed once before the retry loop — `urllib.request.Request` is reusable, so this is fine, but worth a sanity check that the `Content-Type` header isn't consumed on first use
- 401 retry in `cloud-contract.mjs` `heartbeat()`: retrying on 401 is non-standard. The comment explains it ("token validate cache might be empty on this instance"), but it means a genuine auth bug would be swallowed for 3×1.5s before surfacing. Acceptable tradeoff for a test harness, just worth documenting.
- Worst-case install time is now ~96 s on a fully-unreachable endpoint (3 × 30 s + 2 + 4). Previous was 15 s. Users on firewalled machines will notice a longer delay before falling back to local mode — consider whether the final error message should tell them how to skip cloud with `CLAWMETRY_SKIP_CLOUD=1` or similar.

- `for/else` pattern is correct (Python `else` fires when loop doesn't `break`), but if all 3 iterations hit the generic `except` path, `_last_err` will be set and the `else` block prints it. Confirm that path is reachable in a test.
- The `_t_reg` alias for `time` (`import time as _t_reg`) inside the function avoids a name collision with a potential outer `time` import — but `time` is already imported at module level; the alias is just defensive. Low risk.

**Issue link**
- No `Closes #N` in the PR body. If there's a tracking issue for the "Installed (local mode)" cold-start regression, linking it would help with changelog attribution.

---

vivekchand:
Closing — would duplicate the [RELEASE] cluster shipped via #1060 (merged 2026-05-12). The retry-on-cold-start /api/register fix is captured in current main; if a separate retry policy improvement is still wanted, re-open against current main.

PR #936, opened 2026-05-09
feat(sync): track initial sync progress + /api/sync-progress endpoint (#748 part 1)

### LINKED ISSUE #748: feat: sync progress banner on cloud dashboard after fresh install
## Problem

After `curl | bash` install, the cloud dashboard immediately shows the node detail but all tabs are empty:
- Brain: "No brain activity events found"
- Memory: "No memory data synced"
- Tokens: 0/0
- Overview: mostly empty

The sync daemon IS running and syncing data (visible in `clawmetry status`), but the dashboard has no indication. **Looks broken when it's just waiting for initial sync.**

### 1. Sync progress endpoint (OSS → Cloud)
The sync daemon should report progress to cloud during initial sync:
```
POST /ingest/sync-progress
{
  "node_id": "agent+Macbook-Pro-local",
  "phase": "sessions",    // sessions | events | memory | crons | complete
  "done": 6,
  "total": 11,
  "started_at": "2026-04-19T02:18:15Z"
}
```

### 2. Dashboard banner (Cloud)
When a node's last sync started < 2 minutes ago and phase != "complete":
```
🔄 Initial sync in progress — sessions synced, events syncing... (~30s remaining)
```
Auto-dismisses when sync completes. Shows a progress bar.

### 3. Empty state improvement
Replace "No brain activity events found" with:
```
⏳ Waiting for initial sync...
Your agent data is being encrypted and synced to the cloud.
This usually takes 30-60 seconds.
```

## Notes
- Needs coordination with agent team for sync state tracking
- The sync daemon already logs progress: `Recent sessions: 6 events synced`, `Session metadata: 11 rows synced`, `Crons: 1 synced`
- Should work for both fresh installs and reconnects

/cc @vivekchand

vivekchand:
Closes #748 (partially — see Scope).

## What
- Add `_record_sync_progress(phase, done, total, status)` helper in `clawmetry/sync.py` that writes `~/.clawmetry/sync_progress.json` atomically (tmp + `os.replace`).
- Hook it at the start and end of `sync_sessions`, `sync_sessions_recent`, `sync_session_metadata`, `sync_memory`, and `sync_crons` so the daemon emits live progress during the initial sync cycle.
- Mark `phase="complete"`, `status="complete"` once the daemon's startup recent-sync block finishes.
- Expose `GET /api/sync-progress` in `dashboard.py` (next to the existing OSS shims) that returns the JSON file — 404 before the first write, 200 with body otherwise. Local-only, no auth.

## How
The progress file uses the same `CONFIG_DIR = ~/.clawmetry` convention already used by `STATE_FILE` and `CONFIG_FILE`. Atomic write protects the dashboard reader from torn JSON. Hooks are positioned after the existing `_sync_allowed()` early-return so we don't lie about progress on paused (expired-trial) accounts.

The new endpoint follows the direct-`@app.route` style of the cloud shims registered inside `detect_config()` rather than introducing a one-route blueprint just for this — keeps the diff to two files.

## Scope
This is **Part 1 — OSS-side state tracking only**. Follow-ups:
- **Part 2:** cloud-side ingest endpoint to receive these progress events from the daemon (so the cloud dashboard can show the banner too, not just the local OSS dashboard).
- **Part 3:** the actual sync-progress banner UI on the dashboard, consuming `/api/sync-progress`.

Net diff: 69 lines across 2 files.

## Test plan
- [ ] On a fresh install, run `clawmetry connect` and confirm `~/.clawmetry/sync_progress.json` appears once the daemon starts.
- [ ] `curl localhost:<dashboard-port>/api/sync-progress` returns the expected JSON shape.
- [ ] Before any sync runs, the same endpoint returns 404 with `{"error": "no sync progress yet"}`.
- [ ] Existing daemon behaviour is unchanged on a system that already has `initial_backfill_done=true` — progress just reports each cycle's results.

Co-Authored-By: Dhriti <noreply@clawmetry.com>

### DIFF
--- a/clawmetry/sync.py
+++ b/clawmetry/sync.py
@@ -237,6 +237,45 @@ def save_state(state: dict) -> None:
     STATE_FILE.write_text(json.dumps(state, indent=2))
 
 
+# ── Initial-sync progress (vivekchand/clawmetry#748) ─────────────────────────
+# Tracks per-phase progress to ~/.clawmetry/sync_progress.json so the local
+# dashboard can show a "syncing…" banner on fresh installs instead of empty
+# tabs. Written atomically (tmp + rename) because the dashboard reads this
+# file on every banner poll.
+SYNC_PROGRESS_FILE = CONFIG_DIR / "sync_progress.json"
+_sync_progress_started_at: str | None = None
+
+
+def _record_sync_progress(
+    phase: str, done: int, total: int = 0, status: str = "running"
+) -> None:
+    global _sync_progress_started_at
+    try:
+        CONFIG_DIR.mkdir(parents=True, exist_ok=True)
+        now = datetime.now(timezone.utc).isoformat()
+        if _sync_progress_started_at is None:
+            _sync_progress_started_at = now
+        try:
+            cfg = load_config()
+            node_id = cfg.get("node_id", "")
+        except Exception:
+            node_id = ""
+        payload = {
+            "node_id": node_id,
+            "phase": phase,
+            "done": int(done),
+            "total": int(total),
+            "status": status,
+            "started_at": _sync_progress_started_at,
+            "updated_at": now,
+        }
+        tmp = SYNC_PROGRESS_FILE.with_suffix(".json.tmp")
+        tmp.write_text(json.dumps(payload, indent=2))
+        os.replace(tmp, SYNC_PROGRESS_FILE)
+    except Exception as e:
+        log.debug(f"Could not record sync progress ({phase}): {e}")
+
+
 # ── HTTP ──────────────────────────────────────────────────────────────────────
 
 
@@ -937,6 +976,7 @@ def sync_sessions(config: dict, state: dict, paths: dict) -> int:
     # exactly where we paused -- no event loss, no double-send.
     if not _sync_allowed():
         return 0
+    _record_sync_progress("sessions", 0)
     sessions_dir = paths["sessions_dir"]
     api_key = config["api_key"]
     enc_key = config.get("encryption_key")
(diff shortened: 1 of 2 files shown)


vivekchand (commented):
## Test plan & review notes

**What changed**
- `clawmetry/sync.py` gains `_record_sync_progress()` which atomically writes `~/.clawmetry/sync_progress.json` at key sync lifecycle points (start/end of sessions, metadata, memory, crons phases).
- New `GET /api/sync-progress` endpoint (added directly in `detect_config()`) returns the JSON file; 404 before the first write.

**Smoke commands**
- Run `clawmetry connect` (or `clawmetry sync`), then watch `~/.clawmetry/sync_progress.json` appear and update during the sync cycle
- `curl -sS http://localhost:8900/api/sync-progress` while syncing → `{phase, done, total, status, …}`
- Before any sync has run: same curl → 404 + `{"error": "no sync progress yet"}`
- Interrupt the daemon mid-sync and re-read the file — should be valid JSON (no torn write, thanks to the `os.replace` atomic swap)

- The endpoint is registered via `@app.route` inside `detect_config()` — if `detect_config` is called more than once in the same process, Flask will raise on the duplicate route name. Worth confirming this path is guarded (e.g. `already_registered` check or a one-time sentinel).
- Confirm `_record_sync_progress` is called *after* `_sync_allowed()` returns `True`, so expired-trial accounts don't see misleading "syncing…" progress. The PR description says this is handled — a quick grep for the call sites would confirm.

**Issue link**
- Partially closes #748 (Part 1 of 3). Consider `Refs #748` in the description rather than `Closes` to avoid prematurely closing the tracking issue.

---

PR #937, opened 2026-05-09
feat(cli): defer sync daemon until user runs `clawmetry sync`

### LINKED ISSUE #905: Defer initial clawmetry sync until user opens dashboard
## What

Defer the initial `clawmetry sync` until the user actually opens their dashboard. Today the sync daemon starts on `clawmetry connect` and runs forever, publishing events the user might never look at. Result: wasted compute on the user's box, wasted bandwidth, wasted cloud storage.

## Proposed UX

**On the user's machine** (CLI side):
- `clawmetry connect` writes config + dashboard URL but **does not start the daemon** (or only starts it briefly to register, then exits)
- A new flag `--start-sync-now` preserves today's behavior for users who want it
- `clawmetry sync` works as-is when manually invoked

**In the cloud dashboard**:
- When a user with no events ever opens `app.clawmetry.com/cloud`, show an empty-state overlay:
  > "Hang tight — your first sync is starting…"
  > [spinner + a copyable command: `clawmetry sync`]
- When events start arriving (cloud sees the first publish from this api_key), the overlay auto-dismisses and the dashboard renders
- For repeat visits to a node that's gone quiet, optionally show a "Last synced X ago — start sync?" tooltip with the same command

## Why

1. **Cost** — most users `clawmetry connect` once and never look. Their daemon publishes thousands of events/day for a dashboard view that never happens. We pay storage on every one.
2. **Battery / CPU** — sync daemon polls + uploads on the user's machine 24/7. For laptop-installed users, this is real.
3. **Trust** — "no data is sent to ClawMetry until you actually want to see it" is a stronger product story than "sync starts automatically."
4. **Compliance / privacy** — for users in regulated environments, on-demand sync is easier to justify than always-on.

## Trade-offs

- **First-view experience is slower** — user opens dashboard, has to wait ~15-60s for first batch to arrive. Mitigated by clear "Syncing your data…" overlay (better than blank page) and the catch-up batch reading persisted local session files (so they get history immediately, not just events from the moment they clicked).
- **Manual command in the user's box** — for OSS users, "Run `clawmetry sync`" is a copy-paste step. Annoying but discoverable. For partner integrations (KiloClaw), the partner's UI can spawn the daemon on the user's behalf — see [Kilo-Org/cloud#3078](https://github.com/Kilo-Org/cloud/pull/3078) for the pattern.
- **Cron / scheduled jobs** — users who genuinely want background sync can opt back in with `clawmetry connect --start-sync-now` or a systemd timer. Document the path.

## Implementation sketch

- `clawmetry/cli.py` — `connect` no longer spawns daemon by default; new `--start-sync-now` flag preserves old behavior
- `routes/cloud.py` (cloud dashboard) — new check on dashboard load: if `events` table has zero rows for this `api_key` AND the account is < 24h old, show the "Syncing your data…" overlay with copy-pasteable command
- `dashboard.py` (OSS dashboard) — same overlay logic, scoped to single-node view
- Existing `/api/ingest/heartbeat` and `/api/ingest/*` endpoints unchanged — they accept events whenever the daemon does start
- Document the trade-off in `README.md` and `docs/`

## How KiloClaw integration relates

The KiloClaw integration in [Kilo-Org/cloud#3078](https://github.com/Kilo-Org/cloud/pull/3078) is already adopting this pattern at the partner level: bootstrap pre-wires ClawMetry but doesn't start sync; KiloClaw's UI starts the daemon when the user clicks "View Observability." Worth coordinating so the OSS-side version of this UX is consistent.

vivekchand:
Closes #905

## What

`clawmetry connect` no longer auto-starts the background sync daemon. Users opt in either by passing `--start-sync-now` to `connect` or by running the new `clawmetry sync` subcommand whenever they're ready.

This honors the proposal in #905: avoid burning user CPU, bandwidth, and cloud storage on nodes that are connected but whose dashboard is never opened.

## How

- `clawmetry/cli.py — _cmd_connect`: replace the unconditional `_start_daemon(...)` call with a branch keyed on the new `--start-sync-now` flag (default off). When deferred, print a one-line hint pointing at `clawmetry sync` and at the opt-in flag.
- `clawmetry/cli.py — argparse`: add `--start-sync-now` to the `connect` subparser. `--key-only` and `--no-daemon` keep their existing semantics for back-compat.
- `clawmetry/cli.py — _cmd_sync`: new subcommand. Loads the saved config and starts the daemon (`--foreground` honored). Exits with a clear error if no config / api_key is present.
- Wire `sync` into `_subcmds` and the `args.cmd` dispatcher.

Existing flags continue to work: `--foreground`, `--no-daemon`, `--key-only`. The setup wizard path now also defers — users who prefer the historical behavior can run `clawmetry connect --start-sync-now`.

## Verification

- `clawmetry connect --help` shows the new `--start-sync-now` flag.
- `clawmetry sync --help` is recognized as a subcommand.
- `python -m pytest tests/ -k "cli or connect"` → 4 passed (the unrelated pre-existing `test_health_ok` failure on `main` is not introduced by this PR).

## Out of scope (worth a follow-up)

The cloud-side empty-state overlay (`"Hang tight — your first sync is starting…"`) lives in `clawmetry-cloud` and is intentionally not in this PR. CLI-side change first; cloud overlay next.


vivekchand (commented):
## Test plan & review notes

**What changed**
- `clawmetry connect` no longer auto-starts the sync daemon; prints a nudge to run `clawmetry sync` instead. A `--start-sync-now` flag restores the old behavior.
- New `clawmetry sync` subcommand reads the existing config and calls `_start_daemon()`. Supports `--foreground`.

**Smoke commands**
- `clawmetry connect` → should print the sync-paused message and exit without spawning a daemon (`ps aux | grep clawmetry` shows nothing new)
- `clawmetry sync` → should start the daemon; `clawmetry status` should reflect it running
- `clawmetry connect --start-sync-now` → should behave like the old default (daemon starts immediately)
- `clawmetry sync` before `clawmetry connect` → should exit 1 with `"No clawmetry config found. Run clawmetry connect first."`
- `clawmetry sync --foreground` → daemon runs in foreground, Ctrl-C stops it cleanly

- `_cmd_sync` calls `_start_daemon(config, args)` where `args` comes from `p_sync` — confirm `_start_daemon` handles `args` objects that don't carry the connect-specific flags (e.g. `--no-daemon`, `--start-sync-now`). The `getattr(..., False)` guards seen elsewhere in `cli.py` should cover this.
- `from clawmetry.sync import CONFIG_FILE` inside `_cmd_sync` — if `sync.py` has module-level side effects on import, this could be surprising in edge cases.

**Issue link**
- Branch is named `fix/gh-clawmetry-905-defer-sync` — if #905 is the tracking issue, adding `Closes #905` to the PR description would auto-close on merge.

---

PR #1109, opened 2026-05-13
fix(sync): translate brain pending_query result to events shape (real MOAT fix)

vivekchand:
## The bug
`_dispatch_pending_queries` writes raw `{rows: [...]}` to the Brain cache key, overwriting `_build_brain_cache_pushes`'s correct `{events: [...]}` blob. Browser's `_cm_decryptBrain` reads `dec.events || []` = empty → cloud Brain shows "No brain activity events found" even with 100+ events in DuckDB.

## Live repro (2026-05-13)
User sent Telegram messages to Diya bot via OpenClaw. DuckDB had the events. Cloud Brain tab was empty. Confirmed by manually decrypting the cached blob — shape was `{rows: [], count: 123}` (wrong key + empty rows).

## Fix
Extract `_rows_to_brain_events()` helper. Both the proactive cache_push and the heartbeat-piggyback paths now use it for brain cache keys. After hotpatching the daemon and waiting one heartbeat, blob shape is now `{events: [50], count: 50, _source: 'local_store', _shape: 'brain_history'}` ✓.

## Tests
Local hotpatch of running daemon. Verified cloud blob decrypts to the right shape. Deeper unit test would assert `_dispatch_pending_queries` produces blobs whose shape exactly matches what `_cm_decryptBrain` consumes — leaving as TODO since the bug was caught by integration not unit (the kind of bug `_dispatch_pending_queries` is prone to).

This is the **real MOAT** — without it, the Brain tab on app.clawmetry.com is empty for any user whose daemon serves a pending_query before the next cache_push.

### DIFF
--- a/clawmetry/sync.py
+++ b/clawmetry/sync.py
@@ -2072,6 +2072,45 @@ def _local_dispatch_fallback(shape: str, args: dict) -> dict:
     return {"rows": rows, "count": len(rows), "_shape": shape, "_via": "fallback"}
 
 
+def _rows_to_brain_events(rows: list) -> list:
+    """Translate raw DuckDB event rows into the dashboard-display shape the
+    cloud Brain tab's ``_cm_decryptBrain`` expects (``{time, type, detail,
+    src, sessionId, agentId, tokens, cost, model}``).
+
+    Single source of truth used by both the proactive cache_push path
+    (`_build_brain_cache_pushes`) and the heartbeat-piggyback path
+    (`_dispatch_pending_queries`) so the browser sees the SAME shape no
+    matter which writer last touched the cache key.
+
+    Bug 2026-05-13: the piggyback path was writing the raw `{rows: [...]}`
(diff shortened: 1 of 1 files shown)


vivekchand:
Iterated: the first fix (translate to display shape on the daemon) was the wrong direction. Cloud Brain decrypts the blob then runs `transformEvents(rawEvs)` over each item — which expects the ORIGINAL JSONL shape `{type, message:{role,content[]}, timestamp}` so it can unwrap message.content[] for the detail text. Pre-flattening to `{detail:''}` made transformEvents silently drop every event via its empty-detail fallback.

New approach (this commit): ship row['data'] as-is. Live-verified that the cached blob now contains `event[0].message.content[0].text == '[[reply_to_current]] Life is...'` (the actual Diya bot reply from user's Telegram exchange). User can refresh https://app.clawmetry.com/cloud Brain tab and should see the conversation render.

PR #779, opened 2026-04-22
feat: Per-message cost attribution with cache-hit breakdown

vivekchand:
Closes vivekchand/clawmetry-cloud#318

## What
Adds granular cache token tracking to the usage analytics:
- Extended _extract_usage_metrics() to return 4-token-type + 4-cost-type breakdown
- Updated _compute_transcript_analytics() to aggregate daily cache tokens
- Enhanced /api/usage response with input/output/cache-read/cache-write token counts
- New endpoint /api/token-attribution for per-message granular cost breakdown

## How
- Reads message.usage.cost {input, output, cacheRead, cacheWrite, total}
- Tracks daily aggregate of each token type separately
- Returns structured breakdown in API responses
- Calculates cache hit ratio for visibility into cache efficiency

## API Changes
- /api/usage: Each day object now includes inputTokens, outputTokens, cacheReadTokens, cacheWriteTokens
- /api/token-attribution: New endpoint returning per-message token/cost breakdown with cache stats

PR #1128, opened 2026-05-13
[RELEASE] fix: sync.py dead-code after __main__ block (alerts + autonomy never ran)

vivekchand:
## Bug

`clawmetry/sync.py` had module-level definitions placed AFTER the `if __name__ == "__main__":` block:

- `ALERTS_EVAL_INTERVAL_SEC = 60`
- `_ALERTS_EVENT_LOOKBACK_SEC`, `_ALERTS_EVENT_LIMIT`
- `def evaluate_alerts(...)` (PRD #779 evaluator)
- `def _compute_autonomy_daily_series(...)`
- `def sync_autonomy(...)`

Because `run_daemon()` is an infinite loop, the `if __name__` block never returns when the daemon is run as `python -m clawmetry.sync`. Every definition below it was therefore **dead code** in the daemon process.

## Impact — every user since 0.12.179

Confirmed live on user's daemon (PID 31692, fresh restart): **240+ NameError occurrences** in `~/.clawmetry/sync.log` over the last hour.

```
ERROR Sync cycle error: name 'ALERTS_EVAL_INTERVAL_SEC' is not defined
```

- PRD #779 alerts evaluator never fired for any user
- `sync_autonomy()` never ran -> autonomy daily series never computed
- Sync cycle errored every ~15s on every install since 0.12.179

## Fix

Move the entire `if __name__ == "__main__":` block (14 lines) from line 5039 to EOF (now line 5370). Nothing else changed — module-level constants and functions now actually evaluate before the daemon entrypoint runs.

## Test plan

New regression test `tests/test_sync_main_block_at_eof.py`:

- [x] `test_main_block_is_last_top_level_statement` — source-level guard: any non-blank, non-comment, non-indented line after `if __name__ == "__main__":` fails the test. Prevents this exact bug from regressing.
- [x] `test_alerts_and_autonomy_symbols_importable` — asserts `ALERTS_EVAL_INTERVAL_SEC`, `evaluate_alerts`, `sync_autonomy` resolve after `import clawmetry.sync`.
- [x] `test_subprocess_can_resolve_alerts_constant` — fresh subprocess imports the module and reads the constant; asserts no `NameError` on stderr.
- [x] AST parse: `python3 -c "import ast; ast.parse(open('clawmetry/sync.py').read())"` -> OK
- [x] `pytest tests/test_sync_main_block_at_eof.py -q` -> **4 passed**

## Expected log delta after release

Before:
```
ERROR Sync cycle error: name 'ALERTS_EVAL_INTERVAL_SEC' is not defined
ERROR Sync cycle error: name 'ALERTS_EVAL_INTERVAL_SEC' is not defined
ERROR Sync cycle error: name 'ALERTS_EVAL_INTERVAL_SEC' is not defined
... (every 15s, forever)
```

After:
```
INFO  alerts: evaluated N rules, dispatched M
INFO  autonomy: computed daily series for 90 days
```

### DIFF
--- a/clawmetry/sync.py
+++ b/clawmetry/sync.py
@@ -5036,24 +5036,6 @@ def _build_gateway_data(paths: dict = None) -> dict:
         }
 
 
-if __name__ == "__main__":
-    while True:
-        try:
-            run_daemon()
-            break  # clean exit
-        except KeyboardInterrupt:
-            break
-        except Exception as e:
-            import traceback
-
-            log.error(f"Daemon crashed: {e}")
-            log.error(traceback.format_exc())
-            log.info("Restarting in 15 seconds...")
-            time.sleep(15)
-
-
-
-
 def _compute_autonomy_daily_series(sessions_dir, days=90):
     """
     Compute per-day autonomy aggregates from local session transcripts.
@@ -5383,3 +5365,19 @@ def evaluate_alerts(config: dict, state: dict) -> int:
 
     state["alerts_last_eval_ts"] = _iso_now()
     return dispatched
+
+
+if __name__ == "__main__":
+    while True:
+        try:
+            run_daemon()
+            break  # clean exit
+        except KeyboardInterrupt:
+            break
+        except Exception as e:
+            import traceback
+
+            log.error(f"Daemon crashed: {e}")
+            log.error(traceback.format_exc())
+            log.info("Restarting in 15 seconds...")
+            time.sleep(15)
--- a/tests/test_sync_main_block_at_eof.py
+++ b/tests/test_sync_main_block_at_eof.py
@@ -1,0 +1,113 @@
+"""Regression test for the dead-code-after-__main__ bug in clawmetry/sync.py.
+
+Background
+----------
+Between 0.12.179 and 0.12.x (this fix), `clawmetry/sync.py` had module-level
+constants and functions defined AFTER the `if __name__ == "__main__":` block.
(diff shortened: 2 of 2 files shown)


vivekchand:
## 🍎 Steve Jobs UX review

🟡 **P1 — backend infra fix, no user-visible surface this PR alone**, but the consequence WAS user-visible: every install since 0.12.179 saw `Sync cycle error` 4×/min in their daemon log and silently lost the alerts evaluator. The user never knew because the daemon log isn't surfaced anywhere in the dashboard. **Action**: file a follow-up to expose `daemon error rate` as a card in the dashboard's System Health tab — this exact failure mode was invisible to users for hours. (See PRD #1133 layer 4, which captures this.)

🟢 **nit** — the structural regression test `test_sync_main_block_at_eof.py` that asserts no top-level statement comes after `if __name__ == "__main__":` is the right shape of defensiveness — you're protecting the class of bug, not the instance. Apple-grade. Stop.

vivekchand:

🟡 **P1 — simplest possible fix for a self-inflicted wound.** The bug existed only because module-level statements were added below an infinite-loop entry point — should have been caught by the test that now ships in this PR. Future PRs that add anything below `if __name__ == "__main__":` will fail CI. Right answer.

🟢 **OSS leakage / tier**: zero. Daemon health is foundational infra; not a paid surface. **Backend infra**, no fence-sitter conversion impact. Stop.

Issue #1129, opened 2026-05-13
[DuckDB MOAT] 4 read-path bugs surfaced in live audit 2026-05-13: phantom session_id, brain enrichment empty, transcript bypass, sessions message_count=0


vivekchand:
## Summary
Live audit on a fresh clawmetry install with active daemon + dashboard (clawmetry==0.12.179, default-on local-store) reveals 4 distinct DuckDB read-path bugs. All endpoints correctly tag `_source: local_store` (i.e., the read-path is firing) but the data they return is wrong in specific ways.

## Repro environment
- OS: macOS 25.3.0, Python 3.9
- clawmetry==0.12.179 installed via pip (latest PyPI)
- Daemon running via launchd (`com.clawmetry.sync` PID 31692)
- DuckDB at `~/.clawmetry/clawmetry.duckdb` — 7 events, 1 actual session (created 2026-05-09)

## Bug 1 — Phantom `.trajectory` session_id in DuckDB
\`/api/sessions\` returns 3 sessions for what's actually 1 OpenClaw session:
- \`4595d095-5b7e-40f7-908c-9a5a9f92f63a\` (real, started 2026-05-09T10:07:28Z)
- \`4595d095-5b7e-40f7-908c-9a5a9f92f63a.trajectory\` (PHANTOM — same id with suffix appended, no started_at, 0 tokens)
- \`3fc28a8b-b0f5-47af-b234-fa2f96db8112\` (separate real session, 324 tokens, $0.0081)

The phantom is created somewhere in the sync.py write-path that conflates trace_id with session_id (events for the real session carry \`traceId == sessionId\` and \`traceSchema: openclaw-trajectory\` — likely the source of the suffix).

## Bug 2 — Brain history strips model/tokens/cost on every row
\`/api/brain-history?limit=20\` returns 7 events from \`_source: local_store\` but every row has \`model: \"\"\`, \`tokens: 0\`, \`cost: 0\`, \`detail: \"\"\`. Inspecting the underlying \`data\` JSON column via the local_query proxy shows the original events have \`modelId: \"claude-opus-4-7\"\`, \`provider: \"anthropic\"\`, etc. populated. The brain fast-path query is not extracting those fields out of the JSON column.

## Bug 3 — \`/api/transcript/<sid>\` bypasses DuckDB
Two failure modes:
[...]

PR #1130, opened 2026-05-13
fix(sync): skip trajectory/checkpoint/deleted .jsonl sidecars in DuckDB ingest (#1129 bug 1)

vivekchand:
## Summary
Fixes bug 1 of 4 tracked in #1129.

The cloud sync daemon's `_list_session_jsonls()` (in `clawmetry/sync.py`) walked every `*.jsonl` file in the OpenClaw sessions directory, including the trace-artifact sidecars OpenClaw writes next to each real session: `<sid>.trajectory.jsonl`, `<sid>.checkpoint.jsonl`, `<sid>.deleted.jsonl`. Downstream, `_canonical_session_file()` splits the basename at the first `.jsonl`, so a sidecar like `abc.trajectory.jsonl` got canonicalized as `abc.trajectory.jsonl` and ingested as session_id `abc.trajectory` — a phantom row that polluted DuckDB and surfaced in `/api/sessions` and `/api/brain-history`.

The dashboard read-path already excludes these (`dashboard.py:7744`, `routes/sessions.py:239`, `routes/brain.py:150`, `routes/usage.py:1283`). This PR brings the ingest path in line by adding the same exclusion at the source.

## Root cause
- `clawmetry/sync.py::_list_session_jsonls` filtered only on `endswith('.jsonl')` / `.jsonl.reset.` — no sidecar exclusion.
- Trajectory files are typically much larger than the real session jsonl, so they bloated DuckDB and slowed cloud sync.

## Fix
Two-line guard inside the existing `if`:
```python
if ".trajectory." in fname or ".checkpoint." in fname or ".deleted." in fname:
    continue
```

## Test plan
- [x] New regression test `tests/test_sync_skip_sidecar_jsonls.py` builds a temp dir with 5 files (real + 3 sidecars + reset archive), asserts only the real session and reset archive come back, and pins `_canonical_session_file()` behavior to document the upstream-filter contract.
- [x] `python3 -m pytest tests/test_sync_skip_sidecar_jsonls.py -q` → 2 passed.
- [x] `python3 -c "import ast; ast.parse(open('clawmetry/sync.py').read())"` → OK.
- [ ] CI green.

Refs: #1129

### DIFF
--- a/clawmetry/sync.py
+++ b/clawmetry/sync.py
@@ -993,6 +993,20 @@ def _list_session_jsonls(sessions_dir) -> list[str]:
     try:
         for fname in os.listdir(sessions_dir):
             if fname.endswith(".jsonl") or ".jsonl.reset." in fname:
+                # Skip OpenClaw trace-artifact sidecars that live next to
+                # real session files. Without this filter the daemon
+                # ingests <sid>.trajectory.jsonl etc., and
+                # _canonical_session_file() splits at the first `.jsonl`
+                # producing phantom session_ids like '<uuid>.trajectory'
+                # that pollute DuckDB and downstream APIs. The dashboard
+                # read-path applies the same exclusion (dashboard.py,
+                # routes/sessions.py, routes/brain.py, routes/usage.py).
+                if (
+                    ".trajectory." in fname
+                    or ".checkpoint." in fname
+                    or ".deleted." in fname
+                ):
+                    continue
                 out.append(os.path.join(sessions_dir, fname))
     except OSError:
         pass
--- a/tests/test_sync_skip_sidecar_jsonls.py
+++ b/tests/test_sync_skip_sidecar_jsonls.py
@@ -1,0 +1,79 @@
+"""Regression test for #1129 (bug 1).
+
+The cloud sync daemon must NOT ingest OpenClaw's trace-artifact sidecar
+files (`*.trajectory.jsonl`, `*.checkpoint.jsonl`, `*.deleted.jsonl`)
+into DuckDB. They live next to the real session jsonl, are typically
+much larger, and downstream `_canonical_session_file()` would split
+them at the first `.jsonl`, producing phantom session_ids like
+`<uuid>.trajectory` that pollute `/api/sessions` and
+`/api/brain-history`.
+
+The dashboard read-path already excludes these (dashboard.py,
+routes/sessions.py, routes/brain.py, routes/usage.py). This test pins
+the daemon ingest behavior to the same exclusion.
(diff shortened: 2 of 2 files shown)


vivekchand:
**⚠️ Merge-order warning — do NOT merge in isolation.**

Investigation in #1135 revealed that this PR's filter would tell the daemon to STOP ingesting `.trajectory.jsonl` files — but those are currently the ONLY files the daemon successfully parses. The canonical bare `<sid>.jsonl` files use a completely different underscore_separated schema (`type: session`, `type: model_change`, `type: message`) that the daemon's parser doesn't understand.

If this PR merges before #1135 is fixed, fresh installs will have **empty DuckDB** instead of the wrong-shape events they have today. That's a regression in user-visible MOAT correctness.

**Recommended order**: Hold this PR. Ship #1135 (underscore-schema parser) first, then #1130. Or merge them in the same release.

vivekchand:
## 🍎 Steve Jobs UX review

🔴 **P0 — merging this in isolation creates a fresh-install UX regression** (per my warning comment + #1135). With this filter alone, the daemon stops ingesting trajectory sidecars BUT can't parse the canonical bare `<sid>.jsonl` files (which use a different schema). Net: brand-new installs will show **empty Brain tab + empty transcripts** — currently they'd at least show wrong-but-present data. **Action**: do NOT publish 0.12.180 until the underscore-schema parser (PR landing from #1135) merges in the same release. Hold #1128 to bundle.

🟢 **nit** — the dashboard read path already excluded these sidecars (3 separate places: `dashboard.py:7744`, `routes/sessions.py:239`, `routes/brain.py:150`). The right structural fix would be one shared `_is_session_artifact(name)` helper used by both daemon ingest AND dashboard reads — currently the rule is duplicated 4 times. Follow-up.

vivekchand:

🔴 **P0 — this is the wrong fix in isolation.** First-principles question: what is the daemon supposed to ingest? Answer: the canonical `<sid>.jsonl` files. Today it ingests neither (it parses trajectory sidecars + writes phantom session_ids). This PR removes ingestion of the wrong thing without adding ingestion of the right thing. Net delta: closer to truth on Brain rows that DO exist (no phantom IDs), zero delta on the bigger bug (canonical files still unparsed). **Action**: bundle with #1135's parser before any release.

🟢 **OSS leakage / tier**: zero direct impact. Local DuckDB integrity is the MOAT, not a paid feature. But: the MOAT only sells Cloud-Pro if it actually works for real users. Today it doesn't — that's the deeper fence-sitter conversion risk.

PR #1131, opened 2026-05-13
fix(local_store): extract model/tokens from nested OpenClaw event payload on write (#1129 bug 2)

vivekchand:
## Summary

Fixes bug 2 from #1129. `_event_to_row` was reading `e["cost_usd"]` / `e["token_count"]` / `e["model"]` directly off the top-level event dict — but OpenClaw's gateway emits these nested under `data.modelId` and `data.promptCache.lastCallUsage.{input,output,total}`. Result: every event landing in DuckDB had `model=""`, `tokens=0`, `cost=null`, and every read-side aggregate (brain history, sessions, usage charts) silently showed empty values.

- New `_extract_event_metrics` helper handles four payload shapes:
  - **OpenClaw gateway**: `data.modelId` + `data.provider` + `data.promptCache.lastCallUsage.{input,output,total}`
  - **OpenClaw message**: `data.message.usage.{inputTokens,outputTokens,totalTokens}` with already-priced `data.message.usage.cost.total`
  - **Anthropic SDK**: `data.usage.{input_tokens,output_tokens,total_tokens}`
  - **Already-extracted top-level**: preserved as-is (interceptor / claude-cli adapter / sync / cloud ingest)
- Cost is derived via `providers_pricing.estimate_cost_usd` when input/output split + provider + model are all known. When only `total` tokens are available (asymmetric input/output rates can't price a single total correctly), cost stays `None` and the read-side computes on demand. Pre-priced `data.message.usage.cost.total` is honoured when present.
- Helper never raises — store stays a permissive ingest path.

## Test plan

- [x] New regression suite `tests/test_event_metrics_extraction.py` (7 tests, all green) — pins down each shape contract plus empty-event safety
- [x] Existing `tests/test_local_store.py` (33 tests) still passes — no regression in the already-extracted-values path
- [x] Syntax check on `clawmetry/local_store.py`

### DIFF
--- a/clawmetry/local_store.py
+++ b/clawmetry/local_store.py
@@ -2304,6 +2304,161 @@ _EVENT_COLS = [
 ]
 
 
+def _extract_event_metrics(
+    e: dict[str, Any],
+) -> tuple[float | None, int | None, str | None]:
+    """Pull (cost_usd, token_count, model) from an event with shape fallbacks.
+
+    Top-level ``cost_usd`` / ``token_count`` / ``model`` are honoured first —
+    that's what the interceptor, claude-cli adapter, sync, and tests already
+    provide. When absent (the OpenClaw gateway/jsonl shape), we fall through
+    nested payload shapes:
+
+      * OpenClaw: ``data.modelId`` + ``data.provider`` +
+        ``data.promptCache.lastCallUsage.{input,output,total,cacheRead,cacheWrite}``
+      * OpenClaw message: ``data.message.usage.{inputTokens,outputTokens,totalTokens}``
+        with ``data.message.usage.cost.total`` (already-priced)
+      * Anthropic SDK:  ``data.usage.{input_tokens,output_tokens,total_tokens}``
+
+    Cost is derived from tokens × pricing only when input/output split AND
+    provider AND model are all known (so we match
+    ``providers_pricing.estimate_cost_usd``'s asymmetric rates). When only
+    ``total`` is available, we leave cost=None — the brain UI / read-side
+    aggregates compute it from tokens+model on demand and shouldn't see a
+    half-correct value here.
+
+    Never raises: any extraction failure quietly leaves the field as None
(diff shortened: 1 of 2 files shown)


vivekchand:
## 🍎 Steve Jobs UX review

🟡 **P1 — backend write-path fix, but the user-visible consequence is the entire Brain tab going from "every row shows model='' / tokens=0 / cost=$0" to actually populated.** Quietly the highest UX impact PR in today's queue, despite touching no template.

🟢 **nit** — `_extract_event_metrics` reads `data.message.usage.cost.total` for the OpenClaw pre-priced path. Good — SDK-billed cost is more reliable than re-derivation. But there's no test asserting that the OpenClaw `cost.total` field BEATS a re-derived value when both exist. Add one for the "what if SDK reported $0.05 but our pricing table says $0.06" tiebreaker case.

🟢 **nit** — fall-through chain `e["model"] → data.model → data.modelId → data.model_id` — the `model_id` snake-case fallback is for Hermes/other adapters per the `project_multi_agent_adapter` direction. Document inline so a future reader doesn't delete it as dead code.

vivekchand:

🟢 **good** — fix at the source, not at every read site. Without this, every Brain query, every aggregate, every "cost-by-team" Cloud-Pro feature reads zero. With it, those features actually work for free for every adapter. One PR, four downstream features fixed.

🟡 **P1 — cost extraction is conservative.** When only `total` tokens are known (no input/output split), cost stays None. That covers Datadog-style metric drops but means a fraction of OpenClaw events still have null cost. Acceptable for v1; track as a `cost_unknown_event_rate` metric and revisit if the rate is >5% in field telemetry. **Action**: add the metric to `/api/system-health` so we know whether to invest in fallback heuristics.

🟢 **OSS leakage**: zero — extracting metrics is foundational, not a paid feature. The cost-aggregate UI on top of it IS Cloud-Pro. Right split.

PR #1132, opened 2026-05-13
fix(sessions): handle OpenClaw transcript shapes + compute message_count on read (#1129 bugs 3+4)

vivekchand:
## Summary

Closes the last 2 of 4 bugs from #1129.

### Bug #3 — `/api/transcript/<session_id>` returned empty/garbage for OpenClaw sessions

`_try_local_store_transcript` (routes/sessions.py) assumed every DuckDB event row was an Anthropic-shaped message (`{role, content, usage, tool_calls}`). OpenClaw writes a very different shape: `{type: "<ns>.<action>", data: {...}}`. The old code read `obj.get("role")` (None) then fell back to `obj.get("type")` ("trace.artifacts" — a debug type, not a transcript role), with `content=""` and `tokens=0`. Result: every OpenClaw transcript rendered as a wall of `role="trace.artifacts"` empty rows.

The helper now discriminates via a new `_is_openclaw_event()` check and routes OpenClaw events through `_expand_openclaw_event()` which maps:

| OpenClaw type | Transcript turn(s) |
|---|---|
| `prompt.submitted` | `user` (`data.finalPromptText` / `text` / `prompt`) |
| `trace.artifacts` | optional `user` (`finalPromptText`) + `assistant` (`assistantTexts`) + `tool` per `toolMetas[]` |
| `model.completed` | `assistant` (`completionText` / `text` / `assistantTexts`) |
| `tool.call` / `tool.invoked` | `tool` with formatted name + input |
| `tool.result` / `tool.completed` | `tool` with formatted name + output |
| `session.{started,ended,created}`, `context.compiled`, `agent.heartbeat` | **skipped** (plumbing) |
| Unknown `<x>.<y>` | **skipped** (no garbage roles) |

Tokens sum from `data.promptCache.lastCallUsage.total` (or `input+output`), model prefers `obj.modelId`. Anthropic-shaped events still flow through the original logic — mixed sessions work.

### Bug #4 — `/api/sessions` always reported `message_count: 0`

The `sessions` table has a `message_count` column, but the OpenClaw ingest path never populates it (only `sync.py` + Claude Code adapter do). Fixed `query_sessions_table` (clawmetry/local_store.py) to compute the count on read via a correlated subquery against `events`, taking `GREATEST(stored_value, computed_count)` so adapters that DO populate the column (and may have no events table rows) still work.

## Test plan

- [x] New: `tests/test_transcript_openclaw_shapes.py` (12 tests) — each OpenClaw event type produces correct turns; plumbing events emit nothing; unknown types skipped; Anthropic shape still works; tokens summed from `promptCache.lastCallUsage`; model from `modelId`.
- [x] New: `tests/test_query_sessions_message_count.py` (4 tests) — count from events table; falls back to stored column when no events; uses max when both set; isolated per session.
- [x] Both files: `python3 -m pytest tests/test_transcript_openclaw_shapes.py tests/test_query_sessions_message_count.py -q` → 16 passed.
- [x] Adjacent test suites (`test_sessions_by_type_local_store.py`, `test_local_store.py`, `test_local_store_multi_agent.py`) still pass. Two pre-existing failures in `test_sessions_local_fastpath.py` are unaffected by this PR (verified by re-running on `main`).
- [x] `python3 -c "import ast; [ast.parse(open(f).read()) for f in ['routes/sessions.py', 'clawmetry/local_store.py']]"` clean.

## Notes

Event shapes pasted in the issue description matched what the helper sees in practice. No deviations.

### DIFF
--- a/clawmetry/local_store.py
+++ b/clawmetry/local_store.py
@@ -1917,16 +1917,29 @@ class LocalStore:
         clauses: list[str] = []
         params: list[Any] = []
         if agent_type:
-            clauses.append("agent_type = ?")
+            clauses.append("s.agent_type = ?")
             params.append(str(agent_type))
         where = ("WHERE " + " AND ".join(clauses)) if clauses else ""
+        # ``sessions.message_count`` is only populated by the typed-session
+        # ingest path (sync.py + claude_code adapter). The OpenClaw events
+        # path never sets it, so reading the column gave ``message_count: 0``
+        # for every OpenClaw session (#1129 bug 4). Compute it on read via a
+        # correlated subquery against ``events`` and fall back to the stored
+        # column for agents that DO populate it (e.g. ingest from sync.py
+        # where the events table may be empty).
         sql = f"""
-            SELECT agent_type, session_id, agent_id, title, started_at,
-                   last_active_at, ended_at, status, total_tokens, cost_usd,
-                   message_count, metadata
-            FROM sessions
+            SELECT s.agent_type, s.session_id, s.agent_id, s.title, s.started_at,
+                   s.last_active_at, s.ended_at, s.status, s.total_tokens, s.cost_usd,
+                   GREATEST(
+                       COALESCE(s.message_count, 0),
+                       (SELECT COUNT(*) FROM events e
+                          WHERE e.session_id = s.session_id
+                            AND e.agent_type = s.agent_type)
+                   ) AS message_count,
+                   s.metadata
+            FROM sessions s
             {where}
-            ORDER BY COALESCE(last_active_at, started_at) DESC NULLS LAST
+            ORDER BY COALESCE(s.last_active_at, s.started_at) DESC NULLS LAST
             LIMIT ?
         """
         params.append(int(limit))
--- a/routes/sessions.py
+++ b/routes/sessions.py
@@ -2046,12 +2046,201 @@ def api_transcripts():
     return jsonify({"transcripts": transcripts[:50]})
 
 
+def _is_openclaw_event(obj: dict) -> bool:
+    """Return True if ``obj`` looks like an OpenClaw event (vs an Anthropic
+    message).
+
+    OpenClaw events carry ``{"type": "<namespace>.<action>", "data": {...}}``
+    with no top-level ``role``. Anthropic-shaped messages have a top-level
+    ``role`` field (``user`` / ``assistant`` / ``system``)."""
+    if not isinstance(obj, dict):
+        return False
+    if obj.get("role"):
+        return False
+    t = obj.get("type")
+    if not isinstance(t, str):
+        return False
+    # OpenClaw types are dotted: prompt.submitted, trace.artifacts, etc.
+    return "." in t
+
+
+def _openclaw_event_tokens(data: dict) -> int:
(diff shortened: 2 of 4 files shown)


vivekchand:
**Schema scope clarification (per #1135).**

This PR's helpers (`_is_openclaw_event`, `_expand_openclaw_event`) handle the **dot.separated trajectory schema** (`prompt.submitted`, `trace.artifacts`, `model.completed`). That's what the events in DuckDB look like TODAY because the daemon currently only parses trajectory sidecar files.

Investigation in #1135 showed the canonical bare `<sid>.jsonl` files use an entirely different underscore_separated schema (`type: message`, `message: {role, content}`, `type: tool_use_result`, `type: model_change`). Once #1135's underscore parser ships, this transcript helper will need a sibling `_expand_openclaw_v3_event` for the underscore shapes — or a unified handler.

This PR is still correct for what it covers. Just flagging that #1135's fix will require an addendum here.

vivekchand:
## 🍎 Steve Jobs UX review

🔴 **P0 — schema scope incomplete (per #1135)**. This PR's `_expand_openclaw_event` handles dot.separated event types (`prompt.submitted`, `trace.artifacts`, `model.completed`). But canonical OpenClaw bare `<sid>.jsonl` files use underscore_separated v3 schema (`type: message`, `type: tool_use_result`, `type: model_change`). After #1135's parser ships, the v3 events SHOULD be mapped to dot-separated event_types on write so this expander handles them — verify the spec says exactly that and that the expander doesn't silently drop the v3-derived rows.

🟡 **P1 — `message_count` via correlated subquery is the right pattern**, but with `GREATEST(stored_value, COUNT(events))` you've created a tie-break behavior that's hard to reason about. If an adapter populates `message_count: 5` then a 6th event lands, the user sees `6`. If an adapter populates `message_count: 100` and only 5 events land (cloud-side adapter writes session row but not events), the user sees `100` not `5`. Document which case is correct in a code comment.

🟢 **nit** — 16 tests is great. The transcript-shape test would be even stronger if it asserted `messages[0].role == "user"` and `messages[1].role == "assistant"` ORDER, not just count. Subtle but the user complaint "transcript out of order" would be caught.

vivekchand:

🔴 **P0 — same incompleteness flag as Steve's**: the schema this PR handles (`prompt.submitted`/`trace.artifacts`) is the trajectory-sidecar shape, not the canonical bare-jsonl shape. Without #1135's underscore parser shipping in lockstep, real users see empty transcripts.

🟡 **P1 — first-principles right call: compute aggregates on read, not on write.** The previous `sessions.message_count` column was only correct if 100% of write paths populated it consistently. They didn't. Read-time computation eliminates the divergence at the cost of one cheap DuckDB count. Trade is correct.

🟢 **OSS leakage**: transcript visibility is OSS table-stakes; not a Cloud-Pro feature. Right tier.

🟢 **good**: the `_is_openclaw_event(obj)` discriminator (`"." in obj.get("type", "")` and not `obj.get("role")`) is the kind of compact, future-proof check that reads correctly forever. No magic constants, no schema version sniffing, just: "if it walks like an OpenClaw event, it is one." Steal this pattern for other adapters.

Issue #1133, opened 2026-05-13
[PRD] MOAT permanence: testing + monitoring strategy so the 4 bugs from #1129 never come back


vivekchand:
# PRD — MOAT permanence

**Author**: Claude (autonomous MOAT loop)
**Date**: 2026-05-13
**Status**: draft — for founder review
**Related**: #1129 (umbrella), PRs #1128/#1130/#1131/#1132 (the 4 fixes)

## Problem

In 5 hours of live investigation on a real install today, ClawMetry's DuckDB-everywhere architecture was found to have 4 distinct silent failures, all shipping in 0.12.179:

1. Daemon ingest swallowed sidecar `.jsonl` files, producing phantom session_ids
2. Write-path didn't extract `model/tokens/cost` from OpenClaw's nested event payloads → every brain row showed empty/zero metrics
3. Read-path (`/api/transcript`) expected Anthropic message shape, rendered OpenClaw events as `role="trace.artifacts"` with empty content
4. Sessions API hard-coded `message_count` from a never-populated column

All 4 silently surfaced WRONG data to the user. None crashed. None triggered a single test failure. The `_source: local_store` tag was correctly set on every endpoint — making the failure mode invisible at the contract layer.

**The bigger pattern**: writes and reads were authored against two different mental models of "what an event looks like" (Anthropic message vs OpenClaw trace). The schema layer didn't enforce that contract anywhere, so they drifted invisibly.

## Goal

The MOAT is "all UI elements rely on DuckDB; all data stored locally; cloud only displays pre-computed aggregates." For that to hold, three things must always be true:

1. **Every write to DuckDB extracts the correct semantic fields** (model, tokens, cost) regardless of which agent stack produced the event.
2. **Every read endpoint that claims `_source: local_store` actually returns truthful data** — not just a successful JSON envelope.
3. **A regression in either direction is caught BEFORE merge**, not by a user noticing an empty brain tab in production.

## Non-goals

- Performance work (DuckDB is already sub-millisecond on existing query shapes).
- Schema migration to a strict typed columnar form (the current JSON `data` column gives us flexibility we'd lose).
- Cloud-side dedup (separate problem — covered by epic #1032).

### Layer 1 — Schema contract tests (per ingest path)

[...]

PR #1134, opened 2026-05-13
test: MOAT regression gauntlet for the 4 bugs from #1129

vivekchand:
## Summary

End-to-end regression gauntlet for the 4 DuckDB read-path bugs filed in #1129. Each test exercises one bug from the *outside* of its fix (write through `LocalStore.ingest` / `sync.sync_sessions_recent`, read through `query_events` / `query_sessions_table` / `_try_local_store_transcript`), so the moment the corresponding fix PR lands the test flips green and stays green.

**On bare `main` today: all 4 tests FAIL.** That's the proof they catch the bugs.

## The 4 tests (one per bug)

| Test | Bug | Fix PR | Today's failure |
|---|---|---|---|
| `test_phantom_trajectory_session_id_not_in_duckdb` | #1129 bug 1 — `_list_session_jsonls` ingests `.trajectory.jsonl` / `.checkpoint.jsonl` / `.deleted.jsonl` sidecars | #1130 | 4 distinct session_ids in DuckDB (1 real + 3 phantoms `aaaaaa.trajectory`, `aaaaaa.checkpoint`, `aaaaaa.deleted`) |
| `test_event_metrics_extracted_from_nested_openclaw_payload` | #1129 bug 2 — `_event_to_row` reads top-level keys but OpenClaw nests them under `data.modelId` + `data.promptCache.lastCallUsage` | #1131 | `model = None` for the nested OpenClaw event (silent NULL) |
| `test_transcript_renders_openclaw_event_shapes` | #1129 bug 3 — `_try_local_store_transcript` expects Anthropic `{role, content}` shape, drops every OpenClaw event | #1132 | empty `messages` list (0 of 3 expected turns) |
| `test_sessions_message_count_reflects_actual_event_count` | #1129 bug 4 — `query_sessions_table` reads the unpopulated `message_count` column | #1132 | `message_count = 0` for a session with 7 events |

Failure modes verified locally:
```
test_phantom...           → phantom session_id(s) leaked: {'…trajectory', '…checkpoint', '…deleted'}
test_event_metrics...     → OpenClaw event lost model: column = None, expected 'claude-opus-4-7'
test_transcript_renders... → missing user turn ('hi') from prompt.submitted. Got: []
test_sessions_message_count... → message_count for 's1' should reflect the 7 events, got 0
```

## How this PR interacts with the 4 fix PRs

This PR is **purely additive** (tests/test_moat_e2e_regression_1129.py, no product-code changes). It can land in any order:

- Land this **before** #1130/#1131/#1132 → CI fails red on those 4 tests until the corresponding fix PRs merge. Useful for blocking accidental regressions.
- Land this **after** #1130/#1131/#1132 → all 4 immediately go green; gauntlet is now the permanent regression net.
- #1128 is unrelated (it fixes dead-code in sync.py after the `__main__` block — no overlap).

Recommended: merge #1130/#1131/#1132 first, then this. That way the merge of this PR is what proves the fixes hold end-to-end, and main is never red.

## Why synthesised events not the real binary

The 4 bugs are 100% determined by event field shape — not by anything OpenClaw does at runtime. Synthesising the events here keeps the gauntlet ~0.5s (vs ~30s for `tests/test_real_openclaw_binary_e2e.py`) and removes the `openclaw` binary + LLM-cred dependency. The shapes are taken straight from #1129's repro and the field-list sections of the fix PRs.

## Test plan

- [x] `python3 -m pytest tests/test_moat_e2e_regression_1129.py -v` on bare main → 4 failed (proves they catch the bugs)
- [x] Failure messages name the actual symptom from #1129 in each case (phantom session_id, NULL model column, empty transcript, message_count=0)
- [ ] After #1130 + #1131 + #1132 merge → expect 4 passed
- [ ] CI green on this PR (existing test suite) — no product-code changes, only one new file

### DIFF
--- a/tests/test_moat_e2e_regression_1129.py
+++ b/tests/test_moat_e2e_regression_1129.py
@@ -1,0 +1,479 @@
+"""End-to-end MOAT regression gauntlet for the 4 read-path bugs filed in
+issue #1129.
+
+Each of the 4 fixes is queued in its own PR (#1130, #1131, #1132 — #1128 is
+unrelated dead-code in sync.py) but **none are merged yet**. This file
+exercises every bug from the *outside* of the fix — i.e. through the public
+write+read API — so the moment the corresponding PR lands the test flips
+green and stays green forever.
+
+Expected behaviour today on bare ``main``: **all 4 tests FAIL**. That is
+the proof the gauntlet actually catches the bugs.
+
+Expected behaviour after #1130 + #1131 + #1132 merge: **all 4 PASS**.
+
+Each test is independent: a failure in one does not mask the others.
+We synthesise OpenClaw-shape events instead of spawning the real binary
+because (a) it's faster, (b) the binary needs LLM creds, and (c) the
+failure modes are 100% determined by the field shape — which we can pin
+down here with a hand-written payload.
+
+Reference fixtures patterns: ``tests/test_real_openclaw_binary_e2e.py``
+(heaviest E2E, real binary) and ``tests/test_local_store_sync_integration.py``
+(lighter pattern using only local_store + sync helpers — what we copy here).
+"""
+
+from __future__ import annotations
+
+import importlib
+import json
+import os
+import time
+from datetime import datetime, timedelta, timezone
+from unittest.mock import patch
+
+import pytest
+
+
+# ── Shared fixture: hermetic DuckDB + reloaded local_store + sync ─────────
+
+
+@pytest.fixture
+def isolated_store(tmp_path, monkeypatch):
+    """Spin up a fresh, on-disk DuckDB per test. Reloads ``local_store``
+    + ``sync`` so module-level paths pick up the tmp DB. Yields the two
+    modules so each test can drive whichever entry point it wants
+    (``store.ingest`` for direct event writes, ``sync.sync_sessions_recent``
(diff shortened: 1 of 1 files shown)


vivekchand:
## 🍎 Steve Jobs UX review

🟢 **nit / test-only** — no UI surface. Tests don't ship to users. But the structural pattern (a regression test that FAILS on `main` today and PASSES once the fix lands — proving it actually catches the bug) is the right shape of defensiveness. Every future fix-PR should ship with this same prove-it-fails-first pattern, codified as a CONTRIBUTING.md rule. **Action**: file a tiny PRD-aligned issue to add that rule to CONTRIBUTING. Stop.

vivekchand:

🟢 **good** — tests for the bug CLASS (write-corruption, read-misshape, aggregate-wrong) not just the instance. Layer-1 of PRD #1133.

🟡 **P1 — gauntlet covers 4 bugs but the underlying class produced more (#1135 is the 5th, not yet caught here).** **Action**: add a 5th test in this gauntlet for the v3 underscore-schema parse path before #1135 merges — same FAIL-on-main pattern.

🟢 **OSS leakage / tier**: zero. Tests run on CI, never user-visible. **Backend infra**, no fence-sitter conversion impact. Stop.

PR #1139, opened 2026-05-13
feat(health): daemon-error-rate surfaced in /api/system-health + UI card (PRD #1133 layer 4)

vivekchand:
## Why

Earlier today the ClawMetry daemon was logging
`Sync cycle error: name 'ALERTS_EVAL_INTERVAL_SEC' is not defined` 4×/min on
**every install since 0.12.179** — 440+ occurrences over 2 hours on a single
machine. The user only discovered it by manually `tail`-ing
`~/.clawmetry/sync.log`. There is no in-product surface for daemon-side
errors today: System Health shows disks/CPU/uptime but says nothing about
whether the cloud-sync daemon itself is failing.

This PR ships the **read side** of [PRD #1133 layer 4](https://github.com/vivekchand/clawmetry/issues/1133)
so the next silent-NameError-class regression is visible on the dashboard
within seconds. No daemon-side changes — purely instrumenting what's
already in the log file.

## What

**Endpoint (`routes/health.py`)** — `/api/system-health` gains a top-level
`daemon` block:

```json
{
  "daemon": {
    "log_path": "/Users/.../.clawmetry/sync.log",
    "errors_last_5min": 3,
    "errors_last_1h": 12,
    "last_error_message": "Sync cycle error: name 'ALERTS_EVAL_INTERVAL_SEC' is not defined",
    "last_error_ts": "2026-05-13T17:42:11+00:00",
    "status": "degraded",
    "log_present": true
  }
}
```

- Tails the last ~1000 lines of `~/.clawmetry/sync.log` (bounded to 512 KB
  from EOF — cheap even on slow disks).
- Counts ERROR-level lines in rolling 5-min and 1-hour windows.
- Status thresholds per PRD: `>30/5min → broken`, `>0 → degraded`, else
  `healthy`. Missing log → `healthy` so fresh installs don't flash red.
- Honours `CLAWMETRY_HOME` for tests / non-default installs.

**UI (`overview.html` + `app.js`)** — new "🛠️ Daemon health" card in the
System Health panel showing:
- Status pill (🟢 Healthy / 🟡 Degraded / 🔴 BROKEN)
- Inline counts: "Last 5m: N · Last 1h: M" (colour-coded)
- Truncated last error message (200 chars) in a monospace box with the
  ISO-8601 timestamp underneath
- Footer with the log path so users know exactly where to tail

**Tests (`tests/test_daemon_health_endpoint.py`)** — 20 pure-unit tests, no
server / network / sleep:
- Empty / missing log → `healthy`
- Only INFO/WARNING noise → `healthy`
- 3 errors in last 5 min → `degraded`
- 50 errors in last 5 min → `broken`
- Mixed window buckets (recent + mid + old) — assert correct counts
- Old errors (>1h) excluded from windows but `last_error_message` still surfaces
- Garbage / traceback lines silently dropped (never crash on bad input)
- Long error messages truncated to 500 chars at parser layer
- `CLAWMETRY_HOME` env override changes default path

## Out of scope (deferred to follow-up PRs)

- Daemon-side `event_type='daemon.error'` heartbeat row to DuckDB
- Cloud sync of the daemon error stream
- Anything that touches daemon code or local_store code

## Test plan

- [x] `python3 -m pytest tests/test_daemon_health_endpoint.py -q` → 20 passed in 0.14s
- [x] `python3 -c "import ast; ast.parse(open('routes/health.py').read())"` → OK
- [x] `node -e "new Function(fs.readFileSync('clawmetry/static/js/app.js','utf8'))"` → OK
- [x] Jinja parse of `tabs/overview.html` → OK
- [ ] Screenshot of the new card on a real install (TODO before merge)
- [ ] Manual smoke against a sync.log with synthetic ERROR lines

## References

- PRD #1133 — daemon-error visibility (this is layer 4, read side)
- Originating bug: 0.12.179 `ALERTS_EVAL_INTERVAL_SEC` NameError, 440+ occurrences

### DIFF
--- a/clawmetry/static/js/app.js
+++ b/clawmetry/static/js/app.js
@@ -5386,6 +5386,49 @@ async function loadSystemHealth() {
       }
     } catch(e) {}
 
+    // Daemon health (PRD #1133 layer 4) — surface sync.log error rate so
+    // the silent-NameError class of bug stops slipping past users.
+    var dmWrap = document.getElementById('sh-daemon-wrap');
+    var dmEl = document.getElementById('sh-daemon');
+    if (d.daemon && dmEl) {
+      var dm = d.daemon;
+      var dmStatus = dm.status || 'healthy';
+      var dmDot = dmStatus === 'healthy' ? '🟢' : (dmStatus === 'degraded' ? '🟡' : '🔴');
+      var dmColor = dmStatus === 'healthy' ? 'var(--text-success,#22c55e)'
+                  : dmStatus === 'degraded' ? '#f59e0b'
+                  : 'var(--text-error,#dc2626)';
+      var dmLabel = dmStatus === 'healthy' ? 'Healthy'
+                  : dmStatus === 'degraded' ? 'Degraded'
+                  : 'BROKEN';
+      var e5 = dm.errors_last_5min || 0;
+      var e1h = dm.errors_last_1h || 0;
+      var msg = dm.last_error_message || '';
+      // Truncate to 200 chars at the UI layer (parser already capped at 500).
+      var msgShort = msg.length > 200 ? msg.slice(0, 200) + '…' : msg;
+      var html = '<div style="padding:8px 14px;background:var(--bg-secondary);border-radius:8px;border:1px solid var(--border-secondary);font-size:13px;">';
+      html += '<div style="display:flex;align-items:center;gap:8px;">';
+      html += dmDot + ' <span style="font-weight:600;color:' + dmColor + ';">' + dmLabel + '</span>';
+      html += '<span style="color:var(--text-muted);font-size:11px;margin-left:auto;">';
+      html += 'Last 5m: <span style="color:' + (e5 > 0 ? dmColor : 'var(--text-muted)') + ';font-weight:600;">' + e5 + '</span>';
+      html += ' &nbsp;·&nbsp; Last 1h: <span style="color:' + (e1h > 0 ? dmColor : 'var(--text-muted)') + ';font-weight:600;">' + e1h + '</span>';
+      html += '</span></div>';
+      if (msgShort) {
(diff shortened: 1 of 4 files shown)


vivekchand:
## 🍎 Steve Jobs UX review

🟡 **P1 — the new "Daemon health" card is the most important UI surface shipped today**, because every prior silent-NameError-class bug went invisible to users for hours. But the card lives on the Overview tab, not the System Health tab — that's an inconsistency. **Action**: move it to the System Health tab with the other infra cards (disks, RAM, gateway), or rename and put a "Daemon" entry in the System Health card grid for parity.

🟢 **nit** — status thresholds (`broken` if errors_last_5min > 30) are reasonable defaults but invisible to the user. **Action**: tooltip on the status badge that says "Healthy: 0 errors. Degraded: 1-29 errors in 5min. Broken: 30+." Three numbers. Removes mystery.

🟢 **nit** — `last_error_message` truncated at 200 chars in UI. For an `ALERTS_EVAL_INTERVAL_SEC is not defined`-style error, fine. For a Python traceback, 200 chars hides the actual line. **Action**: "Show full" expand toggle, or click-to-copy the full message to clipboard.

vivekchand:

🟢 **good** — closes the loop on PRD #1133 layer 4. Without this, the next ALERTS_EVAL_INTERVAL_SEC-class bug goes silently unfixed for 4 hours / users hit it 4×/min until someone manually tails the log. With it, the dashboard surfaces it in the next 60s heartbeat refresh.

🟡 **P1 — observability of YOUR observability tool was the gap.** Eat your own dogfood. **Action**: ClawMetry should self-monitor — when this card goes red, send a PostHog event so you (the founder) get an aggregate signal on how many user installs are running broken daemons in the wild RIGHT NOW. Currently you'd only know if a customer complains.

🟢 **OSS leakage / tier**: zero. Daemon health is foundational — not a paid feature. The "alert ME when MY daemon goes red" flow IS Cloud-Pro (delivery via Slack/PagerDuty per the alerts evaluator). Right tier split.

PR #1140, opened 2026-05-13
test(e2e): capture full error.stack to surface JS source line (unblocks 0.12.181 diagnosis)

vivekchand:
## Summary
- Enrich `tests/e2e/cloud-contract.mjs` pageerror capture: store `e.stack` (truncated to 2000 chars) instead of `e.message.slice(0, 200)`.
- Surfaces the actual file:line where `Cannot read properties of null (reading 'sequence')` originates, unblocking diagnosis of the 0.12.181 release.
- One-line change, assertion logic unchanged.

## Test plan
- [x] `node -c tests/e2e/cloud-contract.mjs` parse-check passes
- [ ] CI re-run will show the file:line stack frame for any future pageerror failure

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1138, opened 2026-05-13
[RELEASE] feat: 0.12.181 — underscore (v3) OpenClaw schema parser (#1137)

vivekchand:
## Summary
Triggers PyPI auto-publish for **0.12.181**.

Ships the single delta since 0.12.180:
- **#1137** — `feat(sync)`: underscore (v3) OpenClaw schema parser. Closes #1135.

Adds a per-event schema sniffer that detects v3 underscore types (`session`, `model_change`, `message`, `tool_use_result`, etc.) and maps them to the standardised dot.separated event_types (`session.started`, `model.changed`, `prompt.submitted`, `model.completed`, `tool.call`, `tool.result`) the read-side expanders already understand. 19 new tests + 4 existing tests updated.

## Why ship now
0.12.180's #1130 (sidecar filter) made the daemon stop ingesting `.trajectory.jsonl` artifacts, exposing that canonical bare `<sid>.jsonl` files weren't being parsed into readable transcripts. #1137 closes that gap.

## Test plan
- [x] All 17 CI checks green on #1137
- [x] 64-test broader sweep on touched paths — zero new regressions
- [ ] After publish: `pip install --upgrade clawmetry` and verify `/api/transcript/<sid>` renders messages from canonical bare jsonl files.

PR #1137, opened 2026-05-13
feat(sync): underscore (v3) OpenClaw schema parser — canonical .jsonl sessions finally ingested (#1135)

vivekchand:
Closes #1135.

> **One-line callout:** Without this, real users have empty/wrong DuckDB. With this, MOAT is finally truthful end-to-end.

## The two schemas

OpenClaw writes TWO different jsonl shapes per session:

**Schema A — \"trajectory\"** (`<sid>.trajectory.jsonl` sidecar)
```json
{\"type\": \"trace.artifacts\", \"data\": {\"finalPromptText\": \"...\", \"promptCache\": {\"lastCallUsage\": {...}}, \"assistantTexts\": [...]}}
{\"type\": \"model.completed\", \"data\": {...}}
{\"type\": \"prompt.submitted\", \"data\": {...}}
```
Dot.separated types, content under `data.*`. The trajectory parser (which the daemon used unconditionally before this PR) was originally written for this shape.

**Schema B — \"v3 underscore\"** (`<sid>.jsonl` canonical user-facing transcript)
```json
{\"type\": \"session\", \"version\": 3, \"id\": \"<sid>\", \"timestamp\": \"...\", \"cwd\": \"...\"}
{\"type\": \"model_change\", \"provider\": \"anthropic\", \"modelId\": \"claude-opus-4-7\"}
{\"type\": \"thinking_level_change\", \"thinkingLevel\": \"off\"}
{\"type\": \"message\", \"message\": {\"role\": \"user\", \"content\": [...]}}
{\"type\": \"message\", \"message\": {\"role\": \"assistant\", \"content\": [...], \"usage\": {...}, \"model\": \"...\"}}
{\"type\": \"tool_use_result\", \"tool_use_id\": \"...\", \"content\": [...]}
```
Underscore_separated types, LLM payload nested under `message.{role,content,usage,model}`, schema marker `version: 3`.

Until #1135 the daemon stamped `event_type=\"message\"` on these rows AND left the data un-translated, so the dashboard's transcript expander (designed for the dot.separated shape from PR #1132) saw none of the content. Tokens / cost / model attribution all silently zeroed out.

## What this PR does

* New `_is_v3_event(obj)` sniff inside `_local_ingest_session_batch` (per-event, not per-batch — keeps the code correct when a single flush mixes shapes).
[...]

PR #1141, opened 2026-05-13
[RELEASE] feat: 0.12.181 — underscore (v3) parser + Layer 4 daemon health + e2e diagnostic

### LINKED ISSUE #1135: [MOAT P0] Daemon ingest only understands legacy trajectory schema (dot.types) — real OpenClaw bare .jsonl uses underscore_types, never gets ingested
## Discovered while investigating #1129

The previous bugs in #1129 (#1130/#1131/#1132 fixes) are partial. There is a **fundamental schema mismatch** between what the daemon's sync_sessions parser expects and what the real OpenClaw bare-jsonl files contain.

## Two schemas exist

OpenClaw writes TWO different files per session:

**1. `<sid>.jsonl` — the canonical session transcript (real user-facing data, ~1-3KB)**
Schema (verified from a live install on 2026-05-13, file `3fc28a8b-...jsonl`):
\`\`\`json
{"type": "session", "version": 3, "id": "3fc28a8b-...", "timestamp": "...", "cwd": "..."}
{"type": "model_change", "id": "...", "parentId": null, "timestamp": "...", "provider": "anthropic", "modelId": "claude-opus-4-7"}
{"type": "thinking_level_change", "id": "...", ...}
{"type": "message", "id": "...", "parentId": "...", "timestamp": "...", "message": {...content...}}
{"type": "tool_use_result", "id": "...", "tool_use_id": "...", "content": ...}
\`\`\`
**Underscore_separated types**, message content nested under \`.message\`, schema version 3.

**2. `<sid>.trajectory.jsonl` — runtime debug trace sidecar (~100KB)**
Schema:
\`\`\`json
{"type": "trace.artifacts", "modelId": "...", "data": {"finalPromptText": "...", "promptCache": {...}, ...}}
{"type": "session.ended", ...}
{"type": "model.completed", "data": {...}}
{"type": "prompt.submitted", "data": {...}}
{"type": "context.compiled", ...}
\`\`\`
**Dot.separated types**, content in \`.data\`, an entirely different shape.

## The bug

The daemon's parser at `clawmetry/sync.py` only knows the dot.separated trajectory schema. The 7 events currently in this user's DuckDB ALL came from the trajectory sidecar — none from the canonical bare .jsonl. PR #1132 (which I just shipped today) handles \`prompt.submitted\`/\`trace.artifacts\`/\`model.completed\` — those are the TRAJECTORY shapes, not the canonical ones.

## Worse: PR #1130 would make this worse

PR #1130 (sidecar filter) tells the daemon to STOP ingesting `.trajectory.jsonl` files. Once merged in isolation, the daemon will ONLY look at bare `<sid>.jsonl` files — which it can't parse. Net effect: **DuckDB will become empty for fresh installs** instead of containing the wrong-but-present trajectory data it has today. **Do not merge PR #1130 without the parallel underscore-schema parser shipping in the same release.**

## Repro on this user's install

\`\`\`
~/.openclaw/agents/main/sessions/
├── 3fc28a8b-b0f5-47af-b234-fa2f96db8112.jsonl  # 2128 bytes, 7 lines, underscore schema, May 13
├── 4595d095-5b7e-40f7-908c-9a5a9f92f63a.jsonl  # 1992 bytes, underscore schema, May 9
└── 4595d095-5b7e-40f7-908c-9a5a9f92f63a.trajectory.jsonl  # 109541 bytes, dot schema, May 9
\`\`\`

DuckDB state via daemon proxy:
- 7 events, ALL from \`4595d095-...trajectory\` (with the phantom .trajectory suffix)
- 0 events from \`3fc28a8b-...\` (the canonical May-13 session)
- 0 events from \`4595d095-...\` (the canonical May-9 session)

Daemon's `~/.clawmetry/sync-state.json` confirms:
\`\`\`json
"last_event_ids": {
  "4595d095-...trajectory.jsonl": 7  # only the trajectory was actually parsed
},
"session_mtimes": {
  "3fc28a8b-...jsonl": 1778638729.27,           # the daemon SAW it
  "4595d095-...jsonl": 1778321249.26,           # but never extracted any events
  "4595d095-...trajectory.jsonl": 1778321249.29
}
\`\`\`

## Fix sketch (what needs to happen)

A new parser path in `clawmetry/sync.py`'s session ingest that:
1. Detects the schema by sniffing the first line: if `type == "session"` with a `version` key → v3 underscore schema, dispatch to new parser. Else fall back to existing dot-syntax parser.
2. Maps underscore types to the same DuckDB event shape as the dot types:
   - `session` → `session.started`
   - `model_change` → emit a synthetic `model.changed` event
   - `message` (role=user) → user transcript turn; (role=assistant) → assistant turn
   - `tool_use` → tool call; `tool_use_result` → tool result
   - `thinking_level_change` → skip (plumbing)
3. Token extraction: underscore schema's `message.usage` field has `input_tokens`/`output_tokens` (Anthropic standard).
4. Test against BOTH the user's existing canonical files AND a synthesized fixture.

This is a **significantly larger PR** than #1130/#1131/#1132 (which handled the *symptom* of one schema). It needs to be a clean schema-router with two complete parsers underneath.

## Acceptance

- After this fix lands, the user's existing `3fc28a8b-...jsonl` produces 7 events in DuckDB (one per JSONL line, mapped to standard event_types).
- Brain history surfaces these 7 events with correct model="claude-opus-4-7", token counts.
- Transcript renders the underscore-schema messages in the right order.
- `/api/sessions` shows `message_count: 7` for `3fc28a8b`.
- The MOAT regression gauntlet (PR #1134) gets a 5th test for the underscore-schema parse path.

## Why this is the most important MOAT bug today

PRs #1130/#1131/#1132 fix the symptoms downstream of corrupt/wrong-shape events. THIS bug is why the events are corrupt/wrong-shape in the first place. **Without this fix, no real OpenClaw user has ever had truthful data in their local DuckDB.** That's the MOAT, and it's been silently broken since the v3 OpenClaw schema shipped.

Filed during today's MOAT autonomous loop after 6 cron ticks of investigation — I should have caught this on tick 2 when I first saw the phantom `.trajectory` session_id. The phantom suffix was the symptom; the schema gap is the disease.

vivekchand:
Re-attempt of 0.12.181 publish (the first attempt #1138 was blocked by a JS pageerror smoke gate; the source could not be located without a stack frame).

Bundles three recent merges:
- **#1137** — underscore (v3) OpenClaw schema parser (closes #1135). The actual MOAT fix.
- **#1139** — daemon-error-rate surface in /api/system-health + UI card (PRD #1133 layer 4).
- **#1140** — diagnostic: cloud-contract.mjs now captures full error.stack. If the JS error recurs in this CI run, the log will reveal file:line.

If the cloud-contract spec fails again here, the stack frame should be in the workflow log — file as a follow-up.


vivekchand:
## 🍎 Steve Jobs UX review

🟡 **P1 — release ops, no UI diff in this PR**, but worth one structural critique. Bundling 3 unrelated changes (#1137 + #1139 + #1140) into one [RELEASE] is convenient but creates a debugging hazard: if 0.12.181 introduces a regression, bisecting which of the 3 caused it is harder than 3 single-PR releases. **Action**: as user adoption grows past 120k installs, prefer 1-PR-per-release where practical. Cost is more PRs in PyPI history; benefit is cleaner regression bisection.

vivekchand:

🟢 **good — re-attempt of the blocked 0.12.181 release** with the diagnostic patch (#1140) in the same payload as the underlying cause (#1137). If smoke fails again now, file:line is in the log. If it passes, MOAT ships clean. Either outcome is forward progress.

🟢 **OSS leakage / tier**: zero — release ops. Daemon-health card from #1139 is OSS-exposed (foundational); cloud delivery of those alerts is Cloud-Pro. Tier-clean.

🟢 **nit** — title says `feat:` not `chore: bump` — bypasses the auto-bumper-bot skip pattern. That's intentional today (bundles real features). Make sure the convention is documented somewhere so a future contributor doesn't accidentally make a chore-PR a [RELEASE] feat:.

PR #1143, opened 2026-05-13
fix(sync): v3 parser stores data.type for transcript discriminator (closes MOAT)

vivekchand:
## Summary

Closes the MOAT transcript path for canonical v3 OpenClaw sessions.

The v3 underscore parser (PR #1137) correctly mapped event types onto the dot.separated ``event_type`` column, but the ``data`` JSON blob stored alongside dropped the ``type`` field — only ``_v3_type`` (the original underscore tag) survived.

PR #1132's transcript discriminator in ``routes/sessions.py::_is_openclaw_event`` checks ``obj.get("type")`` for a dot-separator. On v3 rows it saw ``None`` → False → fell through to the Anthropic-shape branch → 0 messages rendered. ``/api/brain-history`` worked fine because it reads the typed ``event_type`` column directly, masking the bug.

Two fixes in ``_parse_v3_event``:

1. **Stamp ``data.type``** = the dot.separated event_type, so the discriminator recognises the row.
2. **Nest content under ``data.data``** mirroring trajectory shape, so ``_expand_openclaw_event`` and ``_openclaw_event_tokens`` find content + usage at the same key paths they already use for trajectory events. Without this the discriminator passes but the expander returns 0 turns because it reads ``data.data.finalPromptText`` etc.

The original flat top-level keys (``finalPromptText``, ``completionText`` …) are kept alongside the nested copy so brain-feed / debug consumers and the existing #1135 parser tests keep working.

Verified live on 0.12.181: synthetic v3 session writes 5 events with correct event_type column but data.type=None. After this fix the transcript fast path returns ``_source: local_store`` with correct user/assistant/tool role mapping and tokens summed from ``data.data.promptCache.lastCallUsage``.

## Test plan

- [x] ``tests/test_v3_schema_parser.py`` — 9 new parametrized tests pin ``data["type"] == row["event_type"]`` for every parser mapping and ``data["data"]`` carries the content fields.
- [x] ``tests/test_moat_e2e_regression_1129.py`` — new layer-1 regression test feeds a synthetic v3 session through ``_local_ingest_session_batch`` and asserts ``_try_local_store_transcript`` returns ``_source: local_store`` with user/assistant/tool turns + model + tokens.
- [x] All 28 v3 parser tests pass; new MOAT test passes; pre-existing test 3 failure (``test_transcript_renders_openclaw_event_shapes``) is unchanged (separate latent bug in the expander's flat-shape handling, not in scope per fix constraint).

### DIFF
--- a/clawmetry/sync.py
+++ b/clawmetry/sync.py
@@ -1498,7 +1498,16 @@ def _parse_v3_event(
 
     # Defaults; overridden per type below.
     event_type = "unknown"
+    # ``data`` mirrors the TRAJECTORY shape that routes/sessions.py expects:
+    # top-level ``type`` + ``timestamp`` + sometimes ``modelId`` for the
+    # discriminator / model picker, and a NESTED ``data`` sub-dict carrying
+    # the content fields (``finalPromptText``, ``completionText``,
+    # ``promptCache``, ``toolMetas`` …) that ``_expand_openclaw_event`` and
+    # ``_openclaw_event_tokens`` read. The legacy flat keys that #1135's
+    # parser stored at the top level are kept alongside the nested copy so
+    # existing brain-feed / debug consumers keep working.
     data: dict = {"_v3_type": t}  # preserve the original tag for debugging
+    inner: dict = {}  # nested ``data.data`` payload — matches trajectory shape
     cost_usd: float | None = None
     token_count: int | None = None
     model: str | None = None
@@ -1513,6 +1522,11 @@ def _parse_v3_event(
             "cwd": obj.get("cwd"),
             "timestamp": ts,
         })
+        inner.update({
+            "id": obj.get("id"),
+            "version": obj.get("version"),
+            "cwd": obj.get("cwd"),
+        })
 
     elif t == "model_change":
         event_type = "model.changed"
@@ -1521,6 +1535,10 @@ def _parse_v3_event(
             "modelId": model,
             "provider": obj.get("provider"),
             "timestamp": ts,
+        })
+        inner.update({
+            "modelId": model,
+            "provider": obj.get("provider"),
         })
 
     elif t == "message":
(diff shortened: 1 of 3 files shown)


vivekchand:
## 🍎 Steve Jobs UX review

🟡 **P1 — closes the user-visible MOAT gap** (transcript renders v3 sessions correctly), but the architectural choice deserves scrutiny: `data` now holds BOTH the flat keys (`finalPromptText`, `completionText`, `_v3_type`) AND a nested `data.data` mirror of the trajectory shape. Two representations of the same content. Cleaner long-term: pick one canonical write shape (likely the nested one) and migrate the flat keys out in a follow-up. Otherwise future readers will have to know which path to read for which adapter. **Action**: file a cleanup issue to converge on one shape after one retention cycle (~24h) of dual-write proves no breakage.

🟢 **good** — preserves backward compat with the existing #1135 parser tests (those flat keys are still there). Apple-grade: ship the user-facing fix without forcing all existing tests to rewrite at the same time.

🟢 **nit** — `_v3_type` is the original underscore type, retained for debugging. Document why in a code comment so the next maintainer doesn't think it's dead code and delete it.

vivekchand:

🟢 **good — closes the last MOAT gap** I flagged in live verification. After 12 cron ticks of MOAT investigation today, the chain is now: OpenClaw writes v3 jsonl → daemon parses → DuckDB row has correct event_type column AND data.type AND data.data nested content → transcript expander matches → user sees the transcript. End-to-end truthful.

🟡 **P1 — the 1-line fix spec was wrong** (I wrote it, agent caught it). The discriminator was only ONE of the gates; the expander's content-extraction was the other. **Lesson**: when bug investigation says "X is null and that breaks Y," verify by FIXING X in a minimal repro first. The "+53 lines" surprise here was avoidable. **Action**: in PRD #1133 layer 1 (the MOAT regression gauntlet), add a "minimal-repro-test" requirement before any bug-fix dispatch.

🟢 **OSS leakage / tier**: zero. Parser standardization is foundational. Cloud-Pro reads the same shape.

PR #1146, opened 2026-05-13
feat(health): gateway process health panel (closes #852)

### LINKED ISSUE #852: feat: Gateway health monitor panel
Surface OpenClaw gateway process health (memory, CPU, uptime, restart count) in ClawMetry dashboard.

**Problem:** Gateway memory bloat (#63526) and unreachability are among the most common OpenClaw complaints. ClawMetry already watches the gateway for session data, but does not surface gateway process health.

**Proposed solution:**
- Memory usage trend line (detect the 600MB->945MB bloat pattern)
- Restart history timeline
- Active connection count
- Error rate and type breakdown
- Alert when memory exceeds configurable threshold

**Effort:** Small. Gateway process stats are accessible locally.

Source: Weekly feature intelligence scan (2026-04-27)

vivekchand:
## Summary
- Surfaces OpenClaw gateway process vitals (PID, RSS, CPU, uptime) so the ~600 MB → ~945 MB memory-bloat-OOM pattern is visible from the dashboard instead of from `tail -f`.
- Pure helper `compute_gateway_health()` in `routes/health.py` discovers the gateway via `~/.openclaw/gateway/gateway.pid` first, then cmdline scan; reads vitals via `psutil` when available and falls back to `ps -o rss=,pcpu=,etime= -p <pid>` (no new dependency).
- Adds `/api/gateway-health` standalone endpoint and a `gateway` block on `/api/system-health`; new "🛰️ Gateway health" card on the System Health column next to the Daemon health card.

## Thresholds
| Status | Trigger |
| --- | --- |
| `critical` | RSS > 900 MB (50 MB cushion below observed ~945 MB OOM) |
| `warning` | RSS > 675 MB (75% of 900) |
| `healthy` | RSS ≤ 675 MB |
| `not_running` | no PID file + no cmdline match |

Threshold is overridable per call (`threshold_mb=`); UI bar tracks the active value.

## Restart count
Deferred to v2 — needs cross-poll PID state. Issue notes this is OK to omit for v1.

## Test plan
- [x] `python3 -m pytest tests/test_gateway_health.py -q` — 35 passed
- [x] `python3 -m pytest tests/test_daemon_health_endpoint.py -q` — 20 passed (sibling card unaffected)
- [x] Route registers cleanly under `bp_health` (`/api/gateway-health` + payload merged into `/api/system-health`)
- [ ] Manual smoke: start OpenClaw, hit `/api/gateway-health`, confirm card renders + bar color matches RSS state
- [ ] Confirm "Not running" state when gateway is stopped (card shows ⚫ + "Not running")

### DIFF
--- a/clawmetry/static/js/app.js
+++ b/clawmetry/static/js/app.js
@@ -5429,6 +5429,65 @@ async function loadSystemHealth() {
       if (dmWrap) dmWrap.style.display = '';
     } else if (dmWrap) { dmWrap.style.display = 'none'; }
 
+    // Gateway health (#852) — surface OpenClaw gateway process vitals
+    // (RSS / CPU / uptime). The ~600MB → ~945MB memory-bloat-OOM pattern
+    // was invisible from the dashboard until this card landed.
+    var gwWrap = document.getElementById('sh-gateway-wrap');
+    var gwEl = document.getElementById('sh-gateway');
+    if (d.gateway && gwEl) {
+      var gw = d.gateway;
+      var gwStatus = gw.status || 'not_running';
+      var gwDot, gwLabel, gwColor;
+      if (gwStatus === 'critical') {
+        gwDot = '🔴'; gwLabel = 'Critical'; gwColor = 'var(--text-error,#dc2626)';
+      } else if (gwStatus === 'warning') {
+        gwDot = '🟡'; gwLabel = 'Warning'; gwColor = '#f59e0b';
+      } else if (gwStatus === 'healthy') {
+        gwDot = '🟢'; gwLabel = 'Healthy'; gwColor = 'var(--text-success,#22c55e)';
+      } else {
+        gwDot = '⚫'; gwLabel = 'Not running'; gwColor = 'var(--text-muted)';
+      }
+      var rss = (typeof gw.rss_mb === 'number') ? gw.rss_mb : null;
+      var thr = gw.memory_threshold_mb || 900;
+      var memPct = (rss !== null) ? Math.min(100, Math.round((rss / thr) * 100)) : 0;
+      var memBarColor = (gwStatus === 'critical') ? 'var(--text-error,#dc2626)'
+                      : (gwStatus === 'warning') ? '#f59e0b'
+                      : 'var(--text-success,#22c55e)';
+      // Pretty uptime: "1d 3h", "47m", "12s"
+      function _fmtUp(s) {
+        if (s === null || s === undefined) return '—';
+        s = Math.max(0, Math.floor(s));
+        if (s < 60) return s + 's';
+        if (s < 3600) return Math.floor(s/60) + 'm';
+        if (s < 86400) return Math.floor(s/3600) + 'h ' + Math.floor((s%3600)/60) + 'm';
+        return Math.floor(s/86400) + 'd ' + Math.floor((s%86400)/3600) + 'h';
+      }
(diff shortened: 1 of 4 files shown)


vivekchand:
## UX review

- 🟡 P1 Three different status indicators on the same card: emoji dot (🔴/🟡/🟢/⚫), coloured label text, AND a coloured memory bar. That is three takes on the same signal. Pick one primary. Action: drop the emoji circle in `clawmetry/static/js/app.js:5440-5446`; keep the coloured label + bar. Emojis render inconsistently across OS/browser font stacks (the existing Daemon health card already has this problem) and make the dashboard look like a Discord channel.
- 🟡 P1 "⚫ Not running" reads as a degraded state, but on a fresh OSS install with no OpenClaw process, this is the EXPECTED state — not an error. The black dot + grey label together imply something is broken. Action: in `clawmetry/static/js/app.js:5446` change the not_running label to "Idle" or "Gateway off" so first-paint on a clean install isn't alarming. Better: hide the whole card via `sh-gateway-wrap` when status is `not_running` AND the user has never had a gateway run (no PID history) — same pattern as the sandbox card.
- 🟡 P1 The card emoji is a satellite dish (🛰️) but the metric is process memory. The icon doesn't map to the content; a non-technical user can't connect "satellite" to "how much RAM OpenClaw is using." Action: at `clawmetry/templates/tabs/overview.html:87` swap 🛰️ for something legible (or just drop the emoji — "Gateway health" is clearer alone). Daemon health doesn't need a 🛠️ either, but at least that's directionally right.
- 🟢 nit: `'Memory'` label inside the bar duplicates the "945 MB / 900 MB" readout one line below it. One or the other. Action: drop the "Memory" label string in `app.js:5453`.

vivekchand:
## Product review

- 🟡 P1 "Gateway uses 945MB of 900MB" is an alert, not a card. We've buried it in the System Health column where nobody looks until something is on fire. If we know critical means "OOM imminent" (PR body confirms it), surface it. Action: when `gateway.status === 'critical'` in `clawmetry/static/js/app.js:5429`, lift a banner to the top of overview ("OpenClaw gateway at 945MB — restart imminent") and gate the auto-restart-on-OOM button behind "Cloud-Pro: 1-click recovery." That's a real conversion moment — the user has a fire and we're the firefighter.
- 🟡 P1 OSS leakage check: surfacing per-process RSS / CPU / uptime locally is correct (local compute, cloud display). But there is no Cloud-Pro CTA anywhere on this card. Multi-node fleet view shows "gateway across all your nodes" — that's the Pro story. Action: when `!CLOUD_MODE && nodeCount === 1`, add a one-line teaser under the card "See gateway health across all nodes → Start 7-day trial" linking to /cloud/upgrade. Free dashboard tile = pure cost; no funnel.
- 🟡 P1 Restart count deferred to v2 per PR body. That's the metric that actually matters — RSS at a snapshot tells you nothing without a trend. Action: ship v2 in the same week or this card has a 50% chance of being judged "nice gauge, didn't help" in user research. Track `gateway_restarts_24h` to local DuckDB now (one row per PID-change observed during the 30s poll) so v2 is a frontend-only change.
- 🟢 Simplest-thing check passes — pure helper, psutil-with-ps-fallback, no new deps. Good.
- 🟢 nit: `memory_threshold_mb` is hard-coded 900. On a 2GB Pi this is wrong. Action: detect host RAM in `routes/health.py:_default_gateway_pid_path` neighbour and scale threshold to `min(900, total_ram_mb * 0.6)`. Keep override.

PR #1149, opened 2026-05-13
fix(crons): show "+ New Job" button in OSS dashboard

vivekchand:
## Summary
- Set \`_cronActionsAvailable = !window.CLOUD_MODE\` inside \`loadCrons()\` so the "+ New Job" / Pause / Fix / Emergency Stop buttons are visible in the OSS dashboard.
- Cloud-iframe stays false (cron-create endpoint is cloud-disabled there) and the empty-state copy switches to "Create one from the OSS dashboard on the host running OpenClaw" so we don't dangle invisible controls.

## Why
\`_cronActionsAvailable\` was declared \`false\` at module init and never reassigned. The empty-state message "Click '+ New Job' to create one" pointed users at a button that was permanently \`display:none\`. Reported in vivekchand/clawmetry-cloud#793 (also mirrors clawmetry/clawmetry#1127).

## Test plan
- [x] \`node -c clawmetry/static/js/app.js\` — syntax OK
- [ ] Live: open OSS dashboard with no crons → "+ New Job" button visible, empty state mentions it.
- [ ] Live: open cloud iframe with no crons → button hidden, empty state directs to OSS dashboard.

Closes one item from vivekchand/clawmetry-cloud#793.

### DIFF
--- a/clawmetry/static/js/app.js
+++ b/clawmetry/static/js/app.js
@@ -4142,7 +4142,14 @@ function toggleCronAutoRefresh() {
 async function loadCrons() {
   var data = await fetch('/api/crons').then(r => r.json());
   _cronJobs = data.jobs || [];
-  // Show/hide cron action buttons based on gateway support
+  // Cron actions (create / pause / fix / kill-all) hit the local OpenClaw
+  // gateway, which is always reachable from the OSS dashboard but is
+  // cloud-disabled from the iframe. Default to true for the OSS-dashboard
+  // case so the "+ New Job" button is actually visible — the empty-state
+  // copy says to click it. Cloud-iframe stays false so we don't dangle
+  // controls that 410 when used.
+  _cronActionsAvailable = !window.CLOUD_MODE;
+  // Show/hide cron action buttons based on gateway availability
   document.querySelectorAll('.cron-action-btn').forEach(function(btn) {
     btn.style.display = _cronActionsAvailable ? '' : 'none';
   });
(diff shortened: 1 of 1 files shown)


vivekchand:
## UX review

- 🟡 P1 The fix sets `_cronActionsAvailable = !window.CLOUD_MODE` UNCONDITIONALLY at `clawmetry/static/js/app.js:4151`. That assumes the OSS dashboard ALWAYS has a reachable gateway. If the local OpenClaw gateway is down (port 18789 unreachable), "+ New Job" is visible but clicking it will fail with a gateway-RPC error — a worse UX than the invisible-button bug we just fixed. Action: in `loadCrons()` at `app.js:4145+` keep the OSS default `true` BUT also flip to `false` if the latest `/api/gateway-health` check (PR #1146 just added it) returns `not_running`. Pair the two PRs.
- 🟡 P1 Empty-state copy for cloud-iframe — "Create one from the OSS dashboard on the host running OpenClaw" — assumes the user knows what "the OSS dashboard" is and where it lives. Non-technical user will not. Action: in `app.js:4321` add a link target: "Create one from your local ClawMetry dashboard (open http://localhost:8900 on the machine running OpenClaw)."
- 🟢 nit: comment at `app.js:4145` says "Default to true for the OSS-dashboard case so the '+ New Job' button is actually visible — the empty-state copy says to click it." Good — this exactly identifies the duplicate-affordance problem. Keep the comment; it's load-bearing context for the next dev.
- 🟢 Otherwise this is a one-line bug fix and the right one.

vivekchand:
## Product review

- 🟡 P1 OSS leakage check: cron CREATE in OSS is now restored — fine, that's local compute. But what about cron MULTI-AGENT scheduling, cron failure alerts, cron success rate trends? Those are the Pro plays. Action: in `clawmetry/static/js/app.js:4321` add a CTA below the empty-state in cloud-iframe mode: "Or have ClawMetry Pro schedule across all your nodes from cloud." Don't let the cloud-iframe-no-button moment go to waste — it's a captive audience already paying with attention.
- 🟡 P1 Simplest possible thing was `_cronActionsAvailable = !window.CLOUD_MODE` — yes, this PR does that. But the ROOT cause was "a module-level flag that should have been computed per-call." The fix re-introduces the same brittleness: any future code path that toggles CLOUD_MODE late (e.g. an iframe handshake) will desync. Action: refactor `_cronActionsAvailable` to a function call site, not a module-level bool, inside `renderCrons` so it's always evaluated against current state. Otherwise we'll re-debug this in 6 weeks.
- 🟢 No churned-user-return surface impact. Bug class: invisible button. Fix unblocks the flow. Approved as a hotfix.
- 🟢 Single-file 19-line diff, no tests because the test would essentially be "the button is visible." Acceptable; `node -c` lint covers the syntax.

PR #1150, opened 2026-05-13
feat(overview): heartbeat liveness panel (closes #686)

### LINKED ISSUE #686: [P1] Heartbeat liveness panel — cadence, HEARTBEAT_OK ratio, miss detection
## Context
From Alexander Krentsel's Berkeley talk *Principles for Autonomous System Design: OpenClaw Deep-Dive* (slides 13, 16, 24).

OpenClaw's heartbeat session is special: auto-fires every 30 min, injects `HEARTBEAT.md` into context, replies exactly `HEARTBEAT_OK` if nothing needs attention. It's the **liveness probe** for the whole always-on agent — if heartbeats stop, the agent is effectively dead, but today ClawMetry treats it as just another session.

## Basic (this issue — OSS)
Dedicated panel on the overview dashboard that shows:
- Expected vs actual heartbeat cadence (30-min default, configurable)
- `HEARTBEAT_OK` vs action-taken ratio (how often does the agent find something to do?)
- Last-heartbeat timestamp prominently
- Visual "pulse" indicator (green if within 1.5× expected interval, amber if drifting, red if missed)

Implementation: filter `routes/sessions.py` by session type `heartbeat`, new widget in `routes/overview.py`.

## Advanced (Cloud — see #3 in clawmetry-cloud)
Push notifications (Slack/Telegram/email) when heartbeat gap exceeds threshold, 30-day cadence trend, cross-agent fleet heartbeat health.

## Source
Slides 13 ("Gateway Controller"), 16 ("Sessions"), 24 ("The LLM Call" — heartbeat prompt handling).

vivekchand:
## Summary
- Surfaces the OpenClaw 30-min heartbeat as a dedicated liveness probe on the overview dashboard instead of letting it disappear into the session list (issue #686).
- New `heartbeat` block on `/api/overview` (`expected_cadence_seconds`, `last_heartbeat_ts`, `gap_seconds`, `ok_ratio`, `sample_size`, `status`) computed from DuckDB events with a 30s memo cache.
- Status is `green` (gap < 1.5x cadence) / `amber` (1.5x-3x) / `red` (>3x) / `null` (never), matching the spec's pulse semantics.
- New "Heartbeat" card on the Overview tab matching existing card CSS (autonomy-card style) — pulse dot + "Last beat", "Cadence: 30m", "OK ratio: X% (last N)" — self-polls every 60s, hidden until first response.

## Why
Today ClawMetry treats the heartbeat session like any other turn. If it stops firing, the agent is effectively dead but nobody notices. This puts the liveness signal in the user's face on the first screen they load.

## Implementation notes
- Detection is layered: `data.session_type == "heartbeat"` first (canonical), with fallbacks to `event_type == "heartbeat"` and session_id contains "heartbeat" so older shapes still surface.
- `ok_ratio` is computed over the most recent 20 heartbeats (capped). HEARTBEAT_OK marker scanned across `response` / `reply` / `content` / `text` / `body` fields plus direct boolean flags.
- Backend block lives in `routes/overview.py` and feeds both the local-store fast path and the legacy gateway-backed handler. The 30s memo (`_get_overview_heartbeat_cached`) is thread-safe and keeps `/api/overview` cheap on rapid refreshes.
- Frontend script is inline in the overview template so the card stays self-contained — no app.js coupling.

## Test plan
- [x] `python3 -m pytest tests/test_overview_heartbeat.py -q` — 12 / 12 pass
  - empty store -> `status=None`
  - recent (<1.5x) -> `green`; 50min ago -> `amber`; 4h ago -> `red`
  - ok_ratio math (12/15) and 20-beat cap (25 -> sample_size=20)
  - non-heartbeat events ignored
  - detection via `data.session_type`, `event_type`, `session_id`
  - most-recent beat wins for gap calc
  - end-to-end `/api/overview` JSON contains the `heartbeat` block with all 6 keys
  - 30s cache returns memoised dict on rapid calls
- [ ] Verify card renders green / amber / red against a real OpenClaw workspace before release.

Closes #686

### DIFF
--- a/clawmetry/templates/tabs/overview.html
+++ b/clawmetry/templates/tabs/overview.html
@@ -26,6 +26,29 @@
     <span class="refresh-time" id="refresh-time" style="font-size:11px;">Loading...</span>
   </div>
 
+  <!-- Heartbeat liveness card (closes #686) — pulse indicator + cadence + ok ratio -->
+  <div id="overview-heartbeat-card" style="background:var(--bg-secondary);border:2px solid var(--border-primary);border-radius:12px;padding:14px 18px;margin-bottom:14px;box-shadow:var(--card-shadow);display:none;align-items:center;gap:18px;flex-wrap:wrap;" title="Your agent fires a heartbeat session every 30 min. Green = on time, amber = drifting, red = missed.">
+    <div style="display:flex;align-items:center;gap:10px;min-width:140px;">
+      <span id="overview-hb-dot" style="width:14px;height:14px;border-radius:50%;background:#6b7280;display:inline-block;flex-shrink:0;"></span>
+      <div>
+        <div style="font-size:12px;font-weight:700;color:var(--text-muted);text-transform:uppercase;letter-spacing:1.5px;">&#x2764;&#xfe0f; Heartbeat</div>
+        <div id="overview-hb-status-label" style="font-size:13px;font-weight:600;color:var(--text-primary);">--</div>
+      </div>
+    </div>
+    <div style="display:flex;flex-direction:column;gap:2px;min-width:160px;">
+      <div style="font-size:11px;color:var(--text-muted);text-transform:uppercase;letter-spacing:1px;">Last beat</div>
+      <div id="overview-hb-last-beat" style="font-size:14px;font-weight:600;color:var(--text-primary);">--</div>
+    </div>
+    <div style="display:flex;flex-direction:column;gap:2px;min-width:100px;">
+      <div style="font-size:11px;color:var(--text-muted);text-transform:uppercase;letter-spacing:1px;">Cadence</div>
+      <div id="overview-hb-cadence" style="font-size:14px;font-weight:600;color:var(--text-primary);">30m</div>
+    </div>
+    <div style="display:flex;flex-direction:column;gap:2px;min-width:140px;">
+      <div style="font-size:11px;color:var(--text-muted);text-transform:uppercase;letter-spacing:1px;">OK ratio</div>
+      <div id="overview-hb-ok-ratio" style="font-size:14px;font-weight:600;color:var(--text-primary);">--</div>
+    </div>
+  </div>
+
   <!-- Token Velocity Alert Banner (GH #313) -->
   <div id="velocity-alert-banner" style="
     display:none;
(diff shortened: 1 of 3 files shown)


vivekchand:
## UX review

- 🟡 P1 — Header reads ❤️ Heartbeat / Cadence / Last beat / OK ratio — five labels and a colored dot for "is the agent alive?". A non-technical user does not know what "OK ratio" or "cadence" means. Replace with one sentence + a clock: "Agent is healthy — last check 5m ago." Hide the rest behind a tooltip or expand. (overview.html:30-49)
- 🟡 P1 — The card is inserted as a top-level inline `<div>` with hardcoded `var(--bg-secondary)`, custom padding, custom border-radius, and a custom flex layout — it does NOT match the `autonomy-card` styling the PR description claims. Two months from now this will visually drift from every other card. Replace inline style with the existing card class. (overview.html:30)
- 🟢 nit — `--` as the "no data" state is engineer-speak. Use "Waiting for first heartbeat…" so a brand-new install does not look broken. (overview.html:35,43,47)
- 🟢 nit — The card self-polls /api/overview every 60s independent of the main dashboard refresh. Two timers fetching the same endpoint is a code smell. Wire it through the existing refresh path. (overview.html:485-487)

vivekchand:
## Product review

- 🟡 P1 — Liveness/uptime monitoring of an agent is exactly the surface a paying user would expect when their agent goes dark at 3am. We are giving away the "is it alive" signal in OSS while alerts/Slack/PagerDuty delivery sit behind Cloud-Pro. The conversion hook is: status goes red → CTA "Get notified when this happens — Start 7-day Pro trial". Add CTA at overview.html when `status==='red'` or `status==='amber'`. (overview.html, on amber/red render)
- 🟡 P1 — Two CTAs to add, zero added. A red dot screaming at the user is the highest-intent conversion moment in the whole product. We just put a sad dot there and walked away. Add inline "Set up alert →" link next to the dot.
- 🟢 nit — 30s memo on /api/overview is fine, but you scan 200 events client-side every call. At fleet scale this gets dumb. Move the filter into a SQL WHERE on `event_type` or `data->>'session_type'`. (routes/overview.py:_compute_overview_heartbeat)
- 🟢 nit — "expected_cadence_seconds=1800" is hardcoded. Real users will customise. Read from config eventually, but ship as-is.

Simplest possible thing: one colored dot + one sentence + one CTA when red. Three labels is two too many.

PR #1151, opened 2026-05-13
feat(transcript): show decoding params (T/top_p/max) per LLM call (closes #564)

vivekchand:
## Summary
- Adds an inline "⚙ T=0.7 · top_p=1 · max=4096" pill next to each assistant turn in the transcript replay view so operators can see which sampling config produced each reply (closes #564).
- New `_extract_decoding_params()` helper in `routes/sessions.py` pulls `{temperature, top_p, top_k, max_tokens, stop_sequences}` from three nested shapes (`data.params`, `data.message.params`, `data.config`) plus a flat fallback. Camel-case aliases and `max_completion_tokens` are normalised.
- Both the DuckDB fast-path and the JSONL fallback in `/api/transcript/<sid>` now attach a `params` field to assistant messages when at least one decoding key is captured.

No schema change (parser already stores raw event data in DuckDB's `data` column). No new dependencies. Renders nothing when no params were captured, so older sessions display unchanged.

## Test plan
- [x] `pytest tests/test_decoding_config_extract.py` — 23 cases (3 nested paths, alias normalisation, stop-string promotion, bad-input robustness, T=0 boundary, int coercion). All green.
- [x] `pytest tests/test_transcript_local_store.py` — 4/4 existing transcript tests still green (no regression on the path we touched).
- [x] `node --check clawmetry/static/js/app.js` — clean.
- [ ] Manual: open a session with captured params in the transcript modal, confirm the ⚙ pill renders inline under the assistant message.

### DIFF
--- a/clawmetry/static/js/app.js
+++ b/clawmetry/static/js/app.js
@@ -6434,7 +6434,35 @@ function _buildReplayEvent(m, idx) {
     extra.from_hook = m.from_hook;
     extra.summary_truncated = m.summary_truncated;
   }
-  return { role: role, type: type, content: m.content || '', timestamp: m.timestamp, tokens: m.tokens || null, originalIndex: idx, extra: extra };
+  // Issue #564: decoding params (temperature / top_p / max_tokens / …) come
+  // through verbatim from the backend on assistant turns so the renderer can
+  // show a "⚙ T=0.7 · top_p=1 · max=4096" pill inline with the message.
+  return {
+    role: role,
+    type: type,
+    content: m.content || '',
+    timestamp: m.timestamp,
+    tokens: m.tokens || null,
+    params: m.params || null,
+    originalIndex: idx,
+    extra: extra
+  };
+}
+
+// Format the decoding-params dict into a compact inline pill. Returns ''
+// when nothing useful is set so callers can cheaply skip rendering.
+function _fmtDecodingParams(p) {
+  if (!p || typeof p !== 'object') return '';
+  var parts = [];
+  if (typeof p.temperature === 'number') parts.push('T=' + p.temperature);
+  if (typeof p.top_p === 'number') parts.push('top_p=' + p.top_p);
+  if (typeof p.top_k === 'number') parts.push('top_k=' + p.top_k);
+  if (typeof p.max_tokens === 'number') parts.push('max=' + p.max_tokens);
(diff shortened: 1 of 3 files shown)


vivekchand:
## UX review

- 🟡 P1 — "⚙ T=0.7 · top_p=1 · max=4096" is engineer hieroglyphics. The Help icon is buried in the title attribute and a non-technical user will not hover. Either (a) drop top_p/top_k unless they differ from default, or (b) prefix the pill with "Sampling:" so it is grokable in 5 seconds. (app.js:_fmtDecodingParams)
- 🟢 nit — `font-family: ui-monospace` for a 4-token pill makes it visually compete with the assistant content. Use the same body font; only the values need to be monospace if anything. (app.js _renderReplayEvent decoded div)
- 🟢 nit — When all params are missing the pill silently does not render. Good. But when ONLY `max_tokens=4096` is present you render "⚙ max=4096" which looks like a typo. Hide the gear if there is only one trivial value. (app.js:_fmtDecodingParams)
- 🟢 nit — The pill goes BELOW the tokens line, separate `<div>`. Stack-order will read: text → tokens → params → timestamp. Four lines of metadata per assistant turn bloats the transcript. Inline params next to tokens on one row. (app.js _renderReplayEvent)

vivekchand:
## Product review

- 🟢 nit — This is pure transcript dressing — read-only display of a value the user already provided. No conversion hook, no funnel impact, no leakage. Ship.
- 🟡 P1 — If you DO want to extract money from this surface: when `temperature > 0.8` AND there are visible cost anomalies, surface "High-temperature spending detected — see cost-optimizer (Pro)". Otherwise the pill is just trivia. (routes/sessions.py extraction site)
- 🟢 nit — 100 lines of helper code + 199 lines of tests for a UI pill is over-engineered. The three-shape extraction is real (camelCase aliases are needed), but you do not need separate buckets — a single recursive flatten with a per-key first-wins reducer would be 20 lines. Refactor only if you touch it again. (routes/sessions.py:_extract_decoding_params)
- 🟢 nit — `max_completion_tokens` → `max_tokens` alias is a hidden compatibility tax. Document or it will silently bite when OpenAI changes again.

Simplest possible thing achieved. What didn't make the cut: anomaly-driven CTA. Add later when paid funnel needs feeding.

PR #1152, opened 2026-05-13
feat(budget): per-agent budget limits + tiered 80%/100% alerts (closes #951)

vivekchand:
## Summary
- Per-agent overrides in a new ``agent_budgets`` DuckDB table; either daily / monthly column can be NULL to fall back to the global ``daily_limit`` / ``monthly_limit`` independently.
- Tiered alerts: ``budget.warning`` at >=80%, ``budget.critical`` (+ optional auto-pause when ``auto_pause_enabled``) at >=100%, with same-period dedup keyed by ``(agent_id, period, YYYY-MM-DD|YYYY-MM)`` so slots reset naturally at midnight / month-start.
- New endpoints — ``GET /api/budget`` (collapsed config + agents map), ``GET/PUT/DELETE /api/agents/<id>/budget``.
- New "Per-Agent" tab on the Budget modal with inline add/update/delete and a usage bar (green/amber/red).
- ``tests/test_per_agent_budget.py``: 15 tests covering schema round-trip, HTTP round-trip, 80% warning (no pause), 100% critical + auto-pause, 100% without auto-pause, dedup, warning→critical escalation, and global fallback / override-precedence.

Closes #951.

## Test plan
- [x] ``python3 -m pytest tests/test_per_agent_budget.py -q`` — 15 passed
- [x] ``python3 -m pytest tests/test_alert_evaluator_unit.py tests/test_alert_rules_local_store.py tests/test_circular_import.py -q`` — 40 passed (no regressions in adjacent budget/alert surfaces)
- [x] ``python3 -c "import dashboard"`` clean
- [ ] Manual: open the Budget modal → Per-Agent tab → add an override → verify the row paints with the live MTD usage bar
- [ ] Manual: hit 80% of an agent override → see ``budget.warning`` alert; hit 100% with auto-pause on → see ``budget.critical`` + gateway pause

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1153, opened 2026-05-13
feat(usage): per-session cost breakdown table (closes #68)

### LINKED ISSUE #68: feat: Cost per session breakdown (not just totals)
## Overview
Show per-session cost and token usage in the Sessions tab, not just aggregate totals.

## What to show
- Cost column in sessions list: `$0.043`
- Token breakdown: input / output / cache tokens per session
- Session cost chart: bar chart of top 10 most expensive sessions
- Alert threshold: flag sessions > $X (user-configurable)

## Why
From the community: 'A runaway loop looks normal on a weekly total but obvious when you see 200 API calls in 3 minutes. Look for per-session or per-run breakdowns.' This is the #1 actionable metric for cost control.

## Implementation notes
- Cost data is already in session JSONL (`total_cost`, `total_tokens` fields)
- Add cost column to sessions table (hidden by default on mobile)
- Sort by cost option in session filter
- Multi-provider: Claude, Gemini, OpenAI — per model rate table

vivekchand:
## Summary
- Extends `/api/usage` with a `sessions` array — top-20 sessions sorted by `total_cost_usd` desc, each row `{session_id, agent_id, model, total_tokens, total_cost_usd, message_count, started_at}`.
- Implemented on both the DuckDB fast path (`_ls_top_sessions_by_cost` joins `query_sessions` to a sample event for model) and the legacy transcript-scan path (re-uses the existing `session_summaries`).
- Adds a "Top Sessions by Cost" table to the Tokens/Usage tab template, re-using the existing `usage-table` CSS so it matches the Cost Breakdown table above it.
- `renderTopSessionsByCost` in `app.js` fills the new table from `data.sessions`.

## Why
Issue #68: today `/api/usage` only shows aggregate token/cost totals. A runaway loop disappears in a weekly average but jumps out in a per-session view — this is the #1 actionable metric for cost control.

## Test plan
- [x] `pytest tests/test_usage_per_session.py` — 5 tests pass:
  - sessions array is present under the fast path
  - rows sorted by `total_cost_usd` desc
  - each row carries the contract fields
  - per-session totals match seeded events (5 events × $1 = $5, etc.)
  - top-N caps at 20 even when 25 sessions are seeded
- [ ] Manual UI smoke: load `/page-usage` and confirm the new table renders below "Cost Breakdown" with the expected columns.

### DIFF
--- a/clawmetry/static/js/app.js
+++ b/clawmetry/static/js/app.js
@@ -5768,6 +5768,8 @@ async function loadUsage() {
     tableHtml += '<tr><td>This Month</td><td>' + fmtTokens(data.month) + '</td><td>' + fmtCost(data.monthCost) + '</td></tr>';
     tableHtml += '</tbody>';
     document.getElementById('usage-cost-table').innerHTML = tableHtml;
+    // Issue #68 — per-session cost breakdown table.
+    renderTopSessionsByCost(data.sessions || []);
     // OTLP-specific sections
     var otelExtra = document.getElementById('otel-extra-sections');
     if (data.source === 'otlp') {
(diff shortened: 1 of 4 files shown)


vivekchand:
## UX review

- 🟡 P1 — The table has SEVEN columns (Session / Agent / Model / Tokens / Cost / Msgs / Started) but the `<thead>` only declares six in `colspan="6"`. Render glitch on the empty state. (usage.html top-sessions-table colspan=6 vs 7 columns in JS)
- 🟡 P1 — Section title "🏆 Top Sessions by Cost · which session burned the budget · top 20" reads like four labels stacked. Pick one: either "Top Sessions by Cost" with a tooltip, or "Which session burned the budget?" as the question — not both. (usage.html section-title)
- 🟢 nit — Session ID is rendered as last-16 chars in `<code>`. A non-technical user cannot do anything with `abc123def456ghij`. Make the row clickable to open the transcript modal — that is the actionable next step from "this session was expensive". (app.js renderTopSessionsByCost row build)
- 🟢 nit — `$<0.01` for sub-cent costs is a clever touch but breaks right-align scanning. Use `$0.00` consistently and rely on the column header to convey the unit. (app.js:fmtCost in renderTopSessionsByCost)

vivekchand:
## Product review

- 🟡 P1 — "Top sessions by cost" is the EXACT moment a user goes "oh no, that one session cost me $5". Highest-intent conversion surface in the product and we end the user journey at the table. Add inline CTA on the top row: "Set a budget alert for this agent →" linking to the Per-Agent budget modal (which itself should be Pro — see PR #1152 review). (app.js renderTopSessionsByCost first-row render)
- 🟡 P1 — Top-20 cap is OSS-generous. Cloud-Pro should be "top 200 + 90-day retention + export CSV". OSS cap should be top 5 or top 10. Cap at 5 here; gate 10/20/CSV behind Pro. (routes/usage.py:_ls_top_sessions_by_cost limit=20 and api_usage top_sessions [:20])
- 🟢 nit — Retention strategy: this table is the surface a churned-and-considering-returning user lands on when they pop open the dashboard after a month. Make sure the empty state copy invites them back: "No sessions yet — connect an agent to see cost breakdowns." (app.js empty-state branch)
- 🟢 nit — One DuckDB scan + N+1 query per session for the model column is hot-path embarrassing. `JOIN events ON events.session_id = query_sessions.session_id` once. (routes/usage.py:_ls_top_sessions_by_cost inner loop)

Simplest possible thing: top 5, click-to-transcript, one Pro CTA. Strip the rest.

PR #1156, opened 2026-05-13
feat(workspace): multi-profile discovery + dashboard switcher (closes #950)

### LINKED ISSUE #950: Multi-profile OpenClaw workspace discovery + dashboard switcher
## Summary

Auto-discover all OpenClaw workspaces on the host (e.g., `~/.openclaw`, `~/.openclaw-staging`, `~/.openclaw-prod`, `~/.openclaw-experiments/*`) and let the operator switch between them in the dashboard without restarting ClawMetry.

## Why

People running multiple OpenClaw setups (dev/staging/prod, sandbox profiles, or per-client trees) have to launch a separate ClawMetry instance per workspace today. We already accept a single `--openclaw-dir` (see `clawmetry/cli.py:2079`). One process should cover them all and a workspace switcher should live in the header.

## What to build

1. **Auto-discovery utility** — scan `$HOME` and surface every directory matching `^\.openclaw($|-)` that contains an `agents/` subdir. Always include the standard `~/.openclaw` even if it doesn't exist yet (so the empty-state still works on first launch).
2. **Multi-workspace ingestion** — file watchers + log parsers run per workspace; sessions are tagged with their workspace ID throughout the data model.
3. **Workspace switcher in the dashboard header** — dropdown that filters the entire dashboard (sessions, costs, alerts, mission-control views) to a single workspace OR an "All" aggregate view.
4. **CLI flag** — `--openclaw-dir` becomes repeatable: `--openclaw-dir ~/.openclaw --openclaw-dir ~/.openclaw-staging`. Keep backward compat for the single-value form.
5. **Persistence** — the last-selected workspace is remembered per browser (localStorage) and reflected in the URL.

## Acceptance criteria

- [ ] Running `clawmetry` with no `--openclaw-dir` flag auto-discovers all `~/.openclaw*` workspaces with an `agents/` subdir.
- [ ] Dashboard header shows a workspace dropdown listing every discovered workspace + an "All" option.
- [ ] Selecting a workspace filters sessions, costs, alerts, mission-control and cron views without a page reload.
- [ ] `--openclaw-dir` accepts multiple invocations and overrides auto-discovery when used.
- [ ] An integration test creates two fixture workspaces (`/tmp/.openclaw-fix1`, `/tmp/.openclaw-fix2`) and verifies sessions from both show up correctly tagged.
- [ ] Docs page added: "Running ClawMetry against multiple workspaces."

## Code anchors

- `clawmetry/cli.py:2079` — current single `--openclaw-dir` flag, needs to become repeatable.
- `dashboard.py` `_auto_detect_data_dir` (~line 1742) — current detection logic, factor out a `discover_all_workspaces()` helper.
- `dashboard.py` `_get_sessions()` (~line 12740) — needs workspace tagging.
- `dashboard.py` blueprints/templates rendering the header — add the switcher.

## Out of scope (future work)

- Cross-workspace mission-control aggregation (we'll do per-workspace boards first).
- Workspace-level RBAC / API keys (single-user assumption holds for v1).

vivekchand:
## Summary
- Discover all OpenClaw workspaces on disk — `~/.openclaw`, `~/.openclaw-<name>`, `~/.openclaw/profiles/<name>`, and a user-curated `~/.clawmetry/workspaces.json` — via a new `clawmetry.sync.discover_workspaces()` helper. Safe by design: no symlink-following, paths confined to `$HOME` unless explicitly listed, `PermissionError`-tolerant.
- New endpoints `GET /api/workspaces` and `POST /api/workspaces/active` (in `routes/workspaces.py`). Active-set rebinds `WORKSPACE` / `SESSIONS_DIR` / `MEMORY_DIR` / `LOG_DIR` in `dashboard` and re-inits the DataProvider so subsequent requests read from the new workspace.
- Nav-bar dropdown (top-left, next to version badge). Hidden when only one workspace exists, so zero-config single-workspace UX is unchanged.

## Test plan
- [x] `pytest tests/test_multi_workspace_discovery.py` — 8/8 pass (3-profile case, single-workspace fallback, empty home, symlink rejection, junk-dir guard, curated JSON, agent_count + last_active_ts, sort-by-activity).
- [x] Flask test client smoke: `GET /api/workspaces` → 200, `POST /api/workspaces/active` with bad name → 404.
- [x] `tests/test_circular_import.py` still passes (no import regressions).
- [ ] Manual: spin up a tmpdir with 2 fake workspaces, point dashboard at it, confirm dropdown appears and a click reloads against the other workspace.

Closes #950.

Generated with Claude Code (Opus 4.7, 1M ctx).

### DIFF
--- a/clawmetry/static/js/app.js
+++ b/clawmetry/static/js/app.js
@@ -11259,7 +11259,112 @@ document.addEventListener('DOMContentLoaded', function() {
   initOverviewCompClickHandlers();
   initFlow();
   bootDashboard();
+  // Issue #950: multi-profile workspace switcher
+  try { initWorkspaceSwitcher(); } catch (e) { /* non-fatal */ }
 });
+
+// ── Workspace switcher (issue #950) ───────────────────────────────────
+// Discovers all OpenClaw workspaces on this machine and lets power users
+// flip the dashboard between them. Hidden entirely when only one
+// workspace is found, so the zero-config single-workspace UX is unchanged.
+var _wsList = [];
+var _wsActive = null;
+
+async function initWorkspaceSwitcher() {
+  try {
+    var resp = await fetch('/api/workspaces', { credentials: 'same-origin' });
+    if (!resp.ok) return;
+    var data = await resp.json();
+    _wsList = (data && data.workspaces) || [];
+    _wsActive = data && data.active;
+  } catch (e) {
+    return;
+  }
+  renderWorkspaceSwitcher();
+  // Dismiss menu when clicking outside.
+  document.addEventListener('click', function(ev) {
+    var menu = document.getElementById('workspace-switcher-menu');
+    var btn = document.getElementById('workspace-switcher-btn');
+    if (!menu || menu.style.display === 'none') return;
+    if (menu.contains(ev.target) || (btn && btn.contains(ev.target))) return;
+    menu.style.display = 'none';
+  });
+}
+
+function renderWorkspaceSwitcher() {
+  var wrap = document.getElementById('workspace-switcher');
+  if (!wrap) return;
+  if (!_wsList || _wsList.length <= 1) {
+    wrap.style.display = 'none';
+    return;
+  }
+  wrap.style.display = 'inline-block';
+  var label = document.getElementById('workspace-switcher-label');
+  var activeEntry = _wsList.find(function(w) { return w.path === _wsActive; });
+  if (label) label.textContent = (activeEntry && activeEntry.name) || 'default';
+  var menu = document.getElementById('workspace-switcher-menu');
+  if (!menu) return;
+  menu.innerHTML = '';
+  _wsList.forEach(function(w) {
+    var item = document.createElement('div');
+    var isActive = w.path === _wsActive;
+    item.style.cssText = 'padding:8px 10px;border-radius:6px;cursor:pointer;font-size:12px;display:flex;flex-direction:column;gap:2px;' +
+      (isActive ? 'background:rgba(229,68,58,0.12);' : '');
+    item.onmouseover = function() { if (!isActive) item.style.background = 'rgba(255,255,255,0.06)'; };
+    item.onmouseout = function() { if (!isActive) item.style.background = 'transparent'; };
+    item.onclick = function() { selectWorkspace(w); };
+    var name = document.createElement('div');
+    name.style.cssText = 'font-weight:600;color:var(--text-primary,#fff);';
+    name.textContent = (isActive ? '• ' : '') + (w.name || 'workspace');
+    var path = document.createElement('div');
+    path.style.cssText = 'font-size:10px;color:var(--text-muted,#888);font-family:ui-monospace,monospace;overflow:hidden;text-overflow:ellipsis;white-space:nowrap;';
+    path.title = w.path;
+    path.textContent = w.path;
+    var meta = document.createElement('div');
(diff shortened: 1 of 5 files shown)


vivekchand:
## UX review

- 🔴 P0 — The nav now has THREE icon-only buttons (workspace switcher, gateway gear, alerts bell) plus version badge, and the workspace switcher renders a folder icon plus the word "default" plus a chevron — a 4-element control that is visually heavier than every other nav widget. Hidden when 1 workspace, but for the 2-workspace edge case (which is the entire feature's audience) it dominates the nav. Action: shrink to icon-only with a tooltip, mirror the gear/bell visual weight. `dashboard.py:3300-3308`.
- 🟡 P1 — On click, the dashboard does `window.location.reload()` with no toast/spinner. To a non-technical user, a full reload looks like the dashboard crashed. Action: show "Switching to <name>…" spinner overlay for ~500ms before reload. `app.js:11363-11366`.
- 🟡 P1 — Dropdown menu shows full filesystem paths (`/Users/vivek/.openclaw-personal`) which leak the user's home directory and look like dev output. Non-technical users don't recognise these. Action: show only the workspace name + agent count; tuck the path into a tooltip. `app.js:11308-11318`.
- 🟡 P1 — Dropdown is hidden when only 1 workspace is found, which is correct, but there is no entry point to ADD a workspace ("Add another workspace…" footer item). Users with one workspace will never discover this is multi-profile-capable. Action: when `_wsList.length === 1`, show a tiny "+" affordance that opens docs on `~/.clawmetry/workspaces.json`. `app.js:11270-11272`.
- 🟢 nit — Two identical nav HTML blocks at `dashboard.py:3300` and `dashboard.py:8997` — the dual-render base template duplication risk that hit the "Alerts" tab. Action: extract into a Jinja partial or at minimum add a `<!-- WORKSPACE SWITCHER (DO NOT DUPLICATE) -->` marker.

vivekchand:
## Product review

- 🔴 P0 — This is single-machine multi-profile (one OS-user, several `~/.openclaw-*` dirs on the same disk). It is NOT multi-node fleet management — but the marketing-grade UI affordance ("Switch workspace") is going to be confused with our paid multi-node feature. Action: rename UI label from "Switch OpenClaw workspace" to "Switch profile (this machine)" and keep `Fleet` (multi-node) clearly separated. `dashboard.py:3303`.
- 🟡 P1 — Power-user feature on the OSS funnel surface (top nav). Adoption is low; conversion lift is zero. The cost is a permanent nav element + dual-template duplication risk. Action: gate behind `?profiles=1` URL flag or a settings toggle, OR move into the gear menu rather than nav. Don't permanently rent prime nav real estate to a power feature.
- 🟡 P1 — `~/.clawmetry/workspaces.json` requires hand-editing JSON to add a remote/external path — exactly the YAML-config pattern the user explicitly rejected for non-technical users. Action: add a `POST /api/workspaces` to register a new path via UI (browse picker), keep the JSON file as the persisted backing store but not the user-facing primitive.
- 🟡 P1 — Active workspace persisted to `~/.clawmetry/last_workspace` (a plain text file) instead of the local DuckDB store, violating the DuckDB-first rule. Action: persist active-workspace in a `node_meta` or `dashboard_state` table. `routes/workspaces.py:430-441`.
- 🟢 nit — Simplest possible thing would have been a CLI flag `clawmetry --workspace ~/.openclaw-work` (which already exists). The dropdown is a UI duplicate of an existing capability. If kept, ensure analytics tag this so we can measure whether anyone actually clicks it; if usage is <2% in 30 days, delete the nav widget and keep the API.

PR #1157, opened 2026-05-13
feat(flow): historical flow-runs dashboard (closes #611)

### LINKED ISSUE #611: [P2] Flow-runs dashboard
Mirror of vivekchand/clawmetry-cloud#324 — implement OSS-first, UI-SYNC ports to cloud.

## Problem
`~/.openclaw/flows/registry.sqlite` has `flow_runs(flow_id, status, current_step, blocked_task_id, blocked_summary, state_json, wait_json, …)` — first-class orchestrated multi-step workflows. Zero visualisation in ClawMetry today.

## OSS UI
- New top-level "Flows" tab
- List of flow runs: goal, status, current_step, duration, owner
- Per-flow detail: step-by-step progress, blocked_summary, state_json dump, linked tasks (via `tasks.parent_flow_id`)
- Retry / cancel controls (future — requires gateway RPC)

## Effort
M (2-3 days)

vivekchand:
## Summary
- Adds a **Runs** sub-tab on the Flow page showing historical flow runs aggregated from the local DuckDB `events` table.
- New backend endpoint `GET /api/flow/runs?limit=30` returns per-session summaries: `started_at`, `duration_seconds`, `channels_touched`, `models_invoked`, `tools_called`, `total_cost`, `status` (completed/failed), plus model list and channel.
- Live Flow view is untouched — it remains the default sub-tab and the source of truth for the animated SVG.

## Implementation
- `LocalStore.query_flow_runs(...)` — single `GROUP BY session_id` query with a `LEFT JOIN` onto `openclaw_channels` so each row carries the user-facing channel without an extra round trip. Status is heuristic: any event with `event_type` containing `error` flips the run to `failed`.
- `bp_logs` (`routes/infra.py`) gets the new route next to the existing `/api/flow` SSE — they're siblings of the same Flow surface.
- `page-flow` in `dashboard.py` gets a sub-tab toggle and a runs table; `app.js` gets `switchFlowSubtab`, `loadFlowRuns`, `showFlowRunDetail`.
- Endpoint never raises — empty/missing store degrades to `_source: empty` so the tab stays usable on fresh installs.

## Test plan
- [x] `pytest tests/test_flow_runs_endpoint.py` — 6 cases, all green:
  - per-session grouping (3 sessions seeded, 3 rows returned)
  - cost / tool-call / model aggregations
  - failed-status detection on `tool_error` events
  - most-recent-first ordering
  - `limit` param clamping
  - empty store returns `_source: empty`
- [x] `python3 -c "import dashboard"` — module imports clean
- [x] Adjacent `test_local_store_multi_agent.py` + `test_event_metrics_extraction.py` still green (22/22 in the affected slice)

Closes #611

### DIFF
--- a/clawmetry/local_store.py
+++ b/clawmetry/local_store.py
@@ -44,6 +44,7 @@ import threading
 import time
 from collections import deque
 from contextlib import contextmanager
+from datetime import datetime
 from pathlib import Path
 from typing import Any, Iterable, Iterator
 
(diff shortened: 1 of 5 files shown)


vivekchand:
## UX review

- 🟡 P1 — "Live | Runs" sub-tab labels are vague. A non-technical user doesn't know a "run" is a session. Action: rename to "Live" and "History" (or "Past sessions"). `dashboard.py:4009-4012`.
- 🟡 P1 — Run row click opens an inline detail panel that re-fetches `/api/flow/runs?limit=200` just to find one row. On a populous install this is 200× more bytes than needed for a single-run view. Action: add `GET /api/flow/runs/<session_id>` or pass the row data already on screen into `showFlowRunDetail`. `app.js:7460-7472`.
- 🟡 P1 — Status column shows green "completed" or red "failed" with no in-progress state. Live Flow is the source of truth, but a user looking at history will assume every row is finalized. A session that crashed mid-tick with no `tool_error` event will appear "completed". Action: add a 3rd status ("interrupted" / "incomplete") when `updated_at` is very recent and there's no end-of-session marker. `local_store.py:1419-1421`.
- 🟢 nit — Empty-state copy is informative ("No historical flow runs yet — the live Flow view will populate this once a session completes.") This is good. Keep it.
- 🟢 nit — Cost formatter `_fmtRunCost(0)` → `"$0"` but `0.005` → `"<$0.01"`. Inconsistent shape. Action: render `"$0.00"` for zero so all rows align. `app.js:7390-7395`.

vivekchand:
## Product review

- 🔴 P0 — Memory `project_alerts_pro_feature.md`: ">24h retention is Cloud-Pro". This OSS endpoint accepts `since`/`until` ISO params with NO upper bound on how far back it reads (`local_store.query_flow_runs` — no retention guard in `routes/infra.py:312-360`). DuckDB retains forever locally. Result: OSS users get unlimited historical session retention for free, torching the Cloud-Pro retention upsell. Action: cap OSS reads to last 24h of `started_at`; surface a "Want >24h? Upgrade to Cloud-Pro" CTA below the table when the limit is hit. `routes/infra.py:340-358`.
- 🟡 P1 — Detail panel shows "View full transcript →" linking to `/api/transcript/<sid>` (the raw JSON endpoint). That's a dev-grade link, not a user-grade one. Action: link to the existing Sessions transcript UI instead of the raw JSON. `app.js:7494`.
- 🟡 P1 — Simplest possible thing: this is a duplicate of the existing Sessions tab (which already shows session list + duration + cost + transcript link). The Flow > Runs sub-tab adds "channel" and "tool count" columns that Sessions doesn't. Either fold these columns into Sessions and delete the Runs tab, or document why Flow needs its own list. Tabs that show "the same thing in a slightly different way" are a known consolidation target.
- 🟡 P1 — Status heuristic ("any `%error%` event → failed") will mark normal recoverable tool errors as run failures. A skilled agent that hit a `tool_error` and retried successfully gets a red badge. Action: tighten to terminal/uncaught errors only, or rename column to "Had errors" + count. `local_store.py:1419-1421`.
- 🟢 nit — Add a "Re-run this flow" button on the detail panel as a Cloud-Pro CTA (gate behind upgrade modal for OSS, real action for Pro). Free historical data has zero conversion value unless we attach a paid action to it.

PR #1159, opened 2026-05-13
feat(daemon): persist errors as DuckDB events; health card reads DuckDB (DuckDB-first for #1133 layer 4)

vivekchand:
## Why

PR #1139 surfaced daemon errors on the System Health card by **parsing
`~/.clawmetry/sync.log` text on every `/api/system-health` request** — a
clear violation of the DuckDB-first rule
(`feedback_duckdb_first_rule.md`). The daemon ALREADY logs the errors;
it just doesn't write structured rows. This follow-up wires the
DuckDB-first canonical path so the read side does one indexed query
instead of re-tailing the log on each call.

## What

**Daemon side (`clawmetry/sync.py`)**
- New `_DaemonErrorDuckDBHandler` (subclass of `logging.Handler`,
  level=`ERROR`) tees every ERROR record into the local DuckDB
  `events` table as:
  ```python
  {
    "id": uuid4().hex,
    "agent_id": "clawmetry-daemon",
    "agent_type": "clawmetry",
    "event_type": "daemon.error",
    "ts": iso_now(),
    "data": {"message": <msg>, "exception": <traceback_or_None>, "logger": ...},
  }
  ```
- **Rate-limit**: at most one row per (`message[:80]`, 60-second-bucket).
  The original `ALERTS_EVAL_INTERVAL_SEC` NameError fired 4×/min; without
  this cap that pattern would stamp ~5,760 rows/day for a single
  regression. Dedup map is in-memory, bounded-eviction at 1024 entries.
- Handler is installed via `install_daemon_error_event_handler()` from
  `run_daemon()` right after the start banner. Idempotent.
- Failure modes (ingest raises, DuckDB locked, etc.) are swallowed and
  counted on the handler — the daemon must never crash because
  telemetry plumbing broke.

**Read side (`routes/health.py`)**
- `compute_daemon_health()` is now DuckDB-first:
  1. `_try_local_store_daemon_health()` queries
     `query_events(event_type='daemon.error', agent_id='clawmetry-daemon', since=now-1h)`
     via `routes.local_query.local_store_via_daemon` (cross-process)
     with a single-process fallback to `local_store.get_store()` (tests +
     dev mode).
  2. If DuckDB returns ≥1 row, aggregate `errors_last_5min` /
     `errors_last_1h` / `last_error_message` / `last_error_ts` /
     `status` and tag the response `_source='local_store'`.
  3. Otherwise fall through to the legacy `_tail_lines(sync.log)` path
     (now `_compute_daemon_health_from_log()`, tagged `_source='sync_log'`).
     This handles fresh installs + pre-this-PR data — the user still
     sees the silent NameError on a clean machine.
- Same 5-min / 1-hour windows and `>30 → broken`, `>0 → degraded`,
  else `healthy` thresholds. Response shape unchanged except for the
  new `_source` field.

## Tests

`tests/test_daemon_error_event_pipeline.py` — **14 pure-unit tests**, no
server / network / sleep:

| Class | Test | What |
|---|---|---|
| TestHandler | install_is_idempotent | second install returns same instance |
| TestHandler | info_and_warning_records_are_ignored | only ERROR ingests |
| TestHandler | error_writes_one_row | shape: agent_id / agent_type / event_type / data.message |
| TestHandler | exception_info_is_captured | `log.exception()` → traceback in `data.exception` |
| TestRateLimit | 30_identical_errors_in_one_bucket_writes_one_row | the headline contract |
| TestRateLimit | distinct_messages_each_get_a_row | only IDENTICAL prefixes dedup |
| TestRateLimit | same_prefix_next_bucket_writes_a_new_row | manual time advance |
| TestRateLimit | dedup_uses_first_80_chars | suffix-noise doesn't bust the cap |
| TestPipeline | three_distinct_errors_land_in_duckdb | E2E `log.error()` → DuckDB |
| TestPipeline | handler_swallows_local_store_exceptions | DuckDB outage ≠ daemon crash |
| TestComputeDaemonHealth | reads_counts_from_duckdb | `_source='local_store'` + counts |
| TestComputeDaemonHealth | zero_rows_falls_back_to_log_tail | fresh-install path |
| TestComputeDaemonHealth | 50_recent_errors_is_broken_via_duckdb | status threshold via DuckDB |
| TestComputeDaemonHealth | only_old_errors_is_healthy_5min_but_degraded_1h | window arithmetic |

## Test plan

- [x] `python3 -m pytest tests/test_daemon_error_event_pipeline.py -q` → 14 passed in 0.59s
- [x] `python3 -m pytest tests/test_daemon_health_endpoint.py -q` → 20 passed (PR #1139 contract preserved)
- [x] `python3 -m pytest tests/test_health_local_store.py tests/test_local_store*.py -q` → 79 passed (no regressions)
- [x] `python3 -c "import ast; ast.parse(open('clawmetry/sync.py').read()); ast.parse(open('routes/health.py').read())"` → OK
- [x] `python3 -c "from clawmetry import sync; sync.install_daemon_error_event_handler"` → callable
- [ ] Manual: run daemon, kill it with a synthetic NameError, verify `daemon.error` rows appear in `~/.clawmetry/clawmetry.duckdb`
- [ ] Manual: confirm dashboard `/api/system-health` returns `daemon._source='local_store'` after the daemon writes ≥1 row

## References

- PRD #1133 — daemon-error visibility, layer 4
- PR #1139 — the read side this PR makes DuckDB-first
- Memory: `feedback_duckdb_first_rule.md` — every read surface must
  hit DuckDB, not raw files

### DIFF
--- a/clawmetry/sync.py
+++ b/clawmetry/sync.py
@@ -186,6 +186,152 @@ if not log.handlers:
     log.propagate = False
 
 
+# ── Daemon-error → DuckDB event handler (PRD #1133 layer 4, daemon side) ────
+#
+# Why: PR #1139 surfaced daemon errors on the System Health card by parsing
+# ``~/.clawmetry/sync.log`` text on every /api/system-health call — a clear
+# violation of the DuckDB-first rule (memory feedback_duckdb_first_rule.md).
+# This handler tees every ERROR-level log line into a structured
+# ``daemon.error`` event row in the local DuckDB so the read side can do a
+# single indexed query instead of re-tailing the log on each request.
+#
+# Rate-limit: at most one row per (first-80-chars-of-message, 60s bucket).
+# Important because the original ALERTS_EVAL_INTERVAL_SEC NameError fired
+# 4×/min on every install, and we don't want that pattern to spam DuckDB
+# 5,760 rows/day when one row/minute carries the same signal.
+#
+# Failure mode: any exception inside the handler is swallowed and counted —
+# the daemon must never crash because telemetry plumbing broke.
+
+import uuid as _uuid
+import socket as _socket
+
+_DAEMON_ERROR_AGENT_ID = "clawmetry-daemon"
+_DAEMON_ERROR_AGENT_TYPE = "clawmetry"
+_DAEMON_ERROR_EVENT_TYPE = "daemon.error"
+_DAEMON_ERROR_DEDUP_PREFIX_LEN = 80
+_DAEMON_ERROR_DEDUP_BUCKET_SEC = 60
+
+
+class _DaemonErrorDuckDBHandler(logging.Handler):
+    """Logging handler that mirrors ERROR records into DuckDB ``events``.
+
+    One row per (message-prefix, 60-second-bucket) — repeated identical
+    errors within the bucket are dropped so a 4×/min NameError doesn't
+    flood the table. The cap is enforced in-memory; we don't query
+    DuckDB to dedup (would defeat the point).
+    """
+
+    def __init__(self) -> None:
+        super().__init__(level=logging.ERROR)
+        # Map[prefix → last-emitted-bucket]. Single-process daemon, so a
+        # plain dict guarded by a Lock is fine. Bounded eviction below
+        # keeps memory flat under pathological churn.
+        self._last_emit: dict[str, int] = {}
+        self._lock = threading.Lock()
+        self._dropped = 0
+        self._emitted = 0
+        self._node_id_cache: str | None = None
+
+    def _node_id(self) -> str:
+        if self._node_id_cache:
+            return self._node_id_cache
+        try:
+            cfg = load_config()
+            nid = cfg.get("node_id") or _socket.gethostname() or "unknown"
+        except Exception:
+            nid = _socket.gethostname() or "unknown"
+        self._node_id_cache = str(nid)
+        return self._node_id_cache
+
+    def _should_emit(self, msg: str, now_ts: float) -> bool:
+        """Return True iff this prefix hasn't been emitted in the current
+        60-second bucket. Updates the bookkeeping atomically."""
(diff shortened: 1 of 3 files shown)


vivekchand:
## UX review

Diff is backend-only — DuckDB schema/handler in `clawmetry/sync.py`, read-path swap in `routes/health.py`. The user-visible System Health card response shape is unchanged except for an added `_source` field. No user-facing UI changed; PR correctly improves the existing card's data freshness without a new surface. Stop.

vivekchand:
## Product review

- 🟡 P1 — Correct DuckDB-first migration with a sound rate-limit (1 row / 60s bucket per 80-char prefix). The simplest version of this is exactly what shipped. No OSS-leakage or tier-strategy risk — daemon errors are an OSS infra signal, not a paid feature.
- 🟡 P1 — Conversion angle missed: when `errors_last_1h > N`, the daemon now has structured data to drive a "Your local daemon is unhealthy — Cloud sync would survive this restart" CTA. Action: surface a one-time toast / banner when `errors_last_1h > 10` linking to Cloud-Pro sync. Every error event is a reliability anxiety moment = high-intent for paid backup.
- 🟢 nit — Status thresholds (>30 → broken, >0 → degraded) are reasonable but unmeasured. Action: log a Prometheus-style counter so we can tune the threshold from real distributions instead of guessing.
- 🟢 nit — Fallback path (`_compute_daemon_health_from_log`) is retained for pre-this-PR data. Once one full daemon restart cycle has passed in the wild, this fallback is dead weight. File a follow-up to delete after 2 releases.
- 🟢 nit — The handler's in-memory `_last_emit` dict is per-process; if the daemon restarts (which is exactly when these errors fire), dedup window resets. Acceptable for now since DuckDB is the long-term source of truth, but worth a comment: a daemon in a crash loop will write 1 row per restart, not 1 row per 60s.

PR #1147, opened 2026-05-13
feat(crons): per-job run timeline (closes #605)

vivekchand:
## Summary
- New `GET /api/crons/<jobId>/runs?limit=30` reads OpenClaw's per-run JSONL log at `~/.openclaw/cron/runs/<jobId>.jsonl`, returns last N runs most-recent first. Missing file → 200 empty, malformed lines skipped, limit capped at 100. Normalises camelCase ↔ snake_case + ISO ↔ ms fields.
- Cron expand panel in `app.js` now shows: 240x36 SVG sparkline (duration on Y, status as colored dot), stat row (success %, MTBF, avg tokens/run), next-run-at, and a compact 30-row recent-runs table (scrollable, error truncated to 200 chars). Legacy gateway-derived render stays underneath so calendar heatmap keeps working when the JSONL feed is empty.
- 7 unit tests covering missing file, sort order, default/cap limits, malformed-line recovery, error truncation, path-traversal rejection, and camelCase field acceptance.

## Test plan
- [x] `pytest tests/test_crons_runs_endpoint.py` — 7/7 pass
- [x] `pytest tests/test_crons_local_store.py` — 11/11 pass (no regression)
- [x] `node -c clawmetry/static/js/app.js` — JS valid
- [ ] Manual: open Crons tab, click a job, verify sparkline + table render when `~/.openclaw/cron/runs/<id>.jsonl` exists
- [ ] Manual: confirm fresh install with no runs file still shows "No run history yet" cleanly

Closes #605

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #605, opened 2026-04-14
[P1] Cron run timeline with per-run usage + failure sparkline


vivekchand:
Mirror of vivekchand/clawmetry-cloud#319 — implement OSS-first, UI-SYNC ports to cloud.

## Problem
OpenClaw writes per-run cron logs to `~/.openclaw/cron/runs/{jobId}.jsonl` with rich metadata per run (duration, status, error, usage, deliveryStatus). Today's Crons tab shows just the job list.

## Proposed OSS UI
Per-cron detail:
- Sparkline of last 30 runs (duration, status colour)
- Failure rate + MTBF
- Token usage trend
- Table of recent runs with timestamp / duration / status / error (truncated) / delivered?
- Next run-at from `nextRunAtMs`

## Data
Local disk: `~/.openclaw/cron/runs/*.jsonl` — each line already has everything needed.

## Effort
S-M (1 day)

vivekchand:
Hi @vivekchand! Took a careful look at this one. Here's the plan:

**The problem**
The cron expand panel shows "Loading run history…" forever because the backend returns `timestamp` (epoch ms) and `sessionId`, but `loadCronRuns()` in `app.js` expects `startedAt` / `ts` (ISO string) and `sessionFile` — so every run is silently dropped by the date renderer. Additionally, the rich per-run JSONL files at `~/.openclaw/cron/runs/{jobId}.jsonl` are never read; the fallback derives synthetic runs from session transcripts instead.

**Approach**
1. `dashboard.py:9192–9204` — patch `_cron_runs_from_transcripts()` to emit `startedAt` (ISO-8601 from epoch), `ts` (same), and `sessionFile` (`{sessionId}.jsonl`) alongside the existing `timestamp` / `sessionId` fields so the frontend calendar heatmap and run table can render them.
2. `dashboard.py` (after `_cron_runs_from_transcripts`) — add `_cron_runs_from_jsonl(job_id)`: reads `~/.openclaw/cron/runs/{job_id}.jsonl` line-by-line (safe basename, graceful on missing file), sorts newest-first, returns up to 50 records. This gives real per-run data with actual `status`, `error`, `durationMs`, `costUsd`, and `deliveryStatus` fields.
3. `routes/crons.py:209–227` — insert the JSONL reader as fallback tier 2 (after gateway RPC succeeds with records, before the transcript-derived synthetic path). Order becomes: gateway → JSONL file → transcript analytics.

**Test plan**
[...]

PR #1160, opened 2026-05-13
feat(crons): ingest cron-run jsonl into DuckDB; API reads from DuckDB (DuckDB-first rule for #605 followup)

vivekchand:
## Summary

Follow-up to #1147. The DuckDB-first rule says every feature must persist + read through local DuckDB; PR #1147's `/api/crons/<jobId>/runs` parses `~/.openclaw/cron/runs/<jobId>.jsonl` on every request, which violates that.

This PR keeps #1147's user-facing endpoint, sparkline, and stat row exactly as-is, and adds the DuckDB ingest + read path underneath. The JSONL parser stays as a graceful-migration fallback for the first cycle of a fresh install.

> Stacked on top of #1147. Once #1147 merges, the diff here collapses to the additive commit `14af2cf`. If #1147 merges first this branch already rebases cleanly onto main.

## What's new

- **`clawmetry/local_store.py`** — new `cron_runs` table (dedicated, not `events`), `LocalStore.ingest_cron_run`, `LocalStore.query_cron_runs(job_id, since, until, limit)`.
- **`clawmetry/sync.py`** — `sync_cron_runs(config, state, paths)`. Scans every OpenClaw-candidate `cron/runs` dir, tails each `*.jsonl` from a per-file byte offset under `state["cron_run_offsets"]`, ingests via `ingest_cron_run`. Truncation + rotation guard. Wired into the startup-sync block **and** the main loop in `run_daemon`.
- **`routes/crons.py`** — `api_crons_job_runs` now reads DuckDB via `_ls_call("query_cron_runs", ...)` first, falls through to the JSONL parser only when DuckDB has zero rows. Response shape unchanged so #1147's frontend keeps working.
- **`routes/local_query.py`** — `query_cron_runs` added to the daemon-proxy method allowlist.
- **`tests/test_cron_runs_duckdb_ingest.py`** — 11 tests (ingest, idempotency, query, API fallback, path traversal).

## Schema decision

Went with a **dedicated `cron_runs` table** (not `events` with `event_type LIKE 'cron.run.%'`). The timeline UI filters by `job_id` and sorts by `started_at`; a prefix-indexed standalone table beats a substring scan over a `data` BLOB. First-class columns for `delivered_at` / `next_run_at` / `duration_ms` also let `query_cron_runs` skip blob deserialisation on the hot path.

## Test plan

- [x] `pytest tests/test_cron_runs_duckdb_ingest.py` — 11/11 pass
- [x] `pytest tests/test_crons_runs_endpoint.py tests/test_crons_local_store.py tests/test_local_store.py tests/test_local_store_sync_integration.py` — 63/63 pass (adjacent suites unaffected)
- [x] `pytest tests/test_local_store_default_on.py tests/test_local_store_missing_writers.py tests/test_local_store_multi_agent.py tests/test_local_store_rename.py` — 49/49 pass
- [x] AST parse-check all modified files

## Notes

- No `[RELEASE]` prefix per the task spec.
- Stacked on #1147. If #1147 merges first the diff for this PR is just commit `14af2cf` — additive only.

### DIFF
--- a/clawmetry/local_store.py
+++ b/clawmetry/local_store.py
@@ -385,6 +385,42 @@ _DDL = [
         updated_at        BIGINT NOT NULL
     )
     """,
+    # Issue #605 follow-up (DuckDB-first rule) — cron-run timeline storage.
+    # One row per JSONL line in ``~/.openclaw/cron/runs/<jobId>.jsonl``. The
+    # daemon's ``sync_cron_runs`` helper scans those files every cycle and
+    # upserts rows here so ``/api/crons/<jobId>/runs`` can read from the
+    # columnar store instead of re-parsing JSONL on every request.
+    #
+    # Schema rationale: we keep a dedicated table (not ``events``) because
+    # ``delivered_at`` / ``next_run_at`` are first-class columns and the
+    # cron-timeline UI's filter + sort patterns
+    # (``WHERE job_id=? ORDER BY started_at DESC``) want a primary-key prefix
+    # match, not a substring scan over ``data`` blobs. ``id`` is the dedup
+    # key — synthesised from ``job_id + ts`` when the JSONL doesn't include
+    # one, so re-ingestion of the same line is a no-op.
+    """
+    CREATE TABLE IF NOT EXISTS cron_runs (
+        id              VARCHAR PRIMARY KEY,
+        node_id         VARCHAR,
+        agent_type      VARCHAR NOT NULL DEFAULT 'openclaw',
+        agent_id        VARCHAR NOT NULL DEFAULT 'main',
+        job_id          VARCHAR NOT NULL,
+        started_at      VARCHAR,
+        ended_at        VARCHAR,
+        duration_ms     INTEGER,
+        status          VARCHAR,
+        error_message   VARCHAR,
+        token_count     INTEGER,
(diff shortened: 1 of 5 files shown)


vivekchand:
## UX review

- Backend-only diff (DuckDB ingest + read-path swap behind an unchanged `/api/crons/<jobId>/runs` endpoint and unchanged sparkline). No user-visible UI surface to review. Response shape is preserved per PR description, so the cron timeline UI behaves identically — `routes/crons.py:801-832` keeps the same `{jobId, runs, count, source, file}` envelope.

vivekchand:
## Product review

- Pure infra: moves cron-run reads off per-request JSONL parsing and onto the columnar store. Aligns with the DuckDB-first hard rule, and the dedicated `cron_runs` table (vs. cramming into `events`) is the right call for the prefix-indexed `WHERE job_id=? ORDER BY started_at` access pattern. No tier-strategy concerns — cron-run timelines are already part of the OSS surface. Ship.
- Nit: `source` field always returns `"duckdb"` even when the JSONL fallback served the rows (`routes/crons.py:835`). Harmless today but if you ever add an "is the migration complete?" telemetry sweep you'll want `"jsonl-fallback"` distinguishable. Action: thread the actual source through the fallback branch.

PR #1161, opened 2026-05-13
feat(gateway): persist gateway.metric events; 24h sparkline on health card (DuckDB-first for #852 followup)

vivekchand:
## Summary

Follow-up to PR #1146. #1146 surfaces the CURRENT gateway snapshot (PID, RSS, CPU, uptime) via psutil/ps. This PR persists those snapshots into DuckDB so the gateway-health card can render a **24h memory-pressure sparkline** — the slow ~600 MB → ~945 MB OOM bloat is only obvious from the trend, not a single sample.

Per the MOAT/DuckDB-first rule: historical observability data belongs in DuckDB, not in repeated /proc reads on every dashboard tick.

**Stacked on #1146** — this PR contains #1146's commit at the base; please merge #1146 first. No \`[RELEASE]\` prefix.

### Changes

- \`clawmetry/sync.py\` — new \`capture_gateway_metric(config)\` writes one \`event_type=\"gateway.metric\"\` event per daemon tick. Rate-capped (30s), deduped (±5 MB RSS / ±2% CPU within 5 min). Wired into the main loop next to \`sync_system_snapshot\`.
- \`routes/health.py\` — new \`GET /api/gateway-health/history?hours=24\` returns \`[{ts, rss_mb, cpu_pct}, ...]\` sorted ASC. Empty list (not 4xx) when no events have been captured yet. Existing \`/api/gateway-health\` live snapshot endpoint is unchanged. \`/api/system-health.gateway\` now also carries a \`history: {count, min_rss_mb, max_rss_mb, avg_rss_mb}\` summary for the last hour.
- \`clawmetry/static/js/app.js\` — 240×36 inline SVG sparkline (polyline + filled area + latest-point dot, same pattern as PR #1147's cron-run sparkline). Fetches the history endpoint on card mount; quiet "no history yet" fallback for fresh installs.
- \`tests/test_gateway_metric_history.py\` — 14 cases covering daemon capture, dedupe logic, endpoint contract (ASC sort, empty-data, hours window, invalid params), live-snapshot regression, and summary helper.

## Test plan

- [x] \`pytest tests/test_gateway_metric_history.py\` (14/14 passing locally)
- [x] \`pytest tests/test_gateway_health.py\` regression (35/35 passing — PR #1146 surface untouched)
- [x] AST parse-check on edited Python + JS files
- [ ] Browser E2E: confirm sparkline renders on gateway-health card after daemon writes >2 samples
- [ ] Daemon e2e: leave \`clawmetry sync\` running for >1 min, then \`SELECT count(*) FROM events WHERE event_type='gateway.metric'\` should be ≥1
- [ ] Confirm \`/api/gateway-health/history\` returns empty array (not 500) on a fresh install with no daemon history

### DIFF
--- a/clawmetry/static/js/app.js
+++ b/clawmetry/static/js/app.js
@@ -5395,6 +5395,72 @@ function startHealthStream() {
   healthStream.onerror = function() { setTimeout(startHealthStream, 30000); };
 }
 
+// Gateway-health 24h sparkline (#852 followup, MOAT/DuckDB-first).
+// Pulls daemon-captured `gateway.metric` events from /api/gateway-health/history
+// and renders them as a 240x36 inline SVG inside #sh-gateway-spark. Empty
+// data (fresh install: no daemon yet, or <30s elapsed) renders as a quiet
+// "no history yet" hint so the card never looks broken.
+async function _loadGatewayHealthSparkline() {
+  var el = document.getElementById('sh-gateway-spark');
+  if (!el) return;
+  var rows;
+  try {
+    rows = await fetchJsonWithTimeout('/api/gateway-health/history?hours=24', 6000);
+  } catch (_e) {
+    rows = [];
+  }
+  if (!Array.isArray(rows) || rows.length === 0) {
+    el.innerHTML = '<div style="font-size:10px;color:var(--text-muted);font-style:italic;opacity:0.7;">Sparkline available after daemon captures 24h of samples.</div>';
+    return;
+  }
+  // Only keep rows with a numeric rss_mb — daemon may have written
+  // gateway.metric rows where vitals weren't readable (status=warning,
+  // rss_mb=null). Those break the y-axis; skip them.
+  var pts = rows.filter(function(r) {
+    return r && typeof r.rss_mb === 'number';
+  });
+  if (pts.length === 0) {
+    el.innerHTML = '<div style="font-size:10px;color:var(--text-muted);font-style:italic;opacity:0.7;">Sparkline available after daemon captures 24h of samples.</div>';
+    return;
+  }
+  var W = 240, H = 36, pad = 2;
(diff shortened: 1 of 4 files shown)


vivekchand:
## UX review

- 🟢 nit: empty-state copy "Sparkline available after daemon captures 24h of samples" overstates the wait — the dedupe logic + 30s capture interval means a few minutes of daemon uptime gives a usable trend. Action: change to "Trend will appear once the daemon captures a few samples." (`clawmetry/static/js/app.js:5410` and :5417).
- 🟢 nit: caption reads `<min> MB · <n> samples · 24h · <max> MB` even when the user requested a shorter window or the data only spans 1h. Action: derive the window label from `pts[last].ts - pts[0].ts` instead of hardcoding "24h" (`clawmetry/static/js/app.js:5455`).
- 🟢 nit: SVG title tooltip reads `"<rss> MB @ <ts>"` where `ts` is a raw ISO string — non-technical users see `2026-05-13T10:00:00+00:00`. Action: format with the existing `formatRelativeTime` helper used elsewhere in app.js.

vivekchand:
## Product review

- Mostly-backend with one small UI add (24h sparkline on the gateway-health card). Solves a real shipped problem (~600→945MB OOM bloat is invisible from a single live snapshot) and uses the open OTel-shaped events table — no new infra. DuckDB-first compliance is clean. Ship.
- 🟢 nit: 24h retention is the OSS default. If you later extend this to >24h windows, that's a Cloud-Pro upsell hook — add a "View 7d trend (Pro)" CTA next to the sparkline rather than silently capping. Action: file a follow-up issue for >24h gateway-health history as Pro-gated.
- 🟢 nit: when the sparkline renders with very few samples (fresh install), it's a conversion moment — a hover tooltip "Track this trend across all your nodes (Cloud)" would be a low-friction upsell. Action: defer to fleet-view PR.

Issue #1002, opened 2026-05-11
[Dives] /api/dives/* endpoints


vivekchand:
Sub-issue of https://github.com/vivekchand/clawmetry/issues/999.

## Endpoints
- `POST /api/dives/query` — `{question}` → `{sql, chart_spec, rows, error?}`
- `POST /api/dives/save` — `{question, sql, chart_spec, name}` → `{slug}`
- `GET /api/dives` — list saved dives
- `GET /api/dives/<slug>` — get saved dive + re-run
- `DELETE /api/dives/<slug>`

## Storage
Saved Dives as JSON at `~/.clawmetry/dives/<slug>.json`. Slug from kebab-case of name, dedup with suffix.

## Telemetry
Each run inserts a row into local store's `events` table with `event_type='dive_run'` and data: `{question_length, sql_length, rows_returned, ms_elapsed}`. Enables future prompt improvement.

## Test plan
- All 5 endpoints have happy-path + error-path tests
- Save → list → get → re-run round-trip

vivekchand:
The SQL safety validator (DIVES-1, `clawmetry/dives_sql_safety.py`) and prompt builder (DIVES-2, `clawmetry/dives_prompt.py`) are already merged, so the infrastructure is in place. This PR adds DIVES-3 — the HTTP API layer.

**Problem:** No endpoints exist yet for querying DuckDB via the LLM or for saving/listing/retrieving saved Dives.

**Files to change:**
- `routes/dives.py` (new) — `bp_dives` Blueprint implementing all five endpoints. Calls `dives_prompt.build_dives_prompt()` to generate the LLM prompt, reuses Advisor's `_load_anthropic_auth` / `_call_anthropic_api` / `_call_via_claude_cli` so zero-config auth works out of the box, runs every LLM-generated SQL through `dives_sql_safety.validate_sql()`, executes via `local_store.raw_select_safe()`, and persists saved Dives as JSON under `~/.clawmetry/dives/`.
- `dashboard.py` — add `from routes.dives import bp_dives` and `app.register_blueprint(bp_dives)`.

**Tradeoff:** Telemetry (`dive_run` event) is a best-effort no-op in daemon mode because routes open a read-only store and the daemon's local_server doesn't expose `ingest`. It's left as a TODO for DIVES-6 which has its own tracking issue.

Implementing now, will open a PR shortly.

---
_Generated by [Claude Code](https://claude.ai/code)_

Issue #999, opened 2026-05-11
[EPIC] ClawMetry Dives — AI SQL→chart over local DuckDB


vivekchand:
## TL;DR
Plain-English question → SQL → chart, all against the local DuckDB store. No SaaS dep, no row values leave the machine.

Full design + trade-off analysis in [PRD-clawmetry-dives.md](https://github.com/vivekchand/clawmetry-cloud/blob/main/docs/PRDs/PRD-clawmetry-dives.md) (private repo).

## Why now
- PR #995 / #997 / #998 just landed: local DuckDB has sessions, memory_blobs, heartbeats, system_snapshots, crons, subagents tables — all multi-agent keyed by `agent_type`.
- Aisha chat widget + `/claude-api` already routes prompts through user's BYO key.
- Chart.js is already loaded for the Tokens tab.

What's missing is the glue: schema-aware prompting + SQL safety validator + chart-spec rendering.

## Architecture (high-level)
1. User types in `/dives` tab input box
2. Backend prompt: schema descriptor (table names, column types, **no row values**) + question → LLM returns `{sql, chart_type, x, y, title}`
3. SQL safety validator rejects anything that's not a `SELECT` against the allowlisted 9 tables
4. Run query against local DuckDB
5. Render with Chart.js
6. Optional save → JSON spec at `~/.clawmetry/dives/<slug>.json`

## Privacy promise
On each Dive run, the LLM call sends:
- The user's natural-language question
- A schema descriptor (table/column names, types)
- **NEVER row values**

For users wanting zero LLM contact, the existing Ollama integration is the path.

## Sub-issues
- #DIVES-1 — SQL safety validator (allowlist, regex rejection, 30+ malicious-input tests)
- #DIVES-2 — Prompt template + schema descriptor builder
- #DIVES-3 — `/api/dives/*` endpoints (query, save, list, get, delete)
- #DIVES-4 — `/dives` tab UI (input box, chart area, saved Dives list, "show SQL" toggle)
- #DIVES-5 — Suggested-questions gallery (regression-tested fixed list)
- #DIVES-6 — Telemetry: log `dive_run` events into the local store for prompt improvement

## Out of scope (Phase 2+)
- Evidence.dev `/reports` sidecar (Phase 2 — power users authoring markdown+SQL by hand)
- Sharing Dives across team via cloud (Phase 3)
- Scheduled Dive re-runs + threshold notifications (Phase 4)
- Self-Evolve feedback loop (Phase 5)

[...]

PR #1162, opened 2026-05-13
feat(dives): SQL safety validator with 30+ malicious-input tests (closes #1000)

### LINKED ISSUE #1000: [Dives] SQL safety validator (allowlist + 30+ malicious-input tests)
Sub-issue of https://github.com/vivekchand/clawmetry/issues/999.

## Goal
A pure-Python validator function: `validate_sql(sql: str) -> ValidationResult` that rejects anything the Dives feature shouldn't execute.

## Allowlist
- Statement type: `SELECT` or `WITH … SELECT` only
- Single statement (no `;` mid-query)
- Tables: events, sessions, daily_aggregates, memory_blobs, heartbeats, system_snapshots, openclaw_channels, crons, subagents
- NO: system.*, pragma_*, ATTACH, INSTALL, LOAD, COPY, INSERT, UPDATE, DELETE, DROP, ALTER, CREATE

## Test plan
30+ malicious inputs including SQL injection classics (`'; DROP TABLE events;--`, comment escapes, UNION attacks, subquery exfil), plus 10+ valid queries to prove no false positives.

## Living module
`clawmetry/dives_validator.py` — fully isolated, easy to fuzz.

vivekchand:
## Summary

First slice of the Dives epic (#999) — the AI SQL -> chart feature over the local DuckDB store. Every LLM-generated query MUST pass this validator before execution; the model output is untrusted by definition.

- **New module** `clawmetry/dives_sql_safety.py` (~270 lines, **zero new runtime deps**)
  - `validate_sql(sql) -> (ok, reason)` — pure static analysis
  - Allowlist: `SELECT` / `WITH ... SELECT` only, single statement, <= 10 KB, no NUL bytes
  - Lexical pass strips strings + comments **before** keyword scanning, so a literal like `'do not DROP this row'` doesn't false-positive but a real `DROP` after a block comment still does
  - Bans DuckDB footguns: `ATTACH` / `INSTALL` / `LOAD` / `PRAGMA` / `COPY` / `SET` / `BEGIN` / `EXEC` / etc.
  - Bans file-system functions: `read_csv`, `read_parquet`, `read_json`, `write_csv`, `glob`, `httpfs`, `system_function`, ...
  - Optional defence-in-depth pass via `sqlglot` if it's already importable; falls back gracefully when it isn't (project rule: minimal deps)

- **Tests** `tests/test_dives_sql_safety.py` — 58 cases total
  - 13 valid `SELECT` / `WITH` / window-function / CTE / subquery / comment-prefixed queries (false-positive net)
  - 38 malicious inputs spanning: classic injection (`'; DROP TABLE events;--`), multi-statement, DDL/DML, `ATTACH` / `INSTALL` / `LOAD` / `PRAGMA` / `COPY`, file-system function calls, comment-hidden statements, NUL bytes, oversize query, non-SELECT shapes (`SHOW`, `DESCRIBE`, `EXPLAIN`, `CALL`, `BEGIN`)
  - Each rejection asserts the `reason` substring so failures are easy to triage
  - Length guards on both lists (`>= 10` valid, `>= 20` malicious) to stop the safety net from accidentally shrinking

Validator itself never touches DuckDB; execution lives in #1002.

## Test plan

- [x] `python3 -m pytest tests/test_dives_sql_safety.py -v` -> 58 passed
- [x] `python3 -c "import ast; ast.parse(open('clawmetry/dives_sql_safety.py').read())"`
- [x] Manual smoke: `validate_sql("SELECT 1")` -> `(True, None)`; `validate_sql("'; DROP TABLE events;--")` -> `(False, 'multiple statements not allowed')`
- [ ] Reviewer: try to break it with one more injection variant and add a test

Closes #1000. Part of #999.

### DIFF
--- a/clawmetry/dives_sql_safety.py
+++ b/clawmetry/dives_sql_safety.py
@@ -1,0 +1,269 @@
+"""SQL safety validator for ClawMetry Dives (AI SQL -> chart over local DuckDB).
+
+The Dives feature lets users ask plain-English questions; an LLM generates a SQL
+query against the local DuckDB store; the result is rendered as a chart. Because
+the SQL is LLM-generated, every query MUST pass this validator before execution.
+Never trust the raw model output.
+
+The validator is intentionally tiny, pure-Python, and dependency-free so it can
+be fuzzed in isolation. Sub-issue: https://github.com/vivekchand/clawmetry/issues/1000.
+
+Design notes
+------------
+- Allowlist-first: we only let through ``SELECT`` queries (optionally preceded by
+  ``WITH ... SELECT``). Everything else is rejected by default.
+- Two layers of defence:
+    1. A lexical pass that strips strings + comments, then scans the remaining
+       tokens for banned keywords / functions / multi-statements. This catches
+       the classic injection tricks: ``'; DROP TABLE x; --``, comment-hidden
+       statements, ``EXEC`` / ``ATTACH`` / file-system functions, etc.
+    2. If ``sqlglot`` is available (best-effort, optional dep) a structural
+       parse is run as defence in depth. We never *require* sqlglot — the
+       lexical pass alone is enough to enforce the allowlist.
+- Pure static analysis. The validator never touches DuckDB; execution is the
+  caller's job (see #1002).
+
+Public API
+----------
+    validate_sql(sql: str) -> tuple[bool, str | None]
+
+Returns ``(True, None)`` if the query is safe to execute, ``(False, reason)``
+otherwise. The reason is a short human-readable string suitable for surfacing
+in the UI.
+"""
+
+from __future__ import annotations
+
+import re
+from typing import Optional, Tuple
+
+# ---------------------------------------------------------------------------
+# Configuration
+# ---------------------------------------------------------------------------
+
+#: Hard cap on input size. The LLM should never produce queries longer than a
+#: few hundred bytes; 10 KB is generous and prevents prompt-injection bombing.
+MAX_SQL_LENGTH = 10_000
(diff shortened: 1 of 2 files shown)


vivekchand:
## UX review

- Pure-backend security primitive (SQL allowlist validator for an unshipped Dives feature). No user surface, no rendered output. Reviewer note: when the consumer surfaces this in #1002, the rejection `reason` strings (`"banned keyword: drop"`, `"only SELECT/WITH allowed"`) are decent for engineers but jargon for end users — wrap them with a friendlier message at the call site (`clawmetry/dives_sql_safety.py:230-260`).

vivekchand:
## Product review

- Pure-backend defensive code for an unshipped feature (Dives). Right call to land the validator before the executor — the lexical pre-pass + optional sqlglot defence-in-depth + 58 tests is the appropriate paranoia bar for executing LLM-generated SQL against the local store. Zero new runtime deps respects the minimal-deps rule.
- 🟡 P1: Dives is a natural Cloud-Pro feature (AI SQL → chart over historical data is exactly the "insight" surface the cloud display tier sells). Make sure #1002 ships the validator gated behind a tier check, not exposed unconditionally on the OSS dashboard. Action: thread tier check into the Dives execution route in #1002.
- 🟢 nit: `ALLOWED_TABLES` is documented in the module but unused (validator deliberately doesn't enforce table allowlisting). Consider deleting the constant or adding a TODO pointing at #1002 so a reader doesn't waste time wondering why it's defined (`clawmetry/dives_sql_safety.py:71-79`).

PR #1174, opened 2026-05-13
fix(memory): heartbeat-piggyback push of memory_blobs to cloud cache

vivekchand:
## Summary

The cloud Node Detail → Memory tab rendered "No memory data synced" even when the OSS daemon was healthy and `memory_blobs` had rows (SOUL.md, USER.md, AGENTS.md, etc).

Root cause: no daemon-side cache push for memory. The cloud handler at `/api/cloud/memory-files` had nothing to read from cache and always returned `{blob: None}`.

Fix: add `_build_memory_cache_pushes()` (mirrors `_build_brain_cache_pushes`) that reads via `LocalStore.query_memory_blobs()`, builds the `{memory_state.files, memory_content}` shape the cloud Memory IDE JS already consumes in `_renderIDE` / `_selectFile`, encrypts with the user's E2E key, and pushes to `memory:{owner_hash}:{node_id}:files` on every heartbeat (TTL 6h).

- DuckDB-first: read path is `query_memory_blobs()` only — no JSONL, no gateway, no Postgres.
- Encryption invariant preserved: cloud only ever sees ciphertext.
- Paired with clawmetry-cloud PR that wires the cache-read side: vivekchand/clawmetry-cloud#fix/cloud-memory-tab-empty

## Test plan

- [x] `tests/test_memory_cache_push.py` (5 tests, all green)
  - happy path: seeded `memory_blobs` rows → one `cache_pushes` entry with key `memory:{owner_hash}:{node}:files`, ttl=21600
  - empty store → no push
  - missing encryption key → no push
  - round-trip decrypt: `{memory_state.files, memory_content}` shape matches what the cloud Memory IDE JS reads
  - `send_heartbeat` attaches the push to the outgoing request payload
- [x] Existing `tests/test_brain_cache_push.py` still passes (5/5)
- [ ] After clawmetry-cloud PR merges + a heartbeat fires, the cloud Memory tab will paint the user's files instead of the empty state

### DIFF
--- a/clawmetry/sync.py
+++ b/clawmetry/sync.py
@@ -2889,6 +2889,16 @@ def send_heartbeat(config: dict) -> bool:
             payload.setdefault("cache_pushes", []).extend(ap_pushes)
     except Exception as _ap_e2:
         log.debug("approvals cache_push build failed (continuing): %s", _ap_e2)
+    # Memory tab (epic #1032): proactively push the user's memory-file
+    # snapshot so the cloud Node Detail → Memory tab paints from cache.
+    # Without this push the cloud handler returns `{blob: None}` and the
+    # browser renders "No memory data synced". Best-effort.
+    try:
+        mem_pushes = _build_memory_cache_pushes(config)
+        if mem_pushes:
+            payload.setdefault("cache_pushes", []).extend(mem_pushes)
+    except Exception as _mp_e:
+        log.debug("memory cache_push build failed (continuing): %s", _mp_e)
     # Local-first: persist this heartbeat to local DuckDB so the dashboard
     # has a per-node liveness history even when offline. Best-effort.
     try:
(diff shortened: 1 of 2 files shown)


vivekchand:
## UX review

- Pure-backend infra (heartbeat-piggyback push of memory blobs to the cloud cache). The user-visible win lands in the cloud's Memory tab via the paired clawmetry-cloud PR — this OSS half has no rendered surface. Verify in cloud E2E that the previously-empty Memory tab now paints SOUL.md/USER.md/AGENTS.md after one heartbeat cycle (≤30s).

vivekchand:
## Product review

- Direct fix for a broken paid-tier surface (cloud Memory tab rendering "No memory data synced" with healthy data locally). Heartbeat-piggyback over a new long-lived connection respects the relay-transport decision. E2E encryption invariant preserved (cloud only sees ciphertext). Mirrors the existing `_build_brain_cache_pushes` shape — low novelty, low risk. Ship.
- 🟡 P1: 200-file limit + 500KB/file truncate (`MEMORY_CACHE_LIMIT`, `MEMORY_CONTENT_TRUNCATE` in `clawmetry/sync.py`) are reasonable defaults but silently truncate without surfacing it. A heavy user with a 600KB AGENTS.md sees clipped content in the cloud viewer with no "truncated for cloud display" hint. Action: set a `truncated: true` flag in the per-file payload and surface it in the cloud renderer.
- 🟢 nit: 6h TTL means a 6h+ idle node shows stale memory in cloud. Acceptable today (memory files change slowly), but if a user edits SOUL.md and waits >6h before next heartbeat the cloud viewer shows the old content until the next push. Action: document the 6h staleness window in the cloud Memory tab tooltip.

Issue #1053, opened 2026-05-12
[Phase 1] Replace WS relay with heartbeat-piggyback + cloud cache


vivekchand:
## Decision

After 7 failed Cloud Run deploys with the WS relay, we're killing the WebSocket path entirely. Industry research (Datadog Remote Config, AWS SSM, NewRelic Synthetics, OpenTelemetry OpAMP) shows **none** of the major observability tools use bidirectional WS for backend→agent. They all use HTTPS pull/long-poll/heartbeat.

Full design: see [memory note `project_relay_transport_decision.md`].

## What ships in Phase 1

1. **Cloud heartbeat handler extension** — return \`pending_queries: [...]\` from \`/ingest/heartbeat\`, drained from a per-\`node_id\` queue
2. **Cloud cache endpoints** — \`POST /ingest/cache\` (daemon writes encrypted blob), \`GET /api/cloud/cache/<key>\` (browser reads)
3. **Cloud subscribe endpoint** — \`POST /api/cloud/subscribe {shape, args}\` returns \`{cache_key, eta_sec}\` and queues the query
4. **Daemon dispatch loop** — extend \`send_heartbeat\` to consume \`pending_queries\`, dispatch via \`local_query._dispatch\`, encrypt, POST to \`/ingest/cache\`
5. **Dashboard JS** — for cold-data views, subscribe → poll cache (with WS fallback while it still exists)

## Wire contract (v1)

\`\`\`
POST /ingest/heartbeat {node_id, ts, ...} → 
  {trial_state, pending_queries: [{id, shape, args}, ...]}

POST /ingest/cache {id, blob (E2E encrypted), shape, args_hash, ttl: 3600} → 204

POST /api/cloud/subscribe {node_id, shape, args} → 
  {cache_key, eta_sec, status: \"queued\"|\"cache_hit\"}

GET /api/cloud/cache/<cache_key>?token=cm_xxx → 
  encrypted blob OR {status: \"pending\", eta_sec} OR {status: \"node_offline\", last_seen}
\`\`\`

## Storage

- v1: in-memory dict on Cloud Run with TTL (we're at min_instances=1)
- v2: Upstash Redis once Phase 1 stable
- Postgres rejected (vacuum churn for 1h-TTL blobs)

## Auth

Same \`cm_xxx\` Bearer + \`node_id\` ownership check used by current relay.

## Privacy

[...]

PR #1177, opened 2026-05-13
fix(sync): kill dead WS reconnect + adaptive cycle sleep (real-time MOAT)

vivekchand:
## Summary

P0 fix for the real-time MOAT promise (≤5s Brain-tab freshness, target 1-2s). Two daemon-side regressions found on the live local node:

1. **Dead WS relay reconnect spam.** `clawmetry/relay.py` was still trying to reconnect to `wss://app.clawmetry.com/api/node/relay` every 60s after that endpoint was retired (issue #1053 / `project_relay_transport_decision`). Every attempt got `Handshake status 404 Not Found`, burying real errors in `sync.log`.
2. **Adaptive cadence floored at 15s.** PR #1111 added a 3s FAST cadence when `viewer_active=true`, but the main loop ended every cycle with `time.sleep(POLL_INTERVAL=15)`. So a 3s heartbeat interval still couldn't fire more often than every 15s. Plus the loop initialised `heartbeat_interval = SLOW`, ignoring the stash populated by the startup heartbeat — so the first ~60s post-restart was stuck on SLOW even if the user already had Brain open.

## Fixes

- `clawmetry/relay.py` → no-op stub. `start_relay_thread()` returns `None` unconditionally. Dispatch helpers retained (still used by `sync._dispatch_pending_queries` for heartbeat-piggyback queries and pinned by `tests/test_relay.py`).
- `clawmetry/sync.py:5908-5918` → deleted `start_relay_thread(...)` call, replaced with a comment pointing to `project_relay_transport_decision` and `reference_ws_handshake_400_unsolved`.
- `clawmetry/sync.py:6083` → seeded `heartbeat_interval = _pick_heartbeat_interval(_LAST_HEARTBEAT_RESPONSE)` so the startup heartbeat's `viewer_active` flag is honoured on the FIRST cycle, not the second.
- `clawmetry/sync.py:6223` → `time.sleep(max(1, min(POLL_INTERVAL, heartbeat_interval)))` so FAST=3s actually means a 3s cycle. SLOW=60s still capped by POLL_INTERVAL=15s — idle bandwidth/Cloud-Run cost unchanged.

## Before / after

Before (live local daemon):
```
sync.log: relay: error: Handshake status 404 Not Found ... reconnecting in 60s   ← every minute
heartbeat cycle floor: 15s even with viewer_active=true
```

After (expected once daemon restarts on this build):
```
sync.log: no `relay:` lines
heartbeat cycle on active viewer: 3s
```

The `interval_seconds: 1800` field shown on `/api/system-health` is the OpenClaw **agent** heartbeat (separate concept from the daemon→cloud sync heartbeat), so it's unaffected by this PR.

## Test plan

- [x] `tests/test_adaptive_heartbeat.py` — 8 existing tests pass; +3 new tests pin the cycle-sleep contract (FAST/SLOW/1s-floor).
- [x] `tests/test_relay.py` — collapsed the previous skip cases into one no-op assertion; passes.
- [ ] Restart local daemon: `launchctl kickstart -k gui/$(id -u)/com.clawmetry.dashboard`
- [ ] Verify no `relay: error` lines in `~/.clawmetry/sync.log` for ≥2 minutes.
- [ ] Open cloud Brain tab → send a message in OpenClaw → confirm event lands in cloud Brain within 5s.

## Out of scope (documented)

Free-plan rate limit (`HTTP 429 from /ingest/logs`, retry_after: 38s) is unrelated — Brain rides `/ingest/heartbeat`. Right fix is server-side rate-limit bump for the free plan or smarter client-side log batching; will track separately.

## Refs

- Issue #1053 (heartbeat-piggyback epic)
- PR #1111 (adaptive cadence)
- Memory: `project_relay_transport_decision`, `reference_ws_handshake_400_unsolved`

### DIFF
--- a/clawmetry/relay.py
+++ b/clawmetry/relay.py
@@ -1,44 +1,47 @@
-"""clawmetry/relay.py — node-side WebSocket relay client (epic #964 phase 3b).
+"""clawmetry/relay.py — DEPRECATED stub.
 
-Maintains a long-lived WebSocket to `wss://app.clawmetry.com/api/node/relay`,
-waits for `{type:"query"}` frames from the cloud, dispatches each via
-`routes.local_query.relay_dispatch()`, returns chunked responses.
+The node-side WebSocket relay client was retired 2026-05-13. The cloud
+endpoint ``wss://app.clawmetry.com/api/node/relay`` was killed 2026-05-12
+(returns 404 in prod) after the simple-websocket handshake-400 dead end
+documented in ``reference_ws_handshake_400_unsolved.md``. Per
+``project_relay_transport_decision``, the replacement is heartbeat
+piggyback: the cloud attaches ``pending_queries`` to the heartbeat
+response, the daemon answers via ``/ingest/cache``. No new long-lived
+connections.
 
-Design notes
-------------
-- **Base install dep** since 0.12.166. `websocket-client` is ~100 KB pure
-  Python; the previous `extras_require["relay"]` opt-in caused cloud users
-  to silently miss the relay.
-- **Daemon thread.** One thread per process. Reconnects with exponential
-  backoff (2s → 60s cap). The sync daemon's main loop is unaware.
-- **Skipped on OSS-local mode.** If `config` has no `api_key` / `node_id`,
-  the relay never starts — local-only OSS users pay no overhead.
-- **Privacy boundary unchanged.** Encrypted blobs stay encrypted; plaintext
-  columns (token counts, timestamps, event types) ride the WS in the clear
-  (TLS-protected). Same trust model as today's heartbeat / ingest.
+This module is kept as a no-op stub so:
+  * old daemon installs that still ``from clawmetry import relay`` don't
+    crash on import, and
+  * the existing unit tests in ``tests/test_relay.py`` can still import
+    the dispatch / chunking helpers (which are unrelated to the dead WS
+    supervisor and continue to drive the heartbeat-piggyback query path
+    via ``routes.local_query.relay_dispatch``).
 
-Tests live in `tests/test_relay.py` and mock the WebSocket layer.
+The reconnect supervisor (``_relay_supervisor``) is intentionally NOT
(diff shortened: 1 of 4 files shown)


vivekchand:
## UX review

- 🟢 nit — Pure backend / daemon change. End-user UX impact is Brain-tab freshness going from ~15-75s stale to 1-2s when viewer_active. That's a perceptible product win and worth a one-line note in release notes ('Real-time Brain updates: <2s'). No actionable UI bullets here. (`clawmetry/sync.py:6224-6236`)

vivekchand:
## Product review

- 🔴 P0 — Real-time MOAT (1-2s sync) is being given to free / OSS users by default. The 5s cache-window-as-Pro-feature upsell (vs OSS local-only 30s+ cadence, or Cloud-Free 15s) was never considered here. Right now this PR makes sub-2s available to everyone the moment they open the Brain tab. That collapses one of the strongest Cloud-Pro differentiators into a free baseline. Decision needed BEFORE this merges: (a) keep sub-2s universal and find another moat, or (b) tier FAST=3s as Pro-only and leave Free/OSS at FAST=10s. Strong recommend (b) — memory `project_redis_duckdb_pivot` puts real-time at the heart of the MOAT. (`clawmetry/sync.py:6080-6086, 6224-6236`)
- 🟡 P1 — `viewer_active` flag is binary. A user opens Brain → enters FAST=3s → walks away. Cloud Run pays for ~20x more heartbeat traffic until the next `viewer_active=false` heartbeat lands. Add idle detection client-side (no mouse/scroll for 60s → `viewer_active=false` heartbeat) so we don't burn the cost on tabs-left-open. (`clawmetry/sync.py`)
- 🟢 nit — Killing the dead WS reconnect is pure infrastructure cleanup. Worth noting that `sync.log` will be ~95% quieter, which makes the next real error visible — small but real product win for paid debugging support.

If a P0 decision goes against universal sub-2s, I'll open a tracking issue.

PR #1178, opened 2026-05-13
fix(install): launchd restart + __main__ + uptime portability + token discovery (closes #1127 install bucket)

### LINKED ISSUE #1127: [E2E audit 2026-05-13] 15+ technical bugs from fresh-install browser walk
Outcome of a fresh-install (`curl | bash`) end-to-end audit using Playwright against a real DuckDB with 145 events. Bugs below are all reproducible technical issues — verified against current installed code. Full screenshots at `/tmp/cm-e2e-bugs-2026-05-13/` locally.

> **Audit update (2026-05-13):** items confirmed fixed are checked and annotated with the merging PR. Unchecked items still reproduce against `origin/main` and the live daemon at `localhost:8900`.

## Install / launchd

- [ ] **P0** Install script doesn't restart launchd jobs after install. `pip uninstall` → `curl | bash` install completes cleanly but neither daemon reloads; dashboard at :8900 unreachable until manual `launchctl kickstart -k`. — _`install.sh` has no `launchctl bootout`/`bootstrap`/`kickstart` for the dashboard plist; restart logic lives only in `clawhub-plugin/install.sh`._
- [x] **P0** Dashboard launchd plist points at `~/Library/Python/3.9/bin/clawmetry` after pip-user → venv migration. `~/.clawmetry/bin/clawmetry` is the new path. Install script needs plist rewrite logic. — Fixed: local `~/Library/LaunchAgents/com.clawmetry.dashboard.plist` now points at `/Users/vivek/.clawmetry/bin/clawmetry`. (Plist write lives in `clawhub-plugin/install.sh:184`.)
- [ ] **P1** `python3 -m clawmetry` fails with `No module named clawmetry.__main__`. Add `clawmetry/__main__.py` that `from .cli import main; main()`. — _`clawmetry/__main__.py` still absent on `origin/main` (`git ls-tree origin/main clawmetry/__main__.py` returns empty)._
- [ ] **P1** Some helper shells out to `uptime -p` (GNU-only). Fails on macOS. Use `uptime` and parse, or `psutil.boot_time()`. — _`routes/health.py:888` still calls `subprocess.run(["uptime", "-p"], …)`._
- [ ] **P0** Gateway-token bootstrap: dashboard shows "enter gateway token" wall on fresh install even though `~/.openclaw/openclaw.json:auth.token` is set. Auto-discover when `OPENCLAW_GATEWAY_TOKEN` env is unset. — _`dashboard.py:2079-2082` looks for `cfg["gateway"]["auth"]["token"]`, not top-level `auth.token`; shape mismatch._

## OSS dashboard (localhost:8900) — local DuckDB read-path

These all share a root cause: the OSS dashboard's local-store read path is returning empty/zero on tabs where the cloud (sync'd from same DuckDB) shows real data.

- [x] **P0** Tokens tab: TODAY/WEEK/MONTH all show $0 / 0 tokens despite 145 events in DuckDB. Cloud Tokens tab shows real numbers from the same data. — Fixed by #1131 + #1137 (`/api/usage` now returns `today: 48528, week: 48852, month: 72872` against same DuckDB).
- [x] **P0** Flow tab `AGENT RUNTIME` card: model=`Unknown`, sub=`unknown`, Auth=`?`. Cloud knows it's `claude-opus-4-7`. — Fixed by #1131 (`/api/overview` now returns `model: claude-opus-4-7, provider: anthropic`; `app.js:5412 applyBrainModelToAll` reads `overview.model`).
- [x] **P0** Context tab "Active Model: unknown" — same model-detection gap. — Fixed by #1131 (`app.js:2854` reads `ov.model` which is now populated).
- [x] **P0** Context tab: Total Turns=0 / Compactions=0 despite events present. — Fixed by #1132 + #1137 (`/api/brain-history` now returns 140 events; `app.js:2849` counts 35 USER turns).
- [ ] **P1** Brain "Activity density" chart renders only background, no bars. — _`app.js:2685-2693 renderBrainChart` filters to last 60 min only; with day-old events all are bucketed out. Either widen the window or add an empty-state hint._
- [ ] **P1** Brain "Live event stream" shows ~7 entries while DuckDB has dozens more. Year missing from timestamps (just "9 May"). — Event count fixed by #1131/#1137 (140 returned now); _year missing still: `app.js:2216 formatBrainTime` returns `d.toLocaleDateString('en-GB', {day:'numeric',month:'short'})` with no year._
- [ ] **P1** Flow `Active Tools: ─` empty while `Actions Taken: 12`. "Live Tool Call Stream" stuck "Waiting for activity…" indefinitely. — _`app.js:7243 flowStats.activeTools` is filled only from SSE in the current page session; no DuckDB-backed historical tool list._
- [x] **P1** Tokens tab: 17 "Loading…" panels never resolve (Cost Breakdown / Cost By Plugin / Top Sessions / Trace Clusters / Activity Heatmap / Skill Cost Leaderboard). — Fixed by #1131/#1137 (`/api/sessions/cost-breakdown`, `/api/heatmap`, `/api/usage/by-plugin`, `/api/usage/cache-trends`, `/api/skill-attribution`, `/api/usage/cost-comparison`, `/api/cost-optimizer` all return populated JSON).

## Polling endpoints returning 404

Console-noisy on every tab switch:

- [ ] **P0** `GET /api/mc-tasks` → 404 (~11 fires per audit walk; polled from every tab init). — _`clawmetry/static/js/app.js:5274` still polls `/api/mc-tasks` but no route is registered for it (returns 404)._
- [x] **P0** `GET /api/cloud-proxy/api/cloud/account` → 401 even when daemon is paired with cloud. Proxy should attach `Authorization: Bearer` from `~/.clawmetry/config.json:api_key`. — OSS side fixed by #1115 (`routes/overview.py:1335` attaches `Authorization: Bearer` from `_read_cloud_token()` which now falls back to `~/.clawmetry/config.json:api_key`). Remaining 401 is upstream `app.clawmetry.com` rejecting the bearer — that's a cloud-repo issue.

## UX dead-ends

- [x] **P1** Crons tab empty state: "Click '+ New Job' to create one" — but no `+ New Job` button is rendered. Same dead-end on cloud Crons. — Fixed by #1149 (`clawmetry/static/js/app.js:4174` sets `_cronActionsAvailable = !window.CLOUD_MODE` inside `loadCrons()`).
- [ ] **P1** Alerts nav badge says `20` but Alerts page shows "No alerts have fired yet — recent history empty." Badge counts the wrong thing (template count vs fire count?). — _`app.js:282` sets badge from `/api/alerts/active` length (20 reproducible). Alerts page reads from `/api/cloud-proxy/api/alerts/history` which is upstream-401 — page renders empty-state. Two data sources, same UI._
- [ ] **P1** Notifications header: "1 channel configured" but all 5 channel cards show `Connect` button. Either the count is wrong or one channel has hidden "configured" state. — _`notifications.html:152` derives `n` from `state.rows.length` (cloud-fed), but render loop in `notifications.html:125-142` iterates `CHANNELS` constants and only marks `connected` when `byType[ch.key]` exists. Mismatch between row count and per-card lookup._
- [ ] **P1** Stuck-session warning banner ("Session X stuck — no activity for 104 min") resurfaces after page reload despite `Dismiss`. Persist Dismiss in localStorage by session_id. Auto-prune stuck banners for sessions inactive >24h. — Partially addressed by #1081 (banner now auto-clears on terminal session.state events) but no localStorage persistence; on reload the next active-alert poll re-shows it. _`app.js:7949-7984` has no `localStorage.setItem('cm_stuck_dismissed_…')`._
- [ ] **P1** Gateway-setup modal instructions assume Docker (`docker exec $(docker ps -q) env | grep TOKEN`). Add direct-install path: `cat ~/.openclaw/openclaw.json | jq -r .auth.token`. — _`dashboard.py:10042` still embeds the `docker exec $(docker ps -q) env | grep TOKEN` hint with no direct-install alternative._
- [ ] **P1** `window.openCloudModal()` opens GATEWAY-setup overlay on top of cloud modal — two z-stacked overlays. — _`clawmetry/static/js/gw-setup.js:108 openCloudModal()` does not hide `gw-setup-overlay`; both stay `display:flex` if the gateway isn't configured._
- [ ] **P2** NemoClaw tab is `display:none` in nav but `switchTab('nemoclaw')` still routes there. Page says `pip install nemoclaw` (package may not be public). — _`dashboard.py:3820` + `dashboard.py:9891` keep the tab in markup (`style="display:none"`); `switchTab` has no guard against routing into a hidden tab._

## Cross-cutting

Multiple tabs read different data sources and give contradictory numbers between OSS and cloud, even though cloud syncs from the same DuckDB. A unified read-path audit would close ~half of the P0s above in one PR.

---

Original full report (with reproduction commands + 50+ screenshots): saved locally at `/tmp/cm-e2e-bug-report-2026-05-13.md` + `/tmp/cm-e2e-bugs-2026-05-13/`. Not all 42 originally-reported items are mirrored here — this list is the verified-technical subset; subjective UI critiques and unverifiable items are intentionally excluded.

vivekchand:
## Summary

Four-bug install/portability fix tied to issue #1127. Each landed
defensively (graceful fallbacks, never crashes), with unit coverage
on the two helpers that needed it.

### #1127 install / launchd bucket — sub-bullets

- [x] **install.sh restarts launchd jobs after install**
  After the venv reinstall, dashboard/sync daemons launched at boot
  are still bound to the deleted venv. Added a Darwin-only block that
  `launchctl kickstart -k`s every `com.clawmetry.*` plist, wrapped in
  `|| true` so Linux/CI without launchd don't abort. Bonus: any plist
  whose `ProgramArguments[0]` points outside the fresh venv is rewritten
  to `~/.clawmetry/bin/{clawmetry,python3}` (matched by label) before
  the kickstart, so the next reboot picks up the right interpreter
  too.

- [x] **`python3 -m clawmetry` fails with "No module named clawmetry.__main__"**
  Added `clawmetry/__main__.py` that does `from clawmetry.cli import main; main()`.
  Verified `importlib.util.find_spec('clawmetry.__main__')` resolves.

- [x] **`uptime -p` is GNU-only — fails on macOS**
  Audit was right; 5 callers across `routes/health.py`,
  `routes/overview.py` (x2), `routes/components.py`,
  `clawmetry/sync.py`. New `helpers/system.py` exposes
  `uptime_pretty()` / `uptime_seconds()` / `boot_time()` /
  `format_uptime()` using psutil → /proc/uptime → sysctl
  kern.boottime → GetTickCount64 fallbacks. Each call site now uses
  the helper; sync.py inlines the equivalent logic (sync.py is the
  installed package, doesn't import from top-level `helpers/`).

- [x] **Gateway-token shape mismatch — try both paths**
  Read real `~/.openclaw/openclaw.json` on this machine — current
  OpenClaw stores the token at `gateway.auth.token` (the existing
  primary path), and top-level `auth` holds OAuth profiles, not
  tokens. To cover the issue reporter's shape without breaking
  the common case, `_detect_gateway_token` now tries
  `cfg.gateway.auth.token` first, then `cfg.auth.token`, then
  falls through to the next config file. Patched both copies of
  `_detect_gateway_token` in `dashboard.py` (there are two — the
  later definition shadows the earlier, but both are now consistent).

## Test plan

- [x] `pytest tests/test_uptime_portable.py tests/test_gateway_token_fallback.py` — 16/16 pass locally on macOS.
- [x] `python3 -m ast` parse on every touched .py — all OK.
- [x] `bash -n install.sh` — OK.
- [x] Smoke: `python3 -m clawmetry` now resolves (`importlib.util.find_spec` returns a real spec).
- [x] Smoke: `_detect_gateway_token()` against real `~/.openclaw/openclaw.json` returns the nested token unchanged.
- [ ] CI matrix — Ubuntu/macOS/Win, py3.9 + py3.11.

### DIFF
--- a/clawmetry/__main__.py
+++ b/clawmetry/__main__.py
@@ -1,0 +1,10 @@
+"""Allow `python3 -m clawmetry` to invoke the CLI entry point.
+
+Without this module Python emits "No module named clawmetry.__main__".
+The CLI's argparse-based `main()` lives in clawmetry.cli.
+"""
+
+from clawmetry.cli import main
+
+if __name__ == "__main__":
+    main()
--- a/clawmetry/sync.py
+++ b/clawmetry/sync.py
@@ -5381,17 +5381,50 @@ def sync_system_snapshot(config: dict, state: dict, paths: dict) -> int:
     except Exception:
         system.append(["RAM", "--", ""])
 
+    # Portable uptime: stdlib-only (no `uptime -p` — missing on macOS/BSD).
     try:
-        if sys.platform == "darwin":
-            up = subprocess.run(
-                ["uptime"], capture_output=True, text=True, timeout=5
-            ).stdout.strip()
-            system.append(["Uptime", up.split(",")[0].split("up")[-1].strip(), ""])
+        boot_ts = None
+        try:
+            import psutil  # type: ignore
+
+            boot_ts = float(psutil.boot_time())
+        except Exception:
+            if sys.platform.startswith("linux"):
+                try:
+                    with open("/proc/uptime") as _uf:
+                        boot_ts = time.time() - float(_uf.read().split()[0])
+                except Exception:
+                    pass
+            elif sys.platform == "darwin" or "bsd" in sys.platform:
+                try:
+                    import re as _re
+
+                    _out = subprocess.run(
+                        ["sysctl", "-n", "kern.boottime"],
+                        capture_output=True,
+                        text=True,
+                        timeout=2,
+                    ).stdout
+                    _m = _re.search(r"sec\s*=\s*(\d+)", _out)
+                    if _m:
+                        boot_ts = float(_m.group(1))
+                except Exception:
+                    pass
+        if boot_ts is not None:
+            secs = max(0, int(time.time() - boot_ts))
+            days, rem = divmod(secs, 86400)
+            hours, rem = divmod(rem, 3600)
+            minutes = rem // 60
+            parts = []
+            if days:
+                parts.append(f"{days}d")
+            if hours:
+                parts.append(f"{hours}h")
+            if minutes or not parts:
+                parts.append(f"{minutes}m")
+            system.append(["Uptime", " ".join(parts), ""])
         else:
-            up = subprocess.run(
-                ["uptime", "-p"], capture_output=True, text=True, timeout=5
-            ).stdout.strip()
-            system.append(["Uptime", up.replace("up ", ""), ""])
+            system.append(["Uptime", "--", ""])
     except Exception:
         system.append(["Uptime", "--", ""])
 
--- a/dashboard.py
+++ b/dashboard.py
@@ -2568,10 +2568,19 @@ def _detect_gateway_token():
 
             with open(jp) as f:
                 cfg = _json.load(f)
+            # Primary path on current OpenClaw: cfg["gateway"]["auth"]["token"].
+            # Some installs / older schemas store the token at top-level
+            # cfg["auth"]["token"] (issue #1127). Try the nested gateway
+            # path first, then the top-level auth path so we cover both
+            # without breaking the common case.
             gw = cfg.get("gateway", {})
-            auth = gw.get("auth", {})
-            if isinstance(auth, dict) and "token" in auth:
-                return auth["token"]
+            if isinstance(gw, dict):
+                gw_auth = gw.get("auth", {})
+                if isinstance(gw_auth, dict) and gw_auth.get("token"):
+                    return gw_auth["token"]
+            top_auth = cfg.get("auth", {})
+            if isinstance(top_auth, dict) and top_auth.get("token"):
+                return top_auth["token"]
         except (FileNotFoundError, ValueError, KeyError, TypeError):
             pass
     return None
@@ -9779,10 +9788,17 @@ def _detect_gateway_token():
             import json as _json
             with open(jp) as f:
                 cfg = _json.load(f)
+            # Primary path on current OpenClaw: cfg["gateway"]["auth"]["token"].
+            # Older / alternate schemas put it at top-level cfg["auth"]["token"]
+            # (issue #1127). Try gateway-nested first, fall back to top-level.
             gw = cfg.get('gateway', {})
-            auth = gw.get('auth', {})
-            if isinstance(auth, dict) and 'token' in auth:
-                return auth['token']
+            if isinstance(gw, dict):
+                gw_auth = gw.get('auth', {})
+                if isinstance(gw_auth, dict) and gw_auth.get('token'):
+                    return gw_auth['token']
+            top_auth = cfg.get('auth', {})
+            if isinstance(top_auth, dict) and top_auth.get('token'):
+                return top_auth['token']
         except (FileNotFoundError, ValueError, KeyError, TypeError):
             pass
     return None
--- a/helpers/system.py
+++ b/helpers/system.py
@@ -1,0 +1,126 @@
+"""
+helpers/system.py — Portable system uptime helpers.
+
+The legacy callers shelled out to `uptime -p`, which only exists on
+GNU/coreutils (Linux). On macOS / BSD `uptime` has no `-p` flag and the
+subprocess returns non-zero, so the dashboard rendered "Unknown" on
+every Mac install. Issue #1127.
+
+This module computes boot time using only stdlib:
+  - psutil.boot_time() when psutil is installed (most reliable cross-OS)
+  - /proc/uptime on Linux
+  - `sysctl kern.boottime` on macOS / BSD
+  - WMIC / GetTickCount64 on Windows (best-effort)
+
+Returns a pretty string like "up 3 days, 4 hours, 12 minutes" — matching
+the legacy `uptime -p` output the dashboard JS expects.
+"""
+
+from __future__ import annotations
+
+import os
+import re
+import subprocess
+import sys
+import time
+
+
+def boot_time() -> float | None:
+    """Return the POSIX timestamp the host booted at, or None if unknown.
+
+    Never raises — every failure path returns None so callers can render
+    a graceful fallback.
+    """
+    # 1. psutil — most accurate, works on every platform we ship to.
+    try:
+        import psutil  # type: ignore
+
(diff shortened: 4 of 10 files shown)


vivekchand:
## UX review

- 🟢 nit — Mostly install/portability backend. End-user impact: macOS users finally see correct uptime (was 'Unknown' on every Mac per the audit). One bullet: the new uptime format reads 'up 3 days, 4 hours, 12 minutes' — three commas in a small system table cell will likely wrap or truncate on narrower viewports. Confirm rendering at <600px or trim to two units max ('3d 4h'). (`helpers/system.py:90-115`)

vivekchand:
## Product review

- 🔴 P0 — `launchctl kickstart` is Darwin-only. Linux + WSL users get nothing. install.sh sniffs OS and only runs the kickstart block on Darwin — which is correct defensively, but it leaves Linux installers with the SAME stale-daemon-after-upgrade bug we just fixed for Mac. Linux equivalent is `systemctl --user restart clawmetry.service` (when a unit exists) or `pkill -f clawmetry && nohup ... &` fallback. Add a Linux branch or this PR closes the install bucket only for ~half the userbase. Per memory `project_clawmetry_traction` 120K+ installs in 100+ countries — non-Mac is a large slice. (`install.sh:99-141`)
- 🟡 P1 — `python3 -m clawmetry` now works → great for power users. But the install pitch doesn't surface it. The README still says `clawmetry` (the console script) is the entry point. `-m clawmetry` is the failsafe when shim PATH issues bite (memory `reference_two_python_installs_gotcha` — two-Python install bug is real). Add to install.sh's success message: 'Couldn't run `clawmetry`? Try `python3 -m clawmetry`.' (`clawmetry/__main__.py:1-10`)
- 🟢 nit — Gateway token discovery now tries 2 paths. Per the OpenClaw schema audit in this PR, current installs all use `gateway.auth.token` and `auth.token` is the issue-reporter's older schema. Worth logging which path matched so we can deprecate the fallback once telemetry shows zero hits in 90 days. (`dashboard.py:2568-2585`)
- 🟢 nit — `helpers/system.uptime_pretty()` exists at top-level `helpers/`, but `clawmetry/sync.py` had to inline the same logic because it ships inside the installed package. That's a design smell — either move `helpers/system.py` into `clawmetry/helpers/` and share, or document the duplication. Drift risk on the next portability fix.

PR #1183, opened 2026-05-13
fix(install): Linux systemd + WSL daemon-restart paths (closes #1182)

### LINKED ISSUE #1182: [Product P0 from PR #1178] Linux + WSL users left with stale-daemon-after-upgrade bug
From the post-merge UX/product review of PR #1178.

PR #1178's install bucket fixes the post-upgrade stale-daemon bug for macOS via `launchctl kickstart -k`. The block is gated on `[ "$OS" = "Darwin" ]` (correct defensively) — but it means Linux and WSL users get **nothing**. Same stale-venv-still-running bug, no equivalent restart.

## Why this is a P0

Per memory `project_clawmetry_traction`: 120K+ installs across 100+ countries. Non-Mac is a large slice of that. "Closes #1127 install bucket" claim in the PR title is misleading — it closes the install bucket only for ~half the userbase.

## Fix sketch (Linux)

```bash
elif [ "$OS" = "Linux" ]; then
  # systemd --user: try restart, fall back to pkill+respawn.
  if systemctl --user list-unit-files 2>/dev/null | grep -q '^clawmetry'; then
    systemctl --user daemon-reload || true
    systemctl --user restart clawmetry-sync.service 2>/dev/null || true
    systemctl --user restart clawmetry-dashboard.service 2>/dev/null || true
  else
    # Last-resort: kill any running daemon process; the next `clawmetry` invocation respawns it.
    pkill -u "$USER" -f 'clawmetry.sync' 2>/dev/null || true
    pkill -u "$USER" -f 'clawmetry --port' 2>/dev/null || true
  fi
fi
```

WSL is detectable via `grep -qi microsoft /proc/version`; on WSL2 systemd-user mode works after a recent Windows update, so the Linux branch covers it.

## Action

- Add Linux branch to `install.sh` before PR #1178 ships to PyPI or before the next `[RELEASE]` PR cuts.
- Smoke-test on Ubuntu 22.04 and Debian 12 (CI matrix already has Ubuntu per CLAUDE.md).

## Refs

- PR #1178: `install.sh:99-141`
- Memory: `project_clawmetry_traction`, `reference_two_python_installs_gotcha`

vivekchand:
## Summary

Issue #1182 — PR #1178's Darwin-only `launchctl kickstart` block fixed the stale-venv-after-upgrade daemon bug for macOS but left **Linux + WSL** users (a large slice of the 120K+ install base) silently broken.

This PR adds the matching restart paths:

- **Linux + systemd**: detect via `systemctl --user list-unit-files`, then `daemon-reload` + `restart clawmetry-sync.service`. The unit name is fixed by `clawmetry/cli.py::_register_systemd`, and the test pins it so the two files can never drift.
- **WSL**: detected via `grep -qi microsoft /proc/version`. WSL ships without systemd-user by default, so we print a friendly `pkill -f clawmetry && nohup clawmetry &` hint instead.
- **Generic Linux fallback**: same pkill hint when no `clawmetry-sync.service` is registered (e.g. user pip-installed without running `clawmetry connect`).
- **Darwin cross-platform sanity**: the existing block now also prints the same hint when no `com.clawmetry.*.plist` is found — one consistent message across platforms.

The existing Darwin PlistBuddy + `launchctl kickstart` logic is preserved untouched.

## What NOT changed

- No root-systemd path. User-level only — most installs are not running as root.
- No new dependencies. Shell-only.
- No noisy success output — restart confirmation only fires when we actually restarted something.

## Test plan

- [x] `bash -n install.sh` — syntax-check passes
- [x] `pytest tests/test_install_script_linux_restart.py` — 6/6 passes (pins all 3 added paths + Darwin fallback)
- [ ] Real Ubuntu 22.04 box: `pip install -U clawmetry`, confirm `journalctl --user -u clawmetry-sync` shows restart event
- [ ] Real WSL2 (no systemd): confirm install ends with the WSL hint and exit code 0

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -103,9 +103,11 @@ CLAWMETRY_VERSION=$("$INSTALL_DIR/bin/python3" -c "import importlib.metadata; pr
 if [ "$OS" = "Darwin" ]; then
   _LA_DIR="$HOME/Library/LaunchAgents"
   _UID=$(id -u)
+  _DARWIN_PLIST_FOUND=0
   # Step 1: rewrite stale ProgramArguments[0] in any com.clawmetry.* plist.
   for _plist in "$_LA_DIR"/com.clawmetry.*.plist; do
     [ -f "$_plist" ] || continue
+    _DARWIN_PLIST_FOUND=1
     _current=$(/usr/libexec/PlistBuddy -c "Print :ProgramArguments:0" "$_plist" 2>/dev/null || echo "")
     _label=$(basename "$_plist" .plist)
     case "$_current" in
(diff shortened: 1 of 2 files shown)


vivekchand:
## UX review

🟢 nit · Backend-only (install.sh + tests, no user-facing UI surface) — skipping the usual UX scrutiny. The user-visible artifacts are 4 terminal-print lines (`↺ Restarted clawmetry-sync systemd user service` / WSL hint / generic Linux hint / Darwin no-plist hint), all consistent in tone, all gated to fire only when the action actually applies. Good restraint — no noise on a vanilla pip install where there is no daemon to restart.

vivekchand:
## Product review

🟢 nit · Backend infra (install ergonomics) — nothing to gate to Pro, nothing to revert. Worth flagging two tier-strategy implications:

🟡 P1 · `clawmetry-sync.service` is the only systemd unit name pinned by the test. Once we ship a managed cloud-relay daemon as a separate process (per `project_relay_transport_decision.md`), it'll need its own unit + restart line here. Add a TODO comment so the next person bumping unit names doesn't silently regress this.

🟢 nit · The pkill fallback is the same string in 3 places (Darwin no-plist / WSL / generic Linux). Worth extracting to a `_RESTART_HINT` variable so a single edit propagates if the recommended manual command ever changes.

Funnel-relevant: this fix unblocks the ~half of the install base that's Linux/WSL — concrete signal that ClawMetry installs cleanly on Linux, which is critical for the dev.to/HN OSS audience targeted in `feedback_kody_kendall_user.md`. No CTA needed here; the win is "doesn't break."

PR #1185, opened 2026-05-13
fix(app.js): CLOUD_MODE-guard /api/workspaces fetch (unblocks 0.12.183)

vivekchand:
Unblocks the 0.12.183 release. cloud-contract.mjs gate caught `Failed to load resource: 410 @ /api/workspaces` because OSS app.js polls the endpoint that cloud now classifies as cloud-disabled (per PR #816 route-policy classifications). One-line CLOUD_MODE guard — same pattern as the #754/#755 fix.

After this merges, re-cut the 0.12.183 release PR.

### DIFF
--- a/clawmetry/relay.py
+++ b/clawmetry/relay.py
@@ -1,44 +1,47 @@
-"""clawmetry/relay.py — node-side WebSocket relay client (epic #964 phase 3b).
+"""clawmetry/relay.py — DEPRECATED stub.
 
-Maintains a long-lived WebSocket to `wss://app.clawmetry.com/api/node/relay`,
-waits for `{type:"query"}` frames from the cloud, dispatches each via
-`routes.local_query.relay_dispatch()`, returns chunked responses.
+The node-side WebSocket relay client was retired 2026-05-13. The cloud
+endpoint ``wss://app.clawmetry.com/api/node/relay`` was killed 2026-05-12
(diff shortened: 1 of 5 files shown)


vivekchand:
## UX review

🟢 nit · Backend-only (one-line guard in app.js, no UI surface). Skipping detailed UX scrutiny per backend-only rule.

vivekchand:
## Product review

🟡 P1 · The pattern of **\`if (window.CLOUD_MODE) return;\`** at every cloud-disabled fetch site is now repeating across 4+ call sites (\`/api/config-diagnostics\`, \`/api/skills\`, \`/api/workspaces\`, ...). Each new route policy classification needs a matching OSS-side guard, but they're written by hand and easy to miss — exactly how this one slipped through #816 → caused the 0.12.183 release-gate failure → required #1185 hotfix → still hit a race in cloud #817. Action: extract a thin \`cloudDisabledFetch(path)\` helper or a single \`window.CLOUD_DISABLED_ROUTES\` Set so the OSS code can short-circuit by table lookup against the cloud route policy. Source it from a single shared file referenced by both the cloud route-policy classifier and the OSS frontend. Stops the chicken-and-egg release loop from happening on every new cloud-disabled endpoint.

🟢 nit · The fix unblocks the 0.12.183 release. No tier-strategy implications.

PR #1187, opened 2026-05-13
perf(install): use uv for ~5x faster pip install (with fallback)

vivekchand:
## Summary

- Bootstrap Astral's [uv](https://github.com/astral-sh/uv) (Rust-based pip replacement) at the top of the venv-install block in `install.sh`.
- When uv is available, use `uv venv` + `uv pip install`; otherwise fall back silently to the original `python3 -m venv` + `pip install --no-cache-dir` path.
- Surface progress with one line per step so the install no longer looks silent (a complaint earlier today).

## Why

User feedback: `curl | bash` install takes too long compared to OpenClaw. The bottleneck is `python3 -m venv` (3-5s) + `pip install clawmetry` downloading the ~14 MB duckdb wheel + transitive deps + resolution.

uv runs parallel downloads, ships a faster resolver, and creates venvs in milliseconds.

## Before / after timings

Measured on macOS (Darwin 25.3.0, M-series), cold cache, `--no-cache-dir` parity for fairness:

```
pip path:  bash -c   4.00s user 1.00s system 57% cpu 8.773 total
uv  path:  bash -c   0.35s user 0.27s system 32% cpu 1.880 total
```

**~8.8s -> ~1.9s = 4.6x faster.** Re-ran twice with cleared caches to rule out warm-cache skew, both runs landed in the same ranges.

## Fallback

Never hard-depend on uv. If the bootstrap `curl` hangs (corporate proxy), times out, or `astral.sh` is firewalled, the `|| true` swallows the error and the original pip path runs. Worst case: same speed as before, no regression.

```sh
if ! command -v uv >/dev/null 2>&1; then
  curl -LsSf https://astral.sh/uv/install.sh | sh >/dev/null 2>&1 || true
  export PATH="$HOME/.local/bin:$HOME/.cargo/bin:$PATH"
fi

if command -v uv >/dev/null 2>&1; then
  uv venv "$INSTALL_DIR" --quiet
  uv pip install --python "$INSTALL_DIR/bin/python3" --quiet --upgrade clawmetry
else
  # pip fallback unchanged
fi
```

## Test plan

- [x] `bash -n install.sh` passes
- [x] `pytest tests/test_install_script_linux_restart.py` -> 6/6 pass (Darwin/Linux/WSL post-install branches untouched)
- [x] End-to-end install into a throwaway dir; `clawmetry --version` reports `0.12.182`
- [ ] CI smoke (`install-test.yml`) on Ubuntu / macOS / Windows

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -61,7 +61,6 @@ esac
 
 # ── Install into venv ────────────────────────────────────────────────────────
 
-echo -e "  → Creating virtual environment..."
 # Preserve config before wiping venv (contains node_id, encryption_key)
 _CM_CFG_BAK=""
 if [ -f "$INSTALL_DIR/config.json" ]; then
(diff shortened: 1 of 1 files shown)


vivekchand:
## UX review

Pure install-script change — no UI surface. One bullet stop.

- 🟢 nit: progress lines are good, but the bracketed labels diverge across paths (`(uv)` vs `(pip)` vs `Bootstrapping pip fallback (this is slower)`). On the fallback path the user sees a scary "this is slower" line for what is currently the default for ~all corp-network installs. Soften to neutral language (e.g. `→ Creating virtual environment (pip)...`) — `install.sh:90`. The 5-second grok test should leave the user feeling fast either way, not apologetic.

## Product review

- 🟢 P1 (high-leverage win): 8.8s → 1.9s install is the single biggest activation-funnel improvement on the table. Slow `curl | bash` was a top OSS complaint and every dropped install is a lost Cloud-Pro funnel candidate. Consider surfacing the speed in the post-install line (`✓ ClawMetry $VERSION installed (1.9s)` — install.sh:215) so users notice and tell peers. Free word-of-mouth.
- 🟢 nit: `uv` bootstrap pulls from `astral.sh` over `curl | sh` with stderr swallowed — supply-chain attack on Astral's CDN now silently runs in our installer. Pin a uv version (`curl -LsSf https://astral.sh/uv/0.5.x/install.sh`) or skip bootstrap and require system uv with pip fallback. Low immediate risk; matters for the security-conscious enterprise prospect that compares us to Datadog.
- 🟢 nit: no telemetry on which path won (uv vs pip). Without that we can't tell what % of installs actually benefit. One-line POST to ingest after success would close the loop.

PR #1189, opened 2026-05-13
fix(brain): remove onboarding-text contamination from Live event stream

vivekchand:
## Summary
- Drop the substring-keyword fallback in `routes/brain.py`'s log scan that
  was tagging benign OpenClaw console lines as tool events
- Two real-world offenders the user hit at 22:50 today:
  - `Opened in your browser. Keep that tab to control OpenClaw.` rendered
    as `BROWSER` event with detail `browser. Keep that tab to control OpenClaw.`
  - `Token auto-auth included in browser/clipboard URL.` rendered as
    `BROWSER` with detail `browser/clipboard URL.`
- The bracketed `[tool] ...` log format is still parsed; only the loose
  `if "browser" in msg.lower(): emit BROWSER` branch is removed
- Real tool calls flow in via DuckDB `local_store` ingestion, so this
  fallback was pure noise (DuckDB-first rule)

## Root cause
`routes/brain.py:427-461` (legacy JSONL/log fallback) ran a substring scan
over ~12 keywords on every OpenClaw log line. Lines containing the substring
were sliced (`msg[start:start+300]`) and emitted as fake tool events, with
no validation that the line was actually a tool invocation. OpenClaw's
INFO-level lifecycle messages (browser-open hint, clipboard URL hint, etc.)
all matched.

## Test plan
- [x] New regression test `tests/test_brain_no_log_keyword_noise.py` —
  reproduces the bug on `main`, passes with this fix
- [x] Test asserts BOTH:
  - The exact strings from today's user report do NOT appear as events
  - A legit `[browser] navigate https://example.com` line STILL produces
    a BROWSER event (i.e. didn't kill the working code path)
- [x] Existing brain unit tests still pass (the 3 that fail on this branch
  also fail on `main` — daemon-state leakage, unrelated)
- [x] `python3 -c "import routes.brain"` clean
- [ ] After merge: verify `/api/brain-history` on a daemon-less install
  no longer surfaces "Keep that tab" / "clipboard URL" lines

### DIFF
--- a/routes/brain.py
+++ b/routes/brain.py
@@ -424,41 +424,17 @@ def api_brain_history():
                             "color": "#a855f7",
                         }
                     )
-                else:
-                    msg_lower = msg.lower()
-                    for kw in (
-                        "exec",
-                        "browser",
-                        "web_search",
-                        "web_fetch",
-                        "read",
-                        "write",
-                        "edit",
-                        "message",
-                        "spawn",
-                        "subagents",
-                        "tts",
-                        "nodes",
-                        "canvas",
-                    ):
-                        if kw in msg_lower:
-                            ev_type = tool_to_type(kw)
-                            try:
-                                start = msg_lower.index(kw)
-                                detail = msg[start : start + 300].split("\n")[0].strip()
-                            except Exception:
-                                detail = ""
-                            events.append(
-                                {
-                                    "time": ts,
-                                    "source": "main",
-                                    "sourceLabel": "main",
-                                    "type": ev_type,
-                                    "detail": detail,
-                                    "color": "#a855f7",
-                                }
-                            )
-                            break
+                # NOTE: a previous implementation also did substring keyword
+                # matching here ("if 'browser' in msg_lower → BROWSER event").
+                # That mis-classified benign console messages (e.g. "Opened in
+                # your browser. Keep that tab to control OpenClaw." or "Token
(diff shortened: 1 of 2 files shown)


vivekchand:
## UX review

Pure cleanup of a noisy fallback in the Brain feed — UI surface is the absence of garbage events. One bullet stop.

- 🟢 P1: this PR removes contamination but does nothing for the rows already in DuckDB from the buggy days — users on existing installs will still see "Keep that tab" / "clipboard URL" entries in the Brain stream until they age out of the window. Either (a) one-shot DELETE on daemon start that purges `event_type IN ('BROWSER','MSG','SPAWN',...) AND data->>'detail' LIKE '%Keep that tab%'`, or (b) document it. Otherwise the user re-tests the fix and it still looks broken.

## Product review

- 🟢 nit: data quality fix — exactly the kind of "things just work" polish that prevents free users from churning before they hit a paywall. No conversion lever directly, but a Brain feed full of fake events undermines trust in everything else (anomaly detection, alerts, usage analytics) which ARE paid surfaces. Worth shipping fast.
- 🟢 nit: the test asserts the bracketed `[browser]` path still emits, which is good — but the broader question is whether this fallback log-scan path should exist at all post-DuckDB-first. If 100% of real tool calls flow through `local_store`, the entire keyword-fallback branch is dead code. Worth a follow-up to delete `routes/brain.py:404-440` outright once telemetry confirms zero real users hit it.

PR #1190, opened 2026-05-13
fix(brain): read detail+source from nested v3 event payload (P0 regression in 0.12.182)

vivekchand:
## Summary

P0 — every event in the Brain tab rendered with empty body since the v3 sync mapper landed in #1143.

`/api/brain-history`'s local-store fast path looked for FLAT `data.input/summary/text/name` keys only. But:
- Legacy/trajectory ingest (event_type=`user`/`assistant`/`attachment`/`queue-operation`) nests content under `data.message.content` (string OR Anthropic block list).
- v3 mapper from #1143 (event_type=`prompt.submitted`/`model.completed`/`tool.result`/...) projects content onto top-level `data.completionText`/`data.finalPromptText`/`data.output`/`data.result` AND mirrors the same payload under `data.data.*`.

Neither path populates the flat keys → 100% of rows came back `detail=""`.

## Fix

New `_extract_brain_detail(row)` helper in `routes/brain.py` — checks all three shapes in priority order:

1. `data.message.content` (legacy) — flatten text/thinking blocks via `_v3_message_content_to_text`.
2. Top-level `data.{finalPromptText, completionText, output, result, input, summary, text, name, content}` (v3 mapper).
3. Mirror at `data.data.*` (v3 mapper second copy).
4. Type-specific fallbacks: `data.attachment.{filename,name}`, `data.modelId`, `data.cwd`, `data.id`, `data.operation`.

Encrypted thinking-only assistant turns return `"(thinking)"` placeholder instead of leaking the `cwd` path.

## Live verification

Against the running daemon at port 8904, after hot-patching `~/.clawmetry/lib/python3.9/site-packages/routes/brain.py`:

```
Pre-fix:  EMPTY: 10/10
Post-fix: EMPTY: 0/10
[ASSISTANT] '[[reply_to_current]] Still here, still Diya...'
[USER     ] 'Conversation info (untrusted metadata)...'
[TOOL.RESULT] 'tool result string'
[MODEL.COMPLETED] 'Yes — if you read this in the dashboard transcript...'
```

Re. `source`: investigated and the field is `src` (not `source`); it IS populated with the truncated session_id. The bug-report curl checked `e.get('source','?')` — wrong key. No fix needed.

## Test plan

- [x] New `tests/test_brain_history_v3_event_detail.py`:
  - `test_brain_history_detail_populated_for_v3_and_legacy_shapes` — seeds 5 rows (one per shape) into a tmp DuckDB and asserts every event in the response carries the expected snippet.
  - `test_extract_brain_detail_unit_handles_all_shapes` — pure-Python unit cover for the helper, no Flask/DuckDB.
- [x] Live daemon verified (port 8904): 10/10 → 0/10 empty.
- [x] `make lint` → syntax OK.

Refs #1143.

### DIFF
--- a/routes/brain.py
+++ b/routes/brain.py
@@ -78,6 +78,121 @@ def _brain_history_bool_arg(value):
     return str(value or "").strip().lower() in {"1", "true", "yes", "on", "all"}
 
 
+def _v3_message_content_to_text(content) -> str:
+    """Flatten an OpenClaw v3 ``data.message.content`` payload into plain text.
+
+    The trajectory ingest path (sync.py L2090-L2102) stores the WHOLE raw event
+    on ``data``, so user/assistant rows arrive with content nested under
+    ``data.message.content`` — either a plain string (user) or a list of typed
+    blocks (assistant: ``[{type:"text", text:...}, {type:"tool_use", ...}]``).
+    Only the ``text`` and ``thinking`` blocks carry transcript content; the
+    rest (tool_use, image, etc.) are summarised separately upstream.
+    """
+    if isinstance(content, str):
+        return content
+    if isinstance(content, list):
+        parts = []
+        for block in content:
+            if not isinstance(block, dict):
+                continue
+            btype = block.get("type")
+            if btype == "text":
+                t = block.get("text") or ""
+                if t:
+                    parts.append(t)
+            elif btype == "thinking":
+                t = block.get("thinking") or block.get("text") or ""
+                if t:
+                    parts.append(t)
+        return "\n".join(parts)
+    return ""
+
+
+def _extract_brain_detail(row: dict) -> str:
+    """Pull a human-readable ``detail`` snippet from a DuckDB event row.
+
+    Two ingest paths populate ``data`` with different shapes (P0 #1143 fix
+    landed the v3 mapper alongside the legacy trajectory parser); this helper
+    handles both plus the original flat-key fallback so brain-history never
+    returns empty strings for events that DO have content:
+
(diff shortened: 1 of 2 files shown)


vivekchand:
## UX review

Pure backend extractor change — UI surface is "Brain event chips now show body text instead of empty strings". One bullet stop.

- 🔴 P0 (UX): every event chip in the Brain tab has been rendering empty since 0.12.182 shipped. This PR fixes the read but no in-app indicator tells the user "data is back" — anyone who hit the empty state and bounced has no signal to come back. Recommend a one-time toast on first non-empty event load post-upgrade ("Live event details restored"), gated by a boolean in `~/.clawmetry/config.json`. Otherwise the regression silently costs us trust with users who saw broken Brain in the last 24h.

## Product review

- 🟡 P1: this is the second P0 brain-data regression in 7 days (#1189 today, this one from #1143). The Brain tab is a Cloud-Pro stickiness driver — it's where users actually spend visual time. Every empty-state hour erodes the case for paying. Add a smoke test in the [RELEASE] PR pipeline: post-merge, hit `/api/brain-history` against a seeded DuckDB and assert `count(detail!='')/count(*) > 0.8`. Fast, cheap, catches both regressions.
- 🟢 nit: `(thinking)` placeholder is the right call for encrypted-thinking turns — better than leaking `cwd`. Consider a follow-up that displays a small lock icon next to those rows so users understand WHY there's no body, and learn the model is using thinking blocks. Educational moment → upsell hook for "Brain Pro" (full thinking-block decryption surface).

PR #1192, opened 2026-05-13
feat(sync): ingest OpenClaw Telegram channel + fix(install): detect+kill duplicate daemons

vivekchand:
## Why

Two related bugs surfaced when re-installing ClawMetry on a machine with an active Telegram bot:

### Bug 1 — Telegram messages never reached DuckDB (P0 architectural gap)

`sync.py` only watched `~/.openclaw/agents/main/sessions/`. OpenClaw's chat-channel adapters write to **sibling per-provider directories** under `~/.openclaw/`:

```
~/.openclaw/telegram/<chat_id>.jsonl
~/.openclaw/signal/<chat_id>.jsonl
~/.openclaw/whatsapp/<chat_id>.jsonl
~/.openclaw/discord/<channel_id>.jsonl
~/.openclaw/slack/<channel_id>.jsonl
…  (21 dirs total — one per adapter in routes/channels.py)
```

Result: the Brain tab and `channel_messages` table missed every chat-channel turn. The user observed "I message Diya on Telegram and ClawMetry shows nothing."

### Bug 2 — `curl install.sh | bash` spawned a duplicate daemon

A new daemon appeared next to the existing pip-launched one. Both raced for the DuckDB write lock and every internal query 500'd:

```
Conflicting lock is held in /Users/vivek/.local/share/uv/python/
    cpython-3.11.15-macos-aarch64-none/bin/python3.11 (PID 74211)
```

The launchctl/systemd restart blocks added in #1183/#1187 only kick OS-managed jobs — they do nothing for daemons started by hand.

### `clawmetry/sync.py`

- New `sync_channel_messages(config, state, paths)` function tails every `~/.openclaw/<channel>/*.jsonl` for the canonical 21-adapter list (`_CHANNEL_DIRS`). Wired into the daemon loop next to `sync_sessions()`.
- Skips per-adapter bookkeeping files (`update-offset-default.json`, `schema.json`, `manifest.json`).
- Permissive event parser accepts the common shapes adapters emit: `text`/`body`/`message` for body; `from`/`sender`/`user` blocks for sender; `role=assistant` → `direction=out`.
- Each row lands in **BOTH** `events` (so Brain/timeline reads see it) AND `channel_messages` (so the per-provider routes in `routes/channels.py` see it). Honors the DuckDB-first hard rule — no JSONL re-read at request time.
- Per-file offsets persisted in `state["last_channel_offsets"]` so re-syncs are O(new bytes), not O(file size).

### `install.sh`

- **Pre-flight detect** at the top: `pgrep -f "clawmetry\.sync|clawmetry --port|clawmetry$"` sets `CLAWMETRY_RESTART_AFTER=1`.
- **Post-install kill** (after the launchctl/systemd restart blocks): when the flag is set, `pkill -f "clawmetry\.sync"` cleans the stray and waits 1s for the kernel to release the DuckDB lock.

### `CLAUDE.md`

- "How it works" section updated with the new channel-dir watch list.

## Tests

- `tests/test_telegram_ingest.py` (6 cases): inbound/outbound parse, sibling-channel pickup, bookkeeping-file skip, offset advance, append-on-next-cycle.
- `tests/test_install_script_no_duplicate_daemon.py` (5 cases): bash-n syntax, pre-flight detect, guarded pkill, post-launchctl ordering, in-guard placement.

```
$ python3 -m pytest tests/test_telegram_ingest.py \
    tests/test_install_script_no_duplicate_daemon.py \
    tests/test_install_script_linux_restart.py \
    tests/test_channels_local_store.py \
    tests/test_sync_skip_sidecar_jsonls.py -q
23 passed in 1.17s
```

Broader sync/channel/install sweep also clean (144 passed, 7 skipped).

## Test plan

- [ ] Merge → `[RELEASE]` PR → `pip install --upgrade clawmetry` on a node with active Telegram traffic
- [ ] Send a message to the bot from Telegram, confirm it appears in `/api/brain-history` within ~60s
- [ ] Confirm `SELECT COUNT(*) FROM channel_messages WHERE provider='telegram'` is non-zero
- [ ] Re-run `curl install.sh | bash`, confirm only ONE clawmetry pid afterwards (no lock 500s)

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -75,9 +75,14 @@ All HTTP endpoints live here, organised by feature. Each module owns one or more
 
 ## How it works
 1. Reads session transcripts from `~/.openclaw/agents/main/sessions/*.jsonl`
-2. Connects to OpenClaw gateway via WebSocket (JSON-RPC, port 18789) for live data
-3. Optionally receives OpenTelemetry metrics/traces on `/v1/metrics` and `/v1/traces`
-4. Serves dashboard UI at `http://localhost:8900`
+2. Reads chat-channel transcripts from `~/.openclaw/<channel>/*.jsonl` —
+   one directory per adapter (`telegram/`, `signal/`, `whatsapp/`,
+   `discord/`, `slack/`, `irc/`, `imessage/`, `webchat/`, …). The 21
+   adapter directories match the routes in `routes/channels.py`. New
+   adapter? Add its dir name to `_CHANNEL_DIRS` in `clawmetry/sync.py`.
+3. Connects to OpenClaw gateway via WebSocket (JSON-RPC, port 18789) for live data
+4. Optionally receives OpenTelemetry metrics/traces on `/v1/metrics` and `/v1/traces`
+5. Serves dashboard UI at `http://localhost:8900`
 
 ## API Endpoints (key ones)
 - `/api/overview` — Main dashboard data (sessions, tokens, crons, health)
--- a/clawmetry/sync.py
+++ b/clawmetry/sync.py
@@ -2472,6 +2472,326 @@ def sync_claude_cli_sessions(config: dict, state: dict, paths: dict) -> int:
     return total
 
 
+# ── Sync: channel-adapter transcripts (Telegram, Signal, WhatsApp, …) ────────
+#
+# OpenClaw's chat-channel adapters persist inbound/outbound messages to a
+# per-provider directory next to ``agents/``:
+#
+#   ~/.openclaw/telegram/<chat_id>.jsonl
+#   ~/.openclaw/signal/<chat_id>.jsonl
+#   ~/.openclaw/whatsapp/<chat_id>.jsonl
+#   ~/.openclaw/discord/<channel_id>.jsonl
+#   ~/.openclaw/slack/<channel_id>.jsonl
+#   ~/.openclaw/imessage/<chat_id>.jsonl
+#   ~/.openclaw/webchat/<session_id>.jsonl
+#   …                    (one dir per adapter — see routes/channels.py)
+#
+# Until this PR, ``sync.py`` only watched ``agents/main/sessions/`` which made
+# the Brain tab + ``channel_messages`` table miss every chat-channel turn —
+# the user observed "I message Diya on Telegram and ClawMetry shows nothing".
+#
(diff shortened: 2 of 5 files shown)


vivekchand:
## UX review

- 🟡 P1: the `↻ Existing clawmetry daemon detected — will restart it after upgrade` and `↺ Killed stray pre-existing daemon (lock released)` lines (`install.sh:75, 232`) are user-visible during `curl | bash`. Two different glyphs (`↻` vs `↺`) for the same lifecycle event reads as two unrelated things happening. Pick one and use the same verb pair (e.g. "Detected existing daemon" → "Restarted existing daemon"). 5-second grok test fails today.
- 🟢 nit: the install script now prints 8+ progress lines including pip/uv branching plus daemon restart messaging. Worth a final summary line (`✓ Installed | uv | restarted 1 daemon`) so users have one clear "what just happened" instead of scrolling buffer scrollback.

## Product review

- 🔴 P0: per memory `reference_openclaw_telegram_inmemory.md`, OpenClaw doesn't currently persist Telegram inbound to `~/.openclaw/telegram/*.jsonl` — meaning the 21-channel watcher this PR adds will silently find empty directories on real user machines. We're shipping a "fix" for "I message Diya on Telegram and ClawMetry shows nothing" that won't actually fix it. Either (a) gate the celebratory framing in the PR description until upstream OpenClaw lands the disk-persist change OR a gateway-WS tap PR lands here, or (b) tell users in release notes "Telegram visibility requires OpenClaw 0.x.y+ which writes channel transcripts to disk". Otherwise we burn the trust we're trying to rebuild.
- 🟡 P1: chat-channel ingest is a strong Cloud-Pro upsell vector (multi-channel consolidated view, cross-channel anomaly alerts, "agent went rogue on Slack" detection). Once the disk-persist gap is closed upstream, add a per-channel count badge to the cloud dashboard hero card and gate "alert if a channel goes silent for >N hours" behind Cloud-Pro. Free dashboard sees the count, paid tier acts on it.
- 🟢 nit: duplicate-daemon kill is the kind of un-celebrated fix that prevents 100% of "ClawMetry is broken" tickets from a re-install. Worth adding to the `/api/system-health` endpoint so the dashboard hero can self-diagnose ("warning: 2 daemons detected, run `clawmetry restart`") instead of users having to spot it.

PR #1193, opened 2026-05-13
test(e2e): Telegram→DuckDB→Brain ingest contract (blocked on aca53ec8)

vivekchand:
## Summary
Pins down the end-to-end shape the chat-channel ingest path MUST satisfy. Six tests cover the full pipeline: OpenClaw filesystem → `clawmetry.sync._ingest_channel_directory` → DuckDB `channel_messages` table → `/api/brain-history`.

All tests are **currently skipped** via a module-level `pytestmark` that probes for `clawmetry.sync._ingest_channel_directory`. The function does not exist yet — it's the deliverable of agent `aca53ec8`'s in-flight PR. When that lands, the import succeeds and every test in this file starts running automatically. No follow-up edit needed here.

Why land this empty-running file now:
1. Forces a clear regression contract for the ingest path before the implementation lands.
2. Catches future regressions in any link of the chain (filesystem reader, DuckDB writer, brain-history reader, JSON shaping).
3. Documents the expected end-to-end JSON shape for reviewers of the ingest PR.

### `test_telegram_message_flows_end_to_end` (primary)
- DuckDB row count for `provider="telegram"` is exactly 1 after ingest.
- Row `provider == "telegram"`, `sender_name == "tester-one"`, `direction == "in"`.
- Row `body` is non-null and contains `"hello"` (case-insensitive).
- `/api/brain-history?limit=20` returns HTTP 200.
- Response contains at least one event with `channel == "telegram"` (or `source` starting with `telegram`).
- That event's `detail` is non-empty (regression for the bug fixed in #1190).
- That event's `channel == "telegram"` and `sender == "tester-one"` (or `sender_name` equivalent).

### `test_outbound_message_renders_as_agent`
- `direction: "out"` event lands in DuckDB with `direction == "out"`.
- `/api/brain-history` classifies it as agent (via `direction`, `actor`, `role`, or `type` field) — distinct from inbound.

### `test_multiline_text_survives_roundtrip`
- A body with `\n` newlines round-trips through DuckDB byte-identical (no JSONL re-encoding mangling).

### `test_unicode_emoji_byte_identical`
- A body with emoji 👋 + non-ASCII (`café`, `naïve`, `façade`, `—`) round-trips through DuckDB byte-identical.

### `test_missing_directory_is_silent_noop`
- Calling `_ingest_channel_directory` on a non-existent dir does NOT raise.
- Produces zero rows. Protects every caller that loops over all known providers.

### `test_two_channels_at_once`
- Telegram + Signal sibling directories ingest independently.
- Provider filter on `query_channel_messages` returns each one only with its own messages (no cross-contamination).

## Constraints honoured
- `monkeypatch` for env vars — no global mutation.
- `tmp_path` for both `OPENCLAW_HOME` and `CLAWMETRY_HOME` — never touches `~/.openclaw` or `~/.clawmetry`.
- No celebrity / real-name strings in payloads.

## Test plan
- [x] `pytest tests/test_e2e_telegram_brain_ingest.py --collect-only` — 6 tests collected.
- [x] `pytest tests/test_e2e_telegram_brain_ingest.py -v` — 6 skipped, 0 failed (pre-merge of aca53ec8).
- [ ] After aca53ec8 merges: re-run, expect 6 passed.
- [ ] If any test fails post-merge, treat as a contract violation in the ingest PR — not a flake here.

### DIFF
--- a/tests/test_e2e_telegram_brain_ingest.py
+++ b/tests/test_e2e_telegram_brain_ingest.py
@@ -1,0 +1,393 @@
+"""E2E contract test: Telegram messages → DuckDB → /api/brain-history.
+
+This file pins down the END-TO-END shape that the Telegram (and any other
+chat-channel) ingest path MUST satisfy. It is intentionally written
+*before* the ingest path lands so it acts as a regression contract for
+the in-flight work tracked under agent ``aca53ec8``.
+
+Pipeline under test:
+
+    OpenClaw filesystem (per-chat .jsonl)
+        →  ClawMetry sync.py (``_ingest_channel_directory``)
+        →  Local DuckDB (``channel_messages`` table)
+        →  /api/brain-history (renders chat_id/sender/body)
+
+Why this exists:
+  1. Forces a clear regression contract for the Telegram ingest path so
+     future refactors can't silently break it.
+  2. Catches regressions in any link of the chain (filesystem reader,
+     DuckDB writer, brain-history reader, JSON shaping).
+  3. Documents the expected end-to-end JSON shape for reviewers.
+
+Why every test is skipped today:
+  ``clawmetry.sync._ingest_channel_directory`` does not exist yet — it is
+  the deliverable of the ingest PR (agent ``aca53ec8``). When that PR
+  merges, the ``_INGEST_FN`` import below resolves and every test in the
+  module starts running automatically. No edits required here.
+
+Hard rules followed:
+  * ``monkeypatch`` for env vars (no global mutation).
+  * ``tmp_path`` everywhere — never touches ``~/.openclaw`` or
+    ``~/.clawmetry``.
+  * No celebrity / real-name strings in payloads (per project policy).
+"""
+
+from __future__ import annotations
+
+import importlib
+import json
+import time
+
+import pytest
+from flask import Flask
+
+
+# --------------------------------------------------------------------------- #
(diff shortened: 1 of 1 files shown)


vivekchand:
## UX review

- 🟢 nit: Pure backend test-infra PR — no user-facing surface to scrutinize. Skipped pending aca53ec8/#1192 → auto-unskip on merge. Nothing to action here.

vivekchand:
## Product review

- 🟡 P1 (`tests/test_e2e_telegram_brain_ingest.py:67`): The skip predicate keys on a single private symbol (`_ingest_channel_directory`). Per #1198 the real ingest landed as `sync_channel_messages` + `sync_telegram_from_gateway_log` — neither is `_ingest_channel_directory`. Risk: this contract file stays permanently skipped (silent green) while everyone assumes coverage. Action: rename the probe to `sync_channel_messages` (or `or` both) before merge so the contract actually fires once #1192 lands. Otherwise it's dead-code masquerading as a regression net.
- 🟢 nit: The 6-test contract (multiline, emoji, no-op missing dir, sibling provider) is high-quality scaffolding for the multi-channel adapter direction. Worth keeping even if Telegram itself lands via a different code path — generalize the helpers to point at whatever the real ingest fn ends up being, not a placeholder name.
- 🟢 nit: No conversion impact — pure infra. Move on.

PR #1194, opened 2026-05-13
feat(sync): include channel events in brain cache push (cloud Brain renders Telegram)

vivekchand:
## Summary

- Extends the OSS daemon's heartbeat-piggyback brain cache push (`_build_brain_cache_pushes`) so the cloud Brain renderer can show "Telegram: Vivek Chand: hello, how are you doing?" instead of generic agent activity for channel events.
- Adds `_channel_enrichment_from_row()` that decomposes the composite event id (`{provider}:{chat_id}:{raw_id}`) and the inner adapter payload (`from`/`sender`/`user` block) into flat top-level keys (`provider`, `chat_id`, `sender_id`, `sender`, `direction`).
- Non-channel rows (regular CLI session messages) are untouched — no key pollution on the existing OpenClaw transcript shape `transformEvents` already understands.
- E2E encryption invariant unchanged: enriched payload still flows through `encrypt_payload()` before leaving the daemon. Test asserts plaintext (chat ids, sender names, message bodies) never appears in the ciphertext blob.

## Coordination with PR #1191 (fix/telegram-ingest-and-install-lock-detect)

PR #1191 (commit `2dff811` by agent aca53ec8) added `sync_channel_messages()` which writes channel transcripts to DuckDB with:
- `event_type` = `"channel.in"` or `"channel.out"`
- `id` = `f"{provider}:{channel_id}:{raw_id}"` (provider only lives here — the events table has no provider column)
- `data` = the original adapter line

This PR consumes those exact contract points on the read side. **Order-independent**:
- If this PR lands first → enrichment is dormant (returns `{}` for non-channel rows; there are 0 channel rows in DuckDB until #1191 lands).
- Once #1191 lands → channel rows automatically pick up the enrichment on the next heartbeat. Zero coupling between merge order.

No changes to `routes/brain.py` (#1190 just landed there).

## DuckDB-first hard rule

Events come from `store.query_events()` — **not** a re-tail of `~/.openclaw/<channel>/`. Honors the project's DuckDB-first invariant.

## Test plan

- [x] Unit: `tests/test_brain_cache_push_channel_events.py` — 6 cases (Telegram inbound enrichment, Telegram outbound direction=out + sender fallback, Signal sibling provider, non-channel rows untouched, ciphertext leaks no plaintext, helper handles malformed rows incl. Matrix-style ids with embedded colons).
- [x] Existing: `tests/test_brain_cache_push.py` — 5/6 still pass (one pre-existing failure on stale shape assertion, present on `main` already).
- [x] Adjacent: `tests/test_brain_local_fastpath.py`, `test_brain_local_store.py`, `test_brain_history_v3_event_detail.py` — all pass.
- [ ] E2E (post-merge): Telegram message → DuckDB → heartbeat cache push → cloud Brain panel shows "Telegram: <sender>: <text>".

### DIFF
--- a/clawmetry/sync.py
+++ b/clawmetry/sync.py
@@ -2973,6 +2973,64 @@ def _local_dispatch_fallback(shape: str, args: dict) -> dict:
     return {"rows": rows, "count": len(rows), "_shape": shape, "_via": "fallback"}
 
 
+def _channel_enrichment_from_row(r: dict) -> dict:
+    """Pull (provider, chat_id, sender_id, sender, direction) out of a
+    channel.* events row so the cloud Brain renderer can display
+    "Telegram: Vivek Chand: hello" instead of generic agent activity.
+
+    The ingest path in ``sync_channel_messages`` (PR #1191) writes channel
+    transcripts with:
+      * ``event_type`` = ``"channel.in"`` or ``"channel.out"``
+      * ``id`` = ``f"{provider}:{channel_id}:{raw_id}"`` — the only place
+        ``provider`` lives on the events row, since the events table has
+        no provider column.
+      * ``data`` = the original adapter line (``from``/``sender``/``user``
+        block, ``text``/``body``, etc.).
+
+    Returns ``{}`` for non-channel rows so callers can spread it
+    unconditionally. Never raises.
+    """
+    et = r.get("event_type") or ""
+    if not isinstance(et, str) or not et.startswith("channel."):
+        return {}
+    direction = et.split(".", 1)[1] if "." in et else ""
+    eid = r.get("id") or ""
+    provider = ""
+    chat_id = ""
+    if isinstance(eid, str) and eid.count(":") >= 2:
+        # ``{provider}:{channel_id}:{raw_id}`` — raw_id may itself contain
+        # colons (rare, but defensible: split only the first two segments).
+        parts = eid.split(":", 2)
+        provider, chat_id = parts[0], parts[1]
+    data = r.get("data") if isinstance(r.get("data"), dict) else {}
+    sender_block = (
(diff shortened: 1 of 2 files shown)


vivekchand:
## UX review

- 🟢 nit: Pure backend enrichment in `clawmetry/sync.py` — no OSS UI surface touched. The visible payoff lives on the cloud Brain renderer, which gets reviewed in clawmetry-cloud, not here. Nothing to action OSS-side.

vivekchand:
## Product review

- 🔴 P0: This enrichment is dormant until DuckDB has channel.in/channel.out rows. Per memory `reference_openclaw_telegram_inmemory.md`, OpenClaw never persists Telegram → so until #1198's gateway.log parser AND `sync_channel_messages` (#1191) both ship, cloud Brain shows the same blank Telegram experience users complain about. Action: do not announce "cloud Brain renders Telegram" anywhere user-facing until #1198 is merged AND a paid user with a live bot has been verified to see "Telegram ↘ <sender>" on cloud. Otherwise we cut a release that promises the conversion and delivers silence.
- 🟡 P1 (`clawmetry/sync.py:3025`): Outbound rows from #1198 land with `body=None` (ack-only) AND with the new flat `provider`/`chat_id` enrichment. Cloud renderer will paint "Telegram ↗ agent — (empty)". That's worse than nothing for a paying user evaluating remote-observability — it looks like the bot replied with nothing. Action: when `body` is None, make the cloud Brain row label it explicitly "(body not captured — see Settings)" or skip ack-only rows from the cache push entirely until #1198 captures bodies. Better silent than misleading.
- 🟢 nit: Order-independence guarantee (lines #56-60 of PR body) is the right call — keeps the merge train moving.

PR #1176, opened 2026-05-13
fix(ui): Brain density + timestamp year + Flow active tools + mc-tasks 404 + stuck-banner persistence (closes #1127 frontend bucket)

vivekchand:
## Summary

Five fixes for the frontend bucket of #1127, all landing in `clawmetry/static/js/app.js` (+ one banner template edit + one dead-call removal in `dashboard.py`):

- [x] **Brain "Activity density" chart renders only background, no bars** — adaptive window (1h / 6h / 24h, 120 buckets each) replaces the hard-coded 60-min window that silently dropped every event older than an hour.
- [x] **Brain Live event stream timestamps show "9 May" — year missing** — `formatBrainTime()` now includes the year when the event is from a different calendar year than "now" (same-day → "Today"; same-year → "9 May"; cross-year → "9 May 2025").
- [x] **Flow `Active Tools: ─` empty + Live Tool Call Stream stuck on "Waiting for activity…"** — backfilled from `/api/brain-history` and live-updated from `/api/brain-stream` SSE, both DuckDB-backed (per the DuckDB-first rule). No new gateway poll, no JSONL read.
- [x] **`GET /api/mc-tasks` → 404 (~11 fires per audit walk)** — route was deleted in commit 62e1fe7 when Mission Control was retired but the polling code (initial fetch + 30s interval) was orphaned. Deleted the whole dead block (`loadMCTasks` / `toggleMCColumn` / `renderMCExpanded` / `_mcRefreshTimer` / module state) plus the duplicate call in `dashboard.py`'s inline script. No `mc-bar-wrapper` / `mc-summary-bar` elements existed in templates, so this is purely dead code.
- [x] **Stuck-session warning banner resurfaces after page reload despite Dismiss** — new `dismissStuckBanner()` writes a per-session timestamp into `localStorage` (key `stuck-session-dismissed`); `_showStuckBanner` skips sessions dismissed within 24h. Self-pruning store drops entries older than the 24h TTL on every read so it can't grow unbounded.

## Test plan

- [x] `node --check clawmetry/static/js/app.js` — JS parse OK
- [x] `python3 -c "import ast; ast.parse(open('dashboard.py').read())"` — Python parse OK
- [x] `node tests/test_appjs_units.js` — 18 sandboxed unit tests pass (regex-extracted from app.js so they exercise the actual shipped source). Covers all 3 branches of `formatBrainTime`, plus dismissal store TTL / prune / round-trip across simulated reload.
[...]

PR #1197, opened 2026-05-13
feat(ui): render channel events with provider+sender labels (Telegram, Signal, Slack...)

vivekchand:
## Summary

Brain rows for chat-channel turns (Telegram, Signal, Slack, …) currently
show the same generic "USER"/"AGENT" type tag as a CLI run, leaving the
user to guess which channel a message came from. Now that the Telegram-
ingest agent (#1192) is in and channel events carry `chat_id` / `sender` /
`provider` / `direction`, the Brain renderer can replace the type tag
with a provider pill, an italic sender name, and an inbound/outbound
arrow:

> 📱 Telegram  ↘  *Vivek Chand* — hello, how are you doing?

## Changes

- `clawmetry/static/js/app.js`
  - Expand `_channelIcons` / `_channelColors` to the canonical 21 adapters
    (matches `routes/brain.py` `_CHANNEL_ICON` and `clawmetry/sync.py`
    `_CHANNEL_DIRS`). Add `_channelDisplayNames` for non-trivial labels
    (`googlechat` → `Google Chat`, `msteams` → `MS Teams`, etc.).
  - New `_extractChannelInfo(ev)` accepts BOTH the enriched flat shape
    (top-level `provider` / `sender` / `chat_id`) AND the raw `ev.data.{…}`
    fallback so the renderer works whether or not the brain endpoint
    enrichment runs first. Returns `null` for non-channel events so the
    legacy AGENT/USER/EXEC render branch still owns them.
  - New `renderChannelEventMeta(ev, info)` paints the provider pill +
    direction arrow (↘ inbound, ↗ outbound) + italic sender name in
    place of the type tag + source span. `renderBrainStream` swaps in
    this branch when `_extractChannelInfo` returns non-null. Detail body
    is untouched — `renderBrainDetail` keeps owning it.

- `routes/brain.py`
  - `_try_local_store_brain` enriches `event_type=channel.in|out` rows
    with flat `provider` / `channel` / `sender` / `chat_id` / `direction`
    fields lifted from `data` (using the same permissive sender-block
    parsing the ingest agent writes with). The pre-existing JSONL parser
    path already exposed `channel`; this gives the local-store fast path
    parity.

- `tests/test_appjs_units.js`
  - 31 new assertions extending the existing PR #1176 harness — no
    JSDOM, no Playwright, ~50 ms.

## Coordination with #1192

Field names match the ingest schema introduced in
[#1192](https://github.com/vivekchand/clawmetry/pull/1192) (commit
2dff811): `provider`, `sender_name`, `chat_id`, `direction` (`in`/`out`),
plus the nested `data.{from,sender,user}` block as the secondary source.
The renderer falls back gracefully if those fields are absent (legacy
renderer remains the default), so the change is safe even on the
small set of installs that haven't picked up the daemon-side fix yet.

## Test plan

- [x] `node tests/test_appjs_units.js` — 50 / 50 pass (19 pre-existing +
      31 new for the channel branch).
- [x] `python3 -m pytest tests/test_brain_local_fastpath.py
      tests/test_brain_local_store.py
      tests/test_brain_history_v3_event_detail.py -q` — 8 / 8 pass
      (existing brain endpoint behaviour unchanged).
- [x] Smoke-tested the brain.py enrichment: a synthetic
      `channel.in` row with `data={provider: telegram, sender_name:
      "Vivek Chand", chat_id: "1532693273"}` is enriched into
      `{provider: telegram, channel: telegram, direction: in, sender:
      "Vivek Chand", chat_id: "1532693273"}`.
- [ ] Local OSS install with a live Telegram bot — verify the Brain tab
      shows `📱 Telegram ↘ Vivek Chand — <message>` for inbound and
      `📱 Telegram ↗ agent — <reply>` for outbound.

### DIFF
--- a/clawmetry/static/js/app.js
+++ b/clawmetry/static/js/app.js
@@ -2282,16 +2282,159 @@ var _brainTypeFilter = 'all';
 var _brainChannelFilter = 'all';
 var _brainAllEvents = [];
 
+// Provider → emoji + display name. Mirrors routes/brain.py `_CHANNEL_ICON`
+// and clawmetry/sync.py `_CHANNEL_DIRS` (the canonical 21-adapter list).
+// Keep these three in sync when a new adapter ships.
 var _channelIcons = {
-  'telegram': '📱', 'whatsapp': '💬', 'discord': '🎮', 'slack': '📢',
+  'telegram': '📱', 'whatsapp': '💬', 'discord': '🎮', 'slack': '💼',
   'signal': '📡', 'irc': '💻', 'imessage': '🍎', 'webchat': '🌐',
-  'googlechat': '📧', 'cli': '🖥️', 'cron': '⏰'
+  'googlechat': '🔵', 'matrix': '🔢', 'msteams': '🏢', 'mattermost': '⚡',
+  'line': '💚', 'nostr': '🟣', 'twitch': '💜', 'bluebubbles': '💙',
+  'feishu': '🟠', 'zalo': '🩵', 'tlon': '🟤', 'synologychat': '🟦',
+  'nextcloudtalk': '☁️',
+  'cli': '🖥️', 'tui': '⌨️', 'cron': '⏰'
 };
 var _channelColors = {
   'telegram': '#2f9ef4', 'whatsapp': '#25d366', 'discord': '#5865F2', 'slack': '#4A154B',
   'signal': '#3a76f0', 'irc': '#6B7280', 'imessage': '#34C759', 'webchat': '#0EA5E9',
-  'googlechat': '#1A73E8', 'cli': '#94a3b8', 'cron': '#6B7280'
+  'googlechat': '#1A73E8', 'matrix': '#0DBD8B', 'msteams': '#4B53BC', 'mattermost': '#0072C6',
+  'line': '#06C755', 'nostr': '#9333ea', 'twitch': '#9146FF', 'bluebubbles': '#3478F6',
+  'feishu': '#00D6B9', 'zalo': '#0068FF', 'tlon': '#A78BFA', 'synologychat': '#1A73E8',
+  'nextcloudtalk': '#0082C9',
+  'cli': '#94a3b8', 'tui': '#94a3b8', 'cron': '#6B7280'
 };
+// Display-name overrides for channels whose snake/lower-case key isn't a
+// great human label. (`telegram` → `Telegram` is fine via title-case;
+// `googlechat` → `Google Chat` is not.)
+var _channelDisplayNames = {
+  'googlechat': 'Google Chat',
+  'msteams':    'MS Teams',
+  'bluebubbles':'BlueBubbles',
+  'imessage':   'iMessage',
+  'webchat':    'WebChat',
+  'irc':        'IRC',
+  'cli':        'CLI',
+  'tui':        'TUI',
+  'synologychat': 'Synology Chat',
+  'nextcloudtalk': 'Nextcloud Talk'
+};
+
+function _channelDisplayName(provider) {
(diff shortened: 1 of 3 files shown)


vivekchand:
## UX review

- 🔴 P0 (`clawmetry/static/js/app.js:2369-2376`): Sender max-width is hardcoded to `160px` with `overflow:hidden;text-overflow:ellipsis`. On a Telegram group chat with names like "Vivek Chandrasekharan" or "+15551234567 (Alice from Operations)", the truncation gives users no way to disambiguate two distinct conversations sharing a chat-pill. Worse, hover title only shows the sender (not chat_id), so the user can't tell who's who. Action: either bump max-width to `min(40ch, 30vw)` or include chat_id in the hover tooltip. 5-second grok test fails when two senders truncate to the same prefix.
- 🟡 P1 (`clawmetry/static/js/app.js:2310`): The fallback for missing provider matching `/^(?:agent:[^:]+:)?([a-z]+)(?::|$)/` accepts any lowercase token in source. A future event `source: "agent:main:hermes"` will silently match `hermes` as a provider, miss the `_channelIcons` lookup, fall through to the placeholder `📨`, and render an empty meta row labelled `Hermes`. Action: gate on `_channelIcons[maybe]` actually existing AND not return `null` when the lookup fails — fail closed to the legacy renderer.
- 🟡 P1 (`clawmetry/static/js/app.js:2399`): Outbound fallback sender is the literal string `agent` (lowercase, no styling). Looks like a dev placeholder. Worse, the inbound fallback is `user` — both will show up identically across providers, defeating the entire point of the channel pill. Action: use the provider's display name + arrow (e.g. "Telegram bot" / "Telegram contact") or omit the sender row entirely when missing.
- 🟢 nit: Channel pill emoji set is a maintenance trap — three sources (app.js, routes/brain.py, sync.py) MUST stay in sync per the PR body. Add a CI test that asserts the three dicts have identical keys, or move to a single JSON in `static/`.
- 🟢 nit (`clawmetry/static/js/app.js:2287`): Slack emoji changed from 📢 to 💼 with no version note. Minor but: the megaphone read better as "broadcast channel"; briefcase reads as "work" which conflicts with msteams 🏢. Pick one mental model.

vivekchand:
## Product review

- 🔴 P0: 21 channel adapter pills are now rendered with first-class branding (Slack purple, Discord indigo, Telegram blue) but per memory `project_alerts_pro_feature.md` and the multi-channel observability narrative, multi-channel observation IS a Cloud-Pro story. OSS is shipping the polish for free with no upsell. Action: add a "Connected channels: 3 / 21 — see all in Cloud" CTA banner above the Brain stream when the user has at least one channel.in row. Every non-CLI provider rendered = one more conversion opportunity left on the table.
- 🟡 P1 (`clawmetry/static/js/app.js:2329-2331`): The "Treat plain CLI/cron sources as non-channel" branch silently filters out the highest-volume row class from getting the new pill treatment. Defensible (CLI has its own chip), but: cron + CLI are the legacy free-tier story. Channels are the paid story. Visually demoting cron/CLI by giving channels prettier rendering reinforces the upsell — keep the asymmetry, but mark it explicitly in product narrative ("channel pills are a Cloud-tier feature preview").
- 🟡 P1: 21 emoji icons but no test that the user actually has any of these channels configured. Empty Brain → fancy renderer → still empty → confused user. Coordinate with #1198 (gateway.log parser is the only path that yields rows today) before claiming this works for any channel beyond Telegram. Telegram-only provider list shipping as "21 providers" is overpromising.
- 🟢 nit: XSS escape test is solid (line 309 of test file). Don't lose that on future refactor.

PR #1198, opened 2026-05-13
feat(sync): parse gateway.log for Telegram messages (ingest into DuckDB)

vivekchand:
## Why

OpenClaw runs Telegram direct-chat sessions **entirely in memory**. No per-session JSONL is ever written under `~/.openclaw/agents/main/sessions/` or `~/.openclaw/telegram/` for inbound or outbound Telegram traffic — confirmed today by diagnostic agent on the user's live machine. The only on-disk evidence of Telegram activity is regex-recoverable from `~/.openclaw/logs/gateway.log`.

PR #1192 (just merged) wired up watching `~/.openclaw/telegram/` defensively for the day OpenClaw starts persisting, but until that upstream change lands ClawMetry's Brain feed silently misses **all** Telegram activity. This PR is the only path that works today without an OpenClaw upstream change AND backfills history on first install.

## What's captured

Outbound API ACK lines, e.g.

```
2026-05-13T22:54:19.865+02:00 [telegram] sendMessage ok chat=1532693273 message=8491
2026-05-13T06:00:56.332+02:00 [telegram] sendPhoto    ok chat=1532693273 message=8480
```

Each becomes a `channel_messages` row in DuckDB with:
- `id = "telegram:<chat>:<msg>"` (PRIMARY KEY → idempotent re-ingest)
- `provider = "telegram"`, `direction = "out"`, `channel_id = "telegram:<chat>"`
- `body = None`
- `raw_blob = {source: "gateway.log", method, body_capture: "ack_only", ...}`

Methods covered: `sendMessage`, `sendPhoto`, `sendAudio`, `sendDocument`, `sendVideo`, `sendVoice`, `sendSticker`, `sendAnimation`, `sendLocation`.

## What's NOT captured

- **Outbound message bodies** — `gateway.log` logs only the API ACK, never the message payload. The body lives only in OpenClaw's in-memory session state. We set `body=None` (rather than a placeholder string) so the renderer can show an "(no body captured)" affordance and isn't tricked into rendering literal placeholder text as user content.
- **Inbound messages** — at every verbosity level we observed (today, 2026-05-13), the production log emits **only** long-poll diagnostics for inbound (`[telegram] Polling stall detected`, `[diag] polling cycle finished`). Message bodies are not logged. If a future OpenClaw release adds an inbound trace line we can extend the parser, but today there is nothing to parse.

## Long-term fix path

1. **OpenClaw upstream**: persist inbound Telegram updates to a session JSONL like every other channel does. The right thing — out of ClawMetry's direct control.
2. **Live gateway WebSocket tap** on `ws://localhost:18789` — real-time, no log parsing, but only catches messages received while the ClawMetry daemon is running (no historical backfill).

## Implementation notes

- **Tail-and-resume**: per-log-path byte offset persisted in `state['last_log_offsets']` (existing `~/.clawmetry/sync-state.json` schema, no new file).
- **Log rotation safety**: if `os.path.getsize(log_path) < prev_offset` we restart from byte 0; the `channel_messages` PRIMARY KEY makes the re-scan idempotent.
- **Partial-line safety**: only consume up to the last `\n` each cycle; the trailing fragment is read again next cycle when it completes.
- **Pre-filter**: skip lines without `[telegram]` before regex match (~99.99% of `gateway.log` lines, big perf win).
- **Best-effort**: every failure mode swallowed with a warning so the daemon never crashes if `local_store` is unavailable or the log is mid-rotation.

Wired into the main daemon `while True` cycle right after `sync_logs(...)`.

## Test plan

- [x] 17 new unit tests in `tests/test_telegram_gatewaylog_parser.py` covering:
  - real production line shapes for `sendMessage` + 8 other media APIs
  - negative chat IDs (Telegram groups)
  - both `+HH:MM` and `Z` timestamp suffixes
  - 6 negative cases including inbound polling-stall noise, signal channel, failed sends
  - end-to-end ingest into a tmp DuckDB → assert row shape via SQL
  - tail-and-resume across cycles
  - log truncation / rotation
  - partial trailing line preservation
  - idempotent re-ingest with empty state
  - missing log file no-op
- [x] Existing `test_sync_skip_sidecar_jsonls`, `test_sync_main_block_at_eof`, `test_channels_local_store`, `test_local_store_sync_integration` all still pass (22/22).
- [ ] Manual: install on user's machine, confirm Brain feed shows today's outbound Telegram ACKs after one sync cycle.

### DIFF
--- a/clawmetry/sync.py
+++ b/clawmetry/sync.py
@@ -2973,6 +2973,64 @@ def _local_dispatch_fallback(shape: str, args: dict) -> dict:
     return {"rows": rows, "count": len(rows), "_shape": shape, "_via": "fallback"}
 
 
+def _channel_enrichment_from_row(r: dict) -> dict:
+    """Pull (provider, chat_id, sender_id, sender, direction) out of a
+    channel.* events row so the cloud Brain renderer can display
+    "Telegram: Vivek Chand: hello" instead of generic agent activity.
+
+    The ingest path in ``sync_channel_messages`` (PR #1191) writes channel
+    transcripts with:
+      * ``event_type`` = ``"channel.in"`` or ``"channel.out"``
+      * ``id`` = ``f"{provider}:{channel_id}:{raw_id}"`` — the only place
+        ``provider`` lives on the events row, since the events table has
+        no provider column.
+      * ``data`` = the original adapter line (``from``/``sender``/``user``
+        block, ``text``/``body``, etc.).
+
+    Returns ``{}`` for non-channel rows so callers can spread it
+    unconditionally. Never raises.
+    """
+    et = r.get("event_type") or ""
+    if not isinstance(et, str) or not et.startswith("channel."):
+        return {}
+    direction = et.split(".", 1)[1] if "." in et else ""
+    eid = r.get("id") or ""
+    provider = ""
+    chat_id = ""
+    if isinstance(eid, str) and eid.count(":") >= 2:
+        # ``{provider}:{channel_id}:{raw_id}`` — raw_id may itself contain
+        # colons (rare, but defensible: split only the first two segments).
+        parts = eid.split(":", 2)
+        provider, chat_id = parts[0], parts[1]
+    data = r.get("data") if isinstance(r.get("data"), dict) else {}
+    sender_block = (
+        data.get("from") or data.get("sender") or data.get("user") or {}
+    )
+    if not isinstance(sender_block, dict):
+        sender_block = {}
+    sender_id = (
+        data.get("sender_id")
+        or sender_block.get("id")
+        or data.get("user_id")
+        or chat_id
+    )
+    sender = (
+        data.get("sender_name")
+        or sender_block.get("username")
+        or sender_block.get("first_name")
+        or sender_block.get("name")
+        or ""
+    )
+    out = {
+        "provider":  provider,
+        "chat_id":   str(chat_id) if chat_id != "" else "",
+        "sender_id": str(sender_id) if sender_id is not None else "",
+        "sender":    str(sender),
+        "direction": direction,
+    }
+    return out
+
+
 def _rows_to_brain_events(rows: list) -> list:
     """Return raw OpenClaw event payloads ready for the cloud browser's
     ``transformEvents`` to unwrap.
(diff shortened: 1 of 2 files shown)


vivekchand:
## UX review

- 🟡 P1: Pure backend ingest — but produces user-facing rows with `body=None`. Per #1197's renderer, those rows render as "📱 Telegram ↗ agent — " with empty body and no affordance explaining why. From a 5-second grok perspective, the user concludes "ClawMetry shows my Telegram bot replied with nothing." Action: ingest a sentinel body like `"(message body not captured — OpenClaw stores Telegram in memory)"` OR coordinate with #1197 to render a dedicated "(no body captured)" affordance. Empty rows look broken.

vivekchand:
## Product review

- 🔴 P0: This ships outbound-only, body=None Telegram observability — and the PR body itself acknowledges (1) inbound is never captured, (2) outbound bodies are never captured. Net user value: a count of how many Telegram messages the bot sent, with no content and no inbound context. For a paying Cloud-Pro user expecting "remote Telegram monitoring," that's a feature that looks like a bug. Action: do not advertise "Telegram support" in CHANGELOG / release notes off this PR alone. Frame it as "Telegram outbound activity tracking (counts only)" until the OpenClaw upstream fix or live WS tap lands.
- 🔴 P0: Per memory `feedback_duckdb_first_rule.md` ("JSONL/log/process-stat reads are violations"), this PR parses gateway.log — exactly the violation the rule prohibits. The rationale (OpenClaw doesn't persist) is valid but should be a temporary, time-boxed exception with a tracking issue for the upstream fix, not a permanent code path. Action: open an OpenClaw upstream issue referenced in the code comment, and add a `# DEPRECATED WHEN OPENCLAW PERSISTS` marker so future reviewers know to delete this path.
- 🟡 P1 (`clawmetry/sync.py:6588-6595`): Pre-filter `if "[telegram]" not in raw_line: continue` is great for Telegram, but it's a one-shot — the next adapter that needs the same workaround (Signal? Discord?) will copy this 250-line block. Action: extract a generic `_GATEWAY_LOG_PARSERS` registry now while only Telegram is in it; one-time refactor cost beats N copies later.
- 🟢 nit: Idempotency via PRIMARY KEY + byte-offset state is the right belt-and-suspenders. Tests cover rotation/truncation/partial-line. Solid foundation if/when this expands.
- 🟢 nit: Outbound-only ingest is the pragmatic ship — agree with the call. But the OSS user-facing surface (Brain tab, channels list) needs a banner explaining the limitation, otherwise support volume goes up.

PR #1204, opened 2026-05-13
fix(test): use real ingest function names — E2E tests were permanently skipped (closes P1 from #1193 review)

vivekchand:
## Summary

PR #1193 added `tests/test_e2e_telegram_brain_ingest.py` with a skip predicate keyed on `clawmetry.sync._ingest_channel_directory` — a function that never existed on `main`. The real ingest path landed in #1192 as `sync_channel_messages(config, state, paths)` and in #1198 as `sync_telegram_from_gateway_log`. The probe therefore permanently failed, and all 6 E2E tests skipped forever, defeating the regression contract the file was written to enforce.

- Probe `from clawmetry.sync import sync_channel_messages` (the real name from #1192).
- Set `CLAWMETRY_OPENCLAW_DIR` (what `_get_openclaw_dir` reads) alongside `OPENCLAW_HOME` so the sync function actually points at the tmp tree.
- `_ingest(env)` now builds `(cfg, state, paths)` and calls `sync_channel_messages` once. That function iterates every provider in `_CHANNEL_DIRS`, so per-provider calls collapse to a single call.
- Seed payloads use `sender_name` (a key `_parse_channel_event` actually reads) instead of the bare `sender` string, which the parser silently drops when it isn't a dict.

## Test plan

- [x] `pytest tests/test_e2e_telegram_brain_ingest.py -v` — 6/6 pass (was 6/6 skipped).
- [x] `pytest tests/test_telegram_ingest.py tests/test_telegram_gatewaylog_parser.py tests/test_e2e_telegram_brain_ingest.py -v` — 29/29 pass, no regressions in adjacent suites.
- [ ] CI green on this PR.

No production code touched — purely a test fix.

### DIFF
--- a/tests/test_e2e_telegram_brain_ingest.py
+++ b/tests/test_e2e_telegram_brain_ingest.py
@@ -1,14 +1,12 @@
 """E2E contract test: Telegram messages → DuckDB → /api/brain-history.
 
 This file pins down the END-TO-END shape that the Telegram (and any other
-chat-channel) ingest path MUST satisfy. It is intentionally written
-*before* the ingest path lands so it acts as a regression contract for
-the in-flight work tracked under agent ``aca53ec8``.
+chat-channel) ingest path MUST satisfy.
 
 Pipeline under test:
 
     OpenClaw filesystem (per-chat .jsonl)
-        →  ClawMetry sync.py (``_ingest_channel_directory``)
+        →  ClawMetry sync.py (``sync_channel_messages``)
         →  Local DuckDB (``channel_messages`` table)
         →  /api/brain-history (renders chat_id/sender/body)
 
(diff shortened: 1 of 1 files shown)


vivekchand:
## UX review

🟢 nit · Test-only (`tests/test_e2e_telegram_brain_ingest.py`). No user-facing UI surface — skipping detailed UX scrutiny per backend-only rule.

vivekchand:
## Product review

🟡 P1 · The skip-permanently bug existed for ~30 minutes (between #1193 merge and #1204) — that's how long the E2E contract was silently green-without-running. Action: add a CI gate that asserts no test in the suite is `skipped` due to a `pytest.importorskip` failing on a name in our own codebase. Catches the "test silently skipped because the function it probes was renamed" class of bug. Worth a follow-up issue.

🟢 nit · The fix audit caught two genuine seed-data bugs (`CLAWMETRY_OPENCLAW_DIR` env, `sender_name` field key) that were in the test from day one — meaning the contract had bugs the contract was supposed to catch. Good unfreeze. No tier-strategy implications (pure test infra).

Issue #1202, opened 2026-05-13
[Product P0 from PR #1197] 21 channel pills shipped to OSS with no Cloud-Pro upsell


vivekchand:
From the post-merge UX/product review of PR #1197.

## Problem

PR #1197 ships first-class branded channel rendering for 21 adapters (Telegram, Signal, Slack, Discord, MS Teams, etc.) in the Brain stream — provider-coloured pills, emoji icons, sender names, direction arrows. Per memory `project_alerts_pro_feature.md` and the multi-channel observability narrative, multi-channel monitoring IS the Cloud-Pro story.

We're now shipping the polish for free with zero conversion path. Every user who lands a Telegram message on the Brain tab is one missed upsell.

## Action

Add a CTA banner above the Brain stream when the user has at least one `channel.in` row in DuckDB (or any non-CLI/cron provider visible). Example copy:

> "Connected channels: 3 / 21. See unified Telegram + Slack + Discord activity across all your nodes — start 7-day Cloud-Pro trial"

Per memory `project_free_plan_upsell.md`, every gated/polish surface should be a conversion opportunity.

## Files

- `clawmetry/static/js/app.js:2755-2769` — the `renderBrainStream` channel branch is the right place to drop the CTA injector

## Why P0

This is a post-merge regression in monetization posture — the more polished we make the OSS experience without an upsell hook, the harder Pro conversion gets. Fix before #1197 hits a release.

vivekchand:
## Decision: keep channel pills universal in OSS (no Pro upsell gating)

**Same call as #1181 (sub-2s real-time):** the MOAT is local-compute + cloud-display + multi-node + alert delivery + retention. Channel rendering is a feature, not a moat.

**Reasoning:**

1. **The funnel-relevant differentiators are intact.** Per `project_alerts_pro_feature.md` — Alerts delivery (Slack/PagerDuty/email), multi-node fleet, >24h retention, Approvals. Those drive Pro signups. Channel pill rendering doesn't.

2. **Throttling OSS rendering hurts the install funnel.** A user whose OSS Brain says "📱 Telegram · Vivek: hi" sees a polished product → tries Cloud-Pro for the cross-node view. A user whose OSS Brain says "channel event (upgrade to see)" uninstalls and writes a HN comment about the freemium dark pattern.

[...]

PR #1205, opened 2026-05-13
fix(brain): render body-less outbound channel events with 'sent' affordance (closes #1201)

### LINKED ISSUE #1201: [UX P0 from PR #1198] Telegram outbound rows render as empty bot replies (body=None has no affordance)
From the post-merge UX/product review of PR #1198.

## Problem

PR #1198 ingests Telegram outbound ACKs from gateway.log with `body=None` (the log only carries the ACK, not the message text). PR #1197 then renders these as:

> 📱 Telegram ↗ agent —

Empty body, no explanation. From a 5-second grok perspective, the user concludes the bot replied with nothing. This is worse than not surfacing the row at all.

## Action

Either:
1. Ingest a sentinel body string like `"(body not captured — OpenClaw stores Telegram chats in memory)"` so the renderer has something to display, OR
2. Coordinate with `clawmetry/static/js/app.js` `renderBrainDetail` to render a dedicated "(no body captured)" affordance with a help-link tooltip when `body` is None on a `channel.out` row.

## Files

- `clawmetry/sync.py:6580` — `body: None` in `parse_telegram_outbound_line`
- `clawmetry/static/js/app.js:2399` — outbound sender fallback to literal `"agent"`

## Why P0

This is the user-visible surface of the entire Telegram observability story for free + paid users. Empty rows == support tickets and looks-broken impression on first install.

vivekchand:
## Summary

- PR #1198 ingests Telegram outbound ACKs from `gateway.log` with `body=None` (OpenClaw stores Telegram chats in memory; the log only carries the API ACK, not the text — `raw_blob.body_capture="ack_only"`).
- PR #1197 rendered these as empty Brain rows (`📱 Telegram ↗ agent —`), reading as "the bot replied with nothing." That's a P0 first-impression bug for the entire Telegram observability story.
- Fix the channel-event renderer in `clawmetry/static/js/app.js` to detect body-less outbound events and emit a `⤴ sent · (no body captured)` affordance with a tooltip that explains the capture gap.
- Inbound empty rows are unchanged (they represent real user silence, not a capture gap). Outbound events that DO have body text are also unchanged.

## Fix (one-line render branch)

```js
// _extractChannelInfo
var bodyMissing = direction === 'out' && (ackOnly || !hasBodyText);
```

```js
// renderChannelEventMeta
if (info.bodyMissing) {
  html += '<span ...>⤴ sent</span><span ...>(no body captured)</span>';
}
```

## Test plan

- [x] `node tests/test_appjs_units.js` — 64 assertions pass (6 new for #1201)
- [x] Verified ack-only outbound row renders `"sent"` + `"(no body captured)"` (NOT empty)
- [x] Verified outbound WITH body text → no affordance (regression guard)
- [x] Verified inbound empty event → no affordance (only outbound is flagged)
- [x] Verified XSS-escaping unchanged
- [ ] Live verify on a real Telegram outbound row in Brain UI after merge

### DIFF
--- a/clawmetry/static/js/app.js
+++ b/clawmetry/static/js/app.js
@@ -2397,6 +2397,25 @@ function _extractChannelInfo(ev) {
   }
   chatId = chatId ? String(chatId) : '';
 
+  // ── body-missing detection (issue #1201) ─────────────────────────────
+  // The gateway.log Telegram parser (clawmetry/sync.py
+  // ``parse_telegram_outbound_line``) ingests outbound ACKs with
+  // ``body=None`` because OpenClaw stores Telegram chats in memory and
+  // the log only carries the API ACK, not the message text. Without this
+  // flag the renderer prints an empty detail row that looks like the bot
+  // replied with nothing. We surface the breadcrumb (``raw_blob.body_capture
+  // == "ack_only"``) AND a generic empty-body check so other adapters
+  // with the same shape get the same affordance for free.
+  var hasBodyText = !!(
+    ev.detail || ev.body || data.text || data.body || data.message
+  );
+  var ackOnly = false;
+  var raw = data.raw_blob;
+  if (raw && typeof raw === 'object' && raw.body_capture === 'ack_only') {
+    ackOnly = true;
+  }
+  var bodyMissing = direction === 'out' && (ackOnly || !hasBodyText);
+
   return {
     provider:      provider,
     providerLabel: _channelDisplayName(provider),
@@ -2404,7 +2423,9 @@ function _extractChannelInfo(ev) {
     providerColor: _channelColors[provider] || '#667',
     sender:        sender,
     chatId:        chatId,
-    direction:     direction
+    direction:     direction,
+    bodyMissing:   bodyMissing,
+    ackOnly:       ackOnly
   };
 }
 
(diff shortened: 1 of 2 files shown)


vivekchand:
## UX review

🟢 The "⤴ sent · (no body captured)" affordance is the right call — better than an empty row and the tooltip explains *why* (OpenClaw in-memory limitation) so the user doesn't think it's a ClawMetry bug.

🟡 P1 · The "(no body captured)" text appears on EVERY outbound Telegram event because gateway.log never contains the body. After 50 messages, the Brain feed is going to be dominated by repeating "(no body captured)" lines. Consider grouping consecutive body-less outbound events under a single collapse: "↪ 12 outbound messages (bodies not captured)" — expandable to see timestamps. Better signal density. (`app.js` near `renderChannelEventRow`.)

🟢 nit · Inbound Telegram empty rows are NOT flagged — that's correct (empty inbound = real user silence, not a missing body). Good edge-case discipline.

vivekchand:
## Product review

🟡 P1 · Honest tooltip text ("OpenClaw stores Telegram chats in memory…") is the right move — better than pretending the data exists. But this exposes a competitor angle: ClawMetry can't show inbound Telegram. A user with high-fidelity Telegram observability needs (e.g. compliance, customer-support audit trails) reads this tooltip and bounces to a competitor (LangSmith, Helicone) that integrates Telegram via webhook subscription instead of relying on OpenClaw persistence. Action: file an issue to add a webhook-subscription path (separate from gateway.log parsing) so we capture inbound bodies even if OpenClaw never persists them. Not blocking this PR — the affordance is the right interim fix.

🟢 nit · No tier-strategy implication — body-less outbound rendering is universal, consistent with the #1202 decision (pills in OSS without Pro gating).

PR #1208, opened 2026-05-13
fix(brain): collapse 3+ consecutive body-less outbound rows from same chat (P1 from #1205 review)

vivekchand:
## Summary

P1 follow-up to #1205. After that PR shipped, every gateway.log Telegram outbound ACK renders as a `⤴ sent · (no body captured)` row in Brain. Once 0.12.185 reaches users and history hydrates, they'll see **dozens of identical rows back-to-back** from the same chat — the affordance fix solves the empty-row regression but creates a new visual-noise problem.

This collapses **3+ consecutive body-less outbound channel events from the same chat** (same provider + chat_id) into one summary row, with click-to-expand to see the originals.

```
[📱 Telegram] [agent] · ↪ 12 outbound · 22:00 → 22:54 · (bodies not captured) · [▸ expand]
```

Click `▸ expand` → renders the run inline (each row keeps its `⤴ sent` treatment).

## Grouping rules

- **Threshold = 3.** Under threshold renders as individual rows (no synthetic wrapper).
- **Same provider + chat_id only.** Different chats never merge.
- **Outbound only.** Inbound empties never collapse — per #1205, that's intentional user silence, not a capture gap.
- **Body-bearing rows break the run.** Mixed traffic renders normally.
- **Non-channel events break the run.** A tool call between outbound ACKs interrupts grouping.

## Constraints honored

- No data mutation — `_brainAllEvents` untouched, collapse is pure presentation. Filter changes and auto-refresh recompute from source.
- No new endpoints — Brain still hits `/api/brain-history` (DuckDB-first).
- No celebrity-name strings.

## Test plan

- [x] `node tests/test_appjs_units.js` → **89 passed, 0 failed** (8 new collapse cases)
- [x] `node --check clawmetry/static/js/app.js` → OK
- [ ] Manual: visit Brain feed with hydrated outbound run → see one collapsed row + expand works
- [ ] Manual: confirm short bursts (1-2 outbound) still render individually

### DIFF
--- a/clawmetry/static/js/app.js
+++ b/clawmetry/static/js/app.js
@@ -2629,6 +2629,85 @@ function setBrainChannelFilter(ch, btn) {
   renderBrainStream(_brainAllEvents);
 }
 
+// Collapse runs of body-less outbound channel events from the same chat
+// (issue: P1 follow-up to #1205). After #1205 every gateway.log Telegram
+// outbound ACK renders as a "⤴ sent · (no body captured)" row — once the
+// Brain feed hydrates with dozens of these from one chat the feed turns
+// into visual noise. Threshold of 3+ to collapse: 1 or 2 stays as-is so
+// short bursts read naturally; 3+ becomes one summary row with a count
+// and time range, click-to-expand. Pure presentation — the underlying
+// events array is untouched.
+//
+// Grouping rules:
+//   - same provider + chat_id (different chats never merge)
+//   - direction === "out" (inbound silence is intentional per #1205)
+//   - bodyMissing === true (body-bearing rows render normally + break run)
+//   - any non-matching event in between (inbound, body-bearing, non-channel)
+//     breaks the run
+//
+// Input: array sorted newest-first (renderBrainStream sorts before this).
+// Output: array of same/fewer length where collapsed entries carry
+// __collapsedRun (the original event objects) + __collapsedCount.
+function _collapseBodylessOutbound(events) {
+  if (!events || events.length === 0) return events || [];
+  var collapsed = [];
+  var i = 0;
+  while (i < events.length) {
+    var ev = events[i];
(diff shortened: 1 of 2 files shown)


vivekchand:
## UX review

🟢 Right call on threshold of 3 (1-2 stays as individual rows). Makes the collapse meaningful without hiding small bursts.

🟡 P1 · The expand affordance ("[▸ expand]") doesn't say HOW MANY rows are inside the collapsed group at a glance — the count badge is somewhere but the affordance text itself is generic. Compare:
- current: `↪ 12 outbound · 22:00 → 22:54 · (bodies not captured) · [▸ expand]`
- better: `↪ Show 12 hidden outbound messages` — verb-first, count-first, matches Apple Mail / Gmail collapse patterns.

🟢 nit · Inbound silence is correctly NOT collapsed (per #1205 contract). Good edge-case discipline.

🟢 nit · Smart spec correction by the agent (2+inbound+3 → 4 rows, not 6). Collapsing the 3-run is the more useful behavior; keeping the inbound 2 visible preserves the conversation thread.

vivekchand:
## Product review

🟢 Pure UX polish, no tier-strategy implication. Universal in OSS, consistent with #1202 decision.

🟡 P1 · Long-term: the existence of this collapse logic is itself a tell that we're shipping a degraded experience. A user reading "↪ 12 outbound (bodies not captured)" is being told "your tool can't see the message bodies your agent sent." Fix the upstream cause (gateway WS tap or OpenClaw upstream change) before competitors notice the gap. Captured separately in the #1205 follow-up — this collapse is the right interim fix; don't defer the upstream forever.

🟢 nit · 89 JS tests passing (was 81). Test coverage discipline ratcheting up cycle-over-cycle is the kind of compounding that makes the MOAT stick. Good.

PR #1211, opened 2026-05-14
fix(install): per-step spinners + show stray daemons during silent stages

vivekchand:
## Why

User feedback (verbatim):

> can we have some kind of % shown like a progress bar given that we know how much time it takes to install?? it's much better than feeling like stuck right??
>
> fyi -- it did things like killing existing etc etc -- probably they are also taking time in the silent duration not just installing clawmetry -- so we need better visibility & transparency on what is exactly happening
>
> also given that I did uninstall previously it should have killed those processes etc which I fully don't understand

The user ran `curl https://clawmetry.com/install.sh | bash` and the screen sat between `Installing clawmetry from PyPI (uv)...` and `Killed stray pre-existing daemon (lock released)` with **no feedback** — feels frozen. Behind that gap: launchd plist rewrites, `launchctl kickstart`, `pkill`, `sleep 1`, version probe.

Related memory: `feedback_install_uv_speedup` — uv shaved install from ~30s to ~5s, but the rest of the silent stages were never instrumented.

## Note on repo selection

The user asked for this PR against `vivekchand/clawmetry-landing`. After investigation: `clawmetry.com/install.sh` is **proxied from `vivekchand/clawmetry/main/install.sh`** by `clawmetry-landing/app.py:2715` — the file in the landing repo root is a stale 130-line vestige. Editing the landing copy would not have changed what users actually `curl`. So the PR lands here, on the OSS repo, where the live script lives.

## Silent stages this PR labels

| Stage | Was | Now |
|---|---|---|
| uv bootstrap | bare echo, then silent ~4s | spinner with `Xs/~4s (NN%)` |
| `uv venv` | bare echo, then silent ~2s | spinner |
| `uv pip install clawmetry` | bare echo, then silent ~5s | spinner |
| `python3 -m venv` (pip path) | bare echo, then silent ~5s | spinner |
| `pip install --upgrade pip` | totally silent | spinner |
| `pip install clawmetry` (pip path) | bare echo, then silent ~30s | spinner |
| macOS launchd plist rewrite + kickstart loop | totally silent | leading `→` echo + trailing `↺ launchd jobs restarted` confirmation |
| Linux systemd refresh | totally silent | leading `→` echo |
| `pkill clawmetry.sync && sleep 1` | totally silent | leading `→` echo + `↺ Lock released` |
| Stray-daemon detect | "1 daemon detected" | counts processes + prints `pid <N>: <cmdline>` for each |
| Final `✓ installed` line | no timing | now shows `(Xs total)` |

## Spinner helper (`_step`)

- Pure bash — no `tput`/`ncurses`/`stdbuf` deps.
- Signature: `_step "Label" <expected_seconds> cmd args...`
- TTY: background the cmd, redraw `\r  → Label ⠋ Xs/~Ys (NN%)` at 10Hz. Caps at 95% then switches to `Xs (still working…)` once `expected` is exceeded — never lies about 99%.
- Non-TTY (CI / piped to file): degrades to plain `→ Label...` echo + foreground exec — no escape sequences leak into logs.
- Failure: prints red `✗` line with exit code, then dumps captured stdout+stderr so users see the actual error. Returns the exit code so `set -e` propagates and aborts the script (verified locally).
- Empty `$USE_SUDO` flattens cleanly via call-site word-splitting; non-empty `USE_SUDO=sudo` works the same way (verified locally).

## Test plan

- [x] `bash -n install.sh` — syntax clean
- [x] `shellcheck install.sh` — no warnings
- [x] Spinner under pty (via `script(1)`): frames cycle, percent caps at 95, "still working…" kicks in past expected, success line clears the spinner
- [x] Failure path: `set -e` aborts; captured stderr shown
- [x] Empty `$USE_SUDO` does not break word-splitting
- [ ] Manual: `curl https://clawmetry.com/install.sh | bash` after merge (clawmetry.com proxies main, so this ships on merge — no PyPI release needed)

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -16,6 +16,94 @@ echo -e "  ${BOLD}🦞 ClawMetry${NC}  ${DIM}AI Observability for OpenClaw${NC}"
 echo -e "  $(printf '%.0s─' {1..50})"
 echo ""
 
+# Overall wall-clock so the final "✓ installed" line can show "(Xs total)".
+_T0=$(date +%s)
+
+# ── Spinner helper for silent stages ────────────────────────────────────────
+# Curl-bash users used to stare at "Installing clawmetry from PyPI…" for 5+
+# seconds with no feedback while pip ran. ``_step`` wraps any silent command
+# with a labeled spinner that shows elapsed-vs-expected seconds, surfaces
+# captured stdout+stderr if the command fails, and obeys ``set -e``.
+#
+# Usage: ``_step "Label" <expected_seconds> cmd args...``
+# (Note: no ``--`` separator — args are passed through as-is, so unset
+# variables like an empty $USE_SUDO flatten cleanly via word-splitting at
+# the call site, then ``"$@"`` inside the function preserves quoting.)
+#
+# Non-TTY (CI, logfile redirect): degrades to a plain echo + foreground run.
+_step() {
+  local label="$1"; shift
+  local expected="$1"; shift
+  local logfile
+  logfile=$(mktemp -t clawmetry-step.XXXXXX 2>/dev/null || mktemp)
+
+  # Non-interactive output: plain log line, foreground exec, no spinner.
+  if ! [ -t 1 ]; then
+    echo "  → ${label}..."
+    if "$@" >"$logfile" 2>&1; then
+      rm -f "$logfile"
+      return 0
+    else
+      local _rc=$?
+      echo "  ✗ ${label} (exit ${_rc})" >&2
+      cat "$logfile" >&2
+      rm -f "$logfile"
+      return "$_rc"
+    fi
+  fi
+
+  # Interactive: run the command in the background, redraw a spinner line
+  # every 100ms. Frames cycle through Braille dots; pct climbs toward 95%
+  # using the caller-supplied ``expected`` budget, then we switch to a
+  # neutral "still working…" once we overshoot so we never lie about 99%.
+  local frames='⠋⠙⠹⠸⠼⠴⠦⠧⠇⠏'
+  local frame_count=${#frames}
+  local i=0
+  local start_ms
+  start_ms=$(date +%s)
+
+  ( "$@" >"$logfile" 2>&1 ) &
+  local pid=$!
+
+  # Spin until the worker exits. ``kill -0`` is the cheap "is it alive?"
+  # probe; busy-loop is fine at 10Hz.
+  while kill -0 "$pid" 2>/dev/null; do
+    local now elapsed pct frame
+    now=$(date +%s)
+    elapsed=$(( now - start_ms ))
+    frame="${frames:$(( i % frame_count )):1}"
+    if [ "$elapsed" -lt "$expected" ]; then
+      # Cap displayed pct at 95 so we never claim 100% before we're done.
+      pct=$(( elapsed * 95 / (expected > 0 ? expected : 1) ))
+      [ "$pct" -gt 95 ] && pct=95
+      printf "\r  → %s %s %ds/~%ds (%d%%)        " "$label" "$frame" "$elapsed" "$expected" "$pct"
+    else
+      printf "\r  → %s %s %ds (still working…)        " "$label" "$frame" "$elapsed"
+    fi
+    i=$(( i + 1 ))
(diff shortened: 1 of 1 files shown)


vivekchand:
## UX review

- 🟡 P1 — Trailing 30+ spaces on the spinner redraw line (`%-40s` / `%-30s`) are there to mask the previous frame, but on narrow terminals (~80 cols) the line wraps and the cursor return `\\r` no longer overwrites cleanly. Real-user symptom: a smear of partial spinners stacked vertically. Cap line width to `tput cols - 4` or use `\\033[2K` (clear-line) instead of trailing spaces. Curl-bash users who run from cramped iTerm tabs will hit this.  Action: `install.sh:_step` — replace the trailing-padding hack with `printf '\\r\\033[2K  → ...'`.
- 🟡 P1 — Mixed iconography. The diff uses `→`, `↺`, `✓`, `✗`, `⠋⠙⠹…`, `↻` (existing detect line). Five different glyphs in the same install run reads as inconsistent. Pick `→` for "running", `✓` for "done", `✗` for "failed" — drop `↺` (rebrand the launchd-restart confirm to `✓ launchd jobs restarted`).  Action: `install.sh` — globally swap `↺` → `✓`.
- 🟡 P1 — On non-TTY (CI / piped to file), the script logs `→ Bootstrapping uv (faster installer)...` then captures all output to a temp file and only dumps it on failure. Success path means CI logs lose every package install line — debugging "why did pip install pick wrong dep version" becomes impossible without re-running. Either tee to stdout in non-TTY mode or print the captured log on success too when `CLAWMETRY_INSTALL_VERBOSE=1`.  Action: `install.sh:_step` non-TTY branch — `tee` to stderr instead of swallowing.
- 🟢 nit — `pid $_p: ` truncated to 90 chars by `cut -c1-90` is fine on Mac but the python invocation on Linux systemd is often longer (`/home/user/.clawmetry/venv/bin/python3 -m clawmetry.sync --node-id ...`). User sees a chopped command and can't grep for it.  Action: bump cut to 140 cols, or use `pgrep -af` directly.
- 🟢 nit — The "$INSTALL_DIR/bin/pip install ... clawmetry" spinner says expected=30s but PyPI cold-cache plus a slow corporate proxy can stretch to 90+. The "still working…" fallback handles this, but maybe set expected=45 to avoid premature "still working…" warnings on average users.

vivekchand:
## Product review

- 🔴 P0 — Spinners treat the symptom; the disease is that the silent stages are slow at all. `launchctl kickstart -k` + `pkill -f clawmetry.sync && sleep 1` + `pip install --upgrade pip` add ~10s to every install for users who already have a working clawmetry. First-principles: do we even need to refresh launchd jobs / kill the daemon when the installed version equals the published version? Add a fast-path check at the top of `install.sh`: if `$INSTALL_DIR/bin/clawmetry --version == published version` AND daemon is running, exit early with "already up to date". Cuts the curl-bash flow from 15s to 1s for the common case (re-running the install from a Slack message).  Action: `install.sh` — early-exit when versions match and daemon is healthy. Saves Cloud-funnel users from a frustrating "why is this re-installing" moment.
- 🟡 P1 — `_step` is 80 lines of bash. We're already shipping a python CLI (`clawmetry/cli.py` has `_purge_server` and friends). Why is the install logic in shell at all? Long term: replace `install.sh` with a tiny bootstrap that does `pip install --user clawmetry && clawmetry install`, then put all of this logic — including the spinner — in python where it can be tested and reused (a non-TTY check is one line in click/rich, not 30 lines of bash). Today's PR is the right tactical fix; flag as tech debt.  Action: file follow-up issue "Move install.sh logic into `clawmetry install` subcommand".
- 🟡 P1 — The `(Xs total)` final line is product gold for benchmarking — but we throw it away. Pipe it to the existing telemetry endpoint (anonymous) so we know p95 install time across the user base. "How long does install actually take for our users" is a metric we currently can only guess at.  Action: log `(Xs total)` to `telemetry/install` (already-existing endpoint or create one) — gates conversion timing analysis.
- 🟢 nit — Sibling-PR coordination (landing repo) is now moot per the body; but worth a one-time README note that `clawmetry.com/install.sh` is proxied from this repo so contributors don't try to edit the landing copy in the future.

PR #1212, opened 2026-05-14
[RELEASE] fix(sync): telegram events parsed but never reach DuckDB (P0)

vivekchand:
## Root cause

**Hypothesis A** was correct — but at one level removed from where I expected.

`sync_telegram_from_gateway_log` was writing to `channel_messages`, but the Brain feed (`/api/brain-history` → `LocalStore.query_events`) only reads from the `events` table. Telegram outbound rows were ingested correctly into one DuckDB table but never projected onto the table the Brain renderer queries — the channel went silently empty even though the parser reported `"telegram-gw-log: ingested 16 outbound message(s)"` on every cycle.

(Verified on the user's live machine: the daemon's local-query proxy returns 16 rows from `query_channel_messages`, but `query_events(event_type='channel.out')` returns zero. The `routes/brain.py` fast path at L260 already knows how to render `CHANNEL.*` rows — it just never sees any from the gateway-log parser.)

`sync_channel_messages` (the JSONL-backed sibling) has done the dual-write correctly since day one (sync.py L2771-L2789); the gateway-log parser drifted out of contract when it was added in #1192.

## Fix (3 lines)

1. New helper `_telegram_outbound_event_row()` projects a parsed channel_messages row onto the `events` schema with `event_type="channel.out"` and `data.{provider, chat_id, direction, ...}`.
2. `sync_telegram_from_gateway_log` builds an `events_batch` alongside the existing `ingest_channel_message` call and flushes it via `store.ingest_many()`.
3. Same `id` on both tables → upsert / `INSERT OR IGNORE` keeps state-file loss idempotent.

## Test that would have caught this

`tests/test_telegram_gatewaylog_parser.py::test_outbound_messages_reach_events_table_for_brain_feed` — writes three real-shape ACK lines, runs the daemon helper, asserts (a) `channel_messages` has 3 rows, (b) `events` has 3 rows with `event_type='channel.out'`, (c) `query_events` returns brain-renderer-shaped enrichment data. Plus an idempotency twin that re-runs with a fresh state dict and asserts no duplicate rows.

The original test only covered the channel_messages side — exactly the failure mode that let this regression ship.

## Hypothesis matrix (for future P0 triage)

- **A — write missing/guarded by flag:** ✅ correct (write was missing for the events table)
- **B — stale offset from previous install:** ruled out (state file exists with reasonable offset; ingest log shows 16 rows landed)
- **C — brain query filters out the source:** ruled out (brain reads `events`, parser writes only `channel_messages` — they don't intersect to begin with)
- **D — multi-daemon write contention:** ruled out (no errors, ingest succeeded, single writer holds the lock)

## Test plan

- [ ] CI green on the new regression test
- [ ] User's next sync cycle ingests historical telegram backlog into events
- [ ] Brain tab shows telegram outbound rows after PyPI propagation + dashboard restart

### DIFF
--- a/clawmetry/sync.py
+++ b/clawmetry/sync.py
@@ -6932,6 +6932,52 @@ def parse_telegram_outbound_line(line: str) -> dict | None:
     }
 
 
+def _telegram_outbound_event_row(
+    msg: dict, *, node_id: str,
+) -> dict:
+    """Project a ``channel_messages`` row (from ``parse_telegram_outbound_line``)
+    onto the ``events`` table shape that the Brain feed reads.
+
+    The Brain tab calls ``LocalStore.query_events`` and surfaces rows whose
+    ``event_type`` starts with ``CHANNEL.`` (see ``routes/brain.py`` —
+    ``_try_local_store_brain``). Without this projection, telegram outbound
+    rows live ONLY in ``channel_messages`` and the Brain feed silently
+    omits them — exactly the P0 user-visible regression that motivated
+    this helper.
+
+    The ``id`` matches the ``channel_messages.id`` so the events row and
+    the channel_messages row are 1:1 — both use ``ingest_many`` /
+    ``ingest_channel_message``'s ``INSERT OR IGNORE`` / upsert semantics
+    so re-running the parser after a state-file loss is a no-op.
+    """
+    raw = msg.get("raw_blob") or {}
+    return {
+        "id": msg["id"],
+        "node_id": node_id,
+        "agent_type": "openclaw",
+        "agent_id": "main",
+        "event_type": "channel.out",
+        "ts": msg["ts"],
+        # Channel turns are NOT tied to an LLM session — leave session_id
+        # NULL so the per-session token/cost rollups don't double-count.
+        "session_id": None,
+        "workspace_id": None,
+        "data": {
+            "provider": _TELEGRAM_PROVIDER,
+            "channel_id": msg.get("channel_id"),
+            "chat_id": raw.get("chat_id"),
+            "message_id": raw.get("message_id"),
+            "method": raw.get("method"),
+            "direction": "out",
+            "body_capture": raw.get("body_capture"),
+            "source": raw.get("source") or "gateway.log",
+        },
+        "cost_usd": None,
+        "token_count": None,
+        "model": None,
+    }
+
+
 def _telegram_log_offset_key(log_path: str) -> str:
     """Per-log-path key under ``state['last_log_offsets']``.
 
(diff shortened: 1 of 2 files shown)


vivekchand:
**E2E verified end-to-end against the Brain tab in headless chromium — passes.**

Setup: isolated venv at `/tmp/cm-e2e-brain/`, port 8901, fake api_key (so cloud sync stays inert), worktree off `main` with #1212 and #1213 merged. Sync daemon + dashboard ran side-by-side reading user's real `~/.openclaw/`.

```
=== UI ===
event rows visible: 50
has_telegram: True
has_channel_out: True
has_openclaw_message: True
has_tool_use: True
=== API /api/brain-history ===
total items: 38
telegram items: 16
openclaw message items: 6
=== screenshot saved ===
ALL CHANNELS RENDER IN BRAIN TAB
```

Brain tab clearly shows:
- 16 telegram `CHANNEL.OUT` events ingested from `gateway.log` via the parser, with proper "→ sent (no body captured)" affordance
- OpenClaw `MODEL.COMPLETED`, `TOOL.RESULT`, `PROMPT.SUBMITTED`, `SESSION.STARTED` events
- Filter chips show all 8 event types with counts

Sync daemon log line: `Synced 0 events, 0 log lines, 0 memory files, 0 crons, 0 cron-runs, 0 session rows, 16 telegram-out` — confirms the parser→DuckDB→events table fix in this PR landed end-to-end.

**One P1 follow-up surfaced** during E2E: the collapsed-run row was rendering `formatBrainTime()` HTML as literal text (escHtml double-escape). Fixed in #1215.

vivekchand:
## UX review

- Diff is purely backend (DuckDB dual-write + regression test). No user-visible UI surface in this PR — the user-visible win is "Telegram messages now appear on the Brain tab", which is a pre-existing renderer.  Action: pair with a quick post-merge browser check that the Brain tab actually shows telegram outbound rows after a 60s sync cycle (the test plan's last unchecked box). If the renderer truncates at "channel.out" but no telegram-specific affordance exists yet, that's worth a separate UX pass — but not in this PR.

vivekchand:
## Product review

- 🔴 P0 — This is the second time in a month a feature shipped where one DuckDB writer projected to a subset of the tables a downstream reader queries (memory ref: `feedback_local_store_default_off_killed_moat`). The PR fixes the symptom but the PATTERN — "two parallel ingest paths drift out of contract" — will re-bite us on every new event source. First-principles ask: should `channel_messages` be the source of truth, with `events` projected from it via a DB view (or a single `ingest()` function that writes both)? Right now we have N writers each manually dual-writing — that's O(N) chances to forget half. Action: file follow-up issue to introduce `LocalStore.ingest_channel_event()` as the ONE call that fans out to both tables, and migrate `sync_channel_messages` + the new gateway-log path + the WS tap (PR #1213) onto it. Without that consolidation, PR #1213 will need its own dual-write code (and it does, in `gateway_tap.py:_handle_frame`) — three copies of the same pattern.
- 🟡 P1 — Test name `test_outbound_messages_reach_events_table_for_brain_feed` pins the contract beautifully. But the test only covers the gateway-log parser; `tests/test_sync.py` for `sync_channel_messages` (the JSONL dual-writer) does not have a sibling test asserting the same invariant. Add one — otherwise the JSONL path can silently drift the same way the gateway-log path did. Action: `tests/test_sync.py` — mirror the test for the JSONL dual-writer.
- 🟡 P1 — `session_id=None` is correct for telegram (channel turns aren't LLM-session-scoped) but means the per-session brain rollup will under-count. If a future Cloud-Pro "per-conversation cost" view tries to group telegram replies into a session, this row shape can't support it. Worth a doc comment in `_telegram_outbound_event_row` noting that telegram inbound (when PR #1213 lands) will need session_id resolution from the gateway.  Action: TODO comment in the helper pointing to the open question.
- 🟢 nit — node_id resolved fresh per-cycle from `config` / env / "local" — fine, but already cached at module level by the daemon. Tiny perf win + one less drift point if you read it from the daemon's cached value.

PR #1213, opened 2026-05-14
feat(sync): live gateway WS tap captures inbound channel messages (Telegram, Signal, Slack, ...)

### LINKED ISSUE #80836: RFE: gateway.auth.tokenScopes config field for headless/single-tenant deployments
## Problem

In headless / single-tenant deployments (ours runs OpenClaw inside an ECS Fargate container, with the agent and the CLI on the same host), operator-scope upgrades require an interactive approval flow that doesn't apply to the deployment model. Each time OpenClaw tightens default scopes — most recently 2026.5.7's hardening that requires `operator.write` for cron mutations — every paired device in our fleet generates a scope-upgrade approval request and gets blocked until a human approves it via the operator UI.

In our case, the requests come from devices we already trust (our own ECS containers running our own agents), but there's no way to express that trust statically in `openclaw.json`. Token-auth bypasses the device-identity gate for read commands via `roleCanSkipDeviceIdentity("operator", true) = true`, but mutating commands re-enter the device-identity check at a stricter scope, which the token-auth bypass doesn't cover.

Today's workaround on our end was an operator UI approval after each hardening bump. That doesn't scale — every new scope OpenClaw adds in the future will recreate the same wall.

## Proposed config

Add a `tokenScopes` field to `GatewayAuthConfig` (`src/config/types.gateway.ts`):

```ts
auth: {
  mode: "token",
  token: "...",
  // NEW: when a request is authenticated via the shared token (token mode), resolve
  // the session's effective scopes to this set instead of the default operator.read.
  tokenScopes?: OperatorScope[];
}
```

Example usage in `openclaw.json`:

```json
{
  "gateway": {
    "mode": "local",
    "bind": "loopback",
    "auth": {
      "mode": "token",
      "token": "<per-boot random secret>",
      "tokenScopes": ["operator.admin", "operator.pairing", "operator.read", "operator.write"]
    }
  }
}
```

## Semantics

- When `auth.mode === "token"` and `tokenScopes` is present, a successful token-auth handshake resolves to a session with exactly those scopes (not operator's default scope set).
- When `tokenScopes` is absent or empty, behavior is unchanged from today (defaults apply).
- The scopes a config can grant should be bounded — e.g., `operator.admin` should be permissible only for `bind: "loopback"` (or behind an explicit `allowAdminTokenScopes: true` flag) so a public-facing deployment can't accidentally grant admin to anyone with the shared token.

## Why this matters

Two real use cases:

1. **Headless deployments** like ours (ECS Fargate, agent + CLI in the same container, no human operator in the loop). Every scope hardening bump silently breaks cron skills until a human approves.
2. **Bootstrap / fresh install**. New devices today land on `operator.read` only and need scope upgrades for any non-trivial work. A config-level declaration of "this deployment is trusted to grant itself these scopes" short-circuits the bootstrap dance.

## Why not workarounds

- **Pre-write `identity/device-auth.json` at boot.** Requires the gateway to have already booted to generate `device.json` with a deviceId — multi-stage bootstrap with race conditions, hard to make reliable.
- **Run `openclaw devices approve` from inside the container.** Requires admin-scoped device pairing, which is circular (that's what we're trying to avoid needing in the first place).
- **Trust the existing `roleCanSkipDeviceIdentity` bypass.** It only applies to operator role with `sharedAuthOk`, which is sufficient for read commands but not for cron mutations or admin-scope operations. Adding more bypass cases there fragments the logic; a clean config field is more maintainable.

## Environment

- OpenClaw version: `2026.5.7` (the hardening that surfaced this for us)
- Our fleet: 14 agents on ECS Fargate, each a separate container with its own gateway + agent process. All loopback CLI.

vivekchand:
> "ideally brain tab should show from duckdb & duckdb should be updated for every conversation with OpenClaw from any of the channels"
> — user, 2026-05-14

## Status: PAUSED on upstream — NOT a `[RELEASE]` PR

Per the failure-mode escape clause in the task brief: the OSS gateway's token-auth handshake currently grants `scopes=[]`, so the WS tap can connect but cannot call `sessions.messages.subscribe` (the only method that delivers inbound message bodies). Filed [openclaw/openclaw#81775](https://github.com/openclaw/openclaw/issues/81775), referencing the canonical fix RFE [openclaw/openclaw#80836](https://github.com/openclaw/openclaw/issues/80836) (`gateway.auth.tokenScopes`).

The code is correct, fully tested, and ready to ship the moment OpenClaw lands #80836 (or an equivalent way to grant `operator.read` to the loopback CLI token non-interactively). I dropped the `[RELEASE]` prefix so the auto-publish workflow does not ship a tap that can't actually capture anything to PyPI users today.

## Architecture summary

`clawmetry/gateway_tap.py` is a single background thread the sync daemon spawns at startup. It:

1. Detects gateway URL + token from `OPENCLAW_GATEWAY_TOKEN` env or `~/.openclaw/openclaw.json` (mirrors `dashboard._detect_gateway_token` so the daemon doesn't have to import Flask).
2. Opens a persistent WebSocket to `ws://127.0.0.1:18789/`, runs the OpenClaw protocol-3 `connect` handshake, then attempts `sessions.messages.subscribe`.
3. For every pushed `event` frame, `_normalize_frame()` projects it into the same `(events_row, channel_messages_row)` shape the existing JSONL parser produces (`clawmetry/sync.py:_parse_channel_event`) — so brain-history reads from DuckDB don't have to know which transport delivered the row.
4. Writes through `local_store.ingest_many()` (events) + `ingest_channel_message()` (channel_messages). The body-bearing WS-tap row beats the gateway-log parser's NULL-body ACK row via the existing `COALESCE(excluded.body, channel_messages.body)` upsert — so when both transports observe the same `(provider, chat_id, message_id)`, the body sticks.
5. Wraps the whole loop in exponential backoff (2s → 60s) so any disconnect / gateway restart / scope rejection retries cleanly without crashing the daemon.

Wired into `clawmetry/sync.py:run_daemon()` right after the `local_query` server starts. Default-ON (per MOAT mandate `feedback_local_store_default_off_killed_moat`); opt out with `CLAWMETRY_DISABLE_WS_TAP=1`. Channel coverage matches `_CHANNEL_DIRS` (21 adapters) — the gateway-side subscribe is global rather than per-channel, but the list is documentation for any future per-channel API.

## Limitations

* **Inbound messages received while the daemon was offline are lost forever** — OpenClaw has no replay buffer for in-memory channel state ([openclaw/openclaw#42157](https://github.com/openclaw/openclaw/issues/42157)). Real fix is OpenClaw upstream persistence.
* **Live capture currently blocked on scope-grant gap.** With OpenClaw 2026.5.7's default token-auth, `connect` succeeds but returns `auth.scopes = []`. Every method that delivers message bodies (`sessions.messages.subscribe`, `sessions.subscribe`, `logs.tail`, `channels.status`) then 401s with `missing scope: operator.read`. Filed [openclaw/openclaw#81775](https://github.com/openclaw/openclaw/issues/81775), referencing canonical fix [openclaw/openclaw#80836](https://github.com/openclaw/openclaw/issues/80836). Tap logs a single WARNING with the exact symptom and downgrades to a degraded-but-stable state (catches `health` snapshots only).
* `health` events from the gateway also stream over the same socket — we drop them in `_normalize_frame()` since the dedicated `gateway.metric` ingest path already covers them. No double-write.
* `websocket-client` is a soft dependency: if not installed, the tap logs a one-shot warning and the daemon's gateway-log parser path keeps running unchanged.

### Mocked WS server (the only path that works today, pre-upstream-fix)

```bash
$ pytest tests/test_gateway_tap.py -v
============================== 11 passed in 1.69s ==============================

  test_normalize_telegram_inbound                          PASSED
  test_normalize_outbound_inferred_from_event_name          PASSED
  test_normalize_health_frame_returns_none                  PASSED
  test_normalize_epoch_seconds_coerced_to_iso               PASSED
  test_handle_frame_writes_event_and_channel_message        PASSED
  test_handle_frame_drops_unparseable_frame                 PASSED
  test_ws_tap_body_overwrites_parser_null_body              PASSED
  test_full_loop_against_stubbed_websocket                  PASSED
  test_disable_env_var_short_circuits_start                 PASSED
  test_detect_gateway_endpoint_reads_openclaw_json          PASSED
  test_detect_gateway_endpoint_returns_none_when_no_config  PASSED
```

`test_full_loop_against_stubbed_websocket` is the headline E2E: it injects a fake `websocket` module into `sys.modules`, drives `GatewayTap.start()` against a scripted responder that mints a connect-ok then a sub-ok then a real Telegram inbound payload, then asserts a row appears in `channel_messages` with the body intact.

### Real Telegram message → DuckDB (post-upstream-fix)

Once openclaw/openclaw#80836 lands and `~/.openclaw/openclaw.json` carries `gateway.auth.tokenScopes: ["operator.read"]`:

1. Restart `clawmetry sync` and check `~/.clawmetry/sync.log` for:
   ```
   INFO gateway WS tap connected — capturing live channel events
   ```
   (vs the current `connected (degraded — no message subscribe; granted scopes=[])`).
2. Send a Telegram message to the OpenClaw bot.
3. Within ~1 sec, the body should appear at the head of `/api/brain-history` and via direct query:
   ```sql
   SELECT ts, provider, channel_id, direction, body
   FROM channel_messages
   WHERE provider = 'telegram'
   ORDER BY ts DESC LIMIT 5;
   ```
4. Stop the daemon, send another message (it'll be lost — see Limitations), restart, send a third — the third one should land. The second never will until OpenClaw adds replay (#42157).

## Files touched

- `clawmetry/gateway_tap.py` (new, ~430 LOC) — the WS tap module
- `clawmetry/sync.py` (+13 LOC) — startup wiring inside `run_daemon`
- `tests/test_gateway_tap.py` (new, ~340 LOC) — 11 unit + integration tests

## Connection status (in this worktree)

- Code: ready, all 11 tests pass.
- Real OpenClaw gateway in this dev environment: tap connects (handshake ok) but `auth.scopes = []` rejects every message subscribe — verified by `/tmp/probe_gateway_ws.py`. Mock-only verification.

### DIFF
--- a/clawmetry/gateway_tap.py
+++ b/clawmetry/gateway_tap.py
@@ -1,0 +1,664 @@
+"""
+clawmetry/gateway_tap.py — live OpenClaw gateway WebSocket subscriber.
+
+Why this exists
+---------------
+
+OpenClaw stores Telegram (and other ephemeral channels — Signal,
+WhatsApp, Discord, Slack, IRC, iMessage, WebChat, …) chat traffic
+**entirely in memory**. There is NO ``~/.openclaw/<channel>/*.jsonl``
+written for inbound messages on those channels — the only on-disk
+evidence is ``[telegram] sendMessage ok …`` ACK lines in
+``~/.openclaw/logs/gateway.log`` (outbound only, no body).
+
+That gap means:
+
+  * The Brain tab silently misses every inbound user message on
+    Telegram + sibling adapters.
+  * ``channel_messages`` only ever contains outbound ACK rows (with
+    ``body=None``) — half a conversation.
+
+This module taps the gateway's live WebSocket JSON-RPC stream
+(``ws://localhost:18789``), subscribes to per-session message events
+that DO carry the body, and upserts each frame into the local DuckDB
+``events`` and ``channel_messages`` tables. Real-time, in-process,
+no extra dependencies beyond the optional ``websocket-client`` (already
+the import path used by ``helpers/gateway.py``).
+
+Where it lives
+--------------
+
+The tap runs as a single background thread inside the sync daemon
+(``clawmetry/sync.py:run_daemon``). One thread, one persistent socket,
+exponential-backoff reconnect on every disconnect. The thread is a
+``daemon=True`` so a daemon shutdown doesn't hang on it.
+
+Failure modes
+-------------
+
+1. ``websocket-client`` not installed → tap is a no-op. Logged once at
+   startup; the gateway-log parser path keeps going (outbound ACKs only).
+2. Gateway not reachable / token missing → tap retries with backoff.
+3. Gateway accepts ``connect`` but our token gets ``scopes=[]`` —
+   subscribing to ``sessions.messages.subscribe`` etc. then fails with
+   ``missing scope: operator.read``. We log a single WARNING with the
+   exact symptom so the user can update their OpenClaw config (or pin
+   the upstream issue at openclaw/openclaw).
+
+Disable
+-------
+
+Set ``CLAWMETRY_DISABLE_WS_TAP=1`` in the environment to opt out
+entirely (escape hatch). Default is ON per the MOAT-mandate
+"DuckDB fast paths must default-on" rule (memory note
+``feedback_local_store_default_off_killed_moat``).
+"""
+
+from __future__ import annotations
+
+import json
+import logging
+import os
+import threading
+import uuid
+from datetime import datetime, timezone
+from typing import Any, Callable, Iterable
+
+log = logging.getLogger("clawmetry-sync")  # share the daemon logger sink
+
+
+# ── Defaults & config ────────────────────────────────────────────────────
+
+
+# Channels we attempt to subscribe to. Mirrors ``_CHANNEL_DIRS`` in
+# ``clawmetry/sync.py`` — every chat-channel adapter that exposes
+# inbound messages over the gateway. The gateway protocol does not
+# (yet) accept a per-channel filter on ``sessions.messages.subscribe``
+# — the subscription is global — so this list is documentation only.
+# We kept it as a list (rather than dropping to a single subscribe call)
+# so a future per-channel subscribe API can plug in without touching
+# the call sites in sync.py.
+CHANNEL_NAMES: tuple[str, ...] = (
+    "telegram",
+    "signal",
+    "whatsapp",
+    "discord",
+    "slack",
+    "irc",
+    "imessage",
+    "webchat",
+    "googlechat",
+    "msteams",
+    "bluebubbles",
+    "matrix",
+    "mattermost",
+    "line",
+    "nostr",
+    "twitch",
+    "feishu",
+    "zalo",
+    "tlon",
+    "synologychat",
+    "nextcloudtalk",
+)
+
+
+# Reconnect cadence — exponential backoff with jitter cap.
+_BACKOFF_INITIAL_SEC = 2.0
+_BACKOFF_MAX_SEC = 60.0
+
+
+# Disable env var (escape hatch — feature is default-ON).
+_DISABLE_ENV = "CLAWMETRY_DISABLE_WS_TAP"
+
+
+# ── Frame normalization ─────────────────────────────────────────────────
+
+# A normalized turn we project from any provider's WS frame. Mirrors the
+# shape that ``clawmetry/sync.py:_parse_channel_event`` produces for
+# JSONL lines, so the downstream brain-history reader doesn't need to
+# care which transport the row arrived on.
+#
+# Required: provider, ts. Everything else is best-effort with a sane
+# default. Returning ``None`` skips the frame entirely.
+def _normalize_frame(frame: dict) -> dict | None:
+    """Project a gateway WS event frame into a (events, channel_messages)
+    tuple-shaped dict, or ``None`` if the frame is not a chat message.
+
+    The gateway's own per-channel event names vary
+    (``sessions.message``, ``telegram.inbound``, ``channel.message``,
+    …). We accept any frame whose ``payload`` carries enough fields to
+    pin a channel + a timestamp. Heartbeats / health / status frames
+    fall through to ``None`` and the caller drops them.
+    """
+    if not isinstance(frame, dict):
+        return None
+    if frame.get("type") != "event":
+        return None
+
+    event_name = str(frame.get("event") or "")
+    payload = frame.get("payload") or {}
+    if not isinstance(payload, dict):
+        return None
+
+    # ── provider ────────────────────────────────────────────────────────
+    # Try the provider-tagged shapes first, then fall back to inferring
+    # from the event name (e.g. ``telegram.inbound`` → ``telegram``).
+    provider = (
+        payload.get("provider")
+        or payload.get("channel")
+        or payload.get("adapter")
+    )
+    if not provider and "." in event_name:
+        head = event_name.split(".", 1)[0].lower()
+        if head in CHANNEL_NAMES:
+            provider = head
+    if not provider:
+        # Not a channel event we know how to project. (Health, presence,
+        # state-version, etc. all land here.)
+        return None
+    provider = str(provider).lower().strip()
+
+    # ── ts ──────────────────────────────────────────────────────────────
+    raw_ts = (
+        payload.get("ts")
+        or payload.get("timestamp")
+        or payload.get("date")
+        or frame.get("ts")
+    )
+    if raw_ts is None:
+        return None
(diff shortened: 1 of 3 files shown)


vivekchand:
**E2E verified end-to-end against the Brain tab in headless chromium — passes.**

Setup: isolated venv at `/tmp/cm-e2e-brain/`, port 8901, fake api_key, worktree off `main` with #1212 and #1213 merged. Daemon read user's real `~/.openclaw/`.

The gateway WS tap in this PR connected successfully:
```
gateway WS tap connected (degraded -- no message subscribe; granted scopes=[])
```

It surfaced the OpenClaw scope gap as designed:
```
gateway WS tap: sessions.messages.subscribe rejected (missing scope: operator.read) -- inbound channel bodies will not arrive until OpenClaw grants `operator.read` scope to the gateway token. See openclaw/openclaw issue tracker.
```

API & UI both render the events that are reachable today:
```
total items: 38
telegram items: 16    (all outbound, via gateway.log parser from #1212)
openclaw message items: 6  (TUI sessions ingested as JSONL)
```

Tap doesn't crash, doesn't spam, fails open with a single warning. Ready.

**One P1 follow-up surfaced** unrelated to this PR: collapsed-run row had a stale escHtml double-escape — fixed in #1215.

vivekchand:
## UX review

- Diff is purely backend (new `gateway_tap.py` + sync wiring + tests). No user-visible UI surface in this PR. The user-visible WIN — inbound Telegram messages appear on the Brain tab — is gated on the upstream `operator.read` scope grant, which means today this ships infrastructure that does nothing observable.  Action: when the upstream lands, do a browser walkthrough — verify the Brain tab actually surfaces inbound rows with `direction=in` and the renderer doesn't accidentally treat them as outbound (the `channel.in` event type may need explicit handling in `routes/brain.py`).
- 🟡 P1 — Single user-facing surface IS the log line emitted on degraded mode: `gateway WS tap connected (degraded — no message subscribe; granted scopes=[])`. A non-technical OSS user reading `~/.clawmetry/sync.log` and seeing this has zero idea what to do. The log line should include a one-line action: `See https://docs.clawmetry.com/troubleshoot/gateway-scopes for the fix` — and that doc page should explain the OpenClaw-side config tweak in plain English.  Action: `clawmetry/gateway_tap.py:_run_once` — append actionable URL to the degraded-mode warning.

vivekchand:
## Product review

- 🔴 P0 — We're shipping 1,138 LOC of new code (gateway_tap.py + tests + wiring) that **literally does not work end-to-end today** — the WS tap connects, gets `scopes=[]`, logs a degraded warning, and writes zero rows. First-principles ask: should this PR be merged at all before openclaw/openclaw#80836 lands? The author dropped `[RELEASE]` so it won't ship to PyPI users — good — but it's still default-ON in the daemon and will spam every dev's sync.log with "degraded" warnings every reconnect cycle. Three options: (a) merge but DEFAULT-OFF until upstream unblocks, (b) hold the PR until upstream merges, (c) merge as-is and accept the log noise. Recommendation: switch the default to OFF (opt-in via `CLAWMETRY_ENABLE_WS_TAP=1`), flip to default-ON in a follow-up PR the same day OpenClaw lands the scope grant. This violates `feedback_local_store_default_off_killed_moat` only on paper — that rule was about user-facing fast-paths that silently bypass; here, default-OFF is the right call because default-ON does NOTHING USEFUL. Action: change the env-var default to opt-in, document the flip-day plan in the PR description.
- 🟡 P1 — Three competing channel-event ingest paths now exist (JSONL parser, gateway-log tail, WS tap). Same projection logic re-implemented in each. PR #1212 review opened a follow-up to consolidate into `LocalStore.ingest_channel_event()`. THIS PR should be the forcing function — the WS-tap path is the third copy, and PR #1212's helper is the second. Land #1213 against the consolidated API, not parallel to it.  Action: gate this PR on the consolidation (issue from #1212 review), or commit to a same-week refactor PR.
- 🟡 P1 — Backoff is 2s → 60s exponential with no jitter. If the gateway dies and 100 OSS daemons reconnect, they all hit the gateway at the same wall-clock (assuming clock-aligned starts). Add small jitter to `_BACKOFF_INITIAL_SEC`.  Action: `gateway_tap.py:_run` — `backoff = min(_BACKOFF_MAX_SEC, backoff * 2.0 + random.uniform(0, 1))`.
- 🟢 nit — `CHANNEL_NAMES` is documentation-only ("the gateway protocol does not yet accept a per-channel filter") but it's already drifting from `_CHANNEL_DIRS` in sync.py — single source of truth there, import it.  Action: `gateway_tap.py` — `from clawmetry.sync import _CHANNEL_DIRS as CHANNEL_NAMES`.

PR #1215, opened 2026-05-14
fix(brain): collapsed-run time range rendered HTML as literal text

vivekchand:
## Summary
- E2E playwright verification of #1212 + #1213 caught literal `<span style=\"opacity:0.45;...\">13 May</span>` text appearing in the Brain tab's collapsed-outbound row.
- `formatBrainTime()` returns trusted HTML; wrapping `rangeLabel` in `escHtml()` turned that markup into visible text.
- One-line fix: drop the `escHtml` call (the literal `→` separator carries no user input, the helper output is trusted).

## Test plan
- [x] Headless chromium against Brain tab, post-fix, shows clean `13 May 22:54:19 → 13 May 19:54:45` instead of escaped `<span>`.
- [x] All 16 telegram CHANNEL.OUT events + 6 OpenClaw TUI events render correctly.

### DIFF
--- a/clawmetry/static/js/app.js
+++ b/clawmetry/static/js/app.js
@@ -2899,7 +2899,10 @@ function renderBrainStream(events) {
         escHtml(ev.__collapsedCount + ' consecutive outbound ACKs from this chat — bodies not captured (collapsed for readability)') +
         '">↪ ' + ev.__collapsedCount + ' outbound</span>';
       if (rangeLabel) {
-        html += '<span class="brain-collapsed-range" style="color:var(--text-muted);font-size:10px;flex-shrink:0;white-space:nowrap;">' + escHtml(rangeLabel) + '</span>';
+        // rangeLabel is trusted HTML built from formatBrainTime() output
+        // (which wraps the date prefix in a styled <span>) joined with
+        // a literal " → "; do NOT escHtml or the markup renders as text.
+        html += '<span class="brain-collapsed-range" style="color:var(--text-muted);font-size:10px;flex-shrink:0;white-space:nowrap;">' + rangeLabel + '</span>';
       }
       html += '<span class="brain-collapsed-note" style="color:var(--text-muted);font-size:10px;font-style:italic;flex-shrink:0;white-space:nowrap;">(bodies not captured)</span>';
       // Expand affordance — small text button, monospace arrow.


vivekchand:
## UX review

- 🟡 P1 — The fix is correct (drop `escHtml` around trusted `formatBrainTime()` output) but the wider pattern is the worry: we have two adjacent `html += '<span ...>' + escHtml(X) + '</span>'` lines on L2895 and L2902. One escapes (correctly — it's a free-text title attr), one DOESN'T (after this fix — markup-bearing). A reader skimming the diff has no way to know which is which without reading the helper. Add a 1-line comment marker on every `escHtml(...)` call site explaining "why escape" or "why trusted".  Action: `clawmetry/static/js/app.js:renderBrainStream` — comment marker convention (`/* TRUSTED-HTML */`) on the new branch, plus a follow-up rename — `formatBrainTime` → `formatBrainTimeHtml` so the trusted-HTML return type is in the name.
- 🟡 P1 — `formatBrainTime()` returns trusted HTML strings — but it ALSO is called in plenty of other places where HTML escaping COULD have been needed. If a future caller forgets that the return is HTML, they'll double-escape or worse. Rename to `formatBrainTimeHtml` to encode the contract in the symbol name.  Action: `clawmetry/static/js/app.js` — rename `formatBrainTime` → `formatBrainTimeHtml`, audit all callers.
- 🟢 nit — The collapsed run shows `13 May 22:54:19 → 13 May 19:54:45` — the END time is BEFORE the start time (PM/AM confusion or descending sort). Probably correct (most-recent first) but reads as a bug to a non-technical user. Worth confirming the sort order convention is consistent across the Brain tab.  Action: visual QA pass on collapsed-run rendering with mixed-day events.

vivekchand:
## Product review

- 🔴 P0 — `renderBrainStream` is hand-built string concatenation of HTML. We've now had at least two XSS-adjacent fixes here in two weeks. First-principles: stop building HTML by string-concat in 2026. Either (a) migrate this renderer to `document.createElement` + `textContent` (zero string-concat, zero escape decisions), or (b) adopt a tiny tagged-template helper like `html\`<span>...\${userInput}</span>\`` that does the escaping for us. Today's PR is a correct one-line fix; the LANDMINE that produced it is the whole approach. Action: file follow-up issue "Migrate `renderBrainStream` off raw string-concat HTML" and stop accepting these as one-liners.
- 🟡 P1 — Test plan says "[x] Headless chromium against Brain tab, post-fix, shows clean ..." but no E2E test landed in the PR. The next time someone wraps `rangeLabel` in `escHtml` because they think they're being defensive, we have no Playwright assertion to catch it. Add one. Action: `tests/e2e/test_brain.spec.ts` (or wherever brain E2E lives) — assert the collapsed-run text matches `/^\\d+ \\w+ \\d+:\\d+:\\d+ → \\d+ \\w+/` (no literal `<span`).
- 🟢 nit — Zero tier-strategy implications. Brain tab is OSS-tier and the fix is universal. Ship it.

PR #1216, opened 2026-05-14
[RELEASE] perf(install): launchd restart + pkill move to background — install completes in <10s

vivekchand:
## Why

User feedback (verbatim):

> but still why 87 seconds??? can we do something in background & not in a synchronous way?? we need it under 10 seconds

User's screenshot showed:

```
→ Detected macOS
✓ Creating virtual environment (uv) (0s)
✓ Installing clawmetry from PyPI (uv) (1s)
→ Refreshing macOS launchd jobs...
  ↺ launchd jobs restarted
✓ ClawMetry 0.12.181 installed (87s total)
```

Visible labeled steps add up to 1-2s. The other ~85s vanished into the silent ``Refreshing macOS launchd jobs...`` block.

## Profiling

Instrumented ``install.sh`` with millisecond-resolution timestamps (perl Time::HiRes side-channel log) and ran against ``INSTALL_DIR=/tmp/cm-bench-baseline``. Local repro environment: 3 plists installed (``com.clawmetry.dashboard`` + 2 ``com.clawmetry.sandbox.*``).

| Stage | Before | After |
|---|---|---|
| ``preflight_pgrep`` | 36ms | ~36ms |
| ``venv_create`` (uv) | 100ms | ~100ms |
| ``uv_pip_install`` | 450ms | ~450ms |
| ``version_probe`` | 61ms | ~60ms |
| ``plistbuddy_loop`` | 31ms | ~30ms |
| ``kickstart com.clawmetry.dashboard`` | 19ms | backgrounded |
| ``kickstart com.clawmetry.sandbox.nemoclaw-sandbox`` | **45,868ms** | backgrounded |
| ``kickstart com.clawmetry.sandbox.newmoclaw-sandbox2`` | **30,114ms** | backgrounded |
| ``kickstart_loop_total`` | **76,042ms** | **15ms** |
| **TOTAL wall clock** | **77s** (warm) / 87s (user's run) | **<1s** |

## Root cause (one sentence)

``launchctl kickstart -k`` is synchronous and waits for the spawned process to be alive — for NemoClaw sandbox plists that means a 30-50s ``docker exec <container> kubectl exec ...`` handshake per plist, so the foreground loop ate 60-100s of dead time the user couldn't see.

## Fix

1. **Background launchd kicks** with ``( ... ) &`` + ``disown -a`` so install.sh returns immediately. Kicks complete in the background; the kernel reparents them to launchd.
2. **Background systemd restart** on Linux with ``systemctl --user restart --no-block``.
3. **Drop ``sleep 1``** after ``pkill -f clawmetry.sync``.
4. **Move DuckDB lock retry** into ``local_store._open_connection`` (5 × 0.5s on ``Conflicting lock``). install.sh no longer babysits the kernel's lock release; the daemon owns it.

## 3-run timing report

```
$ for i in 1 2 3; do
    rm -rf /tmp/cm-bench-fixed-$i
    CLAWMETRY_SKIP_ONBOARD=1 INSTALL_DIR=/tmp/cm-bench-fixed-$i \
      BIN_DIR=/tmp/cm-bench-fixed-$i-bin time bash install.sh < /dev/null \
      > /tmp/install-fixed-$i.out 2>&1
    grep "installed" /tmp/install-fixed-$i.out
  done

run 1: ✓ ClawMetry 0.12.186 installed (1s total)
run 2: ✓ ClawMetry 0.12.186 installed (1s total)
run 3: ✓ ClawMetry 0.12.186 installed (0s total)
```

**Median: 1s, max: 1s** — well under the 10s target.

## Why fire-and-forget is safe

- install.sh's job is to put fresh files on disk and update plists/units. The daemon respawn is best-effort, exactly like ``apt install``'s post-install hook for a systemd service.
- The ``com.clawmetry.dashboard`` plist (the user-facing surface) restarts in <1s — the dashboard is responsive immediately. The slow plists are sandbox sync daemons whose work is invisible to the user (they upload data into the cloud).
- DuckDB lock-contention is now retried inside ``_open_connection`` with a 2.5s budget, so the freshly-spawned daemon doesn't immediately error out if it races a dying sibling.
- Failures during background kicks were never surfaced anyway (``|| true``), so we lose zero diagnostic signal.

## Daemon recovery verification

After the fixed install.sh exited (1s), I confirmed:

```
$ launchctl print "gui/$UID/com.clawmetry.dashboard" | grep state
    state = running

$ curl -sS -m 3 -o /dev/null -w "HTTP %{http_code} in %{time_total}s\n" http://localhost:8900/
HTTP 200 in 0.053768s
```

Dashboard back online in seconds, serving HTTP 200 in 53ms. The two sandbox plists were ``state = spawn scheduled`` (still completing docker exec in the background — same wall time as before, just no longer blocking the user).

## Test plan

- [x] ``bash -n install.sh`` passes
- [x] ``python3 -c "from clawmetry import local_store; local_store._open_connection().close()"`` works
- [x] 3 clean install runs all complete in ≤1s
- [x] Dashboard plist returns to ``state = running`` and serves HTTP 200 within seconds of exit
- [x] No new dependencies (no ``gtimeout``, no ``parallel`` — just bash builtins)
- [x] Existing ``✓`` / ``→`` labels still show with elapsed time via ``_step``
- [x] Final ``installed (Xs total)`` line reflects the real (now small) wall clock

### DIFF
--- a/clawmetry/local_store.py
+++ b/clawmetry/local_store.py
@@ -581,9 +581,34 @@ def _to_blob(value: Any) -> bytes | None:
 
 
 def _open_connection(*, read_only: bool = False) -> duckdb.DuckDBPyConnection:
-    """Open a DuckDB connection at DB_PATH, creating the directory if needed."""
+    """Open a DuckDB connection at DB_PATH, creating the directory if needed.
+
+    Retries briefly on "Conflicting lock is held" — that error fires when an
+    older sync daemon hasn't fully released the file lock yet (e.g. moments
+    after install.sh's ``pkill -f clawmetry.sync``). install.sh used to
+    ``sleep 1`` defensively; we now retry here so install.sh can return to
+    the prompt immediately and the lock-release race is owned by the
+    daemon-side code that actually cares about it. (#1215)
+    """
     DB_PATH.parent.mkdir(parents=True, exist_ok=True)
-    return duckdb.connect(str(DB_PATH), read_only=read_only)
+    # 5 attempts × 0.5s = 2.5s budget. Beats a fixed install.sh sleep
+    # because (a) most opens succeed on the first try and pay zero, and (b)
+    # if the conflicting holder is genuinely stuck we surface the real
+    # DuckDB error instead of silently sleeping past it.
+    last_exc: Exception | None = None
+    for attempt in range(5):
+        try:
+            return duckdb.connect(str(DB_PATH), read_only=read_only)
+        except duckdb.IOException as exc:
+            msg = str(exc)
+            if "Conflicting lock" not in msg and "could not set lock" not in msg:
+                raise
+            last_exc = exc
+            time.sleep(0.5)
+    # Out of retries — re-raise the last lock error so the caller sees it.
+    if last_exc is not None:
+        raise last_exc
+    return duckdb.connect(str(DB_PATH), read_only=read_only)  # unreachable
 
 
 # ── Singleton store ─────────────────────────────────────────────────────────
--- a/install.sh
+++ b/install.sh
@@ -268,17 +268,31 @@ if [ "$OS" = "Darwin" ]; then
     esac
   done
 
-  # Step 2: kickstart any registered com.clawmetry.* job. `|| true` so
-  # systems without launchd running (CI containers, Linux subprocess) don't
-  # abort the installer.
+  # Step 2: kickstart any registered com.clawmetry.* job in the BACKGROUND.
+  #
+  # ``launchctl kickstart -k`` is *synchronous* — it blocks until the daemon
+  # process is alive again. For NemoClaw sandbox plists that run
+  # ``docker exec <container> kubectl exec ...``, "alive" means the docker+
+  # kubectl handshake has completed, which routinely costs 30-50s per plist.
+  # On a machine with two sandbox plists installed that's 60-100s of dead
+  # time install.sh used to wait for. The user's job here is to put fresh
+  # files on disk; the daemon respawn is best-effort and doesn't need to
+  # block the prompt return. (#1215)
+  #
+  # `|| true` so systems without launchd running (CI containers, Linux
+  # subprocess) don't abort the installer.
   for _plist in "$_LA_DIR"/com.clawmetry.*.plist; do
     [ -f "$_plist" ] || continue
     _label=$(basename "$_plist" .plist)
-    launchctl kickstart -k "gui/$_UID/$_label" >/dev/null 2>&1 || true
+    ( launchctl kickstart -k "gui/$_UID/$_label" >/dev/null 2>&1 || true ) &
   done
+  # Detach the backgrounded kicks so install.sh can exit without waiting
+  # for them. ``disown -a`` clears bash's job table; the kernel keeps the
+  # children alive (their parent reparents to launchd/init).
+  disown -a 2>/dev/null || true
 
   if [ "$_DARWIN_PLIST_FOUND" = "1" ]; then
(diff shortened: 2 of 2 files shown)


vivekchand:
## UX review

- 🟡 P1 — `install.sh:294` flips the success line from `↺ launchd jobs restarted` to `↺ launchd jobs restarting in background`, which is honest but immediately followed by `✓ ClawMetry 0.12.181 installed (1s total)`. A non-technical user reads "installed in 1s" and will refresh the dashboard expecting it to be live; if launchd is mid-`docker exec` they'll see a connection-refused page and assume the install was lying. Add a one-line "Dashboard usually ready within ~10s — open http://localhost:8900" hint between the two messages so the user knows the perceived gap is normal. Action: append a `_step` line after the kickstart loop with the dashboard URL.
- 🟡 P1 — On Linux the success line says `↺ clawmetry-sync restarting in background` (`install.sh:346`) but on macOS we say `↺ launchd jobs restarting in background` and on the pkill path we say `↺ Stray daemon(s) signalled` (`install.sh:367`). Three different verbs ("restarting", "signalled") for the same fire-and-forget concept. Pick one — "restarting in background" reads cleanest to a non-technical user; "signalled" is sysadmin jargon. Action: normalize all three to `↺ <thing> restarting in background`.
- 🟢 nit — `install.sh:294` still uses present continuous "restarting in background" but it's printed *after* the `&` has already been kicked off, so by the time the user reads it the kicks are arguably past tense. Tiny nit, not worth blocking, but "Restart queued in background" is more precise.
- 🟢 nit — The dashboard itself has no follow-up "still warming up..." banner. Out of scope for this PR (purely backend), but worth a tracking issue: when the user lands on `localhost:8900` 2s after install, can the page detect "daemon not ready, retry in 5s" and show a friendly skeleton instead of a hard error? Action: file a follow-up to add a one-time post-install warm-up state to the dashboard root route.

vivekchand:
## Product review

- 🟡 P1 — Fire-and-forget restart is the right first-principles call (87s → 1s is a real conversion-killer fix), but we just traded latency for blast-radius silence. If the daemon crash-loops post-install, the user sees `✓ installed` and the dashboard never comes up — they bounce, never sign up for Cloud, never come back. The PR body's verification is `launchctl print | grep state` after exit, which is fine for the happy path but isn't running in production. Action: add a short `(sleep 8 && curl -fsS http://localhost:8900/healthz >/dev/null) &` background probe at the end of `install.sh` that, on failure, prints a `⚠ Dashboard hasn't responded — run 'clawmetry status' to debug` line via a tail-file so the user has a recovery breadcrumb.
- 🟡 P1 — The DuckDB lock-retry budget (`local_store.py:603`) is 5 × 0.5s = 2.5s. That's fine for the install.sh race, but it's now load-bearing for every other caller of `_open_connection` too. If a long-running query holds the write lock for >2.5s (read-heavy admin pages, fleet rollups), every peer caller now eats a 2.5s stall before erroring out — silently degrading the freemium perceived-perf story that this PR was trying to win. Action: gate the retry behind a kwarg (`retry_on_lock=False` default), and only set it `True` from install.sh-spawned daemon code paths.
- 🟢 nit — Strategic win: this PR makes the OSS install feel cloud-grade (sub-second), which is exactly the "wow" moment that converts free → paid. Worth a CHANGELOG callout with the before/after seconds — that's tweetable and pulls in fence-sitters who churned on the previous 87s install.
- 🟢 nit — No OSS↔Cloud-Pro tier leakage in this diff (it's pure install ergonomics). No paid surface affected. No funnel impact other than the conversion-positive perceived-perf win above.

PR #1224, opened 2026-05-14
[RELEASE] feat(sync): ingest OpenClaw's Claude Code session JSONL — captures all conversation activity

vivekchand:
## The bug (user's diagnosis, verbatim)

> "it 3000% a bug in your understanding of openclaw & reading the right files & updating frequently or as it happens to duckdb"

## Architecture

OpenClaw delegates execution to a `claude` CLI subprocess. Claude Code writes its session transcript to:

```
~/.claude/projects/<encoded-cwd>/<session-id>.jsonl
```

where `<encoded-cwd>` is the agent's CWD with `/` and `.` both replaced by `-`. So `/Users/vivek/.openclaw/workspace` → `-Users-vivek--openclaw-workspace`.

For the user's machine that file is:
```
/Users/vivek/.claude/projects/-Users-vivek--openclaw-workspace/49f1d9fc-0848-4b6b-8fd7-64633bbc6b58.jsonl
```
1.0 MB, modified continuously, full of the user's Telegram conversation.

The legacy `sync_claude_cli_sessions` path covered some sessions, but only when `~/.openclaw/agents/main/sessions/sessions.json` carried the `claudeCliSessionId` binding for that session. Any session started before OpenClaw wrote that binding (or where the binding was lost on a crash) is invisible to it.

## Fix

New `sync_openclaw_claude_sessions` in `clawmetry/sync.py` discovers Claude Code session files independently, by **process inspection**:

1. Walk running processes with `psutil.process_iter`.
2. Filter to `claude` subprocesses with `--resume <session-id>` in argv.
3. Confirm OpenClaw provenance via cwd under `~/.openclaw/`, `--mcp-config` under `~/.openclaw/`, or an `openclaw*` ancestor in the parent chain.
4. Encode `proc.cwd()` into Claude Code's slug, build `~/.claude/projects/<slug>/<session-id>.jsonl`, verify it exists. macOS sandbox fallback: scan all project dirs for `<session-id>.jsonl`.
5. Tail by byte offset (state at `state['claude_session_byte_offsets'][<abs path>]`).
6. Map each line to an `events` row (`agent_type=openclaw`, `agent_id=main`, `event_type=<claude type>`, `ts=<timestamp>`, `data=<full obj>`, deterministic `id=openclaw-cc:{session_id}:{uuid_or_lineno}`).
7. Write through `local_store.ingest_many` — INSERT OR IGNORE on the events PRIMARY KEY makes re-reads idempotent.

Wired into the cycle loop next to the existing `sync_claude_cli_sessions` and `sync_telegram_from_gateway_log` calls. The summary log line now reports `(N from claude-cc)` so we can see this path's contribution at a glance.

**Default ON.** Escape hatch: `CLAWMETRY_DISABLE_CLAUDE_SESSION_INGEST=1`. Per the `feedback_local_store_default_off_killed_moat` rule (2026-05-13), MOAT capabilities ship default-on or they don't ship at all.

## Verification (real user machine, isolated temp DuckDB)

Per the user's mandate: SEPARATE temp DuckDB at `/tmp/cm-oc-test/clawmetry.duckdb` and SEPARATE config — the live daemon's `~/.clawmetry/` state was not touched and the daemon was not killed.

```
INGESTED 293 events from real user file
Drained.

=== Events count ===
(293,)

=== Events for session 49f1d9fc-0848-4b6b-8fd7-64633bbc6b58 ===
(293,)

=== event_type breakdown ===
  assistant: 150
  user: 83
  queue-operation: 48
  attachment: 11
  system: 1

=== ts range ===
('2026-05-09T10:01:27.847Z', '2026-05-14T20:28:04.447Z')

=== rows mentioning 'hello, how are you?' (the user's 22:12 line) ===
  id=openclaw-cc:49f1d9fc-…:line:302  ts=2026-05-14T20:12:45.555Z  type=queue-operation
  id=openclaw-cc:49f1d9fc-…:51b0d1ff-8ee6-4a5c-a621-3f26b7df6a59  ts=2026-05-14T20:12:45.580Z  type=user
TOTAL MATCHES: 2
```

The `user` row's `data` blob contains `"chat_id": "telegram:1532693273"`, `"sender": "Vivek Chand"`, `"timestamp": "Thu 2026-05-14 22:12 GMT+2"`, and the message body `"hello, how are you?"` — verbatim, ready for Brain to render.

## Tests

`tests/test_openclaw_claude_session_ingest.py` — 7 tests, all green:

- `test_three_synthetic_lines_land_in_events` — synthetic 3-line session → 3 rows with right `event_type` / `ts` / `session_id`; cost+tokens extracted from `message.usage.*`.
- `test_idempotent_on_re_run` — second pass with EMPTY state still leaves 3 unique rows (PK dedup).
- `test_offset_tracking_picks_up_only_new_lines` — append a 4th line → second call ingests only that one.
- `test_escape_hatch_disables_ingest` — `CLAWMETRY_DISABLE_CLAUDE_SESSION_INGEST=1` short-circuits before discovery.
- `test_encode_cwd_matches_claude_code_convention` — slug encoder matches macOS / Linux / single-dot examples.
- `test_translate_handles_lines_without_timestamp` — drops `None`, never raises.
- `test_e2e_against_real_user_file_if_present` — local-only smoke that pulls the user's known 22:12 line from the real file (skipped on CI).

Existing sync test suite (`test_local_store_sync_integration.py` + sidecar suites + trial-gating) — 31 tests, all green.

## Cross-platform

Claude Code uses the `~/.claude/projects/` layout on every OS; the slug encoding is uniform. `_encode_cwd_for_claude_projects` test asserts macOS, Linux, and single-dot CWDs all encode correctly. Windows users with WSL paths fall under the Linux case.

## Test plan

- [x] Unit tests pass (`pytest tests/test_openclaw_claude_session_ingest.py -v`)
- [x] Existing sync suite still green (31 tests)
- [x] `python3 -c "import ast; ast.parse(open('clawmetry/sync.py').read())"` passes
- [x] `bash -n install.sh` passes
- [x] Real-file E2E on user machine: 293 events ingested, 22:12 "hello, how are you?" present
- [ ] Post-merge: confirm Brain shows recent Telegram conversations on user's live install

### DIFF
--- a/clawmetry/sync.py
+++ b/clawmetry/sync.py
@@ -2472,6 +2472,390 @@ def sync_claude_cli_sessions(config: dict, state: dict, paths: dict) -> int:
     return total
 
 
+# ── Sync: OpenClaw → Claude Code session JSONL (process-discovered) ──────────
+#
+# OpenClaw delegates execution to a ``claude`` CLI subprocess. Claude Code
+# writes its session transcript to:
+#
+#     ~/.claude/projects/<encoded-cwd>/<session-id>.jsonl
+#
+# where ``<encoded-cwd>`` is the agent's working directory with ``/`` and ``.``
+# both replaced by ``-``. The user observed that Telegram conversations
+# weren't appearing in Brain even though the file
+# (``-Users-vivek--openclaw-workspace/49f1d9fc-….jsonl``) was being actively
+# written. The legacy ``sync_claude_cli_sessions`` path covers most cases, but
+# only when ``~/.openclaw/agents/main/sessions/sessions.json`` carries the
+# ``claudeCliSessionId`` binding for that session — which is missing for any
+# session that was started before OpenClaw wrote its binding, or where the
+# binding was lost on a crash. This complementary, defense-in-depth path
+# discovers the live session file by inspecting the running ``claude``
+# subprocess directly, so capture is independent of OpenClaw's index file.
+#
+# Discovery works on macOS, Linux, and Windows because Claude Code uses the
+# same ``~/.claude/projects/`` layout everywhere; the encoding is uniform.
+#
+# Default: ON. Escape hatch: ``CLAWMETRY_DISABLE_CLAUDE_SESSION_INGEST=1``.
+# Per the DuckDB-first / never-default-off-a-MOAT-capability rule
+# (feedback_local_store_default_off_killed_moat 2026-05-13).
+
+
+def _encode_cwd_for_claude_projects(cwd: str) -> str:
+    """Reproduce Claude Code's project-dir naming convention.
+
+    Claude Code derives the ``~/.claude/projects/<dir>`` slug by replacing
+    both ``/`` and ``.`` with ``-`` in the agent's CWD. So
+    ``/Users/vivek/.openclaw/workspace`` → ``-Users-vivek--openclaw-workspace``
+    (the leading ``-`` comes from the absolute path's leading ``/``; the
+    consecutive ``--`` comes from ``/.`` collapsing).
+    """
+    return (cwd or "").replace("/", "-").replace(".", "-")
+
+
+def _looks_like_openclaw_process(proc) -> bool:
+    """True if a psutil.Process appears to be a claude-cli subprocess
+    spawned by OpenClaw.
+
+    Heuristics (any one is sufficient):
+      * cmdline contains ``--mcp-config`` whose argument lives under
+        ``~/.openclaw/`` (OpenClaw passes its own MCP config in)
+      * cwd is under ``~/.openclaw/`` (the agent's working directory
+        is the OpenClaw workspace)
+      * any ancestor process name/exe matches ``openclaw`` or
+        ``openclaw-gateway``
+
+    All access is wrapped because psutil throws ``AccessDenied`` /
+    ``NoSuchProcess`` freely on macOS sandbox-restricted processes.
+    """
+    oc_dir = os.path.realpath(_get_openclaw_dir())
+    try:
+        cmdline = proc.cmdline() or []
+    except Exception:
+        cmdline = []
+    # --mcp-config <path-under-~/.openclaw/>
+    for i, tok in enumerate(cmdline):
+        if tok in ("--mcp-config", "--mcp_config") and i + 1 < len(cmdline):
+            try:
+                mcp_path = os.path.realpath(os.path.expanduser(cmdline[i + 1]))
+                if mcp_path.startswith(oc_dir):
+                    return True
+            except Exception:
+                pass
+    # cwd under ~/.openclaw/
+    try:
+        cwd = os.path.realpath(proc.cwd())
+        if cwd.startswith(oc_dir):
+            return True
+    except Exception:
+        pass
+    # ancestor named openclaw* — walk up at most 6 hops to bound cost
+    try:
+        cur = proc
+        for _ in range(6):
+            cur = cur.parent()
+            if cur is None:
+                break
+            try:
+                name = (cur.name() or "").lower()
+            except Exception:
+                name = ""
+            if "openclaw" in name:
+                return True
+    except Exception:
+        pass
+    return False
+
+
+def _discover_openclaw_claude_session_files() -> list[tuple[str, str]]:
+    """Return ``[(session_id, jsonl_path), …]`` for live Claude Code
+    sessions whose owning ``claude`` CLI process appears to belong to
+    OpenClaw.
+
+    Best-effort. Returns ``[]`` if psutil is unavailable, no claude
+    processes are running, or all paths are unreadable. Never raises.
+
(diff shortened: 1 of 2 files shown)


vivekchand:
## UX review

This PR is backend-only — pure ingest plumbing in `clawmetry/sync.py` plus tests. No HTML/CSS/JS surface touched. One UX-adjacent observation, then stop:

- 🟡 P1 — `clawmetry/sync.py:7128` log line: `"Synced N events (M from claude-cc), L log lines, …"`. The new `(M from claude-cc)` parenthetical is appended *only* to the events count, but a non-technical operator who runs `clawmetry` and suddenly sees event volume triple has no in-product explanation of *why*. The log line is the only signal. Two small fixes: (a) use a plain-English label, e.g. `"events (incl. M Claude Code subprocess events)"`, so `claude-cc` doesn't read as a typo; (b) on the first cycle that ingests >0 from this path, emit a one-time INFO line like `"OpenClaw → Claude Code session capture is ON. To disable: CLAWMETRY_DISABLE_CLAUDE_SESSION_INGEST=1"`. That gives the user the off-switch the moment they notice the bump, instead of having to grep the changelog.

- 🟢 nit — Brain UI angle (out of this PR's diff but worth noting before the next one): per the implementing agent's own note in the PR body's test plan ("confirm Brain shows recent Telegram conversations on user's live install"), there is currently no way for a user to *see* whether this path is working other than waiting for messages to appear. A tiny "data sources" line in the dashboard footer ("ingesting from: openclaw/sessions, openclaw/telegram, openclaw/claude-cc (3 live)") would close that loop and double as proof-of-life.

vivekchand:
## Product review

Genuinely high-leverage PR — closes a "the dashboard is silently lying" bug the user explicitly flagged ("3000% a bug in your understanding"). Verified against the live file with 293 real events ingested. That said, four first-principles concerns, worst first:

- 🔴 P0 — Implementing agent's own follow-up finding ships unaddressed. From the diff at `clawmetry/sync.py:241-247`: events get filed with `agent_type="openclaw"`, `agent_id="main"`, `session_id=<Claude UUID>`. There is no corresponding row written to the `sessions` table for that Claude UUID. Brain rendering joins events to sessions; rows whose `session_id` has no matching `sessions` row are likely to render as orphans or be filtered out entirely. We just spent 400 LOC capturing data Brain may still not display. **Action:** open follow-up issue (filing now) and either (a) upsert a synthetic `sessions` row keyed on the Claude UUID at ingest time, or (b) re-key events to the OpenClaw session UUID and stash the Claude UUID in a side column. Pick one before the next release or the user files the same bug again with the same words.

- 🔴 P0 — Privacy/scope creep. `_looks_like_openclaw_process` (sync.py:48-99) is a 3-clause OR, but clause #3 (any ancestor process named `openclaw*`) means **every** `claude --resume` subprocess launched from a terminal that has an `openclaw` ancestor anywhere up the tree gets tailed — including Claude Code sessions the user spawned by hand for unrelated work in an OpenClaw-adjacent shell. Worse, the macOS sandbox fallback (sync.py:176-186) globs `~/.claude/projects/*/<session-id>.jsonl` by session-id alone, with zero OpenClaw provenance check on the *file*. A non-OpenClaw `claude --resume <id>` running anywhere on the box, if its session-id happens to match, gets ingested. **Action:** tighten — require BOTH cwd-under-openclaw AND a positive provenance signal; make the fallback path verify the file's own `cwd` field (already in every line) is under `~/.openclaw/` before accepting any of its rows.

- 🟡 P1 — psutil is an undeclared, soft dependency. `_discover_openclaw_claude_session_files` returns `[]` if `import psutil` fails (sync.py:127-130). `requirements.txt` doesn't list it. CLAUDE.md is explicit: "Minimal dependencies — Flask + waitress + cryptography." Result: the MOAT capability is silently OFF for any user who hasn't installed psutil, which is most of them. This is exactly the failure mode `feedback_local_store_default_off_killed_moat` is about. **Action:** either add `psutil` to `requirements.txt` (it's already a transitive dep via several common stacks, weight is small) and document it, or implement a `pgrep -af claude`/`ps -ef` shell fallback so the discovery works on bare-bones installs. As written, "default ON" is aspirational, not real.

- 🟡 P1 — Tier-strategy implication: this PR ships full Claude Code transcript capture into OSS. That's fine for the funnel (more local data → better local Brain → stickier free tier), but Claude Code session JSONLs are the richest, highest-cardinality data we've ever captured — full reasoning traces, tool-use chains, the works. The conversion CTA opportunity is huge: once a user sees their first reasoning trace render in Brain, that is the moment to surface "Want >24h retention of these traces? Start a 7-day Pro trial." Per `project_free_plan_upsell.md`, every click is a conversion opportunity, and "your first reasoning trace from Claude Code just landed" is the highest-intent moment we'll get this quarter. **Action:** wire a one-time toast/banner into Brain when the first `openclaw-cc:*` event renders. Not in this PR, but file as a P1 follow-up.

PR #1228, opened 2026-05-14
fix(gateway_tap): default WS tap to OFF until OpenClaw scope grant lands

### LINKED ISSUE #1221: [Product P0 from PR #1213] Default WS gateway tap to OFF until OpenClaw scope grant lands
From the post-merge product review of PR #1213.

## Problem
PR #1213 ships `clawmetry/gateway_tap.py` (~664 LOC) plus daemon wiring and 461 LOC of tests. The tap is **default-ON** (per the MOAT mandate `feedback_local_store_default_off_killed_moat`).

Today it does nothing useful end-to-end:
1. Connects to the OpenClaw gateway WS — succeeds.
2. Sends `connect` handshake — succeeds, but `auth.scopes = []` because OpenClaw 2026.5.7's default token-auth grants no scopes.
3. Subscribes to `sessions.messages.subscribe` — rejected with `missing scope: operator.read`.
4. Logs a WARNING per reconnect cycle: `gateway WS tap connected (degraded — no message subscribe; granted scopes=[])`.
5. Receives ZERO inbound message rows.
6. Reconnects on the next disconnect → loops 4-5 forever.

## Why this is P0 (not just P1)
The MOAT default-ON rule was designed for fast-paths that silently bypass DuckDB and lose data. This is the inverse: a default-ON path that produces NO benefit, only log noise. Every dev's `~/.clawmetry/sync.log` will fill with "degraded" WARN lines until upstream openclaw/openclaw#80836 lands. Worse: when a real user files "my sync log is full of warnings", we'll burn support cycles explaining a feature they don't use.

## Proposed fix
Two-line change in `clawmetry/gateway_tap.py:start`:

```python
# Was: default-ON, opt-out via CLAWMETRY_DISABLE_WS_TAP=1
# Now: default-OFF until upstream scope grant lands; opt-in via CLAWMETRY_ENABLE_WS_TAP=1
if os.environ.get("CLAWMETRY_ENABLE_WS_TAP", "").strip() not in ("1", "true", "yes"):
    log.debug("gateway WS tap disabled (set CLAWMETRY_ENABLE_WS_TAP=1 to enable)")
    return None
```

## Flip-day plan
The DAY OpenClaw lands openclaw/openclaw#80836 (or equivalent), open a follow-up PR that:
1. Reverts the env-var name back to `CLAWMETRY_DISABLE_WS_TAP` (opt-out).
2. Makes default-ON.
3. Bumps the OSS `__version__`.
4. Adds a CHANGELOG entry: "Telegram + sibling-channel inbound messages now appear in the Brain tab."

## Acceptance
- [ ] Default behavior in `gateway_tap.start()` is no-op when env var unset
- [ ] Test `test_disable_env_var_short_circuits_start` updated to test the opt-in inverse
- [ ] PR description documents the flip-day plan
- [ ] Issue closes when the flip-day PR merges

## Related
- Upstream blocker: [openclaw/openclaw#80836](https://github.com/openclaw/openclaw/issues/80836)
- Filed-on-our-side: [openclaw/openclaw#81775](https://github.com/openclaw/openclaw/issues/81775)
- The "three parallel channel ingest paths" follow-up from PR #1212 review (issue #1220) should consolidate the projection logic in `gateway_tap._handle_frame` at the same time.

JaiCodes77:
Closes #1221

## What

Swap the WS gateway tap from default-ON (opt-out via `CLAWMETRY_DISABLE_WS_TAP`) to default-OFF (opt-in via `CLAWMETRY_ENABLE_WS_TAP`).

## Why

OpenClaw 2026.5.7 returns `auth.scopes = []` — the `operator.read` scope required by `sessions.messages.subscribe` isn't granted yet. With default-ON, every install enters a reconnect loop that fills `~/.clawmetry/sync.log` with "degraded" warnings and delivers zero messages.

## Changes

- **`gateway_tap.py`**: Renamed `_DISABLE_ENV` → `_ENABLE_ENV`, flipped the guard in `start()` so it's a no-op unless `CLAWMETRY_ENABLE_WS_TAP=1` is set, updated module docstring.
- **`test_gateway_tap.py`**: Updated fixture to clear the new env var, rewrote `test_disable_env_var_short_circuits_start` to verify default-OFF behavior, added `test_enable_env_var_allows_start` for the opt-in path.

## Flip-day plan

Once upstream [openclaw/openclaw#80836](https://github.com/openclaw/openclaw/issues/80836) lands and scopes are granted, a follow-up PR will:
1. Revert env var to `CLAWMETRY_DISABLE_WS_TAP` (opt-out)
2. Make default-ON
3. Bump `__version__`
4. Add CHANGELOG entry

### DIFF
--- a/clawmetry/gateway_tap.py
+++ b/clawmetry/gateway_tap.py
@@ -48,10 +48,9 @@ Failure modes
 Disable
 -------
 
-Set ``CLAWMETRY_DISABLE_WS_TAP=1`` in the environment to opt out
-entirely (escape hatch). Default is ON per the MOAT-mandate
-"DuckDB fast paths must default-on" rule (memory note
-``feedback_local_store_default_off_killed_moat``).
+Set ``CLAWMETRY_ENABLE_WS_TAP=1`` in the environment to opt in.
+Default is OFF until the upstream OpenClaw server grants the
+required ``operator.read`` scope.
 """
 
 from __future__ import annotations
@@ -108,8 +107,8 @@ _BACKOFF_INITIAL_SEC = 2.0
 _BACKOFF_MAX_SEC = 60.0
 
 
-# Disable env var (escape hatch — feature is default-ON).
-_DISABLE_ENV = "CLAWMETRY_DISABLE_WS_TAP"
+# Enable env var (feature is default-OFF until upstream grants scopes).
+_ENABLE_ENV = "CLAWMETRY_ENABLE_WS_TAP"
 
 
 # ── Frame normalization ─────────────────────────────────────────────────
@@ -587,8 +586,8 @@ def start(config: dict) -> GatewayTap | None:
     from the live OpenClaw config (mirroring dashboard's
     ``_detect_gateway_token`` / ``_detect_gateway_port``).
     """
-    if os.environ.get(_DISABLE_ENV, "").strip() in ("1", "true", "TRUE", "yes"):
-        log.info("gateway WS tap disabled (%s=1)", _DISABLE_ENV)
+    if os.environ.get(_ENABLE_ENV, "").strip() not in ("1", "true", "yes"):
+        log.debug("gateway WS tap disabled (set CLAWMETRY_ENABLE_WS_TAP=1 to enable)")
         return None
 
     url, token = _detect_gateway_endpoint()
--- a/clawmetry/sync.py
+++ b/clawmetry/sync.py
@@ -6474,8 +6474,8 @@ def run_daemon() -> None:
     # memory; gateway.log only carries outbound ACKs (no body), and no
     # JSONL is written for inbound. Without this tap, ClawMetry can
     # NEVER show real Telegram conversations on the Brain tab.
-    # Default-ON per the MOAT mandate; opt out via
-    # CLAWMETRY_DISABLE_WS_TAP=1.
+    # Default-OFF until upstream grants scopes; opt in via
+    # CLAWMETRY_ENABLE_WS_TAP=1.
     try:
         from clawmetry import gateway_tap as _gw_tap
         _gw_tap.start(config)
--- a/tests/test_gateway_tap.py
+++ b/tests/test_gateway_tap.py
@@ -20,8 +20,8 @@ What this pins
    same ``(provider, chat_id, message_id)`` already exists — the
    COALESCE upsert in ``ingest_channel_message`` makes the body
    stick.
-6. ``CLAWMETRY_DISABLE_WS_TAP=1`` short-circuits ``start()`` to a
-   no-op (escape-hatch behavior).
+6. ``start()`` is a no-op by default; set ``CLAWMETRY_ENABLE_WS_TAP=1``
+   to opt in.
 """
 
 from __future__ import annotations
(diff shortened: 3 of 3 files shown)


vivekchand:
## UX review

Diff is mostly backend (env-var rename + test rewrite), so scope is narrow — but a few sharp edges that an Apple-grade pass would catch:

- 🟡 P1 — `tests/test_gateway_tap.py:434` — function name `test_disable_env_var_short_circuits_start` is now actively misleading: there is no disable env var anymore, and the body asserts default-OFF behavior. Rename to `test_default_is_no_op` (and update the docstring at line 23 which still says "short-circuits"). Stale test names rot faster than stale code because they are the first thing a future maintainer greps for.
- 🟡 P1 — `clawmetry/gateway_tap.py:589` — case-sensitivity regression: the old guard accepted `"1" | "true" | "TRUE" | "yes"`; the new guard accepts `"1" | "true" | "yes"` only. A user who exports `CLAWMETRY_ENABLE_WS_TAP=TRUE` (perfectly natural for ALL-CAPS env-var conventions) will silently get a no-op. Either lowercase the value before comparing (`.lower() in ("1","true","yes")`) or restore `"TRUE"` to the tuple. Pick one and apply it consistently.
- 🟡 P1 — `clawmetry/gateway_tap.py:590` — log level dropped from `info` → `debug` for the disabled branch. Default-OFF + silent default = invisible feature. A user who set `CLAWMETRY_ENABLE_WS_TAP=1` based on an older doc and made a typo gets zero feedback. Keep `log.info("gateway WS tap is opt-in (set CLAWMETRY_ENABLE_WS_TAP=1 to enable)")` so the no-op is at least audible at default verbosity.
- 🟢 nit — `tests/test_gateway_tap.py:441` — the new `test_enable_env_var_allows_start` asserts `res is None` for two completely different reasons (env-gate vs. missing endpoint). That is the kind of test that passes forever even when the gate breaks. Mock `_detect_gateway_endpoint` to return a value and assert a non-None tap object, otherwise the assertion is a tautology.

vivekchand:
## Product review

First-principles read: the right call (don't ship a guaranteed-noisy feature), but the rollout has gaps that will cost us once.

- 🔴 P0 — Missing `[RELEASE]` prefix means this fix does NOT auto-publish. Every user who pip-installs between now and the next [RELEASE] PR continues to get the default-ON tap that floods `~/.clawmetry/sync.log`. Issue #1221 stays user-visible. Either re-title this PR `[RELEASE] fix(gateway_tap): …` before merge or chain a [RELEASE] bump immediately after — don't let a "fix the noise" PR sit unshipped.
- 🟡 P1 — The flip-day plan in the PR description is excellent and will be lost the moment the PR closes. Move the 4-step revert plan into a `# FLIP-DAY: when openclaw/openclaw#80836 ships, …` block-comment directly above `_ENABLE_ENV` in `gateway_tap.py:110`. PR descriptions are not a maintenance surface; the next person debugging this in 6 months will be reading the file, not the closed PR.
- 🟡 P1 — Default-OFF here is justified (upstream-blocked feature, zero functionality lost because #1224/#1227 cover the conversation-capture gap). Make that explicit in the docstring at `gateway_tap.py:48-53` so the next reviewer doesn't reflexively flip it back citing the `feedback_local_store_default_off_killed_moat` mandate. One sentence: "This is NOT a DuckDB fast-path — it's an upstream-blocked subscription; default-OFF until scopes are granted."
- 🟢 nit — Conversion angle is null on this one (pure bug-noise reduction, no Cloud-Pro surface touched). A quieter `sync.log` does indirectly raise trust at the OSS install step, which is the funnel entry — worth tracking the support-ticket count for "noisy logs" to confirm impact post-release.
- 🟢 — Same author shipped #1223 cleanly. Pattern is forming. Worth a thank-you in the merge comment; external contributors who fix real P0s without being asked are the rarest input we have.

PR #1227, opened 2026-05-14
[RELEASE] feat(sync): walk OpenClaw sessions.json → claude-cli sessions + subagents + tool-results (closes #1226)

vivekchand:
Closes #1226 — the P0 follow-up review issue from PR #1224.

## Diya's diagnosis (verbatim)

The user's OpenClaw bot Diya verified PR #1224 on the live install and
flagged three specific gaps:

> *Sub-agent transcripts and large tool outputs live at*
> *`49f1d9fc-.../subagents/` and `49f1d9fc-.../tool-results/` — alongside the*
> *top-level transcript.*

> *It needs to follow `claudeCliSessionId` into `~/.claude/projects/` to get*
> *the real conversation.*

> *`sessions.json` lists a `sessionFile` at `~/.openclaw/agents/main/sessions/*
> *625c0ad9-….jsonl`, but I checked: that file does not exist. The generic*
> *OpenClaw write path only runs when OpenClaw uses its built-in agent loop.*
> *With the Claude CLI provider, that path is bypassed and the CLI owns the*
> *transcript.*

That last finding is the same root cause issue #1226 flagged: Brain renders
nothing because the `sessions` table indexes by `625c0ad9` but events from
PR #1224 landed under `49f1d9fc`. Without a session row, even the typed
`query_sessions_table` view is blank, and aggregations/UI render orphans.

## Architecture (after this PR)

```
~/.openclaw/agents/main/sessions/sessions.json
  └─► agent:main:telegram:direct:1532693273
        ├─ sessionId:           625c0ad9-…   (OpenClaw UUID — JOIN KEY)
        ├─ claudeCliSessionId:  49f1d9fc-…   (Claude Code UUID)
        ├─ origin:              {label: "Vivek Chand id:1532693273", …}
        ├─ workspaceDir:        /Users/vivek/.openclaw/workspace
        └─ status:              done
                    │
                    ▼  encode CWD: / → -, . → -
~/.claude/projects/-Users-vivek--openclaw-workspace/
        ├─ 49f1d9fc-….jsonl                  (top-level transcript)
        └─ 49f1d9fc-…/
            ├─ subagents/agent-aXXX.jsonl    (one JSONL per sub-agent)
            └─ tool-results/bXXX.txt         (opaque text dumps)
                    │
                    ▼
DuckDB
  ├─ sessions   — 1 row keyed by OPENCLAW UUID 625c0ad9-…
  │              (title, status, started_at, last_active_at,
[...]

PR #1232, opened 2026-05-14
[RELEASE] fix(sync): unify event id derivation across 3 ingest paths + one-time dedup migration

vivekchand:
## User report (verbatim)

> "I see messages are duplicated -- can this be somehow avoided?? preferably before writing to duckdb or when you think is the right place"

The user observed 2x and 3x duplicate rows in Brain. Confirmed against their live DuckDB: 976 total events, 266 of which were duplicates spread across assistant/user/attachment/queue-operation rows.

## Root cause

Three independent ingest paths converged on the same logical Claude Code message but used **different** ``events.id`` derivations, defeating ``INSERT OR IGNORE`` at the local-store boundary:

| # | Source | Function | id scheme |
|---|--------|----------|-----------|
| 1 | Legacy (pre-#1224) | ``sync_claude_cli_sessions`` | bare uuid |
| 2 | Process inspection (#1224) | ``sync_openclaw_claude_sessions`` | ``openclaw-cc:<cc-uuid>:top:<uuid>`` |
| 3 | sessions.json walk (#1227) | ``sync_openclaw_claude_sessions_via_index`` | ``openclaw-cc:<oc-uuid>:top:<uuid>`` |

PR #1227's ``skip_claude_ids`` short-circuit only suppressed path #2 vs #3 collisions. Path #1 was always free to re-write the same logical event under a bare uuid, and historical data already carried the dupes regardless.

Sample evidence (from the user's live DuckDB):

```
ts=2026-05-14T22:03:01.267  type=assistant  session=625c0ad9
  id=a8f0ec6e-47d9-4d31-991c-f12a13886ef0                                ← path 1: bare UUID
  id=openclaw-cc:625c0ad9-...:top:a8f0ec6e-47d9-4d31-991c-f12a13886ef0  ← path 3: PR #1227

ts=2026-05-14T22:02:50.880  type=queue-operation  session=625c0ad9  COUNT=3
  id=625c0ad9-...:2026-05-14T22:02:50.880Z:queue-operation  ← path 1 fallback
  id=openclaw-cc:625c0ad9-...:top:line:1040441               ← path 3 (no uuid)
  id=openclaw-cc:625c0ad9-...:top:line:1040442               ← path 3 (next line)

May 9 attachment events: 4 rows each = 2 distinct uuids x 2 paths.
```

### Half 1: unified id derivation (write side)

New ``_canonical_event_id`` helper in ``clawmetry/sync.py``:

* Source has Claude Code uuid → ``cc-msg:<uuid>``
* Source lacks uuid (queue-operation, summary, …) → ``cc-derived:<session>:<ts>:<type>:<md5(canonical_body)[:16]>``

The MD5 is computed over a normalised projection that excludes per-path noise (``_claude_session_id``, ``_openclaw_session_id``, ``_oc_cc_kind``, ``_subagent_file``, ``parentUuid``/``parentId``, ``uuid``/``id``), so the digest matches across all three ingest paths even when the wrapping translate function injects metadata into ``data``.

Wired into all three paths:
* ``_translate_claude_cli_event`` tags the event with ``_cc_source: True`` so ``_local_ingest_session_batch`` routes it through the canonical helper instead of its legacy session+ts+type fallback.
* ``_translate_claude_session_line`` calls the canonical helper directly for ``kind="top"`` (the source of all the dupes).
* Subagent rows (``kind="subagent"``) intentionally keep the path-scoped ``openclaw-cc:<oc>:subagent:<file>:<uuid>`` scheme — different subagent files legitimately reuse uuids, so the file basename must scope.
* Tool-result rows already had a path-unique scheme (filename-keyed) and are untouched.

The internal ``_cc_source`` marker is stripped before the row reaches the stored ``data`` BLOB.

### Half 2: one-time dedup migration (cleans existing data)

``SCHEMA_VERSION`` bumped 6 → 7. New ``_run_dedup_migration_v7`` runs once when the daemon first opens a v6 store:

```sql
DELETE FROM events WHERE rowid IN (
  SELECT rowid FROM (
    SELECT rowid, MIN(rowid) OVER (
      PARTITION BY session_id, ts, event_type, (<dedup_key>)
    ) AS keep_rowid
    FROM events
  ) WHERE rowid != keep_rowid
)
```

where ``<dedup_key>`` extracts the underlying uuid from any of the three id schemes (``cc-msg:<uuid>``, bare uuid, ``openclaw-cc:...:top:<uuid>``, ``cc-derived:...:<digest>``) and falls back to ``md5(data::VARCHAR)`` for ``line:N``-style ids whose bodies legitimately differ. Idempotent: a second pass after schema_version reaches 7 is gated out and runs zero queries.

Also adds ``CREATE INDEX IF NOT EXISTS idx_events_session_ts_type ON events(session_id, ts, event_type)`` for the dedup query and any future analytical scan.

DuckDB ``md5()`` syntax verified before relying on it — accepts both VARCHAR and BLOB and returns VARCHAR (no cast gymnastics needed).

## E2E verification (against a copy of the user's live DuckDB)

| Metric | Before | After |
|--------|--------|-------|
| Total events | **976** | **710** (266 dupes removed) |
| 22:03:01 assistant rows | 2 | **1** |
| 2026-05-09 attachment | 4 | **2** (2 distinct uuids, each was 2x) |
| 22:02:50 queue-operation | 3 | 3 (3 legitimately distinct events) |
| ``subagent:*`` rows | 302 | 302 (path-scoped, untouched) |
| ``tool-result`` rows | 4 | 4 (path-scoped, untouched) |
| Legitimate single-row events preserved | 351 | **351 of 351** |
| Remaining (ts,type,session,uuid) dup groups | 64 | **0** |
| ``schema_version`` | 6 | 7 |

Migration on a fresh empty store: 0 deletions, schema_version=7, no errors. Idempotency: re-opening the store after migration runs zero deletes. New index present.

## Tests

``tests/test_event_id_unification_and_dedup.py`` — 10 tests, all passing:

* canonical id is deterministic + path-independent
* canonical id is uuid-driven, independent of which ``session_id`` is passed (the bug)
* injected metadata (``_claude_session_id``, ``_openclaw_session_id``, ``_oc_cc_kind``, ``parentUuid``, …) does NOT change the id
* **all 3 translate functions produce identical id for the same source line** (the dedup contract)
* subagent rows keep file-scoped ids (regression-protect against an over-eager future cleanup)
* canonical id falls back to ``cc-derived:`` for events without a uuid
* dedup migration collapses each duplicate group to one row
* migration is idempotent (second pass = no-op)
* migration creates the new index
* migration is safe on a fresh empty store

All pre-existing ``tests/test_openclaw_claude_session_ingest.py`` and ``tests/test_openclaw_sessions_followup.py`` still pass.

## Why future writes can't dupe

After this lands, all 3 ingest paths route the same source jsonl line through the same ``_canonical_event_id`` helper, so they all compute identical ids. The store's existing ``INSERT OR IGNORE INTO events ... PRIMARY KEY (id)`` then collapses re-deliveries silently — no further migration ever needed for this class of bug.

## Hard constraints honoured

- Branch: ``release/unify-event-ids-and-dedup`` off ``main``
- Title prefixed ``[RELEASE]`` — auto-bumps patch + publishes to PyPI on merge
- ``__version__`` NOT touched manually
- User's daemon NOT killed
- ``~/.clawmetry/`` files NOT touched (all E2E ran against ``/tmp/cm-dedup-test/``)
- ``python3 -c "import ast; ast.parse(open('clawmetry/sync.py').read()); ast.parse(open('clawmetry/local_store.py').read())"`` passes

### DIFF
--- a/clawmetry/local_store.py
+++ b/clawmetry/local_store.py
@@ -117,7 +117,7 @@ LOCAL_MAX_BYTES = int(
     float(os.environ.get("CLAWMETRY_LOCAL_MAX_GB", "5.0")) * 1024 * 1024 * 1024
 )
 
-SCHEMA_VERSION = 6
+SCHEMA_VERSION = 7
 
 # ── Two-layer schema (multi-agent) ──────────────────────────────────────────
 #
@@ -156,6 +156,10 @@ _DDL = [
     "CREATE INDEX IF NOT EXISTS idx_events_agent_ts    ON events(agent_id, ts)",
     "CREATE INDEX IF NOT EXISTS idx_events_type_ts     ON events(event_type, ts)",
     "CREATE INDEX IF NOT EXISTS idx_events_atype_ts    ON events(agent_type, ts)",
+    # Speeds up the v7 dedup migration (#1232) and any future analytical
+    # query that wants to scan a single session's timeline by event_type
+    # without hitting the full ts index.
+    "CREATE INDEX IF NOT EXISTS idx_events_session_ts_type ON events(session_id, ts, event_type)",
     """
     CREATE TABLE IF NOT EXISTS sessions (
         agent_type      VARCHAR NOT NULL DEFAULT 'openclaw',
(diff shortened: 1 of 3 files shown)


vivekchand:
## UX review

Backend-only PR (no HTML/CSS/JS surfaces touched), but it has a real user-visible side effect on the very first daemon start after upgrade — calling that out:

- 🟡 **P1 — silent 27% row drop on first open looks like data loss.** The migration deletes 266/976 rows on the user's live data (per the PR body). At ``log.info`` level that's invisible unless someone is tailing the daemon log. A user who opens the dashboard after auto-update sees Brain row counts plummet and has zero context for why. Action: surface a one-shot dashboard banner ("Cleaned up N duplicate events from earlier versions — totals are now accurate") gated on a ``schema_version_v7_dedup_count`` row in ``meta``, dismissible. Same data the log line already has — just plumb it to ``/api/overview`` for one render.
- 🟡 **P1 — log line at ``info`` will be lost in stdout noise.** ``clawmetry/local_store.py:163`` logs the deletion delta only at info; in the user's typical run that's drowned out. Action: also write the delta + before/after counts into the ``meta`` table so we can render it in ``/api/system-health`` and so support can read it back later without log scrubbing.
- 🟢 **nit — wording.** ``"v7 dedup migration removed N duplicate event row(s)"`` is fine for ops; if you do surface it in-product, soften to ``"Removed N duplicate brain events left over from a sync bug fixed in v0.12.x"`` so it reads as a fix, not a deletion.

vivekchand:
## Product review

First-principles pass on the dedup design. No paid-tier leakage concerns — this is a straight correctness fix in the OSS sync path, doesn't alter the OSS↔Cloud-Pro split. Substantive concerns are technical:

- 🔴 **P0 — destructive DELETE with no backup, on every user's daemon, on first start.** ``_run_dedup_migration_v7`` issues an unconditional ``DELETE FROM events`` on the user's only copy of their data. If the regex is wrong on any user's id corpus we've never seen, we silently destroy real events with no recovery. Standard practice for a destructive migration is ``CREATE TABLE events_pre_v7_dedup AS SELECT * FROM events`` first (or ``COPY (SELECT...) TO 'events_pre_v7_dedup.parquet'``), then DELETE, then drop the backup after N daemon starts with no rollback flag set. Action: add a one-shot ``events_pre_v7_dedup`` snapshot before DELETE; gate cleanup on a future ``SCHEMA_VERSION = 8``.
- 🔴 **P0 — concurrent-daemon race on migration.** ``_migrate`` only takes ``self._write_lock`` (an in-process ``threading.Lock`` at ``local_store.py:703``). DuckDB's file lock serialises writes between processes, but the ``schema_version`` SELECT happens before the DDL+migration; two daemons starting simultaneously after upgrade can both read ``current=6``, both run the dedup DELETE, both stamp v7. The DELETE is idempotent at the row level, but you do 2× the work and double-log "removed N rows" — which makes the support story harder if a user shows you logs. Action: wrap the entire ``current<7`` block in a ``BEGIN EXCLUSIVE``-equivalent (DuckDB ``BEGIN TRANSACTION``) and re-read schema_version inside the txn (double-checked locking).
- 🟡 **P1 — ``md5(data::VARCHAR)`` fallback collapses intentional re-emissions.** The dedup key for ``line:N``-style ids hashes the entire ``data`` blob. That correctly handles "same line read twice." It also incorrectly merges two semantically distinct events that happen to have identical bodies — e.g. an automated retry that emits the same payload, or replay-style debug events. ``cc-derived:`` synthesised ids already carry the digest in the id itself, so the partition is fine for those; the risk is the bare-line-N branch. Action: include ``rowid`` ordering metadata in the partition key (e.g. ``MIN(created_at)`` window) so two same-body events at the same ts but inserted seconds apart aren't merged.
- 🟡 **P1 — ``_canonical_event_id`` short-circuits on uuid presence; what if Claude Code regenerates the uuid on edit?** If a user edits a message in CC and the on-disk jsonl gets re-emitted with a new ``uuid`` for the same logical message, the new and old will land as distinct ``cc-msg:`` ids and Brain shows both. Today's bug was 2 ids per 1 logical event; this would be 2 logical events per 1 user intent. Worth a sanity check on a CC edit/regenerate flow before declaring this class of bug closed.
- 🟢 **nit — return value of ``_run_dedup_migration_v7`` is dead code.** The docstring says it returns the deleted count, but the implementation acknowledges DuckDB doesn't expose ``changes()`` and falls back to 0. The caller already does its own before/after COUNT(*). Just return ``None`` and drop the misleading docstring promise.

PR #1235, opened 2026-05-14
[RELEASE] fix(dashboard): heartbeat query returns 0 from non-empty table + SSE/spinner regressions

vivekchand:
## P0 incident — dashboard appears broken to users

User report (Peter Steinberger just starred ClawMetry — traffic incoming):

> "do whatever is required to fix it ASAP"

Live machine showed: "Load failed - retrying..." toast, all KPIs as "—",
"No heartbeats yet", Brain + Memory tabs stuck on "Loading..." — despite
the local DuckDB carrying **210 fresh heartbeats** and **976 events**, all
seconds-fresh, all under the right node_id.

User's live machine state: untouched. All work happened in
`/tmp/cm-dashfix-test/venv` against a separate test daemon on port 8906.
Their `~/.clawmetry/clawmetry.duckdb` and PIDs 25047 / 30550 / 30572
were never modified or stopped.

## Root causes (3 independent bugs collapsed the UI together)

1. **Heartbeat block scanned the wrong table.** `_compute_overview_heartbeat()`
   in `routes/overview.py` only looked for events with
   `session_type=heartbeat` (OpenClaw gateway sessions). When the gateway
   is stopped but the sync daemon is alive (the common Cloud-Free / OSS
   install), the `events` table has zero matching rows even though the
   dedicated `heartbeats` table has hundreds of fresh daemon-liveness
   pings. **Fix:** fall back to reading from the `heartbeats` table when
   no gateway-heartbeat events are found.

2. **Three fast-path helpers called `get_store()` with the default
   `read_only=False`,** triggering DuckDB's cross-process exclusive-lock
   retry budget (~2.5s burned per request) before falling through. The
   daemon already exposes a localhost HTTP proxy
   (`routes/local_query.py::local_store_via_daemon`) that wires past this
   lock — it just wasn't called from `_try_local_store_heartbeat_status`
   (`routes/health.py`), `_try_local_store_memory_files` /
   `_try_local_store_file` (`routes/infra.py`). **Fix:** route through
   the daemon proxy first; fall back to `read_only=True` direct open in
   single-process / dev-mode boots.

3. **`/api/overview` was 12-15 s on the user's machine,** blowing past
   the JS 3 s fetch timeout — that's what triggered the "Load failed -
   retrying..." toast (it's the catch-block in `loadAll()`,
   `clawmetry/static/js/app.js:1815`). **Root cause:** `_detect_anthropic_oauth()`
   ALWAYS called BOTH the daemon proxy AND a direct local_store open on
   every request, paying the lock-retry penalty even when the proxy
   succeeded. **Fix:** short-circuit the direct open when the proxy
   returned data.

## Before / after — `/api/overview` heartbeat block

```diff
- {
-   "expected_cadence_seconds": 1800,
-   "gap_seconds": null,
-   "last_heartbeat_ts": null,
-   "ok_ratio": null,
-   "sample_size": 0,
-   "status": null
- }
+ {
+   "expected_cadence_seconds": 1800,
+   "gap_seconds": 67,
+   "last_heartbeat_ts": "2026-05-14T22:48:25.857969+00:00",
+   "ok_ratio": null,
+   "sample_size": 20,
+   "status": "green"
+ }
```

## Latency improvements (against the user's live DuckDB, test daemon on :8906)

| endpoint | before | after |
|---|---|---|
| `/api/overview` | 12-15 s | 70-400 ms |
| `/api/heartbeat-status` | timeout (>5 s) | 19 ms |
| `/api/memory-files` | 12 s | 16 ms |
| `/api/memory-analytics` | 7 s | 16 ms |

## Screenshots — BEFORE vs AFTER

Captured headless via Playwright/Chromium. Both runs hit the same live
DuckDB (heartbeats table populated by the user's daemon).

### Overview tab
- **Before** — "Load failed - retrying..." red badge, all KPIs "—", flow
  shows generic "AI Model" placeholder:
  `screenshots/dashfix-before-overview.png`
- **After** — "Updated 12:53:20 AM" green, HEARTBEAT row shows
  "Healthy / 54s ago / 30m", MODEL claude-opus-4-7, TOKENS 471K, Flow
  shows real "Claude opus":
  `screenshots/dashfix-after-overview.png`

### Brain tab
- **Before** & **After** look the same here — the tab header renders, but
  the live event stream sub-panel says "Loading…" because it waits on the
  SSE handshake. Pre-existing behaviour, not a regression caused by this
  PR. Documented as a P0 follow-up below.
- `screenshots/dashfix-before-brain.png`,
  `screenshots/dashfix-after-brain.png`

### Memory tab
- **Before** & **After** — Explorer pane "Loading…", same on both. The
  `/api/memory-files` API is now 16 ms (was 12 s), so the spinner is
  unrelated to the API and lives in the JS tab-activation path.
  Documented as a P0 follow-up below.
- `screenshots/dashfix-before-memory.png`,
  `screenshots/dashfix-after-memory.png`

## Playwright assertion result

```
[before] refresh-time text: 'Load failed - retrying...'
[after]  refresh-time text: 'Updated 12:53:20 AM'
```

The end-to-end browser run confirmed the overview hot-path now lands a
successful initial load instead of catching into the retry toast.

## P0 follow-ups (not blocked on this PR)

- Brain tab "Live event stream" — verify the `/api/brain-stream` SSE
  handshake actually completes and the `Loading…` placeholder is replaced
  by the historical backfill from `/api/brain-history` (which the test
  proved is fast: 110 KB in 50 ms).
- Memory tab "Explorer Loading…" — JS audit; loadMemory may not fire on
  tab activation, only on initial render of the dashboard.
- ~30 other `routes/*` helpers still call `local_store.get_store()` with
  default RW — they likely all pay the same lock-contention tax. Sweep
  via the same daemon-proxy-first pattern.

## Verification — user's machine state untouched

- `~/.clawmetry/clawmetry.duckdb` not modified (test venv was read-only;
  test daemon on :8906 hit the daemon proxy on :53843, did not write).
- PIDs 25047 (sync daemon), 30550, 30572 (user's dashboard) — never
  signaled.
- Test installation isolated to `/tmp/cm-dashfix-test/venv`.

## Test plan

- [ ] After merge, `[RELEASE]` workflow auto-bumps to v0.12.190 and
      publishes to PyPI.
- [ ] Bump cloud's pin to 0.12.190 once PyPI propagates (3 min).
- [ ] User refreshes their dashboard; should see real heartbeat / model
      / token data instead of "—" placeholders.


vivekchand:
## UX review

- 🟡 P1 — **The retry toast is the wrong UX for a 12s wait.** `app.js:1815` catches a 3s fetch timeout and shows "Load failed - retrying...". From a user's perspective the dashboard *was* working — it just hadn't responded yet. Toast should distinguish "request errored" from "request still pending after Ns" (skeleton loaders, not red toast). Action: split the catch path — show skeleton until 8s, then a soft "Still loading..." pill, only escalate to red toast on actual exception.
- 🟡 P1 — **Heartbeat block now reads from two sources but the UI doesn't tell the user which.** When `ok_ratio` is None (the new daemon-liveness fallback path) the OK% column will render empty — likely as "—" or blank. A user looking at "Healthy / 54s ago / 30m / —" will think OK% is broken. Action: when `ok_ratio` is None, label the cell "n/a (gateway off)" with a tooltip explaining the daemon is alive but the gateway isn't reporting outcomes.
- 🟢 nit — **Brain + Memory tabs still show "Loading…" in the after-screenshots.** PR notes acknowledge this and the follow-up landed in #1239, but for users who pull v0.12.190 before #1239 propagates, the dashboard looks half-fixed. Action: ensure cloud's pin and the next OSS release coalesce both fixes; add a short note in the changelog calling out the spinner is a separate fix in 0.12.191+.
- 🟢 nit — **Backend-heavy diff with no UI surface change beyond the data populating correctly.** UX impact is real but indirect (perceived speed). The remaining UX debt is the loading-state language in `app.js`, not in the routes.

vivekchand:
## Product review

- 🔴 P0 — **The pattern is systemic, not local.** This PR fixed 3 RW `get_store()` callsites. Grep across `routes/` shows **46 `get_store()` calls vs only 18 `read_only=True`** — i.e., ~28 sites still default to RW and will pay the same ~2.5s lock-retry tax under the standard install (sync-daemon + dashboard, two processes). Every one of those is a latent dashboard-broken-under-real-load bug like the one Peter Steinberger nearly hit. Action: open a sweep PR converting all read-path `get_store()` calls to `get_store(read_only=True)` *and* prefer the daemon-proxy path; add a `flake8` / pre-commit grep check that flags `get_store()` without the `read_only=True` kwarg in `routes/`. Filing as a P0 issue.
- 🔴 P0 — **No CI gate caught a 12-15s `/api/overview`.** This bug existed for months and only surfaced when external traffic arrived. Action: add an API-latency smoke test to CI that times every documented endpoint in `CLAUDE.md` against a seeded fixture DuckDB with the daemon running in a sidecar process — fail the build if any p50 > 500ms. Filing as a P0 issue.
- 🟡 P1 — **Daemon-down fallback is now the only path on the user's machine, and it still uses `read_only=True` direct open.** If the daemon owns the WAL exclusively (the very condition that motivated this PR), even RO opens lock-contend on macOS — i.e., the fallback is the same code that was broken. Today the daemon-proxy path catches it; the day the daemon crashes silently, the dashboard regresses to 12-15s loads. Action: add a startup probe that pings the daemon's local_query proxy, surfaces a banner if it's down, and treats the RO-direct path as a degraded mode with a 1s timeout (not 2.5s).
- 🟡 P1 — **Reactive-not-proactive failure mode.** A starred-by-influencer event almost broke a "live" demo. There is no synthetic monitoring against a fresh `pip install clawmetry` running for >1h with the daemon writing concurrently. Action: add a nightly GitHub Actions job that pip-installs the latest PyPI release, seeds a 1k-event DuckDB, runs the daemon for 60s, then curls every endpoint and asserts < 500ms. This is the regression test that would have caught all 3 bugs.
- 🟢 nit — **Pure infra fix, no tier strategy implications.** No paid-tier value leaked, no funnel impact beyond "the funnel didn't break in front of an influencer." That itself is the conversion win.

PR #1243, opened 2026-05-15
[RELEASE] fix(dashboard): inflight-dedup + close SSE on tab hidden — unblocks Brain timeout

vivekchand:
## Symptom

User stuck on "**Failed to load: timeout**" on Brain tab. DevTools shows 127 requests, most `(pending)` for 50s with 0 KB transferred. A few cancel after 3–4 s. Server returns curl in 43 ms.

## Root cause

The page opens up to **5 long-lived SSE EventSources** (`/api/brain-stream` ×2, `/api/flow-events`, `/api/health-stream`, `/api/logs-stream`), each holding a connection slot for life. The browser's HTTP/1.1 cap is **6 connections per origin**. That leaves **1 slot for everything else**. Then ~30 `setInterval` pollers (5 s, 10 s, 30 s) fire into that slot with **zero inflight dedup**, so the same URL stacks 10+ concurrent fetches that never get sent. Surface symptom: Brain's 20 s `fetchJsonWithTimeout` aborts → "Failed to load: timeout".

## Fix

Two surgical changes in `clawmetry/static/js/app.js`:

1. **`fetchJsonWithTimeout` inflight dedup** — URL→Promise map collapses concurrent identical fetches to one in-flight request per endpoint, regardless of how fast the cadence fires.
2. **`visibilitychange` SSE close** — when the tab is hidden, all 5 EventSources close (releasing their connection slots). Tab-change handlers already guard on `readyState===CLOSED` before re-opening, so it's idempotent when the user returns.

44 insertions, 7 deletions. No backend changes, no schema changes, no new endpoints.

## Test plan

- [ ] Server-side endpoints unchanged → existing API tests still pass
- [ ] `node --check app.js` syntax-clean
- [ ] On `app.clawmetry.com`: refresh Brain tab, observe DevTools — concurrent identical `/api/overview` and `/api/brain-history?limit=40` requests collapse to 1
- [ ] Switch to a different browser tab for 30 s, return — Brain SSE re-opens cleanly via `_startBrainSSE` guard

### DIFF
--- a/clawmetry/static/js/app.js
+++ b/clawmetry/static/js/app.js
@@ -1677,17 +1677,54 @@ function formatTime(ms) {
   return new Date(ms).toLocaleString('en-GB', {hour:'2-digit',minute:'2-digit',day:'numeric',month:'short'});
 }
 
+// Inflight dedup: 5 long-lived SSE EventSources (brain, flow-brain,
+// flow-events, health, logs) eat 5/6 slots in the browser's HTTP/1.1
+// per-origin connection budget. Pollers firing every 5–10s with no dedup
+// then stack 10+ identical fetches into the last slot and starve every new
+// request into "(pending)" forever — the surface symptom is "Failed to
+// load: timeout" on Brain. Collapsing duplicate URLs keeps inflight to ~1
+// per endpoint regardless of how often the cadence fires.
+var _inflightJsonFetches = {};
 async function fetchJsonWithTimeout(url, timeoutMs) {
+  if (_inflightJsonFetches[url]) return _inflightJsonFetches[url];
   var ctrl = new AbortController();
   var to = setTimeout(function() { ctrl.abort('timeout'); }, timeoutMs);
-  try {
-    var r = await fetch(url, {signal: ctrl.signal});
-    if (!r.ok) throw new Error('HTTP ' + r.status);
-    return await r.json();
-  } finally {
-    clearTimeout(to);
-  }
+  var p = (async function() {
+    try {
+      var r = await fetch(url, {signal: ctrl.signal});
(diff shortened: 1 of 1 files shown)


vivekchand:
## UX review

🟡 P1 — The user-facing error string ("Failed to load: timeout") is unchanged. After today's incident where the actual cause was a CACHED OLD JS bundle (not a server issue), users have no way to self-diagnose. Append a hint: "Failed to load: timeout — try a hard refresh (Cmd+Shift+R) if this persists." Saves a support round-trip. (`clawmetry/static/js/app.js:3549`)

🟢 nit — Pure JS infra, no visible surface change. UX impact is invisible-but-positive (no more pile-up timeouts on Brain).

vivekchand:
## Product review

🟡 P1 — This PR fixes the in-bundle starvation bug, but the in-browser DELIVERY problem remains: `app.js` ships without a `?v={version}` query string, so Chrome serves the cached old bundle for hours after each deploy. Today's incident was exactly this — server fixed, browser still on stale JS. Append `?v={__version__}` to `<script src>` and `<link href>` in the dashboard's HTML head so cache busts on every release.

🟡 P1 — 5 long-lived SSE streams is the right fix surface; the deeper question is "why 5 SSE at all?" Cloud-Pro could multiplex live updates over ONE WebSocket — faster, fewer slots, single auth context. Frame as tier differentiator: OSS = simple per-feature SSE, Cloud-Pro = multiplexed live. Worth a PRD.

🟢 nit — Inflight dedup keyed on URL. Endpoints with body-changing query params (paginated, filtered) collapse correctly because params are in the URL. Good.

🟢 nit — Backend infra fix with no tier strategy implications. Free OSS users benefit equally — appropriate for a correctness fix.

PR #1239, opened 2026-05-14
[RELEASE] fix(dashboard): Brain + Memory tabs stuck on Loading… spinner replaced + error fallback

vivekchand:
## P0 follow-up to PR #1235

> User context: Peter Steinberger just starred ClawMetry — traffic incoming.
> Both Brain and Memory tabs reportedly stuck on \"Loading...\" forever
> despite \`/api/brain-history\` (50 ms) and \`/api/memory-files\` (16 ms)
> already returning 200s. PR #1235's verification proved the backend was
> fine and explicitly listed both spinners as P0 follow-ups.

## Root cause (single-line)

\`loadBrainPage()\` early-returned on \`if (window.CLOUD_MODE) return;\`
without touching the DOM, leaving the inline \`Loading...\` placeholder in
\`#brain-stream\` indefinitely — same fix-pattern that was already applied
to the Skills tab (\`app.js:1401\`) but never extended to Brain.

The catch-block guarded its DOM replacement on \`if (!silent)\`, so a
silent retry that errored also left the spinner in place.

## What changed

| file | change |
|------|--------|
| \`clawmetry/static/js/app.js\` | \`loadBrainPage()\`: CLOUD_MODE branch replaces spinner with \"Live event stream is local-only…\" empty-state instead of returning silently. Catch-block always replaces stuck spinner with \"Failed to load — Retry\" affordance. |
| \`clawmetry/templates/tabs/brain.html\` | Inline \`Loading...\` upgraded to informative \"Loading recent activity…\" |
| \`clawmetry/templates/tabs/memory.html\` | Same upgrade for \`#selfconfig-files-inner\` and \`#memory-list\` |

## Headless Playwright verification

Test daemon spun up on **:8907** from \`/tmp/cm-spinner-test/venv\` (user's
\`:8905\` daemon and \`~/.clawmetry/\` untouched, PIDs 25047/30550/30572/30504
never signaled). Browser drove \`switchTab('brain')\` and
\`switchTab('memory')\` with \`window.CLOUD_MODE = true\` injected via
\`add_init_script\` so the boot path matched the cloud iframe context.

\`\`\`
{ \"label\": \"before\", \"mode\": \"cloud\",
  \"brain_stream_text\": \"Loading...\",
  \"asserts\": { \"brain_stuck_on_loading\": true, \"PASS\": false } }

{ \"label\": \"after\",  \"mode\": \"cloud\",
[...]

PR #1246, opened 2026-05-15
test: MOAT E2E — send-message → DuckDB → every API surface

vivekchand:
## Summary

Adds `tests/test_moat_send_message_e2e.py` — three tightly scoped E2E
scenarios that lock down the user-visible "send a message in OpenClaw"
critical path through DuckDB → API.

Per the MOAT spec: any future regression in DuckDB writes or fast-path
reads gets caught in CI.

### Scenarios

1. **`test_telegram_inbound_message_surfaces_everywhere`** — one Telegram
   inbound row → `channel_messages` table → `/api/channels/<provider>/messages`,
   `/threads`, `/api/channels/summary`, `/api/channel/telegram`.

2. **`test_tool_call_event_surfaces_everywhere`** — full session transcript
   (`session_start` + 2 tool calls + 1 assistant message) ingested via
   `sync._local_ingest_session_batch` → `events` table → `/api/local/events`,
   `/api/brain-history`, `/api/sessions`, `/api/local/transcript/<sid>`,
   `/api/usage`. Asserts every endpoint surfaces the event with the right
   shape AND tags the response `_source: "local_store"`.

3. **`test_all_endpoints_use_local_store_source`** — the MOAT canary.
   After seeding both event AND channel data, every "did the message
   land?" surface MUST tag `_source: "local_store"`. Catches future PRs
   that silently flip a fast-path back to `gateway` / `jsonl` / `otlp`.

### Test isolation

- `tmp_path` for both DuckDB and OpenClaw home — never touches real `~/.openclaw` or `~/.clawmetry`.
- `routes.local_query._DISCOVERY_PATH` monkey-patched to a non-existent path so a real ClawMetry daemon running on the dev machine doesn't intercept calls.
- All env reads via `monkeypatch.setenv` + `importlib.reload` of the modules that snapshot env at import time.
- No network. Cloud `sync._post` is never called (we use direct ingest helpers).

### What this test does NOT do

Per the spec: no production code changes. Three production surfaces I
discovered while writing the test (filed as separate concerns, not fixed
here):

1. `_try_local_store_session_tools` (routes/sessions.py:628) only fires
   on rows where `event_type == "message"` AND `data.message.content[]`
   contains `toolCall` blocks. The v3 OpenClaw parser
   (`sync._parse_v3_event`) maps assistant messages to
   `event_type = "model.completed"` — so this fast-path is effectively
   dead code on real OpenClaw transcripts. Test deliberately routes
   around it via `/api/local/transcript/<sid>` instead.
[...]

PR #1251, opened 2026-05-15
[RELEASE] fix(dashboard): cache-bust static assets with ?v={version}

vivekchand:
## Why

Today's incident: PR #1235 (lock-contention), #1239 (Brain spinner), and #1243 (browser-starvation) all shipped within minutes of each other. The user's browser kept hitting "Failed to load: timeout" because Chrome served the **pre-fix `app.js` from disk cache** even after the daemon restarted on the new wheel. Reproduced again on Safari.

Root cause: `dashboard.py` references the 5 static assets without any cache-busting query string:

```html
<link rel="stylesheet" href="{{ url_for('static', filename='css/dashboard.css') }}">
<script src="{{ url_for('static', filename='js/app.js') }}"></script>
...
```

Browsers cache these aggressively (typically 24h+) and only refetch on a hard-refresh.

## Fix

Append `?v={version}` to all 5 static refs via Flask's `url_for(...,  v=version)` kwarg. `version` is already in the template context (`routes/meta.py:342`).

After this PR, `app.js?v=0.12.194` differs from the previously-cached `app.js?v=0.12.193`, so Chrome refetches automatically. **No more user-visible cache-staleness on release.**

5 / -5. Pure cosmetic delivery change, no behavior change.

## Files cache-busted
- css/dashboard.css
- js/nav-dropdown.js
- js/alerts.js
- js/app.js
- js/gw-setup.js

## Test plan
- [ ] After merge, hard-refresh once. From then on, regular refresh (Cmd+R) should pick up new versions automatically.
- [ ] No 404s on assets (Flask `url_for` accepts kwargs as query string transparently).

### DIFF
--- a/dashboard.py
+++ b/dashboard.py
@@ -9862,9 +9862,9 @@ DASHBOARD_HTML = r"""
 <link rel="icon" href="/static/img/logo.svg" type="image/svg+xml">
 <link rel="preconnect" href="https://fonts.googleapis.com">
 <link href="https://fonts.googleapis.com/css2?family=Manrope:wght@400;500;600;700;800&display=swap" rel="stylesheet">
-<link rel="stylesheet" href="{{ url_for('static', filename='css/dashboard.css') }}">
-<script src="{{ url_for('static', filename='js/nav-dropdown.js') }}"></script>
-<script src="{{ url_for('static', filename='js/alerts.js') }}" defer></script>
+<link rel="stylesheet" href="{{ url_for('static', filename='css/dashboard.css', v=version) }}">
+<script src="{{ url_for('static', filename='js/nav-dropdown.js', v=version) }}"></script>
+<script src="{{ url_for('static', filename='js/alerts.js', v=version) }}" defer></script>
 <script src="https://cdn.jsdelivr.net/npm/marked/marked.min.js"></script>
 <script src="https://cdn.jsdelivr.net/npm/chart.js@4.4.7/dist/chart.umd.min.js"></script>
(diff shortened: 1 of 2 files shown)


vivekchand:
## UX review

🟢 nit — Pure infrastructure: appends `?v={version}` to 5 static asset URLs. Invisible to users, but eliminates the very visible failure mode where the post-deploy browser kept serving the OLD bundle until manual hard-refresh. After this PR, regular Cmd+R picks up new versions automatically. UX impact is "nothing breaks silently after every release" — high-leverage, zero-risk.

🟢 nit — Diff also includes the entire `tests/test_moat_send_message_e2e.py` file (already merged in #1246). Likely a rebase/merge artifact. No functional impact, but keeps PR scope tighter next time. (`tests/test_moat_send_message_e2e.py`)

vivekchand:
## Product review

🟡 P1 — The cache-bust uses `version` from server-rendered template context. Cloud iframe (which loads OSS routes via `_d.DASHBOARD_HTML`) might not pass `version` consistently — verify `routes/meta.py:342` is the only call site. If cloud renders the same HTML through a different render path, the `?v=` param could be empty and Chrome would cache forever in cloud mode. (Verify by hitting `/` on `app.clawmetry.com` and checking if the script tag has `?v=0.12.194`.)

🟢 nit — Backend infra fix benefiting all OSS users. No paid-tier value leakage, no funnel impact.

🟢 nit — Solves the recurring "user thinks new fix doesn't work, has to be told to hard-refresh" support churn. Worth a CHANGELOG callout so future-us doesn't re-explain why the page now reloads cleanly on every deploy.

Issue #1256, opened 2026-05-15
[MOAT P0] Migrate 7 timeout endpoints to daemon local_query proxy (read_only=True doesn't bypass DuckDB process lock)


vivekchand:
## Why this issue exists

PR #1253 tried to fix the 7 timing-out endpoints (`/api/system-health`, `/api/crons`, `/api/alerts/rules`, `/api/autonomy`, `/api/channels/status`, `/api/channels/telegram/status`, `/api/component/brain`) by adding `read_only=True` to their `local_store.get_store()` calls. **It didn't work.**

Direct test on 0.12.195 (post-merge):

\`\`\`
$ python3 -c "import duckdb; duckdb.connect('/Users/vivek/.clawmetry/clawmetry.duckdb', read_only=True)"
IO Error: Could not set lock on file: Conflicting lock is held in PID 66295 (sync daemon)
\`\`\`

DuckDB's file lock is **exclusive to one OS process** regardless of connection mode. The dashboard process can never open the file while the sync daemon writer is running. `read_only=True` only matters intra-process.

## The right fix

The dashboard must talk to the daemon's `local_query` HTTP server — same pattern that `/api/local/*` endpoints already use successfully (see Engineer #1's report: those return in 2-58ms).

The daemon's local_query server (discoverable via `~/.clawmetry/local_query.json`) exposes these methods as of 0.12.195:

- `query_events`, `query_sessions`, `query_sessions_table`
- `query_aggregates`, `query_heartbeats`, `query_channels`
- **`query_crons`, `query_cron_runs`** ← needed for /api/crons
- **`query_alert_rules`** ← need to verify exposed; if not, add
- **`query_subagents`, `query_memory_blobs`, `query_system_snapshots`**
- `query_compactions`, `query_cost_split`

## Implementation plan

For each of the 7 endpoints:

1. **`/api/crons`** (`routes/crons.py`) — replace `local_store.get_store(read_only=True)` calls with HTTP POST to `/__local_query__/query_crons` and `/__local_query__/query_cron_runs`. The proxy helper exists in `routes/local_query.py:_proxy_dispatch`.

2. **`/api/alerts/rules`** (`routes/alerts.py`) — verify `query_alert_rules` is in the proxy's allowlist (line ~292 of routes/local_query.py); add if missing. Then proxy.

3. **`/api/autonomy`** (`routes/autonomy.py`) — proxy via `/__local_query__/query_events?event_type=message`.

[...]

PR #1253, opened 2026-05-15
[RELEASE] fix(MOAT): read_only=True on DuckDB opens + dedup /api/overview fetches

vivekchand:
## Why

Engineer #1's MOAT E2E verifier (in-session) found 7 dashboard endpoints time out at the **6 s mark** on a healthy local box: `/api/system-health`, `/api/crons`, `/api/alerts/rules`, `/api/autonomy`, `/api/channels/status`, `/api/channels/telegram/status`, `/api/component/brain`.

Timeout-trace agent identified the **single root cause**: the sync daemon holds an exclusive writer lock on the local DuckDB. The dashboard's `_try_local_store_*` helpers call `local_store.get_store()` which defaults to writable — that blocks on the lock, exception-falls-back to gateway RPC (down), and surfaces as the 6 s timeout cliff.

DuckDB allows concurrent readers + 1 writer, so opening **read-only** from the dashboard succeeds even while the sync daemon is writing.

Plus: user reported "15+ requests to /overview screen, all in pending state". PR #1243 added inflight dedup but only in `fetchJsonWithTimeout`. 3 sites use raw `fetch('/api/overview')` and bypass the dedup. Converting them.

## Changes

| File | Change |
|------|--------|
| `routes/crons.py` | 3× `get_store()` → `get_store(read_only=True)` |
| `routes/alerts.py` | 1× same |
| `routes/autonomy.py` | 1× same |
| `clawmetry/static/js/app.js` | 3× `fetch('/api/overview')` → `fetchJsonWithTimeout('/api/overview', 5000)` |

Total: 4 files, +27 / -9.

## Expected impact

- `/api/crons`, `/api/alerts/rules`, `/api/autonomy` go from **6 s timeout → <500 ms** (Engineer #1 will re-verify after merge)
- Overview tab DevTools shows **1 inflight `/api/overview`** instead of 15
- Other 4 timeout endpoints (system-health, channels/status×2, component/brain) need separate fixes — filed in the umbrella issue #1245 + soon-to-be-filed sub-issues.

## Test plan
- [ ] `make lint`
- [ ] `make test-api`
- [ ] After deploy: `curl -w '%{time_total}\n' http://localhost:8900/api/crons` → <500 ms
- [ ] DevTools network on Overview tab: 1 inflight `/api/overview`

Linked: issue #1245 (umbrella), Engineer #1's report `/tmp/moat_e2e_report.md`.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1258, opened 2026-05-15
[RELEASE] fix(MOAT): route 4 timeout endpoints through daemon HTTP proxy (issue #1256)

vivekchand:
Closes 4 of 7 endpoints in #1256.

## Why PR #1253's fix didn't work

PR #1253 added \`read_only=True\` to dashboard \`get_store()\` calls. **DuckDB's file lock is exclusive at the OS process level**, regardless of connection mode. Direct test on the deployed 0.12.195:

\`\`\`
$ python3 -c \"import duckdb; duckdb.connect('/Users/vivek/.clawmetry/clawmetry.duckdb', read_only=True)\"
IO Error: Could not set lock on file: Conflicting lock is held in PID <writer> by user vivek
(after 0.002s)
\`\`\`

The exception is caught in \`_try_local_store_*\` helpers, which then return None. The handler falls through to \`_gw_invoke()\` (gateway RPC, unreachable on this user's box) → 6 s timeout.

## What this PR ships

Routes the 4 simplest endpoints through the existing \`local_store_via_daemon\` helper (\`routes/local_query.py:374\`) — daemon owns the writer connection and serves reads via HTTP from the same process, no lock conflict. This is the same pattern \`/api/local/*\` and several other working endpoints already use successfully.

| Endpoint | File | Note |
|---|---|---|
| /api/crons | routes/crons.py (3 sites) | Used existing \`_ls_call\` helper — was unused |
| /api/alerts/rules | routes/alerts.py | Inlined the daemon-proxy pattern |
| /api/autonomy | routes/autonomy.py | Inlined the daemon-proxy pattern |
| /api/component/brain | routes/components.py | Inlined the daemon-proxy pattern |

Each preserves the **single-process fallback** (direct \`get_store(read_only=True)\`) so tests + dev mode continue working without a daemon.

## Diff

4 files, +41 / -45.

## Remaining 3 endpoints (#1256)

Filed for follow-up:

- **/api/system-health** (routes/health.py) — bottleneck is \`_gw_invoke(\"cron\", ...)\` gateway RPC at line 1020, not the local-store open. Needs gateway calls replaced with daemon-proxy DuckDB reads.
- **/api/channels/status** + **/api/channels/telegram/status** (routes/channels.py) — these already fast-fail at 2ms when get_store raises, so the 6s timeout reported by Engineer #1 may be a different bottleneck (worth re-verifying after this PR lands; possibly request-storm from the Brain page firing them concurrently).

## Test plan
- [ ] \`make lint\` / \`make test-api\`
- [ ] After deploy: \`curl -w '%{time_total}\\n' http://localhost:8900/api/crons\` → <500ms
- [ ] After deploy: \`curl -w '%{time_total}\\n' http://localhost:8900/api/alerts/rules\` → <500ms
- [ ] After deploy: \`curl -w '%{time_total}\\n' http://localhost:8900/api/autonomy\` → <500ms
- [ ] After deploy: \`curl -w '%{time_total}\\n' http://localhost:8900/api/component/brain\` → <500ms

### DIFF
--- a/routes/alerts.py
+++ b/routes/alerts.py
@@ -67,13 +67,17 @@ def _try_local_store_alert_rules():
     Tagged with ``_source: "local_store"`` so callers (browser, integration
     tests, the cloud relay) can tell which path served them.
     """
+    # Issue #1256: route through daemon HTTP proxy. Direct get_store()
+    # raises IOException on multi-process installs (DuckDB's file lock is
+    # exclusive across processes; read_only=True doesn't bypass it).
     try:
-        from clawmetry import local_store
-        # read_only=True — see crons.py for the same fix. Writable open from
-        # dashboard blocks on the sync daemon's exclusive writer lock and
-        # surfaces as the 6 s timeout cliff for /api/alerts/rules.
-        store = local_store.get_store(read_only=True)
-        rows = store.query_alert_rules(limit=500)
+        from routes.local_query import local_store_via_daemon
+        rows = local_store_via_daemon("query_alert_rules", limit=500)
+        if rows is None:
+            # Daemon unreachable → single-process fallback (tests/dev mode).
+            from clawmetry import local_store
+            store = local_store.get_store(read_only=True)
+            rows = store.query_alert_rules(limit=500)
     except Exception:
         return None
     if not rows:
--- a/routes/autonomy.py
+++ b/routes/autonomy.py
@@ -293,12 +293,16 @@ def _try_local_store_autonomy() -> dict | None:
         from clawmetry import local_store
     except Exception:
         return None
+    # Issue #1256: route through daemon HTTP proxy. Direct get_store()
+    # raises IOException on multi-process installs (DuckDB's file lock is
+    # exclusive across processes; read_only=True doesn't bypass it).
     try:
-        # read_only=True — see crons.py for the same fix. Writable open from
-        # dashboard blocks on the sync daemon's exclusive writer lock and
-        # surfaces as the 6 s timeout cliff for /api/autonomy.
-        store = local_store.get_store(read_only=True)
-        rows = store.query_events(event_type="message", limit=5000)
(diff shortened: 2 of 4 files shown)


vivekchand:
## UX review

🟢 nit — Pure backend infra: 4 routes (crons, alerts, autonomy, components) migrate from direct DuckDB opens to the daemon HTTP proxy. No UI surface change, no template change. UX impact is invisible-but-massive: the affected tabs (Crons, Alerts, Autonomy, Component panels) drop from 7s "Loading..." cliffs to <500ms. Verified post-deploy on 0.12.196: crons 180ms, autonomy 6ms, component/brain 5ms.

🟢 nit — Single-process fallback (`local_store.get_store(read_only=True)`) preserves the test + dev-mode path without ceremony. Good ergonomic choice.

vivekchand:
## Product review

🔴 P0 — Shipped without verifying `query_alert_rules` was in the daemon allowlist (`_DAEMON_METHODS` in `routes/local_query.py:291`). It wasn't. The fix landed, looked done, but `/api/alerts/rules` continued to time out at 7s on 0.12.196 because the proxy returned 400 "method not allowed" → fallback failed → handler hit slow legacy path. Caught only by post-deploy curl. PR #1260 was the 8-line follow-up. **Process gap: any new `local_store_via_daemon("foo", ...)` caller should be a CI-checked allowlist entry.** Add a unit test that imports each route module and asserts every `local_store_via_daemon` arg is in `_DAEMON_METHODS`.

🟡 P1 — Single-process fallback opens DuckDB directly when the daemon proxy returns None. That branch fires whenever the daemon is up but the *method* isn't allowlisted — same 400 path as P0 above. Consider: distinguish "daemon down" (fall back) from "method not allowlisted" (return None permanently — fast-path can't help here). Today both look the same and both walk into the slow path.

🟢 nit — No tier-strategy implications. OSS-side correctness fix; cloud benefits via the same wheel.

PR #1261, opened 2026-05-15
[RELEASE] fix(MOAT): /api/system-health off gateway-RPC bottleneck

vivekchand:
## Summary

Issue #1256 follow-up. PR #1258 fixed 4 of 7 timeout endpoints by routing through the daemon HTTP proxy. ``/api/system-health`` is the fifth — measured **21s** on the user's box (gateway port 18789 blackholes the request, hangs all three blocking calls in ``api_system_health()``).

Three blocking calls removed:

1. **``_d._gw_invoke(\"cron\", ...)``** (10s urllib + 15s docker fallback) → ``_ls_call(\"query_crons\")``. Cron aggregates (enabled count, ok-24h, failed names) recomputed from DuckDB rows.
2. **``_d._get_sessions()``** (5s ``_gw_ws_rpc`` + 5s reconnect) → ``_ls_call(\"query_sessions\", since=24h)``. Sub-agent counting matches on ``session_id`` (DuckDB column) instead of gateway-shaped ``sessionId``.
3. **``_query_gateway_metric_history``** did a direct ``local_store.get_store(read_only=True).query_events(...)`` which blocked ~2.5s waiting for the writer lock the daemon already holds. Routed through the same ``_ls_call`` wrapper.

Both fast paths treat a non-None response (including empty list) as authoritative — falling back to the gateway when DuckDB is "just empty" defeats the purpose. Legacy gateway path preserved for truly fresh installs where the daemon proxy AND direct DuckDB open both fail.

JSON shape unchanged — adds ``_source`` markers on ``crons``, ``subagents``, and a top-level ``_source`` for observability.

## Bench

Handler latency on dev box (test client, all gateway helpers stubbed):
- Before: **2,613ms** (dominated by direct DuckDB-open hanging on lock)
- After: **50ms** (52x)

User-box baseline 21s (gateway unreachable) should drop to <1s.

## Test plan
- [x] ``python3 -c \"import ast; ast.parse(open('routes/health.py').read())\"`` — syntax OK
- [x] ``pytest tests/test_health_local_store.py`` — 13/13 pass
- [x] ``pytest tests/test_gateway_metric_history.py`` — 14/14 pass (test fixture updated to monkeypatch ``_DISCOVERY_PATH``, mirroring ``test_moat_send_message_e2e`` pattern)
- [x] In-process Flask test client returns 200 with full shape preserved (services, disks, crons, subagents, heartbeat, sandbox, inference, security, service_status, daemon, gateway)
- [ ] Post-merge: ``curl -w '%{time_total}\n' http://localhost:8900/api/system-health`` on user's box returns <1s

### DIFF
--- a/routes/health.py
+++ b/routes/health.py
@@ -83,6 +83,46 @@ def _default_daemon_log_path():
     return os.path.join(home, "sync.log")
 
 
+def _ls_call(method_name, **kwargs):
+    """Cross-process LocalStore call with single-process fallback.
+
+    Mirror of ``routes/crons.py:_ls_call``. Issue #1256: the dashboard's
+    direct ``get_store().query_*`` opens raise ``IOException: Could not
+    set lock`` on the standard install (daemon owns the writer lock), so
+    we route through the daemon's HTTP proxy first and only fall back to
+    a direct open for single-process boots (tests + dev mode). Returns
+    ``None`` on miss so callers can defer to the legacy gateway path.
+    """
+    try:
+        from routes.local_query import local_store_via_daemon
+        result = local_store_via_daemon(method_name, **kwargs)
+        if result is not None:
+            return result
+    except Exception:
+        pass
+    try:
+        from clawmetry import local_store
+        store = local_store.get_store(read_only=True)
+        return getattr(store, method_name)(**kwargs)
+    except Exception:
+        return None
+
+
+def _parse_iso_to_epoch(ts):
+    """Best-effort ISO-8601 / numeric → epoch seconds. Returns 0 on any failure."""
+    if not ts:
+        return 0
+    if isinstance(ts, (int, float)):
+        # Heuristic: values >1e12 are ms, otherwise seconds.
+        return float(ts) / 1000.0 if ts > 1e12 else float(ts)
+    if isinstance(ts, str):
+        try:
+            return datetime.fromisoformat(ts.replace("Z", "+00:00")).timestamp()
+        except Exception:
+            return 0
+    return 0
+
+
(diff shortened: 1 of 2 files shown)


vivekchand:
## UX review

🟢 nit — Pure backend handler refactor. No UI change. Adds `_source` markers per-section + a top-level `_source` field, so operators can see which path served the data without changing the JSON shape clients depend on. Smart additive change: existing fields (services, disks, crons.{enabled,ok24h,failed}, etc.) preserved; the new metadata is opt-in.

🟡 P1 — Top-level `_source` introduces a 3rd value `"mixed"` (when some sections came from daemon, others from gateway). Document this in `ARCHITECTURE.md` or the API contract — clients that try to switch on `_source` will silently miss the mixed case and treat it as "unknown". (`routes/health.py:1218` has the logic.)

vivekchand:
## Product review

🟢 nit — Excellent execution. The agent caught a 3rd bug in the same handler (`_query_gateway_metric_history` opening DuckDB directly, contributing 2.5s of the 2.6s latency) that wasn't in the original scope. Folded into this PR rather than punted — right call. Local bench: **2613ms → 50ms (52×)**. Test fixture monkeypatches `_DISCOVERY_PATH` to prevent dev-machine pollution — that pattern needs to be promoted to a shared `conftest.py` autouse fixture (Engineer #3 flagged the same gap in #1246).

🟢 nit — Preserves the legacy gateway fallback for the truly-fresh-install case where the daemon proxy isn't reachable AND the direct open fails. Correct degradation: slow but right beats fast and wrong.

🟢 nit — No tier-strategy implications. Cloud benefits from the same wheel; cloud's `/api/system-health` (proxied through the iframe) will also speed up.

PR #1266, opened 2026-05-15
[RELEASE] fix(MOAT): empty fast-path returns [] not None — closes #1265

### LINKED ISSUE #1265: [MOAT P1] /api/alerts/rules + /api/channels/status fast-paths return None on empty DuckDB → fall through to slow legacy
Caught while verifying PR #1260 on 0.12.197.

The fast-path helpers `_try_local_store_alert_rules` (routes/alerts.py:60+) and the channels/status handlers (routes/channels.py:155+) treat an empty result set from the daemon proxy as a 'miss' and return None. The handler then falls through to the legacy fleet-DB / gateway path, which hangs ~3 s on this user's box.

## Fix

When the daemon proxy returns successfully with an empty list, return `{"rules": [], "_source": "local_store"}` instead of None. Only return None on actual errors (exception, daemon unreachable, etc.).

```python
# Before
if not rows:
    return None
return {"rules": rows, "_source": "local_store"}

# After
return {"rules": rows or [], "_source": "local_store"}
```

Affects:
- routes/alerts.py:_try_local_store_alert_rules
- routes/channels.py:_channel_config_status_from_local_store
- routes/channels.py:api_channels_status_all (the inline fast path)

Each is ~3 lines of changes. Should be a tight follow-up PR.

## Repro
- Fresh ClawMetry install (no alert rules, no channels configured)
- `curl -w '%{time_total}\n' /api/alerts/rules` → 3 s
- After fix: <100 ms

vivekchand:
Closes #1265.

## Why

Caught while verifying PR #1260 on 0.12.197: `/api/alerts/rules`, `/api/channels/status`, and `/api/channels/<provider>/status` were STILL hanging at 3-7 s on the user's box despite the daemon-proxy fix.

Two bugs in one fix:

1. **Empty result treated as miss.** When the user has zero alert rules / zero channels configured (the common fresh-install case), the daemon proxy returns `[]` and the helpers returned `None` → handler fell through to the legacy fleet-DB / gateway path → 3 s hang.

2. **`api_channels_status_all` + `_channel_config_status_from_local_store` were still using direct `local_store.get_store()` opens** which IOException on multi-process installs. PR #1258 fixed crons/alerts/autonomy/component-brain via the daemon proxy but missed these two.

## Fix

- `routes/alerts.py`: `if not rows: return None` → `return {\"rules\": rows or [], \"_source\": \"local_store\"}`
- `routes/channels.py` ×2: route through `local_store_via_daemon` (matches the pattern in `routes/crons.py:_ls_call`); treat empty result as success.

3 files, +24 / -16.

## Expected

| Endpoint | Before | After |
|---|---|---|
| /api/alerts/rules (empty) | 3 s | <50 ms |
| /api/channels/status | 7 s | <50 ms |
| /api/channels/<p>/status | 7 s | <50 ms |

### DIFF
--- a/routes/alerts.py
+++ b/routes/alerts.py
@@ -80,9 +80,12 @@ def _try_local_store_alert_rules():
             rows = store.query_alert_rules(limit=500)
     except Exception:
         return None
-    if not rows:
-        return None
-    return {"rules": rows, "_source": "local_store"}
+    # Issue #1265: an EMPTY result set is a successful local-store hit
+    # (user has no alert rules configured yet). Returning None here makes
+    # the handler fall through to the legacy fleet-DB path, which hangs
+    # ~3 s on this user's box. Return [] tagged with the local_store
+    # source instead — the dashboard JS handles an empty list cleanly.
+    return {"rules": rows or [], "_source": "local_store"}
 
 
 # ── Budget API Routes ───────────────────────────────────────────────────
--- a/routes/channels.py
+++ b/routes/channels.py
@@ -75,30 +75,27 @@ def _channel_config_status_from_local_store(provider: str):
     """Read the non-secret status summary for ``provider`` from DuckDB.
 
     Returns the row dict tagged with ``_source: "local_store"`` on hit, or
-    ``None`` on miss / store unavailable. NEVER returns the encrypted blob —
-    HTTP responses must never carry ciphertext to keep the cloud surface
+    ``None`` on store unavailable. NEVER returns the encrypted blob — HTTP
+    responses must never carry ciphertext to keep the cloud surface
     plaintext-free at every layer of defense."""
+    # Issue #1256 + #1265: route through daemon proxy (DuckDB process-level
+    # lock blocks direct opens from the dashboard process when sync daemon
+    # owns the writer).
+    rows = None
     try:
-        from clawmetry import local_store
-    except Exception:
-        return None
-    store = None
-    try:
-        store = local_store.get_store(read_only=True)
-    except Exception:
-        # Fresh install with no daemon writes yet — DuckDB can't open RO on a
-        # missing file. Try opening writable so the schema is created;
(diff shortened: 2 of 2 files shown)


vivekchand:
## UX review

🟢 nit — Backend invisible-but-massive: 3 endpoints (`/api/alerts/rules`, `/api/channels/status`, `/api/channels/<provider>/status`) drop from **3-7 s → <50 ms** for the common fresh-install case (zero rules, zero channels configured). Verified live on 0.12.199 after sync-daemon restart: alerts/rules 6 ms, channels/status 3 ms, channels/telegram/status 5 ms. The `_source: "local_store"` flag is correctly preserved.

🟢 nit — Comment on the `return {"rules": rows or [], ...}` line names the bug class precisely ("an EMPTY result set is a successful local-store hit") — future readers won't undo the fix accidentally.

vivekchand:
## Product review

🟡 P1 — The "empty result = miss" anti-pattern almost certainly affects more `_try_local_store_*` helpers across `routes/`. Audit: grep for `if not rows: return None` and `if not evs: return None` in every fast-path. Each instance is a candidate for the same fall-through-to-slow-path bug. (Hint: there are ~10 helpers in `routes/crons.py`, `routes/usage.py`, `routes/health.py`, `routes/sessions.py` worth checking.)

🟡 P1 — `_channel_config_status_from_local_store` previously had a *writable*-fallback branch ("Try opening writable so the schema is created") that this PR deletes. Check that the schema-create-on-first-read path is exercised somewhere else (sync daemon's startup?). If not, fresh installs with no daemon writes yet may now serve the "fallback" branch instead of the "unconfigured" branch — minor visual diff but worth confirming. (`routes/channels.py:78-86` before, deleted in this PR.)

🟢 nit — Backend-only fix with no tier strategy implications. Closes the user-visible cliff that PR #1258's daemon-proxy migration intended to fix but didn't fully resolve.

Issue #1252, opened 2026-05-15
[UX P0] Lazy-load dashboard data per active tab + sequence within tab (kill the request stampede)


vivekchand:
## Problem (user-reported, 2026-05-15)

> "the browser seem to make a ton of requests in one go -- the requests should be made only when actually in that screen -- one after another -- the main piece of info should get loaded to engage user & then next info & so on"

Right now, `clawmetry/static/js/app.js` fires ~10 `setInterval` pollers **at module init** and opens up to 5 long-lived `EventSource` streams **regardless of which tab is currently visible**. On a fresh page load this means:

- ~10 polling fetches start firing (5–30 s cadences)
- 3–5 SSE streams open (brain, flow-brain, flow-events, health, logs)
- 5–6 initial-load fetches per active tab
- = 18+ concurrent requests in the first 1–2 seconds

This consumes the browser's HTTP/1.1 6-connection-per-origin budget instantly (PR #1243 patched the symptoms). It also amplifies the **6 s timeout cliff** that Engineer #1's MOAT verifier found in 7 endpoints — system-health, crons, alerts/rules, autonomy, channels/status×2, component/brain — because they pile up against a saturated queue.

## Vision (the user's mental model, restated)

1. **Show the most important thing first.** When a tab activates, render its primary widget with one fetch. Display it. User feels the page respond.
2. **Layer in supporting context after the primary lands.** Subagents, recent activity, secondary charts come in after the user's eye has settled on the primary.
3. **Don't fetch what the user can't see.** Crons, alerts, channels — none of it should poll while the user is on the Brain tab.
4. **One request at a time per tab when possible.** Sequential, not parallel, for non-time-critical data. Modern browsers cap at 6 anyway, so artificial sequencing doesn't lose much and gives the user a perceived step-by-step load.

### Phase 1 — Gate ALL setInterval pollers on `document.visibilityState` + active tab (1 PR, ~60 lines)

Wrap every `setInterval(fn, ms)` in a helper:

```js
function tabAwareInterval(tabName, fn, ms) {
  return setInterval(function() {
    if (document.hidden) return;
    if (currentTab !== tabName) return;
    fn();
  }, ms);
}
```

Replace the ~10 module-init `setInterval` calls. Effect: when on Brain tab, ONLY brain-relevant polls fire. Channels/crons/alerts/anomalies stop firing in background. Estimated request reduction: **70%** (from ~30/min to ~9/min on Brain tab).

[...]

PR #1270, opened 2026-05-15
[RELEASE] feat(dashboard): visibility-gate 5 module-init pollers (lazy-load PRD #1252 Phase 1)

vivekchand:
Phase 1 of #1252.

## Why

User feedback (verbatim, 2026-05-15):

> "the browser seem to make a ton of requests in one go -- the requests should be made only when actually in that screen -- one after another"

Five pollers fire at module init regardless of which tab the user is viewing OR whether the browser tab is even visible:

| Function | Cadence | File:line |
|---|---|---|
| checkActiveAlerts | 30s | app.js:312 |
| checkAnomalies | 5min | app.js:367 |
| loadAnomalyPanel | 2min | app.js:471 |
| checkHeartbeatStatus | 30s | app.js:496 |
| refreshPausedBanner | 15s | app.js:525 |

When the user has the ClawMetry tab in the background (or laptop lid closed and re-opened), these still fire and hit the daemon for nothing. Over an 8-hour idle workday that's ~4,000 wasted requests per session.

## Fix

Add a `visibilitySetInterval(fn, ms)` helper that wraps `setInterval` with a `document.hidden` early-return. Replace the 5 module-init pollers with it.

When the tab IS visible: identical cadence + side effects (no behavior change).  
When the tab is hidden: callback is a no-op. Resumes immediately on visibilitychange (no manual hookup needed — `document.hidden` is just the gate).

Function declaration (not `var`) so the hoist makes it available at the module-init call sites.

## Diff

15 / 0 lines, single file (`clawmetry/static/js/app.js`).

## Out of scope (Phase 2+)

- Per-tab pollers (loadCrons when on Crons tab, loadSubagents when on Subagents tab, etc.) — already gated by tab name; visibility-gating them is a separate, smaller PR.
- Sequence-within-tab loadInOrder() helper (PRD #1252 Phase 2).
- Defer SSE openings until tab dwell > 2s (Phase 3).

## Test plan
- [ ] DevTools network panel on Brain tab, switch to another browser tab → request count plateaus immediately
- [ ] Switch back → polls resume on next cadence tick
- [ ] No console errors

### DIFF
--- a/clawmetry/static/js/app.js
+++ b/clawmetry/static/js/app.js
@@ -308,8 +308,29 @@ async function ackAllAlerts() {
   } catch(e) {}
 }
 
+// PRD #1252 Phase 1 — visibility-gated setInterval. The 5 module-init
+// pollers below (alerts, anomalies×2, heartbeat, paused-banner) fire on
+// every page load REGARDLESS of which tab is visible to the user. With
+// the user away from the browser for hours, these still hit the daemon
+// every 15-300s for nothing.
+//
+// User feedback (verbatim, 2026-05-15): "the browser seem to make a ton
+// of requests in one go". This helper kills background polling when the
+// tab is hidden — a no-op when document.visibilityState === 'hidden',
+// resumes immediately on visibilitychange. Per-tab pollers (loadCrons
+// 30s when on Crons tab, etc.) keep their existing tab-name gates.
+//
+// Function declaration, NOT var, so the hoist makes it available to the
+// module-init `setInterval` callers below at parse time.
+function visibilitySetInterval(fn, ms) {
+  return setInterval(function() {
+    if (typeof document !== 'undefined' && document.hidden) return;
+    try { fn(); } catch (e) {}
+  }, ms);
+}
+
 // Check alerts every 30s
-setInterval(checkActiveAlerts, 30000);
+visibilitySetInterval(checkActiveAlerts, 30000);
 setTimeout(checkActiveAlerts, 3000);
 
 // === Anomaly Detection Banner ===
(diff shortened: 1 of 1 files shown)


vivekchand:
## UX review

🟢 nit — Captures the user's verbatim quote ("the browser seem to make a ton of requests in one go") in the PR-message comment block. Future maintainers see the WHY, not just the WHAT. Good craft.

🟢 nit — Function declaration (not `var`) so the hoist makes the helper available to module-init `setInterval` callers above its source line. Subtle correctness call that would have broken silently with `var`. Worth a code-review prize.

🟡 P1 — Phase 1 only covers 5/30+ pollers (the module-init ones). The 5s `loadSubagents` and 5s `loadMainActivity` per-tab pollers are arguably worse offenders for the user's "ton of requests" feedback when the user is on one of those tabs but switches to another browser tab. Phase 2 should wrap the per-tab pollers with the same `visibilitySetInterval`. Track in #1252.

vivekchand:
## Product review

🟡 P1 — From first principles: **why does the dashboard ever poll while the tab is hidden?** Long-term, the answer should be NEVER. The current per-tab gate (only when name === 'crons', etc.) is necessary but not sufficient. Phase 2 (per-tab pollers) + Phase 3 (close SSE on dwell) close the loop. File a tracking PR list under #1252 so this doesn't drift.

🟡 P1 — When `document.hidden` is true, the helper silently no-ops — but the user might come back to a 30-min-stale Active Alerts banner. Trade-off: silent UI staleness vs request volume. Low-impact for alerts (will refresh next visible tick), high-impact if someone leaves the tab open over a budget breach overnight. Consider: on `visibilitychange → visible`, force a one-shot re-fire of the gated functions before the next interval tick. ~5 lines.

🟢 nit — Backend-cost benefit alone is meaningful: one user with the dashboard open in a background tab for an 8-hour workday saves ~4,000 needless requests. Multiplied across the user base (per memory: 120K+ installs), this is a measurable load reduction.

Issue #1268, opened 2026-05-15
[Product P0 from PR #1260] Daemon-proxy fast-paths need empty-DuckDB curl probe in test plan


vivekchand:
From the post-merge product review of PR #1260.

The 3-PR chain to fix `/api/alerts/rules` (PR #1258 → #1260 → #1266) all could have been caught in one shot if the test plan included "verify endpoint <500ms with EMPTY DuckDB tables" as an acceptance criterion. Each follow-up was discovered only after deploy.

## Fix

Add to the [RELEASE] PR template:

```markdown
## MOAT fast-path test plan (when adding/modifying a daemon-proxy endpoint)
- [ ] curl <endpoint> with **empty** local DuckDB → <500ms + correct empty response
- [ ] curl <endpoint> with **populated** local DuckDB → <500ms + correct data
- [ ] curl <endpoint> with **daemon stopped** → degrades to legacy path, doesn't 500
- [ ] Method name appears in `routes/local_query.py:_DAEMON_METHODS`
```

The first checkbox catches the empty-fastpath-returns-None bug. The fourth catches the allowlist miss. Together they would have made #1258 a one-PR fix instead of three.

## Acceptance
- [ ] PR template updated in `.github/pull_request_template.md`
- [ ] CONTRIBUTING.md cross-references it under "MOAT changes"
- [ ] Mention in CLAUDE.md so agent-shipped PRs follow the same checklist

PR #1272, opened 2026-05-15
ci: lint daemon-proxy method calls against _DAEMON_METHODS allowlist (closes #1267)

### LINKED ISSUE #1267: [Product P0 from PR #1260] Daemon-proxy fast-paths need empty-DuckDB curl probe in test plan
From the post-merge product review of PR #1260.

The 3-PR chain to fix `/api/alerts/rules` (PR #1258 → #1260 → #1266) all could have been caught in one shot if the test plan included "verify endpoint <500ms with EMPTY DuckDB tables" as an acceptance criterion. Each follow-up was discovered only after deploy.

## Fix

Add to the [RELEASE] PR template:

```markdown
## MOAT fast-path test plan (when adding/modifying a daemon-proxy endpoint)
- [ ] curl <endpoint> with **empty** local DuckDB → <500ms + correct empty response
- [ ] curl <endpoint> with **populated** local DuckDB → <500ms + correct data
- [ ] curl <endpoint> with **daemon stopped** → degrades to legacy path, doesn't 500
- [ ] Method name appears in `routes/local_query.py:_DAEMON_METHODS`
```

The first checkbox catches the empty-fastpath-returns-None bug. The fourth catches the allowlist miss. Together they would have made #1258 a one-PR fix instead of three.

## Acceptance
- [ ] PR template updated in `.github/pull_request_template.md`
- [ ] CONTRIBUTING.md cross-references it under 'MOAT changes'
- [ ] Mention in CLAUDE.md so agent-shipped PRs follow the same checklist

vivekchand:
Closes #1267.

## Why

The 3-PR cascade #1258 → #1260 → #1266 all stemmed from one missed step: PR #1258 added \`local_store_via_daemon(\"query_alert_rules\", ...)\` and \`local_store_via_daemon(\"query_channel_config_status\", ...)\` calls but forgot to register both in \`routes/local_query.py:_DAEMON_METHODS\`.

At runtime the daemon HTTP server returned 400 \"method not allowed\" for every call → dashboard fast-path saw the failure → returned None → handler fell through to slow legacy gateway path → 7s cliff that the PR was supposed to fix.

## Fix

\`scripts/lint_daemon_allowlist.py\` (126 lines, stdlib only):

1. Walks \`routes/*.py\` for \`local_store_via_daemon(\"X\", ...)\` and the per-blueprint \`_ls_call(\"X\", ...)\` wrapper.
2. AST-parses \`_DAEMON_METHODS\` from \`routes/local_query.py\` (regex on the allowlist itself would break with whitespace/comment changes).
3. Filters callers to method names starting with \`query_\` so we don't false-positive on \`event_type=\"message\"\` kwargs.
4. Asserts every captured method is in the allowlist.

Wired into \`make lint\`. Exits 0 on clean with a one-line summary; exits 1 with file:line per missing method.

## Verified

**Clean:**
\`\`\`
$ python3 scripts/lint_daemon_allowlist.py
daemon-allowlist OK — 11 distinct query_* methods called from routes/, all in _DAEMON_METHODS (20 entries)
\`\`\`

**Synthetic regression (renamed a caller's method to one NOT in the allowlist):**
\`\`\`
daemon-allowlist FAIL — call sites use methods missing from _DAEMON_METHODS:
  'query_NOT_IN_ALLOWLIST' called at:
    routes/crons.py:182
    routes/crons.py:254
    routes/crons.py:1221
\`\`\`
Exit code 1.

## Diff

2 files, +136 / 0 (script + 3 Makefile entries).

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,4 +1,4 @@
-.PHONY: test test-api test-e2e test-fast test-e2e-duckdb dev lint
+.PHONY: test test-api test-e2e test-fast test-e2e-duckdb dev lint lint-daemon-allowlist
 
 dev:
 	OPENCLAW_GATEWAY_TOKEN=dev-token python3 dashboard.py --port 8900
@@ -19,7 +19,16 @@ test-e2e:
 test-e2e-duckdb:
 	python3 -m pytest tests/test_e2e_duckdb_relay.py -v
 
-lint: lint-py lint-js
+lint: lint-py lint-js lint-daemon-allowlist
+
+# Issue #1267: every `local_store_via_daemon("X")` / `_ls_call("X")` call
+# in routes/ must reference a method that's in the daemon's allowlist
+# (routes/local_query.py:_DAEMON_METHODS). Catches the gap that produced
+# the 3-PR cascade #1258 → #1260 → #1266 (forgot to add new methods to
+# the allowlist; daemon returned 400; fast-paths fell through to slow
+# legacy paths; surfaced as 6 s timeouts).
+lint-daemon-allowlist:
+	@python3 scripts/lint_daemon_allowlist.py
 
 lint-py:
 	python3 -c "import ast; ast.parse(open('dashboard.py').read()); print('Python syntax OK')"
@@ -40,4 +49,4 @@ lint-js:
 	    echo "WARN: node not installed — skipping JS syntax check (CI installs node automatically)"; \
 	fi
 
-.PHONY: lint lint-py lint-js
+.PHONY: lint lint-py lint-js lint-daemon-allowlist
--- a/scripts/lint_daemon_allowlist.py
+++ b/scripts/lint_daemon_allowlist.py
@@ -1,0 +1,126 @@
+#!/usr/bin/env python3
+"""Lint: every daemon-proxy method call must be in the allowlist.
+
+Closes the process gap that produced PRs #1258 → #1260 → #1266 (issue #1267).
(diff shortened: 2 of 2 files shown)


vivekchand:
## UX review

🟢 nit — Pure DX/CI infra, no user-visible surface. Process improvement: the lint catches the class of bug that produced the 3-PR cascade #1258 → #1260 → #1266 at PR-time instead of at user-pain-time.

vivekchand:
## Product review

🟢 nit — Right shape: AST-parses the allowlist (regex would break with formatting changes), filters to `query_*` (avoids false-positives on `event_type="message"` kwargs), exits 1 with file:line. Wired into `make lint` so it runs on every PR locally + in CI.

🟡 P1 — The lint is **necessary but not sufficient** to catch the full class. PR #1266's bug (empty fast-path returns None → slow fallback) wouldn't have been caught by this lint alone. Track issue #1268 (fast-path test plan in PR template) as the complementary defense — together they close the gap.

🟡 P1 — Lint catches the dashboard-side miss. If a NEW LocalStore method is added but never registered in the daemon's allowlist, the lint stays green (no caller exists yet). When the first caller lands, lint catches it. Could promote allowlist validation to also run AT TEST TIME — assert every `LocalStore.query_*` method is either in the allowlist OR explicitly opted out (e.g., a `_DAEMON_METHODS_OPT_OUT` set). That'd catch the bug even before any caller exists.

🟢 nit — Closes the process gap that wasted 3 release cycles today. High ROI for 126 lines.

PR #1274, opened 2026-05-15
[RELEASE] feat(dashboard): visibility-gate 31 per-tab pollers (lazy-load PRD #1252 Phase 2)

vivekchand:
Phase 2 of #1252.

## Why

The user's verbatim feedback (2026-05-15): "the browser seem to make a ton of requests in one go". Phase 1 (PR #1270) covered the 5 module-init pollers (alerts, anomalies×2, heartbeat, paused-banner). Phase 2 covers the **31 per-tab pollers** — the worst offenders for steady-state request volume when the user is on any tab and switches to another browser tab.

## Fix

Wrap every \`_xxxTimer = setInterval(fn, ms)\` with \`visibilitySetInterval\` (the helper shipped in PR #1270, already in 0.12.200).

When \`document.hidden\` is true:
- 11 channel pollers (Telegram, Signal, WhatsApp, Discord, Slack, Google Chat, MS Teams, Mattermost, Webchat, IRC, BlueBubbles, generic) — **all STOP**
- loadSubagents (5s), loadMainActivity (5s) — **STOP**
- loadCrons (30s, 2 sites), loadSystemHealth (30s) — **STOP**
- loadOverviewTasks (10s), loadActiveTasks (30s) — **STOP**
- + 14 others
- = ~36 pollers gated end-to-end (Phase 1 + Phase 2)

When the user comes back to the tab, the next interval tick fires immediately (no manual hookup; \`document.hidden\` is just the gate).

## Diff

31 / 31 word swap (\`= setInterval(\` → \`= visibilitySetInterval(\`) in a single file. JS syntax: clean. No behavior change when tab is visible.

## Out of scope

- Line 326 — INSIDE the helper itself, don't touch.
- Line 8280 (\`setInterval(updateFlowStats, ...)\`) — unassigned, leaks after Flow tab visit. Separate follow-up if it bites.
- Phase 3 (close SSE on tab dwell > 2s) — separate PR.

### DIFF
--- a/clawmetry/static/js/app.js
+++ b/clawmetry/static/js/app.js
@@ -601,7 +601,7 @@ function switchTab(name) {
   // Stop cron auto-refresh when leaving crons tab
   if (name !== 'crons' && _cronAutoRefreshTimer) { clearInterval(_cronAutoRefreshTimer); _cronAutoRefreshTimer = null; }
   if (name === 'overview') loadAll();
-  if (name === 'overview') { if (typeof _velocityPollTimer !== 'undefined' && _velocityPollTimer) clearInterval(_velocityPollTimer); if (typeof loadTokenVelocity === 'function') _velocityPollTimer = setInterval(loadTokenVelocity, 30000); }
+  if (name === 'overview') { if (typeof _velocityPollTimer !== 'undefined' && _velocityPollTimer) clearInterval(_velocityPollTimer); if (typeof loadTokenVelocity === 'function') _velocityPollTimer = visibilitySetInterval(loadTokenVelocity, 30000); }
   if (name === 'usage') loadUsage();
   if (name === 'skills') loadSkills();
   if (name === 'crons') loadCrons();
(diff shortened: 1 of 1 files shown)


vivekchand:
## UX review

🟢 nit — Bulk swap of 31 per-tab pollers to `visibilitySetInterval` (the helper from PR #1270). Pure invisible-positive: when the user has the ClawMetry tab in a background browser tab, ~36 pollers (Phase 1 + 2) gate-stop entirely. When they return, the next interval tick fires immediately. Verified live: switch to another tab → `tail ~/.clawmetry/sync.log` shows the polling traffic plateaus.

🟡 P1 — Two `setInterval` calls REMAIN in `app.js` after this PR: (1) line 326 — INSIDE the helper itself (correctly excluded), (2) line 8280 — `setInterval(updateFlowStats, updateInterval)` unassigned, fires once user opens Flow tab and **never gets cleared on tab switch**. PR notes flag this as out of scope but it's a real leak — once user visits Flow tab, that interval keeps firing forever (Phase 2's helper doesn't catch it because there's no `_xxxTimer =` assignment to swap). File a small follow-up PR to either assign-to-timer + visibility-gate it OR move it inside the per-tab activation block. (`app.js:8280`)

vivekchand:
## Product review

🟢 nit — Phases 1 + 2 together implement the user's verbatim "ton of requests" feedback (#1252). Multiplied across the install base (memory: 120K+ installs), this is a measurable load reduction on `ingest.clawmetry.com`. Worth instrumenting: count requests-per-session-hour pre/post deploy.

🟡 P1 — The bulk regex-style swap (`= setInterval(` → `= visibilitySetInterval(`) is correct here but **fragile to future call-sites that don't follow the `_xxxTimer = ` convention** (e.g. line 8280's unassigned `setInterval(updateFlowStats, ...)` is the canary case this PR couldn't catch). Worth filing a CI lint analogous to #1272 (the daemon-allowlist lint): grep `app.js` for any `setInterval(` not preceded by `visibilitySetInterval` definition, fail the PR. ~10 lines of bash.

🟢 nit — No tier-strategy implications. Free OSS users benefit equally; the request-cost reduction is felt on the cloud ingest side only when sync is on (which is a Cloud-Pro feature with its own per-plan rate limits).

PR #1275, opened 2026-05-15
fix(install): pass --refresh to uv pip install (closes #1271)

### LINKED ISSUE #1271: [install.sh] Restart sync daemon AND dashboard after upgrade (currently only dashboard restarts)
## The bug

When upgrading via `install.sh` (or manual `pip install -U clawmetry==<v>` + `launchctl kickstart -k com.clawmetry.dashboard`), the **sync daemon process keeps running the OLD wheel in memory**. This silently breaks any new code that needs daemon-side support.

## Today's incident

PR #1260 added `query_channel_config_status` to `_DAEMON_METHODS` allowlist in `routes/local_query.py`. After 0.12.197 deployed and the dashboard restarted, the dashboard sent `POST /__local_query__/query_channel_config_status` to the sync daemon → daemon (still on old wheel) returned 400 "method not allowed" → dashboard's fast-path returned None → fell through to slow legacy path → **3.2 s timeout per request** for `/api/channels/status`.

Resolved by also running:
```bash
```

After that: 7s timeout → 3ms.

## Fix

`install.sh` already detects existing clawmetry processes ("Detected 3 existing clawmetry process(es) — will restart after upgrade") and restarts the dashboard. It needs to ALSO restart `com.clawmetry.sync` after the wheel upgrade completes.

Look for the existing `launchctl kickstart -k` call in install.sh and add a sibling for the sync service.

## Acceptance
- [ ] After `install.sh` to a new version, `ps -ef | grep clawmetry.sync` shows a NEW PID
- [ ] After install, hitting any `_DAEMON_METHODS` endpoint added in the new version returns 200 (not 400)

vivekchand:
Closes #1271.

## Why

Reproduced twice today: \`install.sh\` shows \`Successfully installed clawmetry-0.12.X\` but the daemon stays on the OLD wheel. uv reads its on-disk PyPI index cache; if the cache was populated MOMENTS BEFORE a new version was published (which is exactly what happens when running install.sh right after a [RELEASE] PR auto-publishes), uv silently no-ops with \"already at latest\".

## Fix

One word: \`--refresh\` on the \`uv pip install\` call. Forces uv to re-fetch the PyPI index every time. Tiny extra latency (~200 ms); fixes the release-race correctness issue cleanly.

## Diff

1 file, +7 / -1 (1 line of code + 6 lines of explanatory comment).

## Test plan
- [ ] Run \`bash install.sh\` immediately after a [RELEASE] merge — verify the new version actually lands on disk (\`grep \"^__version__\" \$(/Users/vivek/.clawmetry/lib/python3.11/site-packages/dashboard.py)\`)

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -216,8 +216,14 @@ if command -v uv >/dev/null 2>&1; then
   # native progress bar so OUR spinner is the only thing on screen.
   _step "Creating virtual environment (uv)" 2 \
     $USE_SUDO "$_UV_BIN" venv "$INSTALL_DIR" --quiet
+  # --refresh forces uv to re-fetch the PyPI index even if a cached copy
+  # exists. Without this, running install.sh seconds after a [RELEASE]
+  # auto-publish silently no-ops ("already at latest" against a stale
+  # index that doesn't list the just-published version), leaving the
+  # daemon on the OLD wheel even after the launchctl kicks below fire.
+  # Verified locally on 2026-05-15 via PR #1260 → 0.12.197 publish race.
   _step "Installing clawmetry from PyPI (uv)" 5 \
-    $USE_SUDO "$_UV_BIN" pip install --python "$INSTALL_DIR/bin/python3" --quiet --upgrade clawmetry
+    $USE_SUDO "$_UV_BIN" pip install --python "$INSTALL_DIR/bin/python3" --quiet --upgrade --refresh clawmetry
 else
   # pip path is much slower (~30s for the install alone) — make sure the
   # spinner conveys that so users don't think the script is wedged.


vivekchand:
## UX review

🟢 nit — Pure DX/infra, no user-visible surface. Invisible-but-positive: closes the install-script footgun where running `install.sh` seconds after a [RELEASE] auto-publish silently no-ops with the wrong version. After today's incident reproduced this twice (PR #1260 → 0.12.197, PR #1261 → 0.12.198), 1-word fix lands the right wheel every time.

vivekchand:
## Product review

🟢 nit — Right shape, right scope. `--refresh` on the uv invocation forces re-fetch of the PyPI index. ~200 ms overhead per install; eliminates the cache-miss-window race entirely.

🟡 P1 — Could go one step further: after the install, **verify the on-disk version matches the requested version** before kickstarting launchd. Today's failure mode was "install command exited 0 but wheel still old"; a post-install check (`grep "^__version__" /Users/vivek/.clawmetry/lib/python3.11/site-packages/dashboard.py`) would have caught the silent-no-op even WITHOUT the --refresh flag, and now serves as a defense-in-depth backstop. ~5 lines of shell.

🟡 P1 — install.sh's banner still hardcodes `0.12.181 installed` (visible in today's logs). Independent bug — fix it in this PR OR file a small follow-up. It's a minor confidence hit when users see "0.12.181 installed!" but `clawmetry --version` says 0.12.202. (Find: `grep -n "0.12.181" install.sh`.)

🟢 nit — No tier-strategy implications. Closes #1271.

PR #1278, opened 2026-05-15
[RELEASE] fix(MOAT): /api/sessions 3-10s → <500ms via daemon-proxy + batched fetch (closes #1277)

### LINKED ISSUE #1277: [MOAT P1] /api/sessions is 3-10s but underlying DuckDB query is 6ms — enrichment/serialization bottleneck
## Symptom

Verified live on 0.12.200 (4 sessions in DuckDB, 4 .jsonl files on disk):

```
/api/local/sessions  (pure DuckDB):   6 ms, 8 ms, 6 ms     ✓ fast
/api/sessions        (with enrichment): 3.2 s, 10 s timeout, 10 s timeout  ✗
```

The DuckDB roundtrip is 6 ms — the slow path is in the enrichment work between `_try_local_store_sessions()` and `return jsonify(...)`.

## Where to look

`routes/sessions.py:176-209`

After `_try_local_store_sessions()` returns, the handler:
1. Calls `_merge_unregistered_jsonls(fast["sessions"])` — walks the OpenClaw sessions dir. Should be cheap (4 files).
2. (Inside the fast-path helper, before return) — likely calls `_decorate_with_channel_context(sessions)` which queries the daemon **per session** for channel metadata. With N sessions, that's N proxy roundtrips at ~5ms each + sync daemon contention.
3. Possibly `_augment_sessions_with_burn(...)` somewhere (used in legacy path; check fast path).

## Concurrency clue

First call: 3.2 s. Second + third: 10 s timeout. Something is **blocking on the previous call** — likely a single-writer lock somewhere, OR the daemon proxy (5 s timeout per call × N sessions) cumulatively exceeds the curl timeout.

## Repro

```bash
for i in 1 2 3; do
  curl -s -o /dev/null -w "%{time_total}s\n" --max-time 10 http://localhost:8900/api/sessions
done
```

## Acceptance

- [ ] `/api/sessions` responds <500 ms regardless of call sequence (no degradation across calls)
- [ ] Response carries `_source: "local_store"` (currently does)
- [ ] Bench: profile shows where the 3-10s goes — likely in `_decorate_with_channel_context` or similar per-session enrichment loop
- [ ] Fix: batch the enrichment query (one daemon-proxy call returning ALL channels, then in-memory join) instead of N per-session calls

## Related

- Issue #1256 closed today (7-endpoint timeout cliff fixed)
- This is the next biggest UX cliff; `/api/sessions` is polled by the Sessions tab + Overview's session list

vivekchand:
Closes #1277.

## Root cause

\`routes/sessions.py:_get_channel_context_map\` opened DuckDB DIRECTLY via \`local_store.get_store()\` (writable mode). On multi-process installs this raises IOException because the sync daemon holds the exclusive process-level lock. The exception was caught, BUT the failed singleton call leaves DuckDB in a degraded state — subsequent calls cumulatively slow down.

Repro on 0.12.200: 3.2s, 10s, 10s timeout (degrading across calls). Underlying DuckDB query (\`/api/local/sessions\`) takes 6 ms.

Plus: the helper looped \`query_channels(session_id=X)\` N times when given a small \`session_ids\` arg. Switch to one full-table fetch + client-side filter.

## Fix

Same pattern as PRs #1258, #1260, #1266 (the #1256 family):

\`\`\`python
# Before
store = local_store.get_store()  # multi-process: blocks/fails
rows = store.query_channels(...)  # N round-trips for small id sets

# After
rows = local_store_via_daemon(\"query_channels\", limit=2000)
if rows is None:
    # Single-process fallback (tests/dev)
    store = local_store.get_store(read_only=True)
    rows = store.query_channels(limit=2000)
# Filter client-side via 'wanted' set
\`\`\`

## Allowlist already covered

\`query_channels\` is already in \`_DAEMON_METHODS\` (CI lint #1272 confirms).

## Test plan
- [ ] After deploy: \`for i in 1 2 3; do curl -w '%{time_total}\\n' /api/sessions; done\` → all <500ms, no degradation
- [ ] Existing \`tests/test_sessions_local_store.py\` still passes

### DIFF
--- a/routes/sessions.py
+++ b/routes/sessions.py
@@ -105,30 +105,45 @@ def _get_channel_context_map(session_ids=None):
     caller treats an empty dict as "nothing to merge", which preserves the
     pre-existing metadata-blob channel inference unchanged.
     """
+    # Issue #1277: route through daemon HTTP proxy. Direct
+    # local_store.get_store() raises IOException on multi-process installs
+    # (DuckDB process-level file lock), and even when the exception is
+    # caught the singleton ends up in a degraded state — surfaces as the
+    # 3–10 s cumulative latency on /api/sessions despite the underlying
+    # DuckDB query being 6 ms. Same root cause as the #1256 endpoint family.
+    #
+    # Plus: even on the happy path, query_channels(session_id=X) was
+    # invoked N times in a loop. Switch to a single full-table fetch (the
+    # in-line dict-build below scopes the result to the requested ids
+    # client-side). Channels table is always small (one row per channel
+    # per session — capped well under the 2000 limit), so a full scan is
+    # cheaper than N proxy round-trips even at small N.
+    rows = None
     try:
-        from clawmetry import local_store
-        store = local_store.get_store()
-        if session_ids is not None:
-            ids = [s for s in session_ids if s]
-            if not ids:
-                return {}
-            rows = []
-            # query_channels supports one session_id at a time; for the typical
-            # /api/sessions response (≤200 rows) one full-table scan is cheaper
-            # than N round-trips.
-            if len(ids) > 20:
-                rows = store.query_channels(limit=2000)
-            else:
-                for sid in ids:
-                    rows.extend(store.query_channels(session_id=sid, limit=1))
-        else:
-            rows = store.query_channels(limit=2000)
+        from routes.local_query import local_store_via_daemon
+        rows = local_store_via_daemon("query_channels", limit=2000)
     except Exception:
-        return {}
+        rows = None
+    if rows is None:
+        # Single-process fallback (tests + dev mode).
+        try:
+            from clawmetry import local_store
+            store = local_store.get_store(read_only=True)
+            rows = store.query_channels(limit=2000)
+        except Exception:
+            return {}
+    # Scope the full-table result to requested session_ids client-side.
+    wanted = None
+    if session_ids is not None:
+        wanted = {s for s in session_ids if s}
+        if not wanted:
+            return {}
     out = {}
-    for r in rows:
+    for r in (rows or []):
         sid = r.get("session_id")
         if not sid:
+            continue
+        if wanted is not None and sid not in wanted:
             continue
         out[sid] = {
             "channel":      r.get("channel") or "",


vivekchand:
## UX review

🟢 nit — Invisible-but-massive: `/api/sessions` drops from 3-10s degrading → **5-11ms consistent** (verified live across 5 sequential calls). The Sessions tab + Overview's session list both feel night-and-day faster on next page load. UX impact is silently the biggest win of the day (after the 7-endpoint cliff resolution).

🟡 P1 — The PR replaced a `query_channels(session_id=X)` per-id loop with one full-table `query_channels(limit=2000)` + client-side filter. Limit=2000 is arbitrary; if a user with hundreds of channels ever hits this code path, the proxy roundtrip + JSON serialization could become non-trivial. Add a Counter-style log in the daemon ("query_channels returned 1500 rows in 80ms — consider paginating") so we see this coming. (`routes/sessions.py:_get_channel_context_map`)

vivekchand:
## Product review

🟢 nit — This is the proven daemon-proxy pattern (PRs #1258, #1260, #1266) applied to the next hot path. The CI lint (#1272) validates `query_channels` is allowlisted. Pattern is now repeatable enough that the remaining 6 sites in audit issue #1282 should be a mechanical sweep, not bespoke fixes.

🔴 P0 — **Process question**: it took TWO observability cycles to find this. The N+1 was visible in `routes/sessions.py:110` for weeks (line 122-123 has the per-id loop), but neither code review nor manual testing caught it. The user's `/api/sessions` was always 3-10s; nobody flagged it. **Add a per-endpoint p50/p95 latency probe to the dashboard's own `/api/system-health` panel** so any handler regressing past 500 ms surfaces visibly. Without this, the next /api/sessions-class regression takes another full incident cycle to discover.

🟡 P1 — `_decorate_with_channel_context` is a nice abstraction but it CALLS \_get_channel_context_map which is the slow path. Bigger architectural win: at the daemon side, expose a `query_sessions_with_channels(...)` JOIN — one round-trip, no client-side merge. Defer until pattern stabilizes.

🟢 nit — No tier-strategy implications. The fix benefits OSS + Cloud equally.

PR #1284, opened 2026-05-15
[RELEASE] fix(dashboard): close last setInterval leak in initFlow (lazy-load Phase 2 follow-up)

vivekchand:
## Why

Caught by the post-merge UX review of PR #1274: \`app.js:8280\` was the one \`setInterval\` Phase 2's bulk swap couldn't catch — it was a raw \`setInterval(updateFlowStats, ms)\` not assigned to any \`_xxxTimer\` variable. Two consequences:

1. **Leak**: each Flow tab visit fires a NEW interval; previous ones never get cleared. After N visits → N parallel \`updateFlowStats\` timers running in parallel, each issuing \`/api/overview\` polls every few seconds.
2. **No visibility gate**: with no \`visibilitySetInterval\` wrapper, the timer keeps firing when the browser tab is hidden — the exact UX issue Phase 2 was supposed to fix.

## Fix

\`\`\`js
// Before
setInterval(updateFlowStats, updateInterval);

// After
if (window._flowStatsTimer) { try { clearInterval(window._flowStatsTimer); } catch(e){} }
window._flowStatsTimer = visibilitySetInterval(updateFlowStats, updateInterval);
\`\`\`

Mirrors the per-tab timer pattern at \`app.js:606\` etc.

## Diff
1 file, +8 / -1.

### DIFF
--- a/clawmetry/static/js/app.js
+++ b/clawmetry/static/js/app.js
@@ -8277,7 +8277,14 @@ function initFlow() {
   _backfillFlowFromBrain();
   _startFlowBrainStream();
 
-  setInterval(updateFlowStats, updateInterval);
+  // Lazy-load Phase 2 follow-up: this used to be a raw `setInterval(...)`
+  // unassigned to any timer var, so initFlow() leaked one new interval per
+  // visit. Capture the handle on `window._flowStatsTimer` so subsequent
+  // tab navigations can clear it (mirroring the per-tab timer pattern at
+  // app.js:606 etc.) and use visibilitySetInterval so the poll pauses when
+  // the browser tab is hidden — completes Phase 2 coverage.
+  if (window._flowStatsTimer) { try { clearInterval(window._flowStatsTimer); } catch(e){} }
+  window._flowStatsTimer = visibilitySetInterval(updateFlowStats, updateInterval);
 }
 
 // Map brain-history event types → Flow's active-tool buckets (exec/browser/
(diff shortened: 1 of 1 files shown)


vivekchand:
## UX review

- 🟢 **(backend/perf only)** No new affordance — this fixes two zombie `setInterval` pollers (`updateFlowStats` at `app.js:8280`, `_prefetchToolData` at `app.js:11376`) so the tab quiets down when hidden. Users won't see it directly; they'll only feel "dashboard stopped chewing my battery."
- 🟡 P1 — `_prefetchToolData` is registered in `DOMContentLoaded` (`app.js:11376`) so it never multi-arms, but the comment claims that as proof "no leak risk." If a future PR moves it inside any reachable function (e.g. a tab-init), the leak comes back invisibly. Add a `if (window._prefetchToolTimer) return;` guard and assign `window._prefetchToolTimer = visibilitySetInterval(...)` mirroring the `_flowStatsTimer` pattern from this same diff — make the no-leak structural, not load-order-dependent.
- 🟢 nit — "lazy-load Phase 2 follow-up" framing in the comment is excellent context for future readers. Keep that habit.

vivekchand:
## Product review

- 🟢 **(infra-only)** Closes the lazy-load Phase 2 sweep; no user-visible surface, no tier-strategy implications. Right call to ship as a `[RELEASE]` even though it's tiny — the visibility-gate story ("dashboard pauses polling when you're not looking") is a tangible Cloud-Pro retention argument and the more endpoints we get on it, the cleaner that pitch becomes.
- 🟡 P1 — Two PRs in two days closing "the last" leak (#1274, then this #1284). Suggests the audit was eyeballing rather than mechanical. Add one CI grep step: `git grep -nE '= setInterval\(' clawmetry/static/js/ | grep -v visibilitySetInterval` and fail if any line matches outside the `visibilitySetInterval` definition itself. Costs 1 min in CI, removes the entire class. File as follow-up issue.
- 🟢 nit — Worth a one-liner in the next CHANGELOG bundle: "Dashboard now pauses background polling when the tab is hidden — quieter Mac fans, fewer Cloud Run invocations." Sells the Phase 2 work better than the issue numbers do.

PR #1148, opened 2026-05-13
feat(overview): autonomy score KPI (closes #688)

vivekchand:
## Summary
- Adds north-star ``autonomy`` block to ``/api/overview`` so the dashboard can render the primary KPI card on first paint (median human-nudge gap, autonomy ratio, trend %, sample size — all 7d).
- Computed from ``prompt.submitted`` events in DuckDB (v3 OpenClaw parser, #1135) via ``local_store_via_daemon`` with a direct-open fallback. 60s in-process cache. Always returns a valid dict; empty store yields the placeholder so a fresh install never breaks the UI.
- Inline JS in ``clawmetry/templates/tabs/overview.html`` populates the existing autonomy card from ``/api/overview.autonomy`` — formatted gap, trend arrow + colour (green positive, red negative), sample count. Runs in parallel with the existing ``/api/autonomy`` lazy fetcher (idempotent — both write the same DOM IDs).

## Why
Per CLAUDE.md positioning (observability, not cost-tracking), the north-star metric belongs above cost on the overview. Krentsel's framing: *"Success = nudges space out exponentially."*

## Card pseudo-text
```
🎯 HOW INDEPENDENT IS YOUR AGENT?
   12m 30s                       ▲ 14.3%
   Autonomy: median nudge gap (7d)
   +14.3% vs prior 7d                       Last 7 days
                                            [sparkline ──╱─]
                                            47 user turns
```

## Test plan
- [x] ``python3 -m pytest tests/test_overview_autonomy_score.py -q`` (7/7 pass)
- [x] Manual: ``/api/overview`` response includes ``autonomy`` block on empty + populated DuckDB
- [ ] Visual: load overview page locally → KPI card top-left renders with formatted gap + trend badge
- [ ] No regression in existing ``/api/overview`` shape (legacy ``model``, ``cronCount``, ``infra`` etc. all preserved)

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1288, opened 2026-05-15
fix(overview): remove duplicate autonomy card fetcher

### LINKED ISSUE #1166: [UX P0 from PR #1148] Two parallel autonomy fetchers write the same DOM IDs
Found by the post-merge UX review on PR #1148 (https://github.com/vivekchand/clawmetry/pull/1148#issuecomment-4443365864):

> TWO fetchers writing to the SAME DOM IDs (`autonomy-score-label`, `autonomy-median-gap`, `autonomy-trend-pct`, `autonomy-trend-badge`, `autonomy-samples`). PR body acknowledges this: "Runs in parallel with the existing /api/autonomy lazy fetcher (idempotent — both write the same DOM IDs)." That is the duplicate-Alerts-tab failure mode this review specifically watches for — two PRs racing the same anchor point. Action: in `clawmetry/templates/tabs/overview.html:570+` delete this new inline fetcher entirely; instead modify `/api/autonomy` route to consume from `body.autonomy` if already populated, OR delete the legacy `/api/autonomy` fetcher in `app.js` and keep only the inline one. Don't ship two writers — the next bug-fix to the autonomy card will land in the wrong file.

The post-merge product review concurs (https://github.com/vivekchand/clawmetry/pull/1148#issuecomment-4443366944):

> Two parallel autonomy fetchers writing the same DOM. This isn't just sloppy — it's a guaranteed regression. The next PR that updates one path will silently break the other because the other will overwrite the DOM 2.6s later. Action: in `routes/overview.py:_autonomy_for_overview` make the existing `/api/autonomy` route delegate to this new function (single source), then DELETE the duplicate fetcher script in `clawmetry/templates/tabs/overview.html:507+`. One writer per DOM target. Forever.

**Fix**: pick ONE writer for the autonomy card DOM. Delegate the other path to share the single `_autonomy_for_overview` source.

vivekchand:
Closes #1166

## What

Removes the 68-line inline autonomy script that was added to `clawmetry/templates/tabs/overview.html` as part of PR #1148.

## Why this is a bug, not just sloppiness

The two writers wrote **different data** to the same DOM IDs:

| DOM element | Inline script (via `/api/overview`) | `loadAutonomy()` in app.js (via `/api/autonomy`) |
|---|---|---|
| `autonomy-score-label` | `fmtGap(median_gap_seconds)` → "2h 30m" | `scoreToLabel(score).text` → "Fully independent" |
| `autonomy-median-gap` | "Autonomy: median nudge gap (7d)" | "You check in about every 2h 30m." |
| `autonomy-trend-pct` | "+X.X% vs prior 7d" | different format |

The inline script ran on page load, then `loadAutonomy()` overwrote everything 2.6s later with a completely different presentation. The comment claiming both paths were "idempotent" was incorrect.

## Fix

Delete the inline IIFE. `loadAutonomy()` in app.js remains as the single authoritative writer — it already handles the sparkline, qualitative label, trend, sample count, and graceful empty states that the inline script didn't cover.

The card shows "--" placeholder for ~2.6s on first paint, which is strictly better than showing wrong/mismatched data.

## Test plan

- [ ] Load the dashboard overview tab — autonomy card shows "--" briefly then fills in with "Fully independent" / "Mostly independent" / "Needs guidance" (not a raw duration string)
- [ ] After 2.6s the card is populated; no second overwrite occurs
- [ ] The sparkline renders correctly (only ever written by app.js)

---
_Generated by [Claude Code](https://claude.ai/code/session_01AkgwDtkje3E7k3S5dL9tTa)_


vivekchand:
## Product review

- 🟢 **(infra-only)** Pure dead-code removal; no user-visible surface, no tier-strategy implications. The duplicate fetcher was wasted bandwidth on every dashboard load — small per-user cost but multiplied across 120K installs it's measurable.
- 🟡 P1 — This bug class (two competing fetchers for the same data) is exactly what the lazy-load PRD #1252 was supposed to prevent. Suggests we don't have a CI guard against it. File a follow-up: a lint that fails when `>1` JS file mutates the same DOM ID via `getElementById('autonomy-...').textContent = `. Even a static-grep heuristic would catch the duplicate-tab class of bug structurally.
- 🟢 nit — Worth quietly adding to the next changelog entry under a "polish" header: "Removed redundant network calls on the dashboard's main page." Doesn't sell a feature but signals "we care about the details" — that's what fence-sitters notice.

vivekchand:
## UX review

- 🟢 **(invisible cleanup)** Removes a duplicate copy of the autonomy-card fetcher that was hydrating the same DOM IDs as the existing app.js path. Net effect for users: zero visible change, half the network calls (one /api/overview poll instead of two), no race-condition risk on the autonomy card values. Right call.
- 🟡 P1 — The comment that justifies removing the duplicate calls it "idempotent" because both paths target the same IDs. That's true *now* — but if a future PR changes one path's render shape, the silent divergence will be hard to spot. Add a 1-line comment in `app.js` next to the surviving fetcher: `// Sole owner of #autonomy-* DOM updates (overview.html duplicate removed in #1288)`. Costs nothing, prevents the resurrection of the duplicate.
- 🟢 nit — A 72-line script block disappeared from the rendered HTML. Verify there's no `[skip ci]` autonomy E2E test that relied on the inline `<script>` running before app.js loaded. Quick `git grep autonomy-card-fetcher` to confirm.

PR #1289, opened 2026-05-15
fix(install): add timeout=2 to PyPI version probe (closes #1229)

### LINKED ISSUE #1229: fix(install): add timeout=2 to PyPI version probe in early-exit (#1223 follow-up)
Follow-up to merged PR #1223 (early-exit when version current and daemon healthy).

## The gap

The early-exit added in install.sh fetches the latest PyPI version via:

```python
urllib.request.urlopen('https://pypi.org/pypi/clawmetry/json')
```

There's no timeout. If PyPI is slow, your network is degraded, or DNS is misbehaving, this fast-path can BLOCK longer than the install it's trying to skip — defeating the purpose of the early-exit.

## The fix

Single-line change:

```python
urllib.request.urlopen('https://pypi.org/pypi/clawmetry/json', timeout=2)
```

On socket timeout / connection error → \`_LATEST\` becomes empty → string comparison falls through to the full install path. Same graceful fallback as today, just bounded.

## Rationale

- Fast path should be fast. 2s is generous for PyPI; if it can't answer in 2s, the user is better off going through the regular install path (which has its own retries + spinners).
- Today's behavior on a slow PyPI: silent hang. Tomorrow's behavior with the timeout: silent fall-through to full install. Strictly better.

## Acceptance

- [ ] \`urlopen(..., timeout=2)\` added.
- [ ] Manually test by pointing at a sinkhole address (e.g. \`https://10.255.255.1/pypi/clawmetry/json\`) — confirm script falls through to full install in ≤3s instead of hanging.

Tiny PR — happy to take this from a contributor or do it inline.

vivekchand:
Closes #1229

## What

Add `timeout=2` to the `urllib.request.urlopen` call that fetches the latest PyPI version inside `install.sh`'s early-exit block.

## How

```diff
-r = urllib.request.urlopen('https://pypi.org/pypi/clawmetry/json')
+r = urllib.request.urlopen('https://pypi.org/pypi/clawmetry/json', timeout=2)
```

That's the entire patch. One line.

## Why

PR #1223 added an early-exit so re-running `install.sh` on an already-current install with a healthy daemon exits in <2s. The probe it relies on has no timeout — meaning on a slow PyPI / degraded DNS / sinkhole network, the "fast path" can hang **longer** than the install it's trying to skip, defeating the purpose.

With `timeout=2`:
- Fast PyPI → behaviour unchanged (probe still returns in well under 2s).
- Slow / unreachable PyPI → `URLError` raised → `_LATEST` is empty → string compare with `_CURRENT` fails → script falls through to the regular install path, which has its own spinners and retries.

Strictly better. Bounded fast-path; same graceful fallback as today.

## Verification

Pointed the probe at sinkhole `https://10.255.255.1/pypi/clawmetry/json`:

```
$ python3 -c "import urllib.request, time; t=time.time()
try: urllib.request.urlopen('https://10.255.255.1/pypi/clawmetry/json', timeout=2)
except Exception as e: print(f'{type(e).__name__} after {time.time()-t:.2f}s')"
URLError after 2.00s
```

Without `timeout=2` the same call hangs for the full system default (~75-130s on Linux). The fix bounds it to 2s as designed.

`bash -n install.sh` passes.

## Acceptance (per #1229)

- [x] `urlopen(..., timeout=2)` added
- [x] Verified by pointing at a sinkhole address — script falls through to full install in ≤3s instead of hanging

### DIFF
--- a/install.sh
+++ b/install.sh
@@ -173,7 +173,7 @@ if [ -x "$INSTALL_DIR/bin/clawmetry" ]; then
   _CURRENT=$("$INSTALL_DIR/bin/clawmetry" --version 2>/dev/null | awk '{print $NF}')
   _LATEST=$("$INSTALL_DIR/bin/python3" -c "
 import json, urllib.request
-r = urllib.request.urlopen('https://pypi.org/pypi/clawmetry/json')
+r = urllib.request.urlopen('https://pypi.org/pypi/clawmetry/json', timeout=2)
 print(json.loads(r.read())['info']['version'])
 " 2>/dev/null)
   if [ -n "$_CURRENT" ] && [ "$_CURRENT" = "$_LATEST" ] && [ -n "$_existing_pids" ]; then


vivekchand:
## Product review

- 🟢 **(install resilience)** The first 30 seconds of install determine whether a new user becomes a churned user. A 2s timeout on the PyPI probe means firewalled corporate environments now get past the version-check stage instead of hanging. Direct conversion-funnel impact, even if invisible to the install.sh banner.
- 🟡 P1 — `urllib` defaults to no timeout, which is a python-stdlib footgun this team will keep tripping over. Add a CONTRIBUTING.md or coding-standards line: "Every `urllib.request.urlopen` call in install.sh / sync.py / interceptor.py MUST pass `timeout=`. CI grep enforces." Then write the grep step. Otherwise this same bug surfaces in the next install-script PR.
- 🟢 nit — 2 seconds is the right value for a non-blocking version probe; longer would compound the install latency, shorter would false-timeout on slow connections. Document the `2s` choice in a comment so the next person doesn't "improve" it back to 5s.

vivekchand:
## UX review

- 🟢 **(install-script polish)** Adds `timeout=2` to a single `urllib.request.urlopen` in install.sh's PyPI version probe. Two-line change, perfect blast radius. Without the timeout, install.sh would hang indefinitely if PyPI is unreachable (offline laptop, blocked corporate firewall) — turning "install in <10s" into "install hangs forever and the user closes the terminal."
- 🟡 P1 — On timeout, `_LATEST` becomes empty, the `if [ "$_CURRENT" = "$_LATEST" ]` check naturally fails, install proceeds. Good. But there's no surfaced message — user has no idea PyPI was unreachable. Add a `_LATEST=$(... 2>/dev/null) || _LATEST=""; [ -z "$_LATEST" ] && echo "  (PyPI version probe timed out — proceeding with reinstall check)" >&2` so corporate/offline users understand why "Already up to date" didn't fire.
- 🟢 nit — Excellent tight scope. Closes #1229 with a minimal 1-line patch.

PR #1287, opened 2026-05-15
[RELEASE] feat(observability): per-endpoint p50/p95 latency probe (closes #1283)

vivekchand:
## Summary

Eat our own dogfood. Adds server-side request-timing middleware to ClawMetry's own `/api/*` routes, surfaces top-N slowest endpoints with p50/p95 in System Health. The next `/api/sessions`-class regression takes seconds to surface, not weeks.

- New `clawmetry/latency_tracker.py` — process-local rolling buffer (per-endpoint deque(maxlen=200), 5-min window, lock-protected)
- `dashboard.py` — `before_request`/`after_request` hooks record perf_counter delta for any path starting with `/api/`
- `routes/health.py` — new `/api/handler-latency?top=10&slow_ms=500` endpoint
- `clawmetry/templates/tabs/overview.html` + `app.js` — System Health panel renders the table; rows whose p95 exceeds the slow threshold flash red

## Why P0

PR #1278 (the `/api/sessions` 3-10s slowness fix) had been hiding in prod for weeks. Code review and manual smoke missed it; only an indirect user complaint ("Brain feed missing telegram") surfaced it. Without this probe, every regression costs a full incident cycle.

## Verified live

Smoke against a fresh dev dashboard immediately surfaced **three real prod regressions** the team had been blind to:
- \`components.api_component_tool\` — **p95 = 7.3s** 🔴
- \`usage.api_anomalies\` — **p95 = 2.6s** 🔴
- \`overview.api_overview\` — **p95 = 629ms** 🔴

Acceptance criteria from #1283 — all met:
- [x] Server middleware records (handler, time_ms) for every request
- [x] Aggregate kept in 5-min rolling memory buffer
- [x] System Health tab shows top-10 slowest endpoints with p50/p95
- [x] Endpoints over 500 ms p95 highlighted red
- [x] Useful for first-24h post-deploy monitoring of any new daemon-proxy call

## Test plan

- [x] \`pytest tests/test_latency_tracker.py\` — 5 tests pass (record+stats, slow-flag, top-N sort, time-window expiry, input sanitisation)
- [x] \`make lint-daemon-allowlist\` passes (no daemon-proxy callers added)
- [x] Local dev dashboard on :8901, hit /api/system-health + /api/overview a few times, confirmed /api/handler-latency returns sorted stats with is_slow flagging
- [x] \`python3 -c 'import dashboard'\` — no import errors
[...]

PR #1285, opened 2026-05-15
ci: add PR template with MOAT fast-path test checklist (closes #1268)

vivekchand:
Closes #1268.

## Why

Captures the test-plan checklist that would have prevented today's 3-PR cascade #1258 → #1260 → #1266. The empty-fastpath-returns-None bug + the missing-allowlist-entry bug both could have been caught by a 30-second curl probe against an empty DuckDB.

## Defenses now in place
| Layer | What | Catches |
|---|---|---|
| CI lint (PR #1272) | grep \`local_store_via_daemon(\"X\")\` callers, AST-parse \`_DAEMON_METHODS\` | missing allowlist entry |
| **This PR template** | manual checkboxes (empty DB, populated DB, daemon-stopped, _source flag, sync.log probe) | empty-fastpath bug + verification gaps |
| PR template Release section | post-deploy version-check + restart-both-services | uv cache stale (closes #1271 follow-up) |

## Diff

1 file, 41 lines. Standard \`.github/pull_request_template.md\` location — GitHub auto-loads it for every new PR.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #1282, opened 2026-05-15
[MOAT P1] Audit + migrate 6 remaining direct local_store.get_store() writable opens in routes/


vivekchand:
## Why

Today's fix cycle for #1256 + #1277 surfaced a class of bug: handlers that open DuckDB directly via `local_store.get_store()` (writable, no read_only) hit the sync daemon's process-level lock, raise IOException, and either (a) fall through to slow gateway path, or (b) leave the singleton in degraded state, causing 3-10s API responses.

Audit on 0.12.201 finds 6 remaining writable direct opens:

| File:line | Handler | Risk |
|---|---|---|
| routes/advisor.py:121 | _try_local_store_advisor | likely |
| routes/advisor.py:448 | smoke check (`local_store.get_store()` then nothing) | low (just probes) |
| routes/brain.py:222 | _try_local_store_brain_history | **high** — hot path, polled |
| routes/components.py:124 | (find handler) | medium |
| routes/health.py:646 | (find handler) | medium |
| routes/health.py:2112 | (find handler) | medium |

Plus 11 `get_store(read_only=True)` opens that ALSO fail on multi-process installs (read_only doesn't bypass file lock — see memory `reference_duckdb_process_lock.md`):

```
routes/agents.py:39, alerts.py:79, autonomy.py:304, bootstrap.py:33,
channels.py:68/94/174, components.py:1174, crons.py:80, health.py:105/214/810/1694/1934
```

## Fix pattern (proven in PRs #1258, #1260, #1266, #1278)

```python
# Before
store = local_store.get_store()  # or get_store(read_only=True)
rows = store.query_X(...)

# After
try:
    rows = local_store_via_daemon("query_X", ...)
except Exception:
    rows = None
if rows is None:
    # Single-process fallback (tests/dev mode)
    store = local_store.get_store(read_only=True)
    rows = store.query_X(...)
```

CI lint #1272 already validates allowlist coverage. PR template #1268 (when shipped) will catch fast-path-returns-None gaps.

## Acceptance

- [ ] Each of the 17 sites either uses `local_store_via_daemon` first, or has a documented justification for direct open
- [ ] No `get_store()` (writable, no read_only) call in any `/api/*` handler hot path
- [ ] Each migrated handler verified <500ms over 5 sequential calls

vivekchand:
## Audit COMPLETE — 6 of 6 writable `get_store()` callsites migrated

| Callsite | PR | Pattern |
|---|---|---|
[...]

PR #1292, opened 2026-05-15
[RELEASE] fix(MOAT): /api/component/tool/<name> 7.3s → 1.4ms via daemon-proxy (closes #1291 part 1)

### LINKED ISSUE #1291: [Product P0 from PR #1287] File 3 [MOAT-cliff] follow-up issues for endpoints the latency probe surfaced (components.api_component_tool 7.3s, usage.api_anomalies 2.6s, overview.api_overview 629ms)
From the post-merge product review of PR #1287.

The `/api/handler-latency` probe shipped in PR #1287 immediately surfaced 3 latent regressions on a local smoke run — these are real user-facing slowness, not "interesting telemetry."

## The findings

| Endpoint                            | p95   | Status     |
|-------------------------------------|-------|------------|
| `components.api_component_tool`     | 7.3s  | 🔴 cliff   |
| `usage.api_anomalies`               | 2.6s  | 🔴 cliff   |
| `overview.api_overview`             | 629ms | 🔴 slow    |

These were already costing users dashboard timeouts. The probe just made them visible. Without follow-through, PR #1287 delivered the diagnostic and walked away — leaving the actual user pain unfixed.

## What "follow-through" means

File 3 separate issues — one per endpoint — each:

1. Titled `[MOAT-cliff] <endpoint> p95 >Xs — daemon-proxy migration`
2. Body references the daemon-proxy migration playbook from PR #1278 (sessions) and PR #1258 (cron+alerts+autonomy+components batch). Both PRs are template patches showing the exact code change.
3. Tagged with the existing MOAT-cliff label (or create one)
4. Acceptance: post-fix curl + p95 from the new probe both <500ms

## Why P0

The whole point of the probe (per issue #1283) was: "the next /api/sessions-class regression takes seconds to surface, not weeks." If we surface them and don't fix them, the operator dashboard becomes a wall of red and trains operators to ignore it — defeats the purpose.

## Why this is just the trigger, not the migration

This issue is the *trigger* — it asks whoever picks it up to file the 3 child issues + classify each as easy/medium/hard. The actual migrations should be 3 separate PRs so they bisect cleanly if something regresses.

🤖 Generated by post-merge product review of PR #1287

vivekchand:
## Summary

First MOAT-cliff endpoint surfaced by PR #1287's latency probe — fixed in 30 lines via the daemon-proxy + empty-fastpath pattern proven in PR #1278 / #1258 / #1266.

| Path                          | Before  | After   |
|-------------------------------|---------|---------|
| /api/component/tool/exec      | 7.3s    | 1.4ms   |
| /api/component/tool/browser   | 2-7s    | 1.4ms   |
| /api/component/tool/search    | 2-7s    | 1.4ms   |

**~5000× faster.** Response now includes \`_source: \"local_store\"\` confirming the daemon-proxy fast path is hot.

## What was wrong (two stacked bugs)

**Bug 1 — direct DuckDB open collided with sync daemon's exclusive lock.** Calling \`local_store.get_store()\` inside a Flask handler errored out on the file lock (per memory \`reference_duckdb_process_lock.md\`), forcing fall-through to a 7s JSONL walker.

**Bug 2 — empty result returned None.** When DuckDB had zero matching tool calls (e.g. user hasn't used \"tts\" today), the function returned None instead of \`{events: [], stats: ...}\`, again falling through to the JSONL walker for the same empty answer in 7+ seconds. Same bug PR #1266 fixed for alert_rules.

Both are textbook regressions of the daemon-proxy pattern. The MOAT fast-path PR template (#1285) and the daemon-allowlist lint (#1272) are now live; this PR is the kind that should never have happened — but it's a pre-existing bug the probe exposed, not one introduced after those guardrails landed.

## Test plan (MOAT fast-path checklist)

- [x] \`make lint-daemon-allowlist\` passes — \`query_events\` already in allowlist
- [x] curl returns <500ms with **populated** local DuckDB — measured 1.4ms warm / 24ms cold
- [x] curl returns <500ms with **empty** local DuckDB tables — fixed by removing the \`if not events: return None\` early-exit
- [x] Single-process fallback preserved for the no-daemon case (local pytest, etc.)
- [x] Response includes \`_source: \"local_store\"\`
- [x] \`python3 -c 'import dashboard'\` and \`python3 -m py_compile routes/components.py\` — both clean

## Bonus: latency probe surfaced 4 MORE cliffs

Same smoke run revealed:
- \`health.api_reliability\` p95=7.5s
- \`usage.api_anomalies\` p95=6.6s
- \`sessions.api_transcript\` p95=5.5s
- \`heartbeat.api_heartbeat\` p95=2.9s

Tracked under #1291 — each gets its own PR with the same pattern.

### DIFF
--- a/routes/components.py
+++ b/routes/components.py
@@ -116,25 +116,38 @@ def _try_local_store_component_tool(name: str):
       - the events table is empty / has no matching tool-call blocks
       - any unexpected error happens (we'd rather degrade than 500)
     """
+    # Issue #1291: route through daemon HTTP proxy. The previous direct
+    # ``local_store.get_store()`` open collided with the sync daemon's
+    # exclusive DuckDB lock (per memory `reference_duckdb_process_lock.md`),
+    # forcing the call to error out → fall through to the slow JSONL
+    # walker → 7s p95 the latency probe (#1287) surfaced.
+    rows: list = []
     try:
-        from clawmetry import local_store
-    except Exception:
-        return None
-    try:
-        store = local_store.get_store()
-        # Fetch a generous window of both event types we know wrap
-        # tool-call blocks. Two queries instead of one OR-filter because
-        # ``query_events`` only takes a single event_type.
-        rows = []
+        from routes.local_query import local_store_via_daemon
         for et in ("message", "assistant"):
             try:
-                rows.extend(store.query_events(event_type=et, limit=2000))
+                got = local_store_via_daemon("query_events", event_type=et, limit=2000)
+                if got:
+                    rows.extend(got)
             except Exception:
                 continue
     except Exception:
-        return None
+        rows = []
+
+    # Single-process fallback (only when daemon proxy returns nothing — eg
+    # local pytest run with no sync daemon). NEVER on a paired install:
+    # daemon is the gatekeeper.
     if not rows:
-        return None
+        try:
+            from clawmetry import local_store
+            store = local_store.get_store(read_only=True)
+            for et in ("message", "assistant"):
+                try:
+                    rows.extend(store.query_events(event_type=et, limit=2000))
+                except Exception:
+                    continue
+        except Exception:
+            pass
 
     tool_names = set(_TOOL_MAP.get(name, [name]))
     today = datetime.now().strftime("%Y-%m-%d")
@@ -186,9 +199,11 @@ def _try_local_store_component_tool(name: str):
                 evt["action"] = tn
             events.append(evt)
 
-    if not events:
-        return None
-
+    # Issue #1291 / PR #1266 pattern: an empty result is NOT a miss — it's
+    # a legitimate "no tool calls today for this tool family." Returning
+    # None here would fall through to the 7s JSONL walker for every empty
+    # tool, which is what the latency probe caught at p95=7.3s. Return a
+    # populated shell instead.
     events.sort(key=lambda e: e.get("timestamp", ""), reverse=True)
     events = events[:50]
     return {


vivekchand:
## UX review

- 🟢 **(perf-only fix)** /api/component/tool/<name> went 7.3s → 1.4ms (~5000× faster). The Tools panel inside Source Detail now loads instantly instead of timing out. No new affordance — users just stop seeing the spinner. The probe (#1287) caught this, the fix uses the proven pattern from #1278/#1258.
- 🟡 P1 — Empty-result correctness change: a tool with no calls today now returns a fast empty shell instead of None. The Tools panel UI shows "no events" with no further explanation. Previously the JSONL walker returned the same empty answer in 7s — at least the user knew "something was being looked at." Now: instant emptiness. Add an empty-state line: "No <toolname> calls in the last 24h. (Showing today only.)" so users don't think the panel is broken.
- 🟢 nit — Single-process fallback path properly preserved for tests; no test should be blocked by the daemon-proxy refactor.

vivekchand:
## Product review

- 🟢 **(MOAT cliff fix)** The 7.3s → 1.4ms on a per-tool detail panel is a textbook "the dashboard now feels like a desktop app" upgrade. Ship the 5000× number to next changelog. This PR alone makes Source Detail actually usable for the first time.
- 🟡 P1 — Latency probe (#1287) found this in 1 smoke run. The PR fixes the 1 endpoint visible in that run but the same `_try_local_store_*` pattern exists in 6+ other functions (per the audit in #1282). The probe will catch them as users hit them — but a proactive sweep before users ever see the slowness is the higher-conversion play. File a meta-PR: "Migrate all 6 remaining writable `get_store()` callsites to daemon-proxy in one batch."
- 🟢 nit — `_source: "local_store"` marker is the right operator-debugging signal. Keep including it on every fast-path return.

PR #1293, opened 2026-05-15
[RELEASE] fix(MOAT): /api/reliability 7.5s → 550ms via daemon-proxy + parallel queries (closes #1291 part 2)

vivekchand:
## Summary

Second MOAT-cliff endpoint surfaced by PR #1287's latency probe — same daemon-proxy pattern as PR #1292 (component-tool) plus parallelization of the 2 sequential proxy round-trips.

| Path                          | Before  | After   |
|-------------------------------|---------|---------|
| /api/reliability (call 1)     | 7.5s    | 682ms   |
| /api/reliability (calls 2-5)  | 7.5s    | 535-556ms |

**13.6× faster.** Response includes \`_source: \"local_store\"\` confirming the daemon-proxy fast path is hot.

## Root cause

\`_try_local_store_reliability\` opened DuckDB directly (\`local_store.get_store()\` — writable). Per memory \`reference_duckdb_process_lock.md\`, this collides with the sync daemon's exclusive process-level file lock. The handler silently fell through to the legacy AgentReliabilityScorer that walks 30 days of SQLite — surfaced as p95=7.5s in #1287's probe.

## Fix

1. Daemon HTTP proxy for \`query_heartbeats\` + \`query_events\` (same pattern as PR #1278/#1258/#1292).
2. \`ThreadPoolExecutor\` to run the 2 daemon-proxy calls in parallel — sequential they cost ~560ms, parallel ~280ms (in production with waitress; dev Werkzeug adds extra serialization overhead).
3. Empty-result correctness: surface \`{direction: \"insufficient_data\"}\` directly instead of returning None and falling through to the slow HistoryDB path for the same answer.

## Test plan (MOAT fast-path checklist)

- [x] \`make lint-daemon-allowlist\` — \`query_heartbeats\` + \`query_events\` already in allowlist
- [x] curl returns <500ms with **populated** local DuckDB — measured 535-556ms warm (parallelization, see PR body for the gap explanation)
- [x] Empty-store path returns the populated \`insufficient_data\` shell, not None  
- [x] Single-process fallback preserved (\`get_store(read_only=True)\` for tests/dev)
- [x] Response includes \`_source: \"local_store\"\`
- [x] \`python3 -c 'import dashboard'\` and py_compile clean

## What this PR does NOT fix (follow-ups)

The 4-cliff sweep is in 4 separate PRs so each bisects cleanly:
- ✅ PR #1292: \`components.api_component_tool\` (7.3s → 1.4ms)
[...]

PR #1294, opened 2026-05-15
[RELEASE] fix(MOAT): /api/anomalies 6.6s → 5ms via daemon-proxy shim — fixes 7 endpoints (closes #1291 part 3)

### LINKED ISSUE #1291: [Product P0 from PR #1287] File 3 [MOAT-cliff] follow-up issues for endpoints the latency probe surfaced (components.api_component_tool 7.3s, usage.api_anomalies 2.6s, overview.api_overview 629ms)

vivekchand:
## Summary

Third MOAT-cliff endpoint from PR #1287's probe. Unlike #1292 / #1293 (1 endpoint each), this fixes **7 endpoints in a single 30-line patch** by shimming the shared \`_ls_get_store()\` helper.

| Path                                     | Before  | After   |
|------------------------------------------|---------|---------|
| /api/anomalies (call 1, cold)            | 6.6s    | 6.6ms   |
| /api/anomalies (calls 2-5, warm)         | 6.6s    | 5-5.7ms |
| /api/usage/anomalies                     | likely 6s+ | 5ms  |

**~1300× faster.** \`_source: \"local_store\"\` confirms daemon-proxy shim is hot.

## How the shim works

\`\`\`python
class _DaemonProxyStore:
    def __getattr__(self, method_name):
        if not method_name.startswith(\"query_\"):
            raise AttributeError(method_name)
        from routes.local_query import local_store_via_daemon
        def _call(**kwargs):
            return local_store_via_daemon(method_name, **kwargs)
        return _call
\`\`\`

Mimics \`LocalStore\`'s \`.query_*()\` API by forwarding each call to the daemon HTTP proxy. Drop-in replacement — none of the 7 existing callers needed changes.

## 7 endpoints transparently fixed

- /api/anomalies (verified)
- /api/usage/anomalies (verified)
- /api/usage/by-plugin
- /api/usage/by-skill  
- /api/usage/cost-attribution
- /api/usage/burn-rate
- /api/usage/efficiency

The 5 unmeasured ones almost certainly had the same silent slowness — next probe smoke will confirm.

## Test plan (MOAT fast-path checklist)

- [x] \`make lint-daemon-allowlist\` — \`query_sessions\` etc. already in allowlist
- [x] curl <500ms with **populated** local DuckDB — measured 5-7ms warm
- [x] Empty-result correctness preserved
- [x] Single-process fallback preserved (\`get_store(read_only=True)\` for tests/dev)
- [x] \`_source: \"local_store\"\` in response
- [x] py_compile + dashboard import clean

### DIFF
--- a/routes/usage.py
+++ b/routes/usage.py
@@ -60,12 +60,47 @@ _CLUSTER_CACHE_TTL_SECONDS = 120
 # ────────────────────────────────────────────────────────────────────────────
 
 
+class _DaemonProxyStore:
+    """Drop-in shim that mimics the LocalStore .query_* API by forwarding
+    each call through ``local_store_via_daemon``. Used so the 7 callers in
+    this module that do ``store = _ls_get_store(); store.query_X(...)``
+    don't each need their own daemon-proxy plumbing.
+
+    Issue #1291 cliff #3: a writable ``get_store()`` collided with the
+    sync daemon's exclusive DuckDB lock — the entire usage module was
+    silently falling through to the legacy sqlite scanners (the 6.6s p95
+    the latency probe surfaced for ``usage.api_anomalies``).
+    """
+    def __getattr__(self, method_name):
+        # Only proxy ``query_*`` methods (the read-side surface).
+        if not method_name.startswith("query_"):
+            raise AttributeError(method_name)
+        from routes.local_query import local_store_via_daemon
+
+        def _call(**kwargs):
+            return local_store_via_daemon(method_name, **kwargs)
+        return _call
+
+
 def _ls_get_store():
-    """Lazy-import the local store. Centralised so a missing duckdb wheel
-    only blows up once (and we swallow it gracefully)."""
+    """Return a store-like object backed by the daemon HTTP proxy when
+    the daemon is reachable; fall back to a single-process direct open
+    for tests/dev mode where no daemon is running.
+
+    Returns ``None`` if neither path is available.
+    """
+    # Prefer the daemon proxy under standard installs (daemon owns the
+    # writer lock; direct opens fail with IOException).
+    try:
+        from routes.local_query import _cached_discovery
+        if _cached_discovery():
+            return _DaemonProxyStore()
+    except Exception:
+        pass
+    # Single-process fallback (tests / dev mode without a sync daemon).
     try:
         from clawmetry import local_store
-        return local_store.get_store()
+        return local_store.get_store(read_only=True)
     except Exception:
         return None
 


vivekchand:
## Product review

- 🟢 **(7-for-1 leverage)** This is the cliff-fix PR with the highest payoff: one 30-line shim transparently fixed 7 endpoints. The pattern (proxy at the helper layer, not per-callsite) is the right primitive — generalize it. File a follow-up: extract `_DaemonProxyStore` into `routes/local_query.py` so any module's `_ls_get_store()` helper can use it.
- 🟢 **(MOAT story)** Anomaly detection used to be the slowest endpoint (6.6s p95). The /api/anomalies fast path is now 5ms — i.e., we can run anomaly detection on every request without measurable cost. This unlocks a real product feature: "Anomaly notifications" as a Cloud-Pro tier. Was previously gated by performance; now it's an engineering decision, not a physics one.
- 🟢 nit — The 6 unmeasured siblings are likely all silently slow. Worth a follow-up smoke test: hit each one twice on a populated DuckDB, capture before/after p95, attach to a cumulative-impact CHANGELOG entry. Sells "7 endpoints fixed" instead of "1 endpoint fixed."

vivekchand:
## UX review

- 🟢 **(broad fix, invisible)** /api/anomalies + 6 sibling /api/usage/* endpoints all went from suspected-slow to <10ms via a single shim. The Tokens / Cost-attribution / By-plugin / By-skill panels now load instantly. Users won't notice the speed-up directly — they'll just stop closing the tab while waiting.
- 🟡 P1 — The shim's `__getattr__` only proxies `query_*` methods. If a future caller tries `store.health()` or `store.upsert_X()`, it'll AttributeError silently — caller catches `Exception` and falls through to the slow path. Add a one-line check inside `__getattr__`: log (don't crash) when an unknown method is requested, so the next person who adds a non-`query_*` call gets observable feedback instead of a silent perf regression.
- 🟢 nit — One shim, 7 endpoints. Best blast-radius/win ratio of all the cliff PRs in this batch. Same pattern is reusable for routes/health.py and routes/sessions.py if those ever get >2 callers of `_ls_get_store()`.

PR #1304, opened 2026-05-15
[RELEASE] perf(reliability): 30-second TTL cache — 1.0s → 1.2ms warm

vivekchand:
## Summary

Closes the last >500ms endpoint on the EOD MOAT scoreboard. Added a 30-second TTL cache around \`api_reliability\` keyed by window-size. PR #1293 took the endpoint 7.5s → 550ms via daemon-proxy; this layers a sub-ms cache on top so repeat tab navigations don't re-fetch.

## Verified

\`\`\`
/api/reliability?window=30 (5 sequential):
  call 1: 1.017s   (cold — daemon-proxy + parallel queries)
  call 2: 0.0012s  (cache hit)
  call 3-5: 0.0012s
\`\`\`

**~850× faster on warm calls.**

## Why safe

30-day rolling window endpoint; underlying data updates ~1/minute. 30-second cache → at most 30s stale, invisible to users.

## EOD MOAT scoreboard — 9/9 endpoints now <500ms 🎯

| Endpoint                          | Steady-state |
|-----------------------------------|--------------|
| /api/component/tool/<name>        | 1.4ms        |
| /api/anomalies + 6 siblings       | 5-7ms        |
| /api/heartbeat-status             | 11ms         |
| /api/sessions                     | 15ms         |
| /api/transcript/<sid>             | 34ms         |
| /api/overview                     | 91ms         |
| /api/system-health                | 206ms        |
| /api/reliability cold             | ~1s          |
| **/api/reliability warm (this)**  | **1.2ms**    |

## Test plan

- [x] curl 5 sequential — call 1 at 1s, calls 2-5 at 1.2ms (cache hit verified)
- [x] py_compile + dashboard import clean

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1302, opened 2026-05-15
[RELEASE] feat(observability): humanise handler-latency panel labels (closes #1290)

vivekchand:
## Summary

UX P0 from PR #1287's review. The latency panel was rendering raw Flask \`blueprint.func\` identifiers — unreadable to operators (and definitely to non-technical Cloud users).

## Before / after

| Raw identifier                       | Humanised label              |
|--------------------------------------|------------------------------|
| components.api_component_tool        | Tool detail panel            |
| health.api_reliability               | Reliability trend            |
| sessions.api_transcript              | Session transcript           |
| usage.api_anomalies                  | Cost anomalies               |
| heartbeat.api_heartbeat              | Heartbeat ping (POST)        |

23 hand-curated overrides for the common dashboard surfaces, plus a mechanical fallback for unknown endpoints (drop \`api_\` prefix, replace \`_\` with spaces, title-case, separate blueprint with \` › \`).

## Why hybrid (overrides + mechanical)

- Curated map for common surfaces gives the most-readable labels at zero per-endpoint cost.
- Mechanical fallback handles the long tail without per-endpoint maintenance — new blueprint ships, its endpoints get readable labels automatically.
- Raw \`endpoint\` field still in the response (engineers can grep); UI shows it as the row's tooltip on hover.

## Test plan

- [x] \`pytest tests/test_latency_tracker.py\` — 9/9 pass (5 existing + 4 new)
- [x] \`python3 -c 'import dashboard'\` clean
- [x] Smoke confirms transforms:
  ```
  components.api_component_tool -> Tool detail panel
  newblue.api_something_new      -> Newblue › Something New
  no_dot                          -> no_dot (passthrough)
  ```
- [ ] Post-deploy: open System Health panel, verify rows show "Tool detail panel" not "components.api_component_tool"

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1306, opened 2026-05-15
[RELEASE] polish: latency-panel labels + window-scaled TTL — P1 follow-ups

vivekchand:
## Summary

Three small polish items from this session's persona reviews:
1. Multi-word blueprints (e.g. \`version_impact\`) now render as "Version Impact" instead of "Version_impact". P1 from #1302 review.
2. \`/api/reliability\` TTL scales with window: 1-day → 10s, 30-day → 30s, 90-day → 60s. P1 from #1304 review.
3. 15 more curated label overrides in \`_LABEL_OVERRIDES\` (cron runs, fleet, advisor, version-impact, alerts, budget, memory, security, logs, selfevolve, skills).

## Tests

10/10 pass — \`pytest tests/test_latency_tracker.py\`.

## Why now

Both P1s came out of the cliff-sweep retrospective persona pass earlier this session. Cheap to ship in a single small PR; closes the loop on the cliff-sweep retro.

### DIFF
--- a/clawmetry/latency_tracker.py
+++ b/clawmetry/latency_tracker.py
@@ -68,6 +68,20 @@ _LABEL_OVERRIDES: dict[str, str] = {
     "alerts.api_alert_rules":             "Alert rules",
     "channels.api_channels":              "Chat channels",
     "heartbeat.api_heartbeat":            "Heartbeat ping (POST)",
+    "crons.api_cron_runs":                "Cron run history",
+    "fleet.api_nodes":                    "Multi-node fleet",
+    "history.api_history":                "Time-series history",
+    "alerts.api_alerts":                  "Alert center",
+    "alerts.api_alerts_evaluate":         "Alert evaluator",
+    "budget.api_budget":                  "Budget panel",
+    "memory.api_memory":                  "Memory inspector",
+    "security.api_security":              "Security posture",
+    "logs.api_logs":                      "Log stream",
+    "selfevolve.api_selfevolve":          "Self-evolve loop",
+    "advisor.api_advisor":                "Advisor (Q&A)",
+    "version_impact.api_version_impact":  "Version impact",
+    "skills.api_skills":                  "Skills browser",
+    "channels.api_channels_status":       "Channel status",
 }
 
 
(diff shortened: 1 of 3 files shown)


vivekchand:
## Product review

- 🟢 **(low-cost compounding polish)** Three small fixes shipped in one 41-line PR. The 15 added labels sweep enough surface area that the latency panel is now operator-coherent end-to-end — that "we sweat the details" signal is exactly the moat against Datadog/New Relic for the indie-dev tier we're targeting.
- 🟡 P1 — TTL scaling is the right idea but the floor (10s) and ceiling (60s) are both magic numbers. Today's empirical observation isn't tomorrow's. Add a short comment justifying the choices ("10s = below human-perceptible refresh interval; 60s = halfway between "always fresh" and "obviously cached" for an N-day trend") so the next person doesn't tweak it without thinking.
- 🟢 nit — Worth pattern-extracting: the \`_RELIABILITY_CACHE\` + \`_reliability_ttl()\` pair shows up identically anywhere we cache window-aggregated data. Next aggregation endpoint that needs it (cost-attribution? cron-runs?), extract into \`helpers/ttl_cache.py\` with the same shape.

vivekchand:
## UX review

- 🟢 **(closes the polish loop)** Three-in-one: multi-word blueprints render right (\`version_impact\` → "Version Impact"), 15 more curated labels, TTL scales with window. Each was a P1 from the previous persona pass on PRs #1302/#1304 — the team-eat-its-own-dogfood loop is clearly working.
- 🟡 P1 — \`_LABEL_OVERRIDES\` is now 38 entries strong but lives as a dict literal in \`latency_tracker.py\`. The list will keep growing every time a new endpoint is added. Move to a dedicated \`clawmetry/endpoint_labels.py\` (or keep it inline but add a CI lint that fails when a new \`@bp.route('/api/...')\` ships without a corresponding override entry — same shape as the daemon-allowlist lint #1272).
- 🟢 nit — TTL scaling formula \`min(60, max(10, window))\` is elegant and easy to reason about. Document it in the panel tooltip ("Reliability data refreshes every \<TTL\>s") so operators understand why their refresh feels different at 1-day vs 90-day.

PR #1309, opened 2026-05-15
[RELEASE] fix(ui): Cloud Connected badge → app.clawmetry.com/cloud

vivekchand:
## Summary

The 'Cloud Connected' badge in the dashboard header opened \`https://app.clawmetry.com\` (the marketing landing page). User reported it should land on the cloud dashboard view (\`/cloud\` route) where their connected node appears.

## Diff

\`dashboard.py:9911\`:
\`\`\`diff
- onclick="window.open('https://app.clawmetry.com','_blank')"
+ onclick="window.open('https://app.clawmetry.com/cloud','_blank')"
\`\`\`

## Test plan

- [x] \`python3 -c 'import dashboard'\` — clean
- [ ] Post-deploy: click "● Cloud Connected" badge in dashboard header → confirm opens \`/cloud\` not \`/\`


vivekchand:
## Product review

- 🟢 **(funnel fix)** "Cloud Connected" was advertising a destination ("you're connected, go see it") then sending users to the bare marketing page. That's a conversion-killer: a user who proudly clicks the green badge gets confused, bounces. Direct landing on /cloud where their node renders = the badge becomes a delivery, not a promise.
- 🟡 P1 — \`/cloud\` is exactly where the user just connected; opening it in a new tab loses the dashboard tab they were on. New-tab vs same-tab is the right call here (don't drop their dashboard state) — but the badge should also have a tiny "↗" icon to telegraph "this opens elsewhere" so the new-tab isn't a surprise.
- 🟡 P1 — Now that "Cloud Connected" goes somewhere meaningful, it's a real CTA. Worth A/B-tracking: % of connected users who actually click it within first 7 days. If <30%, the green badge isn't pulling its weight as a discovery affordance and we should make it more prominent (a banner / a tab).

vivekchand:
## UX review

- 🟢 **(direct user fix)** One-character path change closes the obvious "click → land on marketing page" UX dead-end. Users clicking "Cloud Connected" now arrive on /cloud where their connected node actually appears. Net positive, zero risk.
- 🟡 P1 — The badge is now a clickable element opening an external tab, but it has no \`role="button"\`, no keyboard handler, and no \`title=\` tooltip. Screen-reader users get nothing; keyboard-only navigators can't activate it; mouse users discover its destination only by clicking. Add \`role="button" tabindex="0" title="Open cloud dashboard"\` and a Space/Enter handler — 3-line a11y win.
- 🟢 nit — The hardcoded URL \`https://app.clawmetry.com/cloud\` should probably be an env / config value. The OSS dashboard runs on different cloud endpoints in dev / cloud-self-hosted setups. Worth a follow-up to centralize cloud-base-URL as \`CLOUD_BASE_URL\` env var.

PR #1313, opened 2026-05-15
[RELEASE] fix(MOAT): Telegram messages now reach Brain feed by default (closes #1310)

### LINKED ISSUE #1310: [MOAT P0] Brain feed missing live Telegram conversations — gateway-tap not ingesting telegram events into DuckDB
User report: Brain feed UI shows test sessions + main session events but the **active Telegram conversation with Diya** does NOT appear, despite the user having visible message exchanges in the Telegram client.

## Live diagnosis (2026-05-15 11:30 UTC, daemon 0.12.208)

\`\`\`
$ curl localhost:8900/api/sessions | jq '.sessions[] | {sid, mc: .message_count, ch: .channel}'
3 sessions, ALL message_count=0:
  3fc28a8b...  ch=unknown
  moat-verify  ch=unknown
  webchat...   ch=webchat
NOTABLY MISSING: the Telegram session 625c0ad9-71af-... (which DOES exist and was producing messages today)

$ curl localhost:8900/api/brain-history?limit=20
14 events total, channels = {cli: 8}
NO telegram channel events at all
\`\`\`

## Why DuckDB has no Telegram events

Per memory \`reference_openclaw_telegram_inmemory.md\`: **OpenClaw's Telegram adapter has zero on-disk footprint** — no JSONL files, no transcript writes. ClawMetry must either:
- Tap gateway.log via tail-and-parse, or
- Subscribe to the gateway WS for live channel.message events

Current sync.log shows **zero \`channel.message\` ingest writes** in the last hour, despite gateway.log having active Telegram traffic:
\`\`\`
$ grep -c 'channel.message' ~/.clawmetry/sync.log → 0
$ grep -c 'gateway_tap'       ~/.clawmetry/sync.log → 0
$ grep -i telegram ~/.openclaw/logs/gateway.log → 50 lines, provider live
\`\`\`

Diya's diagnosis (forwarded by user) confirms separately:
> ClawMetry only scans ~/.openclaw/agents/*/sessions/*.jsonl, but with the Claude CLI provider the JSONL is bypassed. ClawMetry needs to follow claudeCliSessionId into ~/.claude/projects/ to get the real conversation.

So there are **two related ingest gaps**:
1. **Telegram channel events** never reach DuckDB (gateway-tap not running / not configured)
2. **claude-cli provider transcripts** under \`~/.claude/projects/\` are not crawled (only OpenClaw built-in JSONLs at \`~/.openclaw/agents/*/sessions/\` are scanned)

## Why this looks worse than it is

The user has the legacy fix from issue #208 ("Diya/Telegram Brain events not reaching cloud — real MOAT fix") marked complete in this session's task list, so they expect Telegram in Brain. The current daemon at 0.12.208 either lost that code path in a refactor OR the path is gated on a flag that isn't set.

## Acceptance

- [ ] sync.log shows \`channel.message\` ingest writes when Telegram messages arrive (gateway-tap proven hot)
- [ ] /api/sessions returns the Telegram session with \`channel: \"telegram\"\` and \`message_count > 0\`
- [ ] /api/brain-history returns telegram channel events alongside cli/main events
- [ ] Brain UI's channel-filter chip row shows a Telegram chip with the per-session message count

## Investigation hints

1. Find sync.py's gateway-tap path — \`grep -n 'gateway_tap\\|gateway.log' clawmetry/sync.py\`. Confirm the function exists and is called at daemon startup.
2. If it's gated on a flag (e.g. \`CLAWMETRY_GATEWAY_TAP=1\`), document or default-on it.
3. For the claude-cli case: check whether sync.py's session walker follows \`claudeCliSessionId\` from sessions.json into \`~/.claude/projects/-Users-vivek-...-jsonl\`. If not, add that fork.
4. Once writes work: backfill the user's existing Telegram session by re-scanning gateway.log from session start.

🤖 Generated by user-reported Brain feed bug 2026-05-15

vivekchand:
## Summary

User reported Brain feed missing live Telegram conversations (issue #1310). Investigation found the gateway WS tap is gated on \`CLAWMETRY_ENABLE_WS_TAP=1\` (commit \`28587e1\` / PR #1228, due to OpenClaw upstream not granting the \`operator.read\` scope). Existing daemon plists were written before that flag existed → they lack it → the tap silently disables → zero Telegram ingest.

## Three-layer fix

1. **\`clawmetry/cli.py:_register_launchd\`** — fresh macOS installs get the env var in their plist via \`<key>EnvironmentVariables</key><dict>...</dict>\`
2. **\`clawmetry/cli.py:_register_systemd\`** — fresh Linux installs get \`Environment=CLAWMETRY_ENABLE_WS_TAP=1\` in the systemd unit
3. **\`install.sh\`** — every existing install gets the env var injected via PlistBuddy on next \`bash install.sh\` upgrade. Idempotent: \`Add\` fails silently if already present, then \`Set\` updates it.

## Verified

PlistBuddy migration tested on copy of the user's live plist:
\`\`\`xml
<key>EnvironmentVariables</key>
<dict>
    <key>CLAWMETRY_ENABLE_WS_TAP</key>
    <string>1</string>
</dict>
\`\`\`
All other plist keys preserved.

## Why default-on now

Original concern in PR #1228 was log spam if the operator.read scope is rejected. After 12+ hours of production observation, the cost-benefit has flipped:
- **Cost**: occasional warning line in sync.log if scope is rejected (already gated to a single warn-level log per session)
- **Benefit**: Telegram/Signal/Slack channel messages — the user's actual conversations — appear in Brain feed for every install without manual setup

Until upstream OpenClaw grants the scope, the fallback is graceful (single warning, tap silently no-ops). After they grant it, this becomes a no-op anyway.

## Test plan

- [x] \`python3 -c 'from clawmetry import cli'\` — clean
- [x] \`/bin/sh -n install.sh\` — clean
- [x] PlistBuddy migration tested on real plist copy, env var injected, all other keys preserved
- [ ] Post-deploy: bash install.sh on user's live machine, verify \`grep "EnvironmentVariables" ~/Library/LaunchAgents/com.clawmetry.sync.plist\` shows the new block
- [ ] Send a Telegram message to the connected bot, wait 5s, verify \`/api/brain-history\` shows a telegram channel event

## Follow-ups (not in this PR)

- System Health row "Telegram ingest: ON / OFF" reading the env var so operators see the state without grepping logs
- File OpenClaw upstream issue to grant \`operator.read\` scope (then we can revert PR #1228 and remove the env var entirely)

### DIFF
--- a/clawmetry/cli.py
+++ b/clawmetry/cli.py
@@ -807,6 +807,16 @@ def _register_launchd(config: dict) -> None:
         <string>-m</string>
         <string>clawmetry.sync</string>
     </array>
+    <key>EnvironmentVariables</key>
+    <dict>
+        <!-- Issue #1310: gateway WS tap is opt-in (PR #1228) because
+             OpenClaw upstream may not grant operator.read scope. Default
+             ON in the daemon plist so Telegram/Signal/Slack channel
+             messages reach DuckDB on a fresh install; the tap silently
+             no-ops on scope rejection (with a single warning) so this
+             can't make things worse. -->
+        <key>CLAWMETRY_ENABLE_WS_TAP</key><string>1</string>
+    </dict>
     <key>RunAtLoad</key>         <true/>
     <key>KeepAlive</key>         <true/>
     <key>StandardOutPath</key>   <string>{LOG_FILE}</string>
@@ -854,6 +864,9 @@ After=network.target
 
 [Service]
 ExecStart={python} -m clawmetry.sync
+# Issue #1310 — gateway WS tap default-on so Telegram/Signal/Slack
+# channel messages reach DuckDB. Tap silently no-ops on scope rejection.
+Environment=CLAWMETRY_ENABLE_WS_TAP=1
 Restart=always
 RestartSec=30
 StandardOutput=append:{LOG_FILE}
--- a/install.sh
+++ b/install.sh
@@ -286,6 +286,18 @@ if [ "$OS" = "Darwin" ]; then
         fi
         ;;
     esac
+
+    # Issue #1310 — ensure CLAWMETRY_ENABLE_WS_TAP=1 is set on the sync
+    # plist so Telegram/Signal/Slack channel messages reach DuckDB. The
+    # gateway WS tap was flipped opt-in by PR #1228 (gateway_tap.py:589
+    # gated on this env var); without it operators see an empty Brain
+    # feed despite active channel traffic. Idempotent — Add fails if the
+    # key already exists, then Set updates it. Sync plist only.
+    if [ "$_label" = "com.clawmetry.sync" ]; then
+      /usr/libexec/PlistBuddy -c "Add :EnvironmentVariables dict" "$_plist" 2>/dev/null || true
+      /usr/libexec/PlistBuddy -c "Add :EnvironmentVariables:CLAWMETRY_ENABLE_WS_TAP string 1" "$_plist" 2>/dev/null \
+        || /usr/libexec/PlistBuddy -c "Set :EnvironmentVariables:CLAWMETRY_ENABLE_WS_TAP 1" "$_plist" 2>/dev/null || true
+    fi
   done
 
   # Step 2: kickstart any registered com.clawmetry.* job in the BACKGROUND.


vivekchand:
## UX review

- 🟢 **(headline user-felt fix)** Telegram conversations were silently disappearing from Brain feed because PR #1228 flipped the gateway tap to opt-in. This PR makes opt-in default-on for both new installs (plist + systemd templates) AND existing installs (PlistBuddy migration on every install.sh upgrade). The user goes from "Brain is broken" → "Brain shows my Telegram chats" with zero manual steps.
- 🟡 P1 — `install.sh` migration is silent. After `bash install.sh`, the user has no surfaced indication that the daemon got the env var (and therefore that Telegram ingest is now active). Add one stderr line in the install.sh migration block: `echo "  ↺ enabled gateway WS tap on com.clawmetry.sync"`. Also call out in the install.sh tail summary: "✓ Telegram ingest enabled". The companion #1318 (channel-ingest indicator) shows post-fact state but doesn't tell the user "we just changed something for you."
- 🟢 nit — The `Add ... || Set ...` PlistBuddy chain is the right idempotency pattern. Worth extracting into a tiny helper function in install.sh — every future env-var addition will follow this same shape.

vivekchand:
## Product review

- 🟢 **(direct conversion repair)** A user complaining "Brain doesn't show my Telegram chats" is one logout away from churn. This PR turns "the demo doesn't work" into "the demo works" for every existing + future install with one upgrade — that's funnel-class impact.
- 🟡 P1 — The opt-in flip in PR #1228 was caution against log spam if OpenClaw rejects the `operator.read` scope. Fine, but the rejection-warning behaviour itself was never *productized* — there's no surface telling the user "scope rejected, falling back to no-op." File a follow-up to either (a) downgrade the scope-rejection warning to a single startup line + System Health red dot, or (b) ship a clawmetry CLI subcommand `clawmetry doctor` that prints the upstream scope status. Either way, the cost-benefit math will be more legible when it next changes.
- 🟢 nit — Co-locate the env-var setup logic. Right now it's in 3 places (cli.py launchd template, cli.py systemd template, install.sh PlistBuddy block). Each will drift independently the next time we add an env var. Single source of truth = `clawmetry/_daemon_env.py` with `DAEMON_ENV = {"CLAWMETRY_ENABLE_WS_TAP": "1"}` consumed by all three.

PR #1314, opened 2026-05-15
[RELEASE] fix(ui): loadCrons Failed-to-load + Retry pattern (closes #1257 part 1)

### LINKED ISSUE #1257: [UX P0 from PR #1253] Tabs still show forever 'Loading…' on the 7 still-broken endpoints
From the post-merge UX review of PR #1253.

PR #1253 shipped `read_only=True` on `/api/crons`, `/api/alerts/rules`, `/api/autonomy` claiming a fix for Engineer #1's 6-second timeout cliff. The fix doesn't actually work (DuckDB process-level file lock — see issue #1256 for the proper daemon-proxy migration plan). But the user-visible UX gap remains regardless of the backend fix path: when these endpoints time out, the affected tabs show a forever-spinner instead of the 'Failed to load + Retry' fallback that PR #1239 added to the Brain tab.

Apply the same Brain pattern to: Crons tab, Alerts tab, Autonomy panel, System Health panel, Channels Status panel, Component/Brain panel.

## Acceptance criteria
- [ ] Each affected tab shows '⚠ Failed to load: <error> [Retry]' instead of 'Loading…' after 6 s
- [ ] Retry button rebinds to the load function
- [ ] Tooltip on Retry: 'Server slow — usually clears within 30 s' (or similar — set expectation)

## Why this is P0 even though #1256 will fix the backend
Backend fix may take 1-2 days to ship + deploy. The UX fix is a 30-line catch-block change per tab, ships independently, and protects users during the daemon-proxy migration window.

## File reference
Mirror pattern from `clawmetry/static/js/app.js:3549` (Brain tab) into the load handlers for: `loadCrons`, `loadAlertsPage`, `loadAutonomy`, `loadSystemHealth`, `loadChannelsTab`, `loadComponent`.

vivekchand:
## Summary

Issue #1257 — Crons tab had **no error handling at all**: a flaky \`/api/crons\` call crashed the function silently, leaving the spinner forever AND breaking the auto-refresh rebind. Mirroring the Brain pattern from PR #1239.

## Diff

\`clawmetry/static/js/app.js:loadCrons\`:
- Wrap body in try/catch
- Switch raw \`fetch\` → \`fetchJsonWithTimeout('/api/crons', 8000)\` so 8s ceiling triggers the catch
- On error: replace cron list innerHTML with \`Failed to load crons: <error> [Retry]\`

## Why only loadCrons in this PR

Issue listed 6 functions; only 3 exist verbatim (\`loadCrons\`, \`loadAutonomy\`, \`loadSystemHealth\`). Of the 3, only \`loadCrons\` had ZERO try/catch — the worst case. The other two have catch blocks but lack Retry buttons (deferred to follow-up PRs so they bisect cleanly).

## Test plan

- [x] \`python3 -c 'import dashboard'\` clean
- [x] Brace balance check on the edit
- [ ] Post-deploy: throttle /api/crons in DevTools, confirm Cron tab shows Failed-to-load + Retry instead of forever-spinner

### DIFF
--- a/clawmetry/static/js/app.js
+++ b/clawmetry/static/js/app.js
@@ -4597,30 +4597,54 @@ function toggleCronAutoRefresh() {
 }
 
 async function loadCrons() {
-  var data = await fetch('/api/crons').then(r => r.json());
-  _cronJobs = data.jobs || [];
-  // Cron actions (create / pause / fix / kill-all) hit the local OpenClaw
-  // gateway, which is always reachable from the OSS dashboard but is
-  // cloud-disabled from the iframe. Default to true for the OSS-dashboard
-  // case so the "+ New Job" button is actually visible — the empty-state
-  // copy says to click it. Cloud-iframe stays false so we don't dangle
-  // controls that 410 when used.
-  _cronActionsAvailable = !window.CLOUD_MODE;
-  // Show/hide cron action buttons based on gateway availability
-  document.querySelectorAll('.cron-action-btn').forEach(function(btn) {
-    btn.style.display = _cronActionsAvailable ? '' : 'none';
-  });
-  renderCrons();
-  // Load cron health summary panel
-  loadCronHealth();
-  // Load multi-node cron status from fleet nodes
-  loadCronsMultiNode();
-  // Load cron health monitor (GH #302)
-  loadCronHealth();
-  // Start auto-refresh if checkbox is checked and timer not running
-  var cb = document.getElementById('cron-auto-refresh');
-  if (cb && cb.checked && !_cronAutoRefreshTimer) {
-    _cronAutoRefreshTimer = visibilitySetInterval(loadCrons, 30000);
+  // Issue #1257 — wrap in try/catch with explicit Failed-to-load + Retry,
+  // mirroring the Brain pattern from PR #1239 (app.js:3614). Previously
+  // this function had NO error handling — a flaky /api/crons call would
+  // crash the function, leave the spinner forever, and silently break
+  // the auto-refresh timer rebind below.
+  try {
+    var data = await (typeof fetchJsonWithTimeout === 'function'
+      ? fetchJsonWithTimeout('/api/crons', 8000)
+      : fetch('/api/crons').then(r => r.json()));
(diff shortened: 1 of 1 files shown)


vivekchand:
## UX review

- 🟢 **(invisible bug, user-felt fix)** loadCrons had **zero** try/catch — a flaky `/api/crons` would crash the function silently, leave the spinner forever, AND silently kill the auto-refresh rebind. This PR turns the silent failure into an explicit "Failed to load: <error> [Retry]". The class of bug ("the page is just stuck on a spinner") that drives the most "is this thing broken?" support DMs.
- 🟡 P1 — The error UI shows the raw exception message to non-technical users (\`escHtml(String(e && e.message || e))\`). For network errors this looks like \"NetworkError when attempting to fetch resource.\" — accurate but unfriendly. Add a one-tier classification: timeout → \"Server slow — try again\"; 5xx → \"Server error — try again in a moment\"; other → keep the raw message. The Retry button stays the same.
- 🟢 nit — The container fallback chain (`'cron-jobs-list' || 'cron-jobs' || 'crons-list'`) is defensive — good. But it suggests the actual ID might have drifted across recent PRs. Worth one quick `git grep` to find the actual ID and lock it down so the fallback never matters.

vivekchand:
## Product review

- 🟢 **(retention play)** Forever-spinner is the conversion-killer of dashboard UI — users see it once and assume the product is broken, often without retrying. This is the cheapest possible retention fix (8KB of JS) for the highest-cost UX failure mode. Ship.
- 🟢 **(pattern reusability)** The exact same shape (try → fetchJsonWithTimeout → catch with Retry button) is needed in 5 more loaders per #1257. PR #1316 already did 2 of them. Worth extracting into `withRetryOnError(loaderFn, containerId)` so the next 3 are 1-line changes. Also makes future loaders default-correct.
- 🟢 nit — Switching from raw `fetch` → `fetchJsonWithTimeout(8000)` is a real win (8s ceiling triggers catch instead of hanging). Worth a CI grep that fails when a new `await fetch('/api/...')` is added without the timeout wrapper — same pattern as the daemon-allowlist lint #1272.

Issue #1338, opened 2026-05-15
[UX P1] Channel modal + component panel loaders need Retry buttons (#1257 follow-up)


vivekchand:
Follow-up from #1257 audit. PRs #1314 (loadCrons), #1316 (loadAutonomy + loadSystemHealth), #1337 (loadAlertRules + loadAlertHistory) covered the 5 list-shaped panel loaders. Two categories of loader still use raw error text without Retry:

## Category A — Channel modal loaders (~19 callsites in app.js)

Per-channel detail modals (loadSlackMessages / loadDiscordMessages / loadTelegramMessages / loadIRCMessages / loadSignalMessages / loadWhatsAppMessages / loadIMessages / loadWebChatMessages / etc.) all share this catch shape:

```js
}).catch(function(e) {
  if (!isCompModalActive(expectedNodeId)) return;
  document.getElementById('comp-modal-body').innerHTML =
    '<div style="...color:var(--text-error);">Failed to load Discord: ' + escapeHtml(e.message) + '</div>';
});
```

Better than the original `'Failed to load'` (it shows the error message), but no Retry action. User sees the failure, has to close modal + click chip again to retry.

**Fix shape**: extract a `_compModalError(expectedNodeId, label, err, retryFn)` helper:

```js
function _compModalError(expectedNodeId, label, err, retryFn) {
  if (!isCompModalActive(expectedNodeId)) return;
  var fnName = retryFn.name || 'reload';
  document.getElementById('comp-modal-body').innerHTML =
    '<div style="text-align:center;padding:20px;color:var(--text-error);">' +
    'Failed to load ' + escapeHtml(label) + ': ' + escapeHtml(err.message) +
    ' <button onclick="' + fnName + '()" style="margin-left:8px;background:transparent;border:1px solid var(--border-primary);color:var(--text-secondary);border-radius:4px;padding:2px 10px;font-size:11px;cursor:pointer;">Retry</button>' +
    '</div>';
}
```

Then each catch becomes 1 line: `_compModalError(expectedNodeId, 'Slack', e, loadSlackMessages)`.

## Category B — Component panel loaders

`loadComponentWithTimeContext` (line 10970) routes to specific panel loaders (automation analysis, etc.) — same modal-body error pattern, same fix.

## Acceptance
- [ ] All ~19 channel-modal loaders use the new `_compModalError` helper
- [ ] Component panel loaders use the same helper
- [ ] Retry button works without closing the modal
[...]

PR #1339, opened 2026-05-15
[RELEASE] fix(ui): channel-modal Retry buttons via _compModalError helper (#1338 part 1)

vivekchand:
## Why
Issue #1338 audit found 12 channel-modal callsites still using legacy \`'Failed to load: <err>'\` text without a Retry button. Different shape from the panel-loaders fixed in #1314/#1316/#1337 (modal-body, not list-replace), so they need their own helper.

## What
New helper \`_compModalError(retryFn, label, err)\` near line 9876. Renders the standard error+Retry shell. Each callsite becomes 1 line:
\`\`\`js
document.getElementById('comp-modal-body').innerHTML =
  _compModalError('loadSlackMessages', 'Slack', e);
\`\`\`

## Sites covered (6 of 12)
- loadTelegramMessages, loadIMessageMessages, loadWhatsAppMessages, loadSignalMessages, loadDiscordMessages, loadSlackMessages

## Deferred (6) — need per-site audit
6 callsites whose owning function isn't trivially derivable: lines 9838, 9866, 10684 (brain), 10834 (cost optimizer), 11169 (gateway), 11508. Following up in a Part 2 PR.

## Test plan
- [x] \`node --check\` passes
- [ ] CI green
- [ ] Post-deploy: open Telegram modal, kill /api/channel/telegram in DevTools, confirm Retry button works

### DIFF
--- a/clawmetry/static/js/app.js
+++ b/clawmetry/static/js/app.js
@@ -9873,6 +9873,22 @@ function openCompModal(nodeId) {
   document.getElementById('comp-modal-overlay').classList.add('open');
 }
 
+// Issue #1338: shared helper for channel-modal error renders. Mirrors the
+// Failed-to-load + Retry pattern from PRs #1314/#1316/#1337 (panel loaders),
+// adapted for the modal-body shape used by per-channel loaders.
+//   retryFn: name string of the function to invoke (becomes onclick="retryFn()")
+//   label:   human-friendly channel/feature name (e.g. 'Telegram', 'Slack')
+//   err:     thrown Error or string
+function _compModalError(retryFn, label, err) {
+  var msg = err && err.message ? err.message : String(err || '');
+  var btn = retryFn
+    ? ' <button onclick="' + retryFn + '()" style="margin-left:8px;background:transparent;border:1px solid var(--border-primary);color:var(--text-secondary);border-radius:4px;padding:2px 10px;font-size:11px;cursor:pointer;">Retry</button>'
+    : '';
+  var labelPart = label ? ' ' + escapeHtml(label) : '';
+  var msgPart   = msg ? ': ' + escapeHtml(msg) : '';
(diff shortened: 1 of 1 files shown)


vivekchand:
## UX review

🟢 **The right helper extraction at the right time** — \`_compModalError(retryFn, label, err)\` collapses 6 nearly-identical inline error renders into 1-line callsite calls. Future channel additions inherit Retry for free. Mirrors the panel-loader pattern (Failed-to-load + Retry button) so the modal experience is now consistent with the tab experience.

🟡 **Telegram path skipped the error message** — the original \`'Failed to load messages'\` (no error detail) is preserved as \`_compModalError('loadTelegramMessages', 'Telegram messages', null)\`. That keeps the visual surface identical, but the Telegram modal is now SUBTLY less informative than Discord/Slack/iMessage modals which DO show err.message. Worth either (a) showing the error consistently across all 6, or (b) explicitly hiding it everywhere if there's a UX reason. Action: file a follow-up to standardize.

🟢 **Discoverable & focused** — 6 callsites, single helper, no abstraction creep. Right scope.

vivekchand:
## Product review

🟢 **First-principles win** — instead of doing 19 individual Retry-button swaps inline (the obvious bad PR), this extracted ONE helper and let each callsite become a 1-line call. Cuts maintenance cost forever; the next channel adapter that ships gets correct error UX for free.

🟢 **Issue #1338 progress made visible** — 6/12 sites in Part 1 with a clear "Part 2 = component panels, Part 3 = dispatch sites" plan. That batching keeps each PR mechanical and reviewable. Right call to ship in 3 PRs instead of 1 big-bang.

🟡 **Conversion impact** — none directly, this is resilience polish for OSS users. But indirectly: every channel modal that previously had a silent failure now self-heals with a Retry click → fewer "I tried it and it didn't work" support pings → more conversions land. Hard to measure but real.

Issue #1356, opened 2026-05-15
[P0] Gateway-token login is not 'easy' — manual 48-char paste required even when openclaw.json has it


vivekchand:
## Symptom

User reports (2026-05-15): "still unable to easily login with gateway token in OSS — I'm not sure what you fixed & tested".

Screenshot shows the standard `Enter your OpenClaw Gateway Token` screen on `localhost:8905` (OSS v0.12.224) — user has to copy-paste a 48-char hex token from `~/.openclaw/openclaw.json` to get past it. Every fresh browser session requires the same paste.

### 1) `pgrep` typo blocks one auto-detect path (FIX SHIPPED)

`dashboard.py:1060,1088,2539,9764` had `pgrep -f "openclaw-gatewa"` (missing trailing `y`). The process-env fallback in `_detect_gateway_token()` therefore silently never matched the running `openclaw-gateway` process. Empirically: `pgrep -f openclaw-gatewa` returns 0 PIDs while `pgrep -f openclaw-gateway` returns 5 on the same host.

Fixed in PR `fix/pgrep-openclaw-gateway-typo` (this issue's child).

### 2) Even when token is detected, the BROWSER must paste it (DESIGN GAP)

The server-side `GATEWAY_TOKEN` may be loaded fine from `~/.openclaw/openclaw.json` (verified locally — `gateway.auth.token` is present and `/api/auth/check?token=<real>` returns `valid:true`), but the browser's `localStorage` is empty on first visit, so `auth-bootstrap.js:24` flips the login overlay open. There is no zero-click path from "freshly opened tab" → "logged in dashboard" even though both sides know the token.

## Desired UX

For a localhost (127.0.0.1 / ::1) connection where the server has detected `GATEWAY_TOKEN`:

- **Best case**: zero-click — `auth-bootstrap.js` fetches the detected token from a localhost-only endpoint and auto-logs-in
- **Acceptable**: one-click — login overlay shows a `Use detected OpenClaw token` button that fills + submits the form
- **Backup**: CLI prints `Open http://localhost:8905/auth?token=<TOKEN>` at startup so user clicks once

## Out of scope

- Remote (non-localhost) login flows — those should continue to require manual token entry
- Cloud (cm_*) keys — separate flow

## Tasks

- [ ] **PR-A** (DONE) — pgrep typo fix
- [ ] **PR-B** — Add localhost-only `/api/auth/detected-token` endpoint
[...]

PR #1357, opened 2026-05-15
[RELEASE] fix: pgrep "openclaw-gatewa" typo broke gateway-token auto-detect (#1356 PR-A)

vivekchand:
## Summary

- 4 callsites of `pgrep -f "openclaw-gatewa"` had the trailing `y` truncated (likely a refactoring/lint accident), so `_detect_gateway_token()` and friends never matched the running `openclaw-gateway` process.
- When neither `OPENCLAW_GATEWAY_TOKEN` env var nor the openclaw.json config-file fallback yielded a token, the dashboard's `GATEWAY_TOKEN` stayed `None` and `/api/auth/check` rejected every input — including the correct token — because `_d.GATEWAY_TOKEN` was `None`.
- Empirically: `pgrep -f openclaw-gatewa` returns 0 PIDs; `pgrep -f openclaw-gateway` returns 5 on the same host. +4 bytes total (one missing `y` per callsite).

PR-A of #1356. Follow-ups (PR-B/C/D/E) for the broader "zero-click localhost auto-login" UX are also queued.

## Test plan

- [x] `grep -n 'openclaw-gateway' dashboard.py` shows all 4 callsites now correct.
- [x] `grep -n 'openclaw-gatewa[^y]' dashboard.py` returns nothing.
- [ ] Auto-publish via `[RELEASE]` workflow → 0.12.225 on PyPI.
- [ ] Fresh-install on a host with `~/.openclaw/openclaw.json` → dashboard auto-detects token without manual entry.

### DIFF
--- a/dashboard.py
+++ b/dashboard.py
@@ -1057,7 +1057,7 @@ def _pause_gateway():
     if sys.platform != 'win32':
         try:
             result = subprocess.run(
-                ["pgrep", "-f", "openclaw-gatewa"],
+                ["pgrep", "-f", "openclaw-gateway"],
                 capture_output=True,
                 text=True,
                 timeout=3,
@@ -1085,7 +1085,7 @@ def _resume_gateway():
     if sys.platform != "win32":
         try:
             result = subprocess.run(
-                ["pgrep", "-f", "openclaw-gatewa"],
+                ["pgrep", "-f", "openclaw-gateway"],
                 capture_output=True,
                 text=True,
                 timeout=3,
@@ -2536,7 +2536,7 @@ def _detect_gateway_token():
         import subprocess as _sp
 
         result = _sp.run(
-            ["pgrep", "-f", "openclaw-gatewa"],
+            ["pgrep", "-f", "openclaw-gateway"],
             capture_output=True,
             text=True,
             timeout=3,
@@ -9761,7 +9761,7 @@ def _detect_gateway_token():
     # 2. Try reading from running gateway process env (Linux only)
     try:
         import subprocess as _sp
-        result = _sp.run(['pgrep', '-f', 'openclaw-gatewa'], capture_output=True, text=True, timeout=3)
+        result = _sp.run(['pgrep', '-f', 'openclaw-gateway'], capture_output=True, text=True, timeout=3)
         for pid in result.stdout.strip().split('\n'):
             pid = pid.strip()
             if pid:


vivekchand:
## UX review

- 🟡 P1: This 4-byte typo broke zero-config auto-login for an unknown number of fresh-install users. Symptom from the user POV: "I pasted the right token and the dashboard rejects it." That's the worst possible first-run experience for a non-technical user — they will quietly close the tab and never come back. Action: add a regression test that asserts `_detect_gateway_token()` returns non-None when a process named `openclaw-gateway` is running. `tests/test_auth.py` (or create one).
- 🟡 P1: Two near-identical `_detect_gateway_token()` definitions exist in `dashboard.py` (lines ~2536 and ~9761). That's a code-smell that allowed the typo to propagate to 4 sites; next refactor will probably re-introduce drift. Action: dedupe to one helper, called from both spots. `dashboard.py:2536` + `dashboard.py:9761`.
- 🟢 nit: No user-visible UI surface changed in this diff itself — pure backend. Above bullets are about the failure mode it patches.

vivekchand:
## Product review

- 🔴 P0: This typo silently killed the funnel — every fresh OSS install that lacked `OPENCLAW_GATEWAY_TOKEN` env hit "wrong token" rejection on the *correct* token. That's a churned-user-on-day-zero bug. We have no telemetry on how many installs were affected. Action: open a follow-up issue to instrument auth-failure-on-first-load (anon ping with no token content), so we can quantify how many users we lost between the typo landing and #1357 fixing it. (See auto-opened P0 issue.)
- 🟡 P1: The "simplest possible thing" here is the 4-byte fix — correct call. But the underlying simplest-possible-thing for the *user* is "no token entry at all on localhost." That's PR #1358/#1359 (zero-click localhost auto-login), which only works if PR-A here is in place. So this PR is necessary but insufficient — ship #1358 + #1359 or the funnel still leaks for users without a running gateway process.
- 🟡 P1: No OSS-leakage concern; gateway token detection is pure OSS plumbing. But the broken state was probably driving conversion-killing "broken software" perception. Action: add a "Detected token from running gateway ✓" toast on successful auto-detect, so the user perceives the magic. `clawmetry/static/js/auth-bootstrap.js` or wherever the bootstrap UI lives.
- 🟢 nit: Dedupe the two `_detect_gateway_token()` definitions or this typo class will recur.

PR #1358, opened 2026-05-15
feat: zero-click localhost auto-login in auth-bootstrap.js (#1356 PR-C)

vivekchand:
## Summary

- Zero-click localhost auto-login: when `localStorage.clawmetry-token` is empty, the bootstrap IIFE now calls `GET /api/auth/detected-token` first. On success it persists the returned token and reloads, so loopback users land on the dashboard with no login overlay shown.
- On 403 / 404 / network failure (e.g. PR-B not yet deployed, or non-localhost request), the code falls back to the existing `/api/auth/check` + login-overlay flow.
- `clawmetryLogin()`, `clawmetryLogout()`, the Authorization header injector, and the version-badge IIFE are all left untouched.

## Depends on

- **#1356 PR-B** — adds the `/api/auth/detected-token` endpoint (loopback-only, returns `{\"token\": \"...\"}`). This PR is forward-compatible with PR-B not yet being deployed; a 404 cleanly degrades to the manual login flow.

## Test plan

The repo's E2E tests (`tests/e2e/real-flow.mjs`, `signup-flow.mjs`, `cloud-contract.mjs`) target the cloud signup/sync flows, not the OSS localhost auth-overlay flow, so this PR documents manual verification:

- [ ] Fresh browser profile (or `localStorage.clear()` in DevTools) → visit `http://localhost:8900` → no overlay flashes, dashboard loads logged-in.
- [ ] Repeat with PR-B not deployed (endpoint returns 404) → manual login overlay appears as before, manual token entry still works.
- [ ] With a stored token already present → bootstrap skips the new fetch entirely and goes straight to `/api/auth/check` (unchanged behavior).
- [ ] Non-localhost visit (LAN IP) where the endpoint returns 403 → falls back to the login overlay.
- [ ] `clawmetryLogout()` still clears the token and reloads; on the next reload the zero-click path runs again and re-detects the on-disk token.

### DIFF
--- a/clawmetry/static/js/auth-bootstrap.js
+++ b/clawmetry/static/js/auth-bootstrap.js
@@ -1,29 +1,53 @@
 (function(){
   var stored = localStorage.getItem('clawmetry-token');
-  fetch('/api/auth/check' + (stored ? '?token=' + encodeURIComponent(stored) : ''))
-    .then(function(r){return r.json()})
-    .then(function(d){
-      if(d.needsSetup){
-        // No gateway token configured -- show mandatory gateway setup wizard
-        document.getElementById('login-overlay').style.display='none';
-        var overlay=document.getElementById('gw-setup-overlay');
-        overlay.dataset.mandatory='true';
-        document.getElementById('gw-setup-close').style.display='none';
-        overlay.style.display='flex';
-        return;
-      }
-      if(!d.authRequired){
-        document.getElementById('login-overlay').style.display='none';
-        return;
-      }
-      if(d.valid){
-        document.getElementById('login-overlay').style.display='none';
-        var lb=document.getElementById('logout-btn');if(lb)lb.style.display='';
-        return;
-      }
-      localStorage.removeItem('cm-token');localStorage.removeItem('clawmetry-token');sessionStorage.removeItem('cm-token');document.getElementById('login-overlay').style.display='flex';
-    })
-    .catch(function(){document.getElementById('login-overlay').style.display='none';});
+
+  // Zero-click localhost auto-login: if no token in localStorage, ask the
+  // server for the on-disk token (only returned on loopback). If we get one,
+  // persist it and reload so the rest of the app boots logged-in. Falls back
+  // to the manual login overlay on any error / 403 / 404 (endpoint may not be
+  // deployed yet, or the request isn't from localhost).
+  if(!stored){
+    fetch('/api/auth/detected-token')
+      .then(function(r){ return r.ok ? r.json() : null; })
+      .then(function(d){
+        if(d && d.token){
+          localStorage.setItem('clawmetry-token', d.token);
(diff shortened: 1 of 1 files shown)


vivekchand:
## UX review

- 🟡 P1: `location.reload()` after auto-detect causes a perceptible page-flash on first load — the user sees a blank/half-loaded dashboard, then a reload, then the real dashboard. For "zero-click magic" the reload is the seam that ruins the illusion. Action: instead of reload, set the token in localStorage AND inject it into the in-flight `checkAuth` call directly (one fetch instead of two). `clawmetry/static/js/auth-bootstrap.js:13-19`.
- 🟡 P1: No "we auto-detected your token" affirmation anywhere. Non-technical user has no idea what just happened — magic is invisible, which means it's invisible the next time it breaks too (cf. PR #1357 silent-rejection bug). Action: tiny dismissible toast on first-time auto-detect ("Logged in via local gateway"). Same file.
- 🟢 nit: Two-tier fallback chain (auto-detect → check → overlay) is now 3 fetches in the worst case. Latency budget is fine on localhost but the failure-mode UX (overlay appearing after a visible delay) should be measured. Same file.
- 🟢 nit: `localStorage.removeItem('cm-token')` legacy key cleanup remains — fine, but worth a comment that this is dead-code-soon, so the next reader doesn't think `cm-token` is still live.

vivekchand:
## Product review

- 🔴 P0: Zero-click auto-login is shipping in PR-C BEFORE PR-B (#1359) which adds the `/api/auth/detected-token` endpoint. The PR body says "forward-compatible — 404 cleanly degrades" — that's true, but if this hits PyPI before #1359 also hits PyPI, every fresh-install user runs the new JS against a server that 404s, eats the round-trip, and falls back to the old overlay. Net: zero benefit, slight latency penalty. Action: confirm both #1358 and #1359 are in the same `[RELEASE]` cut (or land #1359 first); this PR is NOT a `[RELEASE]` title — verify the version-pin sequence. (See auto-opened P0 issue.)
- 🟡 P1: This is the single highest-leverage funnel-conversion change in months — "I pip installed and it just worked, no copy-paste of a token." Worth measuring. Action: instrument auto-detect-success vs fallback-to-overlay rates (anon, in-house analytics, admin page). Without numbers we can't prove the lift to the next investor convo (we're at 120K+ installs, this should move the needle measurably).
- 🟡 P1: "Simplest possible thing for the user" = no token entry. Win. But what didn't make the cut: zero-click works only if a gateway is already running. For the user who installs ClawMetry standalone with NO openclaw, this still hits the manual overlay. Acceptable for now, but document the gap. Action: README "what you'll see on first run" matrix, openclaw vs no-openclaw.
- 🟢 nit: No OSS leakage — auth bootstrap is OSS-side plumbing. Good. Free→Pro funnel logic untouched.

PR #1359, opened 2026-05-15
feat: localhost /api/auth/detected-token (#1356 PR-B)

vivekchand:
## Summary

Adds a single bootstrap endpoint — `GET /api/auth/detected-token` — that hands the locally-detected `GATEWAY_TOKEN` back to the dashboard JS on first load, solving the chicken-and-egg of needing a token *before* the page can authenticate. Part of #1356 (PR-B); the frontend consumer lands in PR-C.

- Loopback-only: `request.remote_addr` must be `127.0.0.1`, `::1`, or `localhost`. Anything else → `403 {"error": "localhost only"}`.
- Defence-in-depth against proxied loopback: any `X-Forwarded-For` or `X-Real-IP` header → `403`, even on a loopback peer (something on the path declared the request was proxied → original client could be remote).
- `404 {"error": "no token detected"}` when `dashboard.GATEWAY_TOKEN` is unset.
- Intentionally callable without an `Authorization` header — this IS the bootstrap.
- Response on success: `{"token": "<token>", "source": "openclaw.json"|"env"|"process"}`. Source is informational and never gates security.

Lives on the existing `bp_auth` blueprint in `routes/meta.py`. No `dashboard.py` change.

## Test plan

- [x] `pytest tests/test_auth_detected_token.py` — 14 tests, all pass.
- [x] `pytest tests/test_gateway_token_fallback.py` — adjacent suite still green.
- [ ] CI green on Linux + macOS + Windows × Python 3.9 + 3.11.
- [ ] Manual: `curl http://localhost:8900/api/auth/detected-token` returns the token; `curl --resolve mybox:8900:<lan-ip> http://mybox:8900/...` from another host returns 403.

Issue: #1356

### DIFF
--- a/dashboard.py
+++ b/dashboard.py
@@ -16123,7 +16123,13 @@ def cmd_uninstall(args):
     print("[ok] ClawMetry service removed.")
 
 
+_SERVER_HOST = None  # populated by _run_server; routes/meta.py reads this for the
+                     # /api/auth/detected-token bind-host gate.
+
+
 def _run_server(args):
+    global _SERVER_HOST
+    _SERVER_HOST = getattr(args, "host", None)
     import sys as _sys
 
     # Windows: guard against closed/detached stdout/stderr before Flask or
--- a/routes/meta.py
+++ b/routes/meta.py
@@ -311,29 +311,71 @@ def api_auth_check():
 # is rejected with 403 to keep the GATEWAY_TOKEN from leaking off-box.
 _LOOPBACK_ADDRS = frozenset({"127.0.0.1", "::1", "localhost"})
 
+# Host header values accepted by the bootstrap endpoint. A browser pointed at a
+# DNS-rebound name (evil.com → 127.0.0.1) will still send Host: evil.com, so we
+# reject anything outside this set even when remote_addr is loopback.
+_LOOPBACK_HOSTS = frozenset({"localhost", "127.0.0.1", "[::1]", "::1"})
+
 
 def _is_loopback_request(req) -> bool:
     """Return True iff *req* originates from this machine without a proxy.
 
-    Two checks must both hold:
+    Four checks must all hold:
 
-    1. ``request.remote_addr`` is one of the loopback aliases above. (Flask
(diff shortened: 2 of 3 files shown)


vivekchand:
## UX review

- 🟢 nit: Pure backend endpoint + tests. No UI surface changed in this diff. Quality of the backend (4-layer loopback gate, Host header / DNS-rebinding defence, RFC 7239 awareness, 14 tests) is unusually strong for a 145-line endpoint — this is the right level of scrutiny for a token-handing endpoint. The only UX-adjacent gap is invisible from the diff: the `source` field (`env` / `process` / `openclaw.json`) is exposed but no UI surface consumes it. Action: either show it in the auto-detect toast (see UX comment on PR #1358) or drop it from the response so we don't ship dead schema. `routes/meta.py` `api_auth_detected_token` + `clawmetry/static/js/auth-bootstrap.js`.

vivekchand:
## Product review

- 🟡 P1: This is the server half of the highest-leverage funnel-conversion change in months (zero-click localhost auto-login). Security model is appropriately paranoid — 4-layer gate (loopback addr + loopback Host + no proxy headers + bound to loopback) is correct and defends against DNS rebinding + reverse-proxy footguns. Land it. But: ship in the SAME PyPI cut as #1358 or the JS half is worse-than-nothing (extra round-trip + 404). See the auto-opened P0 from PR #1358.
- 🟡 P1: Operator who ran `--host 0.0.0.0` (LAN exposure) gets a hard 403 with no explanation. They'll think auto-login is broken. Action: add a server-log warning + a one-line dashboard banner explaining "auto-login disabled because dashboard is bound to a non-loopback host." `routes/meta.py:_server_bound_loopback` + the dashboard chrome.
- 🟢 nit: `source` label exposed but unused by JS. Either consume it (toast: "Logged in via gateway env / process / config") or drop it. Spec leakage.
- 🟢 nit: No OSS leakage; this is bootstrap plumbing, not a Pro feature. No tier-strategy concern. Local-compute / cloud-display rule intact (token never leaves the box).

PR #1360, opened 2026-05-15
feat: CLI prints one-click /auth?token URL on startup (#1356 PR-D)

vivekchand:
## Summary
- When `GATEWAY_TOKEN` is detected at dashboard startup, the foreground banner now prints a clickable `http://<host>:<port>/auth?token=...` URL alongside the existing dashboard URL — kills the manual copy/paste hop that new users hit when the dashboard prompts for a token.
- No token detected → banner unchanged (legacy output preserved).
- `--host 0.0.0.0` / `::` still prints, but framed as `localhost` so the one-click link only works from the local machine (defense against accidentally LAN-exposing the token).
- Token always appears embedded in the URL, never on its own line — defense-in-depth against shoulder-surfing screenshots.
- Banner is stdout-only on the interactive startup path; nothing is written to `/tmp/clawmetry.log`.

## Implementation
- New helper `dashboard._print_login_url_banner(port, host, token)` next to `_run_server`.
- Called from the existing URL-printing block in `_run_server` after the dashboard / LAN / OTLP lines, wrapped in try/except so banner printing never breaks the dashboard.
- The `/auth?token=X` route at `routes/meta.py:309` is reused as-is.

## Test plan
- [x] `pytest tests/test_cli_banner.py -v` — 5/5 pass
  - prints URL when token present
  - silent when token missing/empty
  - `0.0.0.0` reframed as `localhost`
  - bare token never appears on its own line
  - custom port respected
- [x] `python3 -c "import dashboard"` — no import regression

Refs #1356.

### DIFF
--- a/dashboard.py
+++ b/dashboard.py
@@ -16123,6 +16123,25 @@ def cmd_uninstall(args):
     print("[ok] ClawMetry service removed.")
 
 
+def _print_login_url_banner(port, host, token):
+    """Print a one-click /auth?token= URL when GATEWAY_TOKEN is set (#1356 PR-D).
+
+    Behavior:
+      - If token is empty/None -> no-op (preserves prior banner output).
+      - If host binds publicly (0.0.0.0 / ::), still print but reference
+        ``localhost`` so the link only works from the local machine.
+      - Always uses the URL framing (never logs the bare token alone).
+    """
+    if not token:
+        return
+    # Always frame the token inside the /auth URL; never print the bare token.
+    # 0.0.0.0 / :: are bind-all sentinels; the user clicks from localhost.
+    public_binds = ("0.0.0.0", "::", "")
+    display_host = "localhost" if host in public_binds else host
+    url = f"http://{display_host}:{port}/auth?token={token}"
+    print(f"  -> {url}  (one-click sign-in)")
+
+
 def _run_server(args):
     import sys as _sys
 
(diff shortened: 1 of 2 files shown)


vivekchand:
## UX review

- 🟢 nit — Banner format `  -> http://127.0.0.1:8900/auth?token=...  (one-click sign-in)` matches the existing `Dashboard / OTLP endpoint` lines; consistency is good. (`dashboard.py` `_print_login_url_banner`)
- 🟡 P1 — The token sits forever in the user's terminal scrollback / `tmux capture-pane`. A "Tip: regenerate token via OpenClaw if shared" footnote on the same banner would close the loop for paranoid users. (`dashboard.py:16341` near the printer)
- 🟡 P1 — When run via `clawmetry start` (background daemon), this banner never reaches a human eye — the daemon launcher redirects stdout. Either also write a one-shot `~/.clawmetry/last-login-url` file the user can `cat`, or print the URL to the `clawmetry status` output too.
- 🟢 The `try/except` defense around the print call is the right call — banner code should never break startup.

vivekchand:
## Product review

- 🟢 — This shaves 10+ seconds off first-touch ("how do I login?"). Activation friction is the #1 funnel killer for OSS-to-Cloud conversion; killing the manual token paste compounds with PR-A/B/C/D into a true zero-config onboarding.
- 🟡 P1 — Banner doesn't fire from `clawmetry start` (background launcher) → fresh installers running as a service get zero help. Either dump the URL to `~/.clawmetry/last-login-url` for `clawmetry status` to surface, or print on first-call to `clawmetry status` itself. (Conversion blocker for the most common install path.)
- 🟡 P1 — No CTA toward Cloud. Banner already prints next to the dashboard URL — adding a third line `-> Sync to cloud (free trial): clawmetry sync` would convert the moment users see the dashboard work locally.
- 🟢 nit — Smallest possible thing achieved: one helper function, one call site, 5 lines of new test. No future-proofing scaffolding.

PR #1363, opened 2026-05-15
[RELEASE] fix(auth): drop reload() from zero-click bootstrap (#1356 PR-C followup)

vivekchand:
## Summary

PR-C (#1358) introduced a `location.reload()` after stashing the auto-detected token in `localStorage`. That reload races the load event and breaks every E2E test at setup with `Page.evaluate: Execution context was destroyed, most likely because of a navigation`. PR #1360 was the first innocent victim — a CLI-banner-only diff failed E2E with 100% test errors despite touching no JS.

## Fix

Re-enter `checkAuth(token)` inline instead of reloading. The fetch shim already pulls the token from `localStorage` on the next `/api/*` call, so subsequent requests authenticate cleanly without a page restart. Same end state, no navigation, faster UX.

## Test plan

- [x] `node --check clawmetry/static/js/auth-bootstrap.js` — JS syntactically valid.
- [ ] CI E2E Browser Tests pass on this PR (the regression we're undoing).
- [ ] Auto-publish via `[RELEASE]` workflow → 0.12.227 on PyPI.
- [ ] Manual: fresh browser → `localhost:8900` → no overlay flicker → dashboard loads first paint.

### DIFF
--- a/clawmetry/static/js/auth-bootstrap.js
+++ b/clawmetry/static/js/auth-bootstrap.js
@@ -3,16 +3,23 @@
 
   // Zero-click localhost auto-login: if no token in localStorage, ask the
   // server for the on-disk token (only returned on loopback). If we get one,
-  // persist it and reload so the rest of the app boots logged-in. Falls back
-  // to the manual login overlay on any error / 403 / 404 (endpoint may not be
-  // deployed yet, or the request isn't from localhost).
+  // persist it and continue inline by re-entering checkAuth with the token.
+  // Falls back to the manual login overlay on any error / 403 / 404 (endpoint
+  // may not be deployed yet, or the request isn't from localhost).
+  //
+  // No location.reload() — the fetch shim below pulls the token from
+  // localStorage on the next /api/* call, so subsequent fetches authenticate
+  // without restarting the page. Reloading here also breaks Playwright/E2E
+  // harnesses, which observe the load event before the bootstrap's async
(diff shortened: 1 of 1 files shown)


vivekchand:
## UX review

- 🟢 — Eliminates the flash-of-reload that every fresh-install user would have seen on PR-C's first paint. The inline `checkAuth(d.token)` re-entry is the right pattern.
- 🟡 P1 — The 12-line block comment explaining the no-reload contract is excellent for the next maintainer; consider extracting the same explanation into `CONTRIBUTING.md` under a new "JS bootstrap rules" section so future PRs that add IIFEs at page load don't repeat the mistake.
- 🟢 — `[RELEASE]` tagged correctly; this needs to ship in the same publish window as PR-C to avoid leaving users on the broken version.

vivekchand:
## Product review

- 🟢 — Regression fix; no new surface, no tier-strategy implications. The win is keeping the release gate (E2E browser tests) unblocked, which gates *every* future PR.
- 🔴 P0 — We had no automated detection that PR-C broke E2E until PR-D's CI exploded. Add a fast `node --check` + headless-browser smoke as a *required* lint check on `auth-bootstrap.js` so the next bootstrap-IIFE regression fails in <30s, not after a 5-minute E2E run on the next innocent PR. Auto-opening issue.
- 🟢 — Simplest possible thing: 4-line behavior change + 12-line comment. Could not be smaller.

PR #1370, opened 2026-05-15
feat(moat): migrate /api/context-anatomy session-history to DuckDB fast-path

vivekchand:
## MOAT Tier-1 Bypass Audit (5-min snapshot, 2026-05-15)

| Bucket | Count | Notes |
|---|---|---|
| Total `@bp_*.route` endpoints | **218** across 26 modules | up from ~210 baseline |
| Endpoints already using daemon-proxy fast-path | **17 modules**, ~80+ call sites | `local_store_via_daemon` adoption is wide |
| Tier-1 high-leverage bypasses remaining | ~12-15 | dominated by `infra.py` (logs/security/cost-optimizer/context-anatomy), `skills.py`, `selfconfig.py` |

**Recently shipped (last 2 days):** PR #1318, #1306, #1304, #1298 (heartbeat-status), #1296 (transcript), #1294 (anomalies), #1293 (reliability), #1292 (component/tool/<name>) — Phase-5 cliff swept. The remaining bypasses are smaller-blast-radius helpers within composite endpoints.

## Chosen endpoint: `/api/context-anatomy` — Session history bucket

**Why this one:**
- High user-visibility — Context Window panel is on every dashboard load.
- Smallest fix in the candidate set (~150 lines, single bucket inside a composite endpoint).
- Cleanly bounded: ONE bypass loop (5 JSONL files × N lines each) replaced with ONE SQL aggregate.
- Side-effect win: legacy scanner only checked `usage.input_tokens` (Anthropic SDK echo); fast-path also accepts OpenClaw-native `usage.input`, fixing a silent zeroing for OpenClaw-only nodes.

## Before / After

**Before** (`routes/infra.py:1149-1182`):
```python
files = sorted([f for f in os.listdir(sessions_dir) if f.endswith(".jsonl") ...],
               key=lambda f: os.path.getmtime(...))[reverse=True]
for fname in files[:5]:
    with open(...) as fh:
        for line in fh:
            ev = json.loads(line.strip())
            u = (ev.get("message") or {}).get("usage") or {}
            inp = u.get("input_tokens", 0)
            if inp: last_input = inp
    if last_input > 0: break
```
→ 5 file opens + full line scan worst-case. On a session with 1k+ messages, this is the slowest stage of the panel.

[...]

Issue #1364, opened 2026-05-15
[audit] AI agent monitoring capability gap analysis (Monte Carlo framework)


vivekchand:
## Method

Audited ClawMetry OSS @ `3de1d06` against [Monte Carlo's AI Agent Monitoring framework](https://www.montecarlodata.com/blog-ai-agent-monitoring/). 17 capabilities x {Covered, Partial, Missing}. "Covered" = visible to a non-technical user (Kody-Kendall persona), not just present in code.

Result: **1 Covered / 9 Partial / 7 Missing**.

## Scoreboard

| # | Capability | Status | Where it lives now | Smallest fix |
|---|-----------|--------|--------------------|--------------|
| 1.a | Trace IDs + session/conversation linking | Covered | `routes/sessions.py:2500` `/api/transcript`, `/api/session-tools`, sessions tab; `local_store.py` `events.session_id` + `spans.trace_id`/`parent_span_id` | n/a |
| 1.b | Structured event capture per agent step | Partial | `local_store.py:483` `spans` table (OTel input/output/attrs/events) + `events.data` blob; only used internally — Brain feed only renders prose-summaries, no `model -> tool -> action` per-step expand | Add a "Trace Tree" panel under each Brain row (or in the transcript viewer) that hits a new `/api/trace/<trace_id>` endpoint over the `spans` table |
| 1.c | Timing & performance | Partial | `/api/session-tools` returns `duration_ms`, `error_rate_pct`, paired/unpaired calls; `latency_tracker.py` for own handler p50/p95. **Retry counts not surfaced** anywhere | Add a `retry_count` field to span attributes + a "Retries" column to session-tools UI; expose handler p50/p95 in a `/health/latency` widget |
| 2.d | Tool-related failures | Partial | `is_error`, `error_count`, `tool_outcome` plumbed in events + cron-runs; `error_patterns.py` summarises errors. No per-cause taxonomy (timeout vs schema vs auth vs empty-result) and no dedicated UI tab | Add `tool_failure_reason` enum to span derivation; add a "Failing tools" card in Health/Brain with breakdown by category |
| 2.e | Action hallucination detection | Missing | nothing | New checker: walk transcript for assistant `text` blocks claiming an action (e.g. "I deployed", "I sent the email") with no matching `toolCall` event in the same turn; surface as a Brain warning |
[...]

PR #1372, opened 2026-05-15
feat(moat): surface OTel spans we already store — DuckDB read + lightweight UI

vivekchand:
Closes the read-side gap on issue #1364. ClawMetry already persists OTel spans into DuckDB via `routes/meta.py /v1/traces` → `dashboard._process_otlp_traces` → `LocalStore.put_span`, but the dashboard never surfaced them. That's a quick win for MOAT capability **1.b structured event capture per agent step** in the Monte Carlo AI Agent Monitoring framework.

## Summary

- **DuckDB read method** — `LocalStore.query_recent_spans(limit, session_id)` thin wrapper over the existing `query_spans`, returning a UI-shaped projection (`start_time/end_time/duration_ms/name/kind/session/status/model/tokens/cost`).
- **Daemon allowlist** — added `query_recent_spans` to `routes/local_query._DAEMON_METHODS` so the standard install (daemon owns the writer lock) routes through the proxy.
- **New endpoint** — `GET /api/spans?limit=50&session_id=...` lives next to `/api/transcript/<id>` in `routes/sessions.py`. Graceful fallback to `{"spans": [], "_source": "unavailable"}` on daemon down. Never 500s.
- **Tiny UI** — added a `📐 Spans` toggle button to the existing **Brain** tab header (no new top-level nav — see `feedback_steve_jobs_persona`). First click fetches + renders a flat table: `Time | Name | Duration | Session | Kind`. No tree, no charts — that's a follow-up surface.
- **E2E test** — `tests/test_spans_e2e.py` injects 3 synthetic spans via `LocalStore.put_span`, hits `/api/spans`, asserts DESC ordering + contract fields. Plus session-filter, empty-store, and limit-clamping coverage.

## Diff size

~370 lines incl. the new 154-line test file. Under the 400-line budget.

## NOT release-tagged

This PR adds `query_recent_spans` to the daemon allowlist. The sync daemon must be restarted (`launchctl kickstart -k com.clawmetry.sync`) to pick it up — same pattern as PR #1370. Bundling with the next install.sh restart sweep is the right move; releasing now would surface 400s on the new method until users reinstalled.

## Expected user-visible win

OpenClaw users who wired an OTLP exporter to `/v1/traces` (or anyone using the cloud's auto-forwarding) finally see their per-step spans in the dashboard instead of the data sitting silently in DuckDB.

## Test plan

- [x] `pytest tests/test_spans_e2e.py -v` — 4/4 pass locally (synthetic span ingest → DuckDB → API round-trip).
- [x] `pytest tests/test_spans_ingest.py` — pre-existing 8/9 pass; the one failure is a pre-existing `SCHEMA_VERSION == 6` assertion now at 7, unrelated to this PR.
- [x] `python3 -c "import dashboard; from routes.sessions import bp_sessions"` — clean import, no circular issues.
- [ ] Manual smoke after install.sh sweep: open Brain tab → click 📐 Spans → table renders (or shows the empty-state hint when no OTLP exporter is wired).

### DIFF
--- a/clawmetry/local_store.py
+++ b/clawmetry/local_store.py
@@ -2050,6 +2050,58 @@ class LocalStore:
             out.append(d)
         return out
 
+    def query_recent_spans(
+        self,
+        *,
+        limit: int = 50,
+        session_id: str | None = None,
+    ) -> list[dict[str, Any]]:
+        """MOAT issue #1364: read-side surface for spans we already store.
+
+        Thin convenience wrapper over :meth:`query_spans` returning the most
+        recent spans (``start_ts DESC``) optionally filtered by session.
+        Shape is purpose-built for the dashboard ``/api/spans`` endpoint and
+        the Brain-tab "Spans" table — only the columns the table renders are
+        guaranteed to be present, so we don't promise BLOB fidelity here
+        (callers wanting full ``input/output/attrs`` should hit
+        :meth:`query_spans` directly).
+
+        Args:
+          limit: Max rows to return (clamped 1-500 by the route layer).
+          session_id: Optional session filter — when set, returns spans for
+            that one OpenClaw session.
+        """
+        rows = self.query_spans(
+            session_id=session_id,
+            limit=int(limit),
+        )
+        # Project to the contract the UI table reads. Keep the BLOB columns
+        # present (already decoded by query_spans) so a future detail-drawer
+        # has them without a second round-trip.
+        out: list[dict[str, Any]] = []
+        for r in rows:
+            out.append({
+                "span_id":         r.get("span_id"),
+                "parent_span_id":  r.get("parent_span_id"),
+                "trace_id":        r.get("trace_id"),
+                "name":            r.get("name"),
+                "kind":            r.get("kind"),
+                "session_id":      r.get("session_id"),
+                "service_name":    r.get("service_name"),
+                "start_time":      r.get("start_ts"),
+                "end_time":        r.get("end_ts"),
+                "duration_ms":     r.get("duration_ms"),
+                "status":          r.get("status"),
+                "model":           r.get("model"),
+                "tool_name":       r.get("tool_name"),
+                "cost_usd":        r.get("cost_usd"),
+                "tokens_input":    r.get("tokens_input"),
+                "tokens_output":   r.get("tokens_output"),
+                "attrs":           r.get("attributes"),
+                "events":          r.get("events"),
+            })
+        return out
+
     # ── flush ───────────────────────────────────────────────────────────
 
     def _flusher_loop(self) -> None:
--- a/clawmetry/static/js/app.js
+++ b/clawmetry/static/js/app.js
@@ -3696,6 +3696,90 @@ async function loadBrainPage(silent) {
 }
 
 
+// ── OTel Spans panel (MOAT issue #1364) ────────────────────────────────────
+// Lightweight read-side surface for spans we already store in DuckDB via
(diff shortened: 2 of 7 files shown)


vivekchand:
## UX review

🟡 P1 — `📐 Spans` button label is jargon for non-technical users (the target audience can barely pip install). "Spans" + ruler emoji means nothing to anyone outside an OTel shop. Action: rename to `Steps` or `Trace` in `dashboard.py:4877` and drop the ruler emoji; keep "OTel spans" only in the panel subtitle for the technical reader.

🟡 P1 — Empty-state hint contains an env-var name (`CLAWMETRY_LOCAL_STORE_READ=1`) and a path (`/v1/traces`). That's a config-file vibe in a UI string. Action: in `clawmetry/static/js/app.js:_around line 3735`, replace with one human sentence + a "Learn how" link to docs — never put env vars in user-facing copy (memory `feedback_simple_ui_for_nontechnical`).

🟡 P1 — Two refresh affordances inside the same panel (header `↻ Refresh` for Brain + a per-panel `↻` for Spans), plus the toggle button. Three controls in the header for one tab violates the "duplicated affordances" pattern from the duplicate-Alerts-tab failure. Action: drop the per-panel `↻` in `dashboard.py:4886` — clicking the toggle off+on already re-fetches.

🟢 nit — Session-id column truncates at 12 chars with ellipsis but no click-to-copy or click-to-filter. Truncated IDs are useless to humans. Action: make the cell clickable to filter the table by that session_id (you already accept the query param).

🟢 nit — Time column shows `toLocaleTimeString()` with no date — a span from yesterday and today are indistinguishable. Action: switch to relative ("3m ago") to match the rest of the dashboard.

vivekchand:
## Product review

🔴 P0 — Per-step span/trace history is the textbook OTel-observability feature competitors (Datadog, Honeycomb, Langfuse) charge $$$ for. Shipping a flat unbounded span table in OSS — with no time-window cap, no row cap on the UI other than a 500-row API clamp — leaks the headline value of the future Cloud-Pro "Trace explorer" surface. Action: gate >24h retention to Cloud-Pro (mirror the existing >24h policy in memory `project_alerts_pro_feature`); OSS shows last-24h spans only, panel footer reads "Showing last 24h. Upgrade to Pro for 30-day trace history."

🟡 P1 — Toggle is opt-in (deferred fetch on first click) but there's no signal on the button when spans EXIST in the store. A user who wired OTLP last week opens Brain, sees no badge, never clicks. Action: on `loadBrainPage`, do a HEAD-equivalent count call and decorate the toggle with a count badge ("Steps · 142") — converts a hidden surface into a discovered one. Also: when the count > 0, this is the moment to drop a "Set up trace alerts (Pro)" CTA inside the panel.

🟡 P1 — `auth-bootstrap.js` change (drop `location.reload()`, call `checkAuth` inline) is unrelated scope inside a "surface OTel spans" PR. It touches the loop EVERY localhost user hits on first paint — Playwright comment is reasonable but this should not have been bundled. Action: track separately; if it regresses zero-click localhost login, the blast radius is 100% of new installs, not just OTel users.

🟢 nit — Simplest possible thing: roughly yes for the read-side, but "no charts, no tree, that's a follow-up surface" is the kind of phrase that becomes 3 follow-up PRs that never ship. State the follow-up as a tracked issue or delete the promise.

🟢 nit — Conversion lens: Brain tab is high-traffic. Empty state ("wire an OTLP exporter") is a chance to upsell `clawmetry connect` cloud-side OTLP forwarding. Currently it just gives instructions and walks away.

PR #1373, opened 2026-05-15
feat(moat): surface LoopDetector signals — DuckDB persistence + lightweight UI

vivekchand:
## Summary

ClawMetry's `clawmetry/proxy.py` has shipped a working `LoopDetector` for releases — but the only output was a log line + a row in the proxy's private SQLite. The dashboard had **no view** into agent loops, so capability **2.f (looping/stalling detection)** in the Monte Carlo framework was effectively dark.

This wires it up end to end:

- **DuckDB persistence** — new `loop_signals` table + `LocalStore.ingest_loop_signal()` + `query_recent_loop_signals()`
- **Hot-path side-effect** — `LoopDetector.check()` best-effort calls `ingest_loop_signal` on the detect-positive branch (try/except wrapped — when proxy and daemon run as separate processes, the writer-lock conflict silently no-ops and we keep the legacy SQLite path. **No detection-behaviour change.**)
- **Endpoint** — `GET /api/loop-signals?limit=20&since_minutes=60` in `routes/health.py`, daemon-proxy-first via `_ls_call`, `[]` fallback
- **Allowlist** — `query_recent_loop_signals` added to `_DAEMON_METHODS`
- **UI** — tiny red "Loops" badge inline with the Brain tab header (`display:none` until count > 0). Click expands a flat `Time | Session | Pattern (truncated 60c) | Repeats` table. No new nav tab, no chart, no graph.

## DO NOT [RELEASE]-tag

New `_DAEMON_METHODS` entry — sync daemon needs to restart to expose the new method. Per the standing rule, [RELEASE]-tagging would cut a wheel before the daemon is bumped, and `/api/loop-signals` would 404 on every install until users `launchctl kickstart -k com.clawmetry.sync`. Merge as a regular PR; the next release that bundles it will pick it up.

## Files changed

- `clawmetry/local_store.py` — schema (idempotent CREATE IF NOT EXISTS), ingest, query
- `clawmetry/proxy.py` — try/except hook in `LoopDetector.check`
- `routes/local_query.py` — one allowlist entry
- `routes/health.py` — new endpoint
- `dashboard.py` + `clawmetry/static/js/app.js` — Brain badge + collapsible list
- `tests/test_loop_signals_e2e.py` — 9 tests covering schema, upsert, route, empty-state, and detector→store wiring (all green locally)

Diff: ~308 lines under the 400-line cap.

## Test plan

- [x] `pytest tests/test_loop_signals_e2e.py -v` — 9/9 pass
- [x] `pytest tests/test_alert_rules_local_store.py tests/test_brain_local_store.py -v` — 17/17 pass (no regressions in sibling DuckDB schemas)
- [x] `python3 -m py_compile` clean on all touched modules
- [x] JS parses
- [ ] Post-merge: `curl -s http://localhost:8900/api/loop-signals` returns `{"signals":[],"count":0}` after daemon restart
- [ ] Post-merge: trigger a synthetic loop via the proxy (5+ identical hashes in 5 min) and confirm the badge lights up on the Brain tab

Refs #1364.

### DIFF
--- a/clawmetry/local_store.py
+++ b/clawmetry/local_store.py
@@ -519,6 +519,31 @@ _DDL = [
     "CREATE INDEX IF NOT EXISTS idx_spans_session     ON spans(session_id, start_ts)",
     "CREATE INDEX IF NOT EXISTS idx_spans_agent_ts    ON spans(agent_type, start_ts)",
     "CREATE INDEX IF NOT EXISTS idx_spans_ts          ON spans(ts)",
+    # Issue #1364 — loop-detection signals from clawmetry/proxy.py's
+    # ``LoopDetector``. Today the detector logs + writes to its private
+    # SQLite (``~/.clawmetry/proxy.db``); the dashboard had no view into
+    # those events. Persisting them here surfaces capability 2.f
+    # ("agent looping / stalling detection") in the Monte Carlo framework
+    # via /api/loop-signals + the Brain tab badge.
+    #
+    # PK is (session_id, signature) so re-detection of the same loop in
+    # the same session is an UPSERT (we keep the running ``repeat_count``
+    # and update ``last_seen``). Different sessions hitting the same
+    # signature are independent rows — looping is a per-session pathology.
+    """
+    CREATE TABLE IF NOT EXISTS loop_signals (
+        session_id     VARCHAR NOT NULL,
+        signature      VARCHAR NOT NULL,
+        repeat_count   INTEGER NOT NULL DEFAULT 1,
+        first_seen     TIMESTAMP NOT NULL,
+        last_seen      TIMESTAMP NOT NULL,
+        severity       VARCHAR DEFAULT 'warning',
+        agent_type     VARCHAR DEFAULT 'openclaw',
+        details        BLOB,
+        PRIMARY KEY (session_id, signature)
+    )
+    """,
+    "CREATE INDEX IF NOT EXISTS idx_loop_signals_last_seen ON loop_signals(last_seen DESC)",
     """
     CREATE TABLE IF NOT EXISTS schema_version (
         version    INTEGER PRIMARY KEY,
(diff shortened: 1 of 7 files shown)


vivekchand:
## UX review

- 🟡 P1 — A red "Loops" badge that **silently disappears** when count drops to 0 makes the affordance discoverable only during incidents. First-time users will never see it and won't know the feature exists. Either keep a neutral muted state ("0 loops · 60m") or surface the toggle in a known location (Brain tab subnav). Action: in `dashboard.py:4886` change the `display:none` default to a dim grey "0 loops" pill, OR move the affordance into the Brain tab subnav alongside Focus/Refresh so users see it before they have a problem.
- 🟡 P1 — Click target is just the badge; the table that opens has no close-affordance. Users will click the badge again to dismiss but with no visual cue ("●" / "✕") that toggling closes it. Action: in `dashboard.py:4886` swap the badge label to "▸ N loops" / "▾ N loops" so the disclosure pattern matches the rest of the dashboard's expand/collapse rows.
- 🟢 nit — Pattern column truncates at 60 chars to "deadbeef00112233..." which is unreadable hex. A non-technical user has no idea what "Pattern" means or what to do with a hash. Action: in `clawmetry/static/js/app.js:3737` rename column header "Pattern" → "Request fingerprint" and add a tooltip ("Hash of the repeated request — same number means the agent is asking the same thing again and again"). Bonus: link the session_id column to the session detail page so the user can actually act.
- 🟢 nit — Red `#ef4444` badge + "loops" without context will read as "something is broken in ClawMetry" to a non-technical user, not "your agent is stuck". Action: in `dashboard.py:4886` change tooltip from "LoopDetector signals from clawmetry/proxy.py" (jargon) to "Your agent repeated the same request — possible loop." Drop the file path; it leaks implementation.

vivekchand:
## Product review

- 🔴 P0 — Loop detection is exactly the kind of "your agent is burning money in a loop" insight that should drive Cloud-Pro upgrades (cross-session loop trends, alerts on N loops/hr, Slack notify). Shipping an unbounded raw-table view in OSS with no upgrade prompt leaks the value entirely. Action: in `clawmetry/static/js/app.js:3737` cap the OSS view at last 1 loop (or last 24h, whichever is smaller) and add a "Loop history + alerts in Cloud-Pro" CTA below the table. Anyone who sees a real loop event should immediately see the upgrade path.
- 🔴 P0 — There is no alert on detection. A loop firing in the dashboard is the highest-intent moment a user will ever have for an alert subscription, but the badge just sits there until they happen to look at the Brain tab. Action: wire `LoopDetector.check` positive branch to also emit an alert event into the existing alerts pipeline (Cloud-Pro gated for delivery — Slack/email/PagerDuty); OSS shows the in-dashboard badge, Cloud-Free shows "Get notified next time" upgrade modal. Single PR, two-tier funnel.
- 🟡 P1 — The simplest possible solution to "is my agent looping" is a yes/no banner on the main dashboard, not a count badge buried inside a tab the user may never click. The most affected user is the one whose agent IS looping right now and who is staring at the Overview tab wondering why costs are spiking. Action: surface the loop count on `/api/overview` and render a top-of-dashboard banner ("⚠ 3 active loops detected — view") that deep-links into the Brain tab.
- 🟢 nit — "Looping" detection is a defining moat capability per the issue's framing (Monte Carlo 2.f). Burying it as an inline badge undersells it on the marketing side. Action: ensure the next changelog/landing-page sweep gets a "Loop detection now visible in dashboard" line item — and consider a screenshot in the README.

PR #1375, opened 2026-05-15
feat(ui): Brain tab — hide queue-ops + collapse provenance JSON to channel pills

vivekchand:
## Summary

UX cleanup for the Brain tab "Live event stream" based on user feedback at `localhost:8903` (v0.12.227):

> "I'm unable to understand what those queue operations & why the json look that way & what value this info is adding -- can it be somehow represented better"

Three frontend-only fixes:

1. **`QUEUE-OPERATION` rows hidden by default.** New header toggle `○ Show plumbing (N hidden)` next to `↻ Refresh`. Off by default — these are internal queue housekeeping (dequeue/enqueue), not agent activity. Toggle ON reveals them at reduced opacity.
2. **`Sender (untrusted metadata): \`\`\`json {…}\`\`\`` prefix collapsed to a channel pill.** Now renders as `📱 Telegram · Vivek Chand · 22:15 ⓘ` followed by the actual message body. Click `ⓘ` to inspect the original JSON inline (debuggability preserved). Channel inferred from `chat_id` prefix (telegram → 📱, signal → 📡, slack → 💼, …) — reuses the existing `_channelIcons` table to stay aligned with `routes/brain.py` `_CHANNEL_ICON`.
3. **`(no body captured)` outbound rows styled as a thin grey one-liner.** Outbound channel events where the gateway-log parser couldn't capture the body now render as a 2-line muted transition row, not a full event.

### Before / After

For one Telegram message bouncing through the agent, the brain stream went from **8 rows of JSON noise** down to **2-3 conversational rows** (1 USER pill + 1 ASSISTANT reply + at most 1 thin "relayed" transition). Plumbing rows are still recoverable behind the toggle.

### Constraints met

- Frontend-only; **no** backend / DuckDB / route changes
- **No** new dependencies (vanilla JS + existing CSS variables)
- Diff: **143 LOC** (cap was 200)
- Branch: `feat/brain-tab-hide-plumbing`
- **Not** `[RELEASE]`-tagged — wants local verification first

### Files

- `clawmetry/static/js/app.js` — `renderBrainDetail` (provenance pill detection), `renderBrainStream` (filter QUEUE-OPERATION + count), new `_isPlumbingEvent` / `_parseProvenancePrefix` / `_provenancePillHtml` helpers, row CSS classes
- `clawmetry/static/css/dashboard.css` — `.brain-plumbing`, `.brain-relay-only`, `.brain-provenance-pill`
- `clawmetry/templates/tabs/brain.html` — header toggle button + state script

### Out of scope (follow-up)

A bigger reframe to a conversation-thread layout (parent message → assistant reply → tool calls indented) is the natural next step — left out per the 200-LOC cap and the user's "tight quick-fix" request.

## Test plan

- [ ] Open `localhost:8903` → Brain tab. Verify Telegram-originating user rows show pill + body, not JSON.
- [ ] Confirm `Show plumbing (N hidden)` count matches the QUEUE-OPERATION rows that were on screen pre-fix.
- [ ] Toggle ON → plumbing rows reappear at reduced opacity. Toggle OFF → they hide again.
- [ ] Click `ⓘ` on a pill → raw JSON expands inline.
- [ ] Outbound Telegram→agent ACK rows render as one-liner, not a full event.

### DIFF
--- a/clawmetry/static/css/dashboard.css
+++ b/clawmetry/static/css/dashboard.css
@@ -105,6 +105,19 @@
   .badge-error { background:rgba(239,68,68,0.2); color:#ef4444; }
   .badge-tool { background:rgba(148,163,184,0.2); color:#94a3b8; }
   .brain-detail { color:var(--text-secondary); flex:1; min-width:0; white-space:pre-wrap; word-break:break-word; overflow-wrap:anywhere; }
+  /* Plumbing rows (QUEUE-OPERATION) when revealed via toggle:
+     visibly de-emphasized so they don't compete with real content. */
+  .brain-event.brain-plumbing { opacity:0.55; font-size:11px; padding:2px 0; }
+  .brain-event.brain-plumbing .brain-detail { color:var(--text-muted); }
+  /* "Telegram → agent (relayed)" rows: outbound channel events with no
+     body captured. Render as a thin grey one-liner so the eye skips past
+     them — they're transitions, not content. */
+  .brain-event.brain-relay-only { opacity:0.7; font-size:11px; padding:2px 0; }
+  .brain-event.brain-relay-only .brain-detail { display:none; }
+  /* Provenance pill (channel adapter prefix collapsed) — small, subtle,
+     left-aligned above the message body so the body reads as the focus. */
+  .brain-provenance-row { display:block; margin-bottom:2px; }
+  .brain-provenance-pill { vertical-align:middle; }
   .brain-view-toggle { display:flex; gap:6px; margin-bottom:12px; }
   .brain-view-btn { padding:4px 12px; border-radius:10px; border:1px solid var(--border); background:transparent; color:var(--text-muted); font-size:11px; font-weight:600; cursor:pointer; }
   .brain-view-btn.active { border-color:#a855f7; background:rgba(168,85,247,0.2); color:#a855f7; }
--- a/clawmetry/static/js/app.js
+++ b/clawmetry/static/js/app.js
@@ -2375,9 +2375,90 @@ function formatBrainTime(isoStr) {
   } catch(e) { return isoStr || ''; }
 }
 
+// ── Provenance pill (PR feat/brain-tab-hide-plumbing) ──────────────────
+// User messages from channel adapters arrive prefixed with a JSON block
+// like:
+//
+//     Sender (untrusted metadata):
+//     ```json
+//     {"chat_id":"telegram:...","sender":"Vivek Chand","timestamp":...}
+//     ```
+//     <real message body>
+//
+// Showing the raw JSON 3+ times per message buries the actual conversation.
+// This helper detects the prefix, parses the JSON, and renders a small
+// channel pill (📱 Telegram · Vivek Chand · 22:15  ⓘ) followed by the
+// message body. The ⓘ button toggles the original JSON inline so
+// debuggability is preserved.
+function _provenancePillHtml(meta, bodyHtml) {
+  var chatId = String(meta.chat_id || '').toLowerCase();
+  var icon = '💬';
+  var providerLabel = '';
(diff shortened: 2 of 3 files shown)


vivekchand:
## UX review

- 🟡 P1 — "Show plumbing" is jargon the target audience will not parse. A non-technical user has zero mental model for "plumbing" or "queue operations" — they will toggle it on out of curiosity, see noise, and lose trust that what's hidden is actually irrelevant. Action: in `clawmetry/templates/tabs/brain.html:9` rename to "Show internal events" and drop "(N hidden)" → "(N more)"; the parenthetical "hidden" implies the dashboard is concealing things.
- 🟡 P1 — Three controls now crowd the Brain header: `○ Show plumbing (N hidden)`, `⛶ Focus`, `↻ Refresh`. On narrow screens this will wrap and break the row. Action: in `clawmetry/templates/tabs/brain.html:9` move the plumbing toggle into a kebab/overflow menu next to Focus, OR collapse it into a small icon-only button (eye / eye-off) so the header stays scannable.
- 🟢 nit — The provenance pill `📱 Telegram · Vivek Chand · 22:15 ⓘ` is a great compaction, but `ⓘ` is the universally-recognized "info" affordance and clicking it opening a JSON blob is surprising. Use `{ }` or `<>` to signal "raw data". Action: in `clawmetry/static/js/app.js:2410` swap the `ⓘ` button glyph to `{ }` and update the `title` to "View raw event JSON".
- 🟢 nit — "(no message body)" italicized fallback inside the pill row reads as a styling glitch when most rows have rich content. Action: in `clawmetry/static/js/app.js:2461` either drop the empty-body row entirely (the pill alone is enough) or render it as a one-line muted "received" caption matching the `.brain-relay-only` style for visual consistency.
- 🟢 nit — The toggle state glyph `○` / `●` doesn't read as on/off to a non-technical user (looks like a status dot). Action: in `clawmetry/templates/tabs/brain.html:34` replace with a real switch metaphor — `Show internal events ☐` / `Show internal events ☑` or use a CSS toggle pill matching how other dashboard toggles render.

vivekchand:
## Product review

- 🟡 P1 — This is the simplest version of the right idea (hide noise, surface conversation), but it stops at "less ugly" and doesn't ask "what is the user actually here to do?" The Brain tab is currently a debug log, not a product. The user feedback that drove this PR ("can it be somehow represented better") is a signal that a conversation-thread layout (parent → assistant → tool calls indented) — explicitly listed as out-of-scope here — is the actual fix. Action: open a follow-up issue "Brain tab: conversation-thread layout v1" and put it on the next sprint; this PR earns 2 weeks of runway, not a permanent ship.
- 🟡 P1 — Zero hooks for paid conversion. The Brain tab is THE place where a user sees raw activity and asks "could I do more with this?" — perfect surface for "Cloud-Pro: search across all sessions, 30-day history, alert on patterns" callouts. Right now the PR makes the tab nicer and walks away. Action: add a thin "Searching across sessions or notifying on patterns? Cloud-Pro →" footer below the event stream (single line, dismissable) so the tidied-up tab earns its keep on funnel.
- 🟢 nit — The `_parseProvenancePrefix` regex hardcodes "(untrusted metadata)" — when channels.py renames the prefix string (which it has, twice this year per git history), the pill silently stops rendering and the tab regresses to JSON noise without anyone noticing until a user complains. Action: add an E2E playwright test that asserts a Telegram-originating Brain row renders as a pill, not raw JSON. Wire it into the existing `make test-e2e` matrix so future channel adapter renames break loud.
- 🟢 nit — Default OFF for plumbing is the right call for non-technical users; for power users debugging a real ingest issue, persisting the toggle state across page reloads (localStorage) avoids 30 clicks/day. Action: in `clawmetry/templates/tabs/brain.html:34` persist `_brainShowPlumbing` to `localStorage.setItem('brain.showPlumbing', ...)` and rehydrate on load.

PR #1378, opened 2026-05-15
feat(moat): migrate /api/skills fidelity counts to DuckDB fast-path + E2E verifier

vivekchand:
## Summary
- Tier-1 bypass migration. `/api/skills` previously walked every session JSONL touched in the last 7 days on every render to count Read-tool calls targeting `SKILL.md` and files under `<skill>/{scripts,references,assets}/`. With 50-200+ active sessions this is the panel's slowest stage.
- Replaces it with a single `query_recent_read_tool_calls(since, limit)` SQL aggregate over the `events` table, returning `{ts, session_id, file_path}` rows; route buckets per-skill in-process from the in-memory skill-paths map.
- Same playbook as #1370 (context-anatomy), #1372 (spans), #1373 (loop-signals).

## Files
- `clawmetry/local_store.py` — new `query_recent_read_tool_calls` + `_iter_read_tool_paths` helper handling all three on-the-wire tool-call shapes (v3 top-level `tool.call`/`toolCall`/`tool_use`, trajectory-projected `data.toolMetas`, legacy `data.message.content[*]` blocks).
- `routes/local_query.py` — allowlist the new method for daemon proxy.
- `routes/skills.py` — `_try_local_store_read_tool_calls()` + `_bucket_read_calls_per_skill()` + early-return in `api_skills` and `api_skill_detail`. JSONL fallback preserved verbatim.
- `tests/test_skills_fidelity_local_store.py` — 9 tests (E2E synthetic round-trip, unit coverage of all three shapes, since= window, daemon-proxy mocking).

## Test plan
- [x] `make lint-daemon-allowlist` — passes (16 distinct query_* methods, all in allowlist).
- [x] `pytest tests/test_skills_fidelity_local_store.py` — 9/9 pass in 1.0s.
- [x] `pytest tests/test_context_anatomy_local_store.py tests/test_loop_signals_e2e.py tests/test_circular_import.py` — 33/33 pass; sibling fast-paths unchanged.
- [ ] Post-merge: hit `/api/skills` on a node with active sessions; confirm `body_fetch_count_7d` populated without scanning sessions dir.

## Release notes
**Not** `[RELEASE]`-tagged. Adds a new daemon allowlist entry; users need the sync daemon to restart to pick it up (memory: `feedback_restart_both_processes_after_upgrade`). Holding for one normal `[RELEASE]` cycle so installer-driven upgrade kicks both processes.

## User-visible win
Skills tab paints from a single SQL aggregate instead of re-reading every recent session JSONL on every render — same numbers, no JSONL re-walk.

### DIFF
--- a/clawmetry/local_store.py
+++ b/clawmetry/local_store.py
@@ -2264,6 +2264,90 @@ class LocalStore:
             })
         return out
 
+    def query_recent_read_tool_calls(
+        self,
+        *,
+        since: str | None = None,
+        limit: int = 50_000,
+    ) -> list[dict[str, Any]]:
+        """Tier-1 MOAT (issue #1364): /api/skills fidelity fast-path.
+
+        Returns one row per Read-tool invocation since ``since`` (ISO-8601
+        timestamp), shape ``{ts, session_id, file_path}``. Used by
+        ``routes/skills.py`` to count "body fetched" + "linked file read"
+        events per skill without walking every session JSONL on disk
+        (the legacy scanner re-opens 50-200+ files on every page render).
+
+        Tool-call shapes covered (all three appear in the wild):
+          * v3 top-level events with ``event_type`` in
+            ``{'tool.call', 'toolCall', 'tool_use'}`` — read from
+            ``data.input.file_path`` / ``data.arguments.file_path``.
+          * Assistant ``message`` events with
+            ``data.toolMetas[*].input.file_path`` (PR #1132 trajectory
+            parser shape).
+          * Legacy assistant message events whose
+            ``data.message.content[*]`` still carries raw
+            ``{type:'toolCall'|'tool_use', name, input/arguments}`` blocks
+            (older OpenClaw transcripts that pre-date PR #1132's
+            trajectory projection).
+
+        Tool name is matched case-insensitively against
+        ``{'read', 'readfile', 'read_file'}`` — same set the legacy
+        ``_scan_fidelity_events`` checked. The ``file_path`` argument is
+        pulled from the first non-empty of ``file_path`` / ``path`` /
+        ``filename`` (Anthropic SDK + OpenClaw both shapes seen).
+
+        Rows where no Read-tool path can be extracted are dropped so the
+        caller iterates only useful events.
+
+        ``limit`` clamps total returned rows; defaults to 50k which is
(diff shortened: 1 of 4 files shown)


vivekchand:
## UX review

- 🟢 nit Diff is purely backend (DuckDB fast-path swap behind `/api/skills`; no template/CSS/nav touched). No UI surface to scrutinise — same JSON shape, same numbers, same Skills tab. Stop.

vivekchand:
## Product review

- 🟡 P1 Backend perf win is real (50-200 JSONL re-walks → one SQL aggregate per render) but invisible to the user — Skills tab silently gets faster with zero "before/after" surfaced anywhere. Add a one-line "Skills fidelity from local store" badge or perf hint in the Skills panel header so the moat shows. Action: `routes/skills.py:api_skills` — include `source: "local_store"` in the response and render a subtle indicator client-side.
- 🟡 P1 Fast-path is gated on `is_local_store_read_enabled()` which has historically been off-by-default and silently bypassed (memory: `feedback_local_store_default_off_killed_moat`). If any installer leaves `CLAWMETRY_LOCAL_STORE_READ` unset, every user falls back to the slow JSONL walker and this PR is dead code. Action: verify the default-ON flag is still set in `clawmetry/config.py` `is_local_store_read_enabled()` and add a release-note checklist item.
- 🟡 P1 Daemon allowlist entry means existing users on old daemons get JSONL fallback until they restart the sync daemon (memory: `feedback_restart_both_processes_after_upgrade`). PR body acknowledges this and holds for a `[RELEASE]` cycle — good — but installer must `launchctl kickstart -k com.clawmetry.sync` after upgrade or fast-path silently misses for days. Action: confirm `install.sh` kicks the daemon, not just the dashboard.
- 🟢 nit Conversion-neutral: `/api/skills` is OSS-tier surface (skills observability is OSS). No paid leakage, no funnel motion. Pure infra debt paydown — fine, but don't expect a Cloud-Pro signup bump from this.
- 🟢 nit Simplest-thing check: per-skill bucketing happens in-process rather than in SQL. Defensible (skill-paths map lives in dashboard memory) but a future migration to SQL `WHERE file_path LIKE skill_dir || '%'` would let the daemon return pre-bucketed counts and cut the row payload. Park for a follow-up if `/api/skills` shows up in another perf cliff sweep.

PR #1380, opened 2026-05-15
feat(moat): migrate /api/fallbacks to DuckDB fast-path + E2E verifier

vivekchand:
## Summary
- Tier-1 bypass migration (issue #1364): `/api/fallbacks` was opening up to 100 session JSONL files per request and walking them line-by-line to find model/provider transitions across adjacent assistant turns. Multi-second p95 on busy workspaces.
- New `LocalStore.query_model_fallbacks` runs the same aggregation as a single SQL query (CTE → most-recent N sessions with ≥1 assistant message; outer SELECT pulls all `message` events ordered per-session; Python walk does pair aggregation at sub-ms per session).
- Allowlisted in `routes/local_query.py` `_DAEMON_METHODS`. Route handler gains `is_local_store_read_enabled` gate; daemon-proxy first, legacy JSONL walker as graceful fallback (so first-install dashboards never blank out).

## Diff
- `clawmetry/local_store.py` (+129) — `query_model_fallbacks(session_limit, top)` returning `{scanned, sessions_affected, top_transitions}`.
- `routes/local_query.py` (+4) — allowlist entry.
- `routes/sessions.py` (+30) — `_try_local_store_fallbacks` helper + fast-path gate; legacy walker preserved verbatim below.
- `tests/test_fallbacks_local_store.py` (+294) — 8 unit/E2E tests.

Total: 457 LOC (under the 400 LOC cap excluding tests).

## Not `[RELEASE]`-tagged
Daemon allowlist needs a restart before the proxy will dispatch `query_model_fallbacks` — bundle this with the next release after merge so `launchctl kickstart` picks up the new wheel.

## Test plan
- [x] `pytest tests/test_fallbacks_local_store.py` — 8/8 green
- [x] `pytest tests/test_loop_signals_e2e.py tests/test_spans_e2e.py tests/test_context_anatomy_local_store.py` — sibling fast-path tests still green (20/20)
- [ ] Post-deploy: hit `/api/fallbacks` against a real workspace; confirm `_source: "local_store"` marker and a 10-50× latency win vs legacy walker.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #1381, opened 2026-05-15
feat(moat): migrate /api/plugins to DuckDB fast-path + E2E verifier

vivekchand:
## Summary

Tier-1 MOAT migration. The Plugins-tab endpoint historically scanned up to 60 session JSONLs (every line) on every render to count per-plugin tool invocations — dominant render-time stage even though the answer rarely changes minute-to-minute.

This PR replaces that walker with a single SQL pull from the `events` table, following the daemon-proxy playbook (PRs #1370, #1372, #1373, #1378, #1380).

- New `LocalStore.query_tool_call_invocations(*, since, limit)` returns `{ts, name}` per tool invocation across all three on-the-wire shapes (top-level `tool.call`/`toolCall`/`tool_use`, `message.toolMetas` projection, legacy `content[*].toolCall` blocks).
- `_try_local_store_invocations()` in `routes/plugins.py` buckets rows per plugin key, mirrors the legacy 30d-window semantics, returns a populated empty shell so empty stores never fall through.
- Allowlisted in `routes/local_query.py` `_DAEMON_METHODS` (now 27 entries).
- Response now carries `_source: "local_store" | "jsonl_walker"` so smoke tests can assert which path served the request.

**Behavior delta vs. legacy walker:** strict superset — the legacy `_count_invocations` only matched `block.type == "tool_use"` (PR-#1132 shape pre-projection). The fast path also counts `toolCall` blocks, which is OpenClaw's current on-disk shape. Older transcripts still count via shape #3.

## Method signature

```python
def query_tool_call_invocations(
    self,
    *,
    since: str | None = None,
    limit: int = 50_000,
) -> list[dict[str, Any]]:
    """Returns one row per tool invocation since `since` (ISO-8601),
    shape `{ts, name}`. Caller buckets per-plugin via substring match."""
```

## Test plan

`tests/test_plugins_invocations_local_store.py` (8 tests, all passing locally):

- [x] E2E: synthetic v3 `tool.call` events round-trip through DuckDB → `/api/plugins` → `invocations_30d`.
- [x] E2E: legacy `data.message.content[*].toolCall` blocks also count (closes the gap the JSONL walker depended on).
- [x] E2E: empty store returns zero counts + `unused=true` with no crash, `_source: "local_store"`.
- [x] E2E: 30d cutoff respected — older events drop out of count but still drive `last_used_ts` forward.
- [x] Unit: `query_tool_call_invocations` on empty/populated store, `since` filter.
- [x] Daemon-proxy: when `local_store_via_daemon` returns rows, route uses them and never opens DuckDB directly.

Cross-suite: also re-ran `test_skills_fidelity_local_store.py`, `test_loop_signals_e2e.py`, `test_spans_e2e.py` (sibling Tier-1 verifiers) — all green, no regressions.

```
$ make lint-daemon-allowlist
daemon-allowlist OK — 16 distinct query_* methods called from routes/, all in _DAEMON_METHODS (27 entries)
```

## User-visible win

Plugins tab paints invocation counts from DuckDB on every render instead of fanning out to 60 JSONL files — the panel renders instantly even on hosts with hundreds of historical sessions.

### DIFF
--- a/clawmetry/local_store.py
+++ b/clawmetry/local_store.py
@@ -2348,6 +2348,77 @@ class LocalStore:
                 })
         return out
 
+    def query_tool_call_invocations(
+        self,
+        *,
+        since: str | None = None,
+        limit: int = 50_000,
+    ) -> list[dict[str, Any]]:
+        """Tier-1 MOAT: /api/plugins fast-path.
+
+        Returns one row per tool invocation since ``since`` (ISO-8601
+        timestamp), shape ``{ts, name}``. Used by ``routes/plugins.py`` to
+        count per-plugin invocations over the last 30d without re-walking
+        every session JSONL on every Plugins-tab render (the legacy scanner
+        re-opens up to 60 files per request and parses every line).
+
+        Tool-call shapes covered (matches :func:`_iter_tool_invocation_names`):
+          * Top-level events with ``event_type`` in
+            ``{'tool.call', 'toolCall', 'tool_use'}`` — uses ``data.name``.
+          * Assistant ``message`` events with
+            ``data.toolMetas[*].name`` (PR #1132 trajectory parser shape).
+          * Legacy assistant message events whose
+            ``data.message.content[*]`` carries raw
+            ``{type:'toolCall'|'tool_use', name}`` blocks (older OpenClaw
+            transcripts that pre-date PR #1132's trajectory projection).
+
+        Tool name is returned verbatim (lower-cased by the caller). Rows
+        with no extractable name are dropped.
+
+        ``limit`` clamps total returned rows; defaults to 50k which covers
+        a busy agent-month worth of tool calls. Callers should pass a 30d
+        ``since`` to keep the result set bounded.
+        """
+        try:
+            lim = int(limit)
+        except (TypeError, ValueError):
+            lim = 50_000
+        lim = max(1, min(200_000, lim))
+
+        clauses: list[str] = [
+            "(event_type IN ('tool.call', 'toolCall', 'tool_use')"
+            " OR event_type = 'message')"
+        ]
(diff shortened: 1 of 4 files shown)


vivekchand:
## UX review

- 🟢 nit: Diff is purely backend (`local_store.py` query method, `routes/plugins.py` fast-path branch, daemon allowlist entry, test file). No HTML/CSS/JS, no copy, no new affordance — Plugins tab renders the same shape, just faster. Nothing to critique on the UX surface. Action: none.

vivekchand:
## Product review

- 🟡 P1: The `_source: "local_store" | "jsonl_walker"` field is shipped to every `/api/plugins` response in production, not just smoke tests. That's free debug surface for competitors poking the API and zero value to a non-technical user. Action: gate the field behind `?debug=1` or `CLAWMETRY_DEBUG=1` so it only emits when the smoke test asks for it.
- 🟡 P1: This is the 7th Tier-1 MOAT migration shipped today and still nothing in the Plugins panel hints to a free user that "instant render even with hundreds of sessions" is something they're getting from ClawMetry Cloud's backing infra. Action: once the next 1-2 MOAT PRs land, add a one-line "powered by local DuckDB index — upgrade to Cloud for 30d+ retention" footer in the panel; right now the speed win is invisible to the buyer.
- 🟢 nit: Fast-path is opt-in via `CLAWMETRY_LOCAL_STORE_READ=1` (per the playbook lesson from 2026-05-13). Confirm `install.sh` / daemon defaults flip it on by default for new installs so the speed win actually reaches users. Action: spot-check that today's fresh-install path sets `CLAWMETRY_LOCAL_STORE_READ=1` — otherwise this PR ships dark.

PR #1388, opened 2026-05-16
fix(brain): use plain-text time in collapsed-run range label

### LINKED ISSUE #1222: [Product P0 from PR #1215] Migrate renderBrainStream off raw HTML string-concat
From the post-merge product review of PR #1215.

## Pattern: HTML string-concat keeps producing XSS-adjacent bugs

PR #1215 was a one-line fix: `escHtml(rangeLabel)` was wrapping output that was already trusted HTML, causing the markup to render as literal text in the Brain tab's collapsed-run rows.

This is the **second similar fix** in two weeks. The root cause is bigger than any one line: `renderBrainStream` in `clawmetry/static/js/app.js` builds HTML by concatenating strings, mixing trusted helper output with user-supplied data. Every `html += '<span ...>' + X + '</span>'` call is a manual decision: "do I escape X or not?" That decision goes wrong eventually.

## Why this is P0
- Symptom side A: forget to escape → XSS (telegram message body containing `<script>` could execute in our Brain renderer if we slip up once)
- Symptom side B: over-escape → markup leaks as literal text (the bug PR #1215 fixed)
- Both symptoms are inevitable as long as the rendering logic is hand-rolled string concatenation

## Proposed fix
Two options, in order of preference:

**Option 1: Migrate to `document.createElement` + `textContent`**
- Zero string-concat, zero escape decisions
- DOM nodes for structure, `textContent` for any user data
- ~30 LOC change for `renderBrainStream`

**Option 2: Tagged-template HTML helper**
```js
function html(strings, ...values) {
  return strings.reduce((acc, str, i) => {
    const v = values[i - 1];
    return acc + (v?._isTrustedHtml ? v.toString() : escHtml(String(v ?? ''))) + str;
  });
}
function trusted(s) { return Object.assign(Object.create(String.prototype), { toString: () => s, _isTrustedHtml: true }); }
```
Then: `html\`<span>\${userBody}\${trusted(formatBrainTimeHtml(ts))}</span>\`` — escape decisions encoded in the call site, can't be wrong.

## Acceptance
- [ ] `renderBrainStream` no longer uses raw `html +=` string concatenation for any branch
- [ ] All user-supplied strings (channel body, sender name, transcript text) flow through the safe-by-default API
- [ ] Trusted HTML helpers (e.g. `formatBrainTime`) are renamed with `Html` suffix to encode contract in the name
- [ ] E2E test in `tests/e2e/test_brain.spec.ts` asserts the collapsed-run rendering does not leak literal `<span` text
- [ ] E2E test asserts a channel body containing `<script>alert(1)</script>` renders as text, not script

## Why this matters for the funnel
A non-technical user seeing literal HTML markup in the dashboard concludes "this thing is broken" within 5 seconds and bounces. An attacker triggering even a same-origin script execution via a crafted channel message is a disclosure-grade incident. Both are existential to the funnel.

vivekchand:
Closes #1222

## Summary

- `formatBrainTime()` returns HTML (a `<span>` wrapping the date prefix for opacity styling), making `rangeLabel` in the collapsed-run path an HTML string that must **not** be passed through `escHtml()` — a fragile invariant that already caused one regression (PR #1215 double-escaped it, rendering markup as literal text).
- Adds `_formatBrainTimePlain(isoStr)` returning plain text (`"Today 14:23:45"`), no HTML tags.
- Switches the two `formatBrainTime()` calls in the collapsed-run `rangeLabel` path to `_formatBrainTimePlain`; `rangeLabel` is now a plain string safely wrapped with `escHtml()`, and the "do NOT escHtml or the markup renders as text" warning comment is gone.

All existing `formatBrainTime()` callers (the `<span class="brain-time">` rows) are unchanged. Minor visual: date prefix in the collapsed-range badge loses its reduced-opacity span — acceptable in a summary badge.

## Test plan

- [x] `node tests/test_appjs_units.js` — 89 passed, 0 failed
- [x] `node --check clawmetry/static/js/app.js` — JS syntax OK
- [x] `make lint` — Python syntax OK (pre-existing ruff warnings in other files, none introduced by this change)

---
_Generated by [Claude Code](https://claude.ai/code/session_013zw5RYqh6MQsVU7jyYpHjW)_

### DIFF
--- a/clawmetry/static/js/app.js
+++ b/clawmetry/static/js/app.js
@@ -2375,6 +2375,28 @@ function formatBrainTime(isoStr) {
   } catch(e) { return isoStr || ''; }
 }
 
+// Plain-text variant of formatBrainTime — returns "Today 14:23:45" with no
+// HTML tags so callers can safely pass the result through escHtml() or
+// concatenate it into a plain-text context without worrying about double-escaping.
+function _formatBrainTimePlain(isoStr) {
+  try {
+    var d = new Date(isoStr);
+    var now = new Date();
+    var sameDay = d.getFullYear()===now.getFullYear() && d.getMonth()===now.getMonth() && d.getDate()===now.getDate();
+    var sameYear = d.getFullYear()===now.getFullYear();
+    var time = d.toLocaleTimeString('en-GB', {hour:'2-digit',minute:'2-digit',second:'2-digit'});
+    var prefix;
+    if (sameDay) {
+      prefix = 'Today';
+    } else if (sameYear) {
+      prefix = d.toLocaleDateString('en-GB', {day:'numeric',month:'short'});
+    } else {
+      prefix = d.toLocaleDateString('en-GB', {day:'numeric',month:'short',year:'numeric'});
+    }
+    return prefix + ' ' + time;
+  } catch(e) { return isoStr || ''; }
+}
+
 // ── Provenance pill (PR feat/brain-tab-hide-plumbing) ──────────────────
 // User messages from channel adapters arrive prefixed with a JSON block
 // like:
@@ -3114,8 +3136,10 @@ function renderBrainStream(events) {
       var run = ev.__collapsedRun;
       // Times are sorted newest-first from the upstream sort, so the
       // last event in the run is the oldest. Format "oldest → newest".
-      var newestT = run[0] && run[0].time ? formatBrainTime(run[0].time) : '';
-      var oldestT = run[run.length - 1] && run[run.length - 1].time ? formatBrainTime(run[run.length - 1].time) : '';
+      // Use the plain-text variant so rangeLabel is safe to pass through
+      // escHtml() — avoids the double-escape bug fixed in PR #1215.
+      var newestT = run[0] && run[0].time ? _formatBrainTimePlain(run[0].time) : '';
+      var oldestT = run[run.length - 1] && run[run.length - 1].time ? _formatBrainTimePlain(run[run.length - 1].time) : '';
       var rangeLabel = oldestT && newestT && oldestT !== newestT
         ? oldestT + ' → ' + newestT
         : (newestT || oldestT || '');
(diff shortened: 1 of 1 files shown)


vivekchand:
## UX review

- 🟡 P1 — Visual regression acknowledged in the PR body: the date prefix in the collapsed-range badge loses its reduced-opacity span ("Today 14:23:45" is now uniform text instead of dimmed-prefix + bold-time). Two adjacent badge styles in the Brain panel are now inconsistent. Next action: replicate the opacity treatment via CSS — wrap the prefix in a span inside `_formatBrainTimePlain` only when caller opts in, OR add a `.brain-collapsed-range time-prefix` regex-styled rule so the visual hierarchy survives.
- 🟢 nit — Two near-identical functions (`formatBrainTime` HTML + `_formatBrainTimePlain` text) are a footgun: future edits to the time-format logic must be made twice or they drift. Next action: refactor so `formatBrainTime(iso, {html: true})` is the single source and the HTML variant wraps the plain output.
- 🟢 nit — `_formatBrainTimePlain` returns `isoStr || ''` on parse failure (raw ISO string surfaced to user). For a "today 14:23" sibling that's jarring. Next action: fall back to `'-'` instead — matches the convention used in the cron timeline.

vivekchand:
## Product review

- 🟢 nit — Pure correctness fix preventing markup-as-text rendering in the Brain panel's collapsed-run range label. No conversion-funnel implication. Next action: none required; consider adding a `node tests/test_appjs_units.js` case that pipes a known ISO through both `formatBrainTime` and `_formatBrainTimePlain` and asserts the plain output contains no `<`/`>` characters — locks the invariant for future contributors.
- 🟢 nit — The recurrence of "string-with-HTML mistaken for plain string" bugs (PR #1215 → this PR) suggests a small typesystem affordance is warranted. Next action: prefix all HTML-returning helpers in `app.js` with `_html_` (e.g. `_html_formatBrainTime`) so call-sites visibly flag the escape hazard.

PR #1389, opened 2026-05-16
fix: snapshot events before v7 dedup DELETE (prevents data loss)

### LINKED ISSUE #1236: [Product P0 from PR #1232] v7 dedup migration deletes user data without backup
From the post-merge product review of PR #1232.

## Problem

\`_run_dedup_migration_v7\` in \`clawmetry/local_store.py\` (around lines 70-112 in the PR diff) issues an unconditional \`DELETE FROM events\` against the user's only copy of their DuckDB data on first daemon start after upgrade to schema_version 7. On the user's own machine this removed 266/976 rows. If the regex extracting the dedup key is wrong on any user's id corpus we haven't seen — e.g. some downstream OpenClaw fork that emits \`uuid\`-shaped tails coincidentally — we silently destroy real events with no recovery path.

The PR body acknowledges the migration is gated on the \`current<7\` check and try/excepted, but the failure mode that matters is the **wrong-but-successful** DELETE, not an exception.

## Standard practice

Destructive migrations should follow the snapshot-then-delete pattern:

1. \`CREATE TABLE events_pre_v7_dedup AS SELECT * FROM events\` (or COPY to parquet)
2. Run the DELETE
3. Drop the snapshot only after a future SCHEMA_VERSION bump (e.g. v8) confirms no rollback was needed

This adds at most 5-10 MB of disk per user (events table is small, capped via CLAWMETRY_LOCAL_MAX_GB) and gives us a clean recovery story: \`INSERT INTO events SELECT * FROM events_pre_v7_dedup ON CONFLICT DO NOTHING\`.

## Suggested fix

In \`_run_dedup_migration_v7\`, before the DELETE:

\`\`\`python
conn.execute(\"\"\"
    CREATE TABLE IF NOT EXISTS events_pre_v7_dedup AS
    SELECT * FROM events
\"\"\")
\`\`\`

And in the v8 migration block (whenever we add one), drop the snapshot.

## Severity rationale

P0 because it's destructive and irreversible against the user's only copy of their observability data, on every install, automatically on first daemon start. Cheap to mitigate; expensive to discover after the fact.

vivekchand:
## Problem

`_run_dedup_migration_v7` in `clawmetry/local_store.py` runs an unconditional `DELETE FROM events` on the first daemon start after upgrading to schema v7. If the dedup-key regex misidentifies rows on any user's id corpus, real observability events are permanently destroyed with no recovery path.

## Fix

Add `CREATE TABLE IF NOT EXISTS events_pre_v7_dedup AS SELECT * FROM events` immediately before the DELETE:

- **Idempotent**: `IF NOT EXISTS` means repeated daemon starts (restarts) don't error or re-snapshot
- **Small footprint**: events table is capped via `CLAWMETRY_LOCAL_MAX_GB`, so ≤10 MB per user
- **Clean rollback**: `INSERT INTO events SELECT * FROM events_pre_v7_dedup ON CONFLICT DO NOTHING`
- **Fresh installs unaffected**: the `has_events` early-return fires before this line

The snapshot table can be dropped in a future v8 migration once we're confident no rollbacks are needed.

## Test plan

- [ ] Run daemon with existing v6 data → verify `events_pre_v7_dedup` table is created and row count matches original
- [ ] Run daemon twice (restart) → verify second start doesn't error on `IF NOT EXISTS`
- [ ] Fresh install (no prior events) → verify early-return still fires, no empty snapshot created
- [ ] `make lint` passes

Closes #1236

---
_Generated by [Claude Code](https://claude.ai/code/session_01DCvgDWyEzmaaq9Ado4i8pN)_


vivekchand:
## UX review

- 🟢 nit — Backend-only schema migration safety net. No surface affordance changes. The snapshot table `events_pre_v7_dedup` is silent — a power user who notices DuckDB file size increased will have no idea what the new table is for. Next action: add a one-line comment header in the file (already done in the docstring) plus surface the snapshot's existence + row count in `/api/system-health` debug output, so support can confirm "rollback table present" without opening DuckDB.

vivekchand:
## Product review

- 🟡 P1 — Snapshot exists but the rollback path is documented in PR body, not code. If a real user reports data loss next week the on-call engineer will have to dig through commit history to find `INSERT INTO events SELECT * FROM events_pre_v7_dedup ON CONFLICT DO NOTHING`. Next action: add a `_rollback_v7_dedup(conn)` helper in `local_store.py` (even if never wired to a CLI flag) so the rollback path is grep-able and tested.
- 🟡 P1 — "Future v8 migration can drop the snapshot" is a TODO that will rot. Storing duplicate event-table copies indefinitely costs disk on every install. Next action: open a tracking issue for v8 to drop `events_pre_v7_dedup` after the 0.12.230 release ages out (~30 days), with a date deadline; otherwise the snapshot becomes permanent dead weight.
- 🟢 nit — `CREATE TABLE IF NOT EXISTS` on a fresh install where `has_events` is true but the table is already snapshotted is fine, but on a daemon that crashes mid-DELETE and restarts, the snapshot will contain the *pre-crash* state. That's actually correct behavior — worth a one-line comment confirming the invariant.

PR #1390, opened 2026-05-16
fix(local_store): prevent schema migration race between concurrent daemons

### LINKED ISSUE #1237: [Product P0 from PR #1232] schema_version migration races between concurrent daemons
From the post-merge product review of PR #1232.

## Problem

\`LocalStore._migrate\` in \`clawmetry/local_store.py\` (line ~728) takes only \`self._write_lock\` — an in-process \`threading.Lock()\` declared at line ~703. That serialises writes within one Python process. It does NOT serialise migrations across two daemon processes opening the same DuckDB file.

Reference: \`reference_two_python_installs_gotcha.md\` already documented that ClawMetry routinely has two Python installs running in parallel (\`~/.clawmetry/bin/python3\` daemon + \`~/Library/Python/3.9/\` dashboard). After the v7 schema bump, both will hit \`_migrate\` near-simultaneously on the next user start.

## Race window

\`\`\`
Process A                          Process B
SELECT MAX(version) → 6
                                   SELECT MAX(version) → 6
DELETE FROM events WHERE ...
                                   DELETE FROM events WHERE ...
INSERT schema_version (7, ...)
                                   INSERT schema_version (7, ...)
\`\`\`

DuckDB's file lock will serialise the actual writes — but both processes still believe they need to run the migration, so the DELETE runs twice. The DELETE is idempotent at the row level (second pass finds no dupes), so data is safe, but:

1. We waste cycles re-scanning every user's events table twice on upgrade
2. The log line \"v7 dedup migration removed N duplicate event row(s)\" is emitted twice with non-matching deltas (first pass: real N, second pass: 0). Support reading these logs will be confused.
3. If we later add a destructive migration that ISN'T idempotent (the snapshot-then-delete suggestion in the sibling P0 issue would be one), this race becomes a corruption bug.

## Suggested fix

Wrap the \`current<7\` block in a DuckDB transaction with double-checked locking:

\`\`\`python
self._conn.execute(\"BEGIN TRANSACTION\")
try:
    cur = self._conn.execute(\"SELECT MAX(version) FROM schema_version\")
    current = cur.fetchone()[0] or 0
    if current < 7:
        # ... migration body ...
        self._conn.execute(\"INSERT INTO schema_version VALUES (7, ?)\", [now_ms])
    self._conn.execute(\"COMMIT\")
except Exception:
    self._conn.execute(\"ROLLBACK\")
    raise
\`\`\`

DuckDB's transactional isolation will make the second process see the v7 stamp on its re-read inside the txn and skip the migration.

## Severity rationale

P0 because the daemon+dashboard concurrent-process pattern is the default install (not an edge case), and the migration framework in this PR is the template for every future schema migration. Fixing it here is cheap; retrofitting after a non-idempotent migration ships is much harder.

vivekchand:
Closes #1237

## Problem

`LocalStore._migrate` read `MAX(version) FROM schema_version` outside any transaction. The daemon process and the dashboard process (two separate Python installs, the default ClawMetry setup) could both observe `current = 6` simultaneously and both enter the v7 migration body — running the dedup DELETE twice.

## Fix

- Wrap Steps 3 + 4 (version check → migration body → version stamp) in an explicit `BEGIN TRANSACTION` / `COMMIT` block.
- DuckDB serialises writers at the file level, so the second concurrent process blocks on `BEGIN TRANSACTION` until the first commits; it then re-reads `schema_version` inside its own transaction, sees version 7 already stamped, and skips the migration body entirely.
- Steps 1 and 2 (column-add and DDL) stay outside the transaction — they're `CREATE IF NOT EXISTS` and are idempotent regardless.
- The inner `try/except` around `_run_dedup_migration_v7` is preserved: opportunistic cleanup still stamps the version even if dedup fails.
- Outer `except` rolls back and re-raises only for unexpected failures of `BEGIN`/`COMMIT` itself.

## Test plan

- [ ] `python3 -m py_compile clawmetry/local_store.py` passes (done locally)
- [ ] `make lint` passes
- [ ] Manual: start two ClawMetry processes pointing at the same DuckDB on a v6 store; confirm "v7 dedup migration removed N rows" log line appears exactly once across both processes

---
_Generated by [Claude Code](https://claude.ai/code/session_0161wGnqms9jm73r9tnwDUjX)_

### DIFF
--- a/clawmetry/local_store.py
+++ b/clawmetry/local_store.py
@@ -839,9 +839,11 @@ class LocalStore:
           2. Full v2 DDL — CREATE TABLE/INDEX IF NOT EXISTS, no-op on
              already-migrated tables, creates the new tables on fresh stores.
           3. Version-gated data migrations (v6→v7 dedup, etc.) — only run on
-             stores that haven't seen this version yet. Idempotent at the
-             SQL level too, but the version gate makes the common (already-
-             migrated) case a single SELECT.
+             stores that haven't seen this version yet. Wrapped in an explicit
+             DuckDB transaction so concurrent daemon + dashboard processes
+             can't both pass the version gate: the second process blocks on
+             BEGIN TRANSACTION until the first commits, then re-reads the
+             committed stamp and skips the migration body.
           4. Stamp the schema_version row.
         """
         with self._write_lock:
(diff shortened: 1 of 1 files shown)


vivekchand:
## UX review

- 🟢 nit — Backend-only concurrency fix wrapping schema migration in `BEGIN TRANSACTION`. Zero UI surface change. The user-visible failure mode (dedup DELETE running twice and racing with a peer process) was silent corruption, not a visible error — so the fix is also silent. Next action: on success of the v7 dedup path, log the deleted-row count at INFO level (already done) and include it in a once-per-install banner shown the first time the dashboard loads post-upgrade ("Local store upgraded to schema v7 — N duplicate rows removed") so users have one human-readable record that the migration ran exactly once.

vivekchand:
## Product review

- 🟡 P1 — The two-Python-installs gotcha (daemon venv vs dashboard `--user`) is the root cause of this race and has caused at least one other documented data-loss-class bug. Wrapping a transaction here is the right tactical fix, but the underlying duplication (two processes both holding write capability on the same DuckDB file) is a recurring pattern. Next action: open a tracking issue to enforce single-writer architecture — only the daemon ever opens DuckDB writable; the dashboard always goes through the daemon's HTTP proxy (consistent with the existing "DuckDB locks at PROCESS level" finding).
- 🟡 P1 — Inner `try/except` around `_run_dedup_migration_v7` swallows exceptions and still proceeds to stamp the version. Combined with the new outer `try/except` that rolls back on `BEGIN`/`COMMIT` failure, the failure modes are subtle: a regex bug in the dedup migration logs `log.exception` but the version still stamps — meaning the rollback path documented in PR #1389 is never triggered. Next action: clarify the contract — if `_run_dedup_migration_v7` raises, do we stamp v7 (current behavior, "we tried") or skip stamping (next attempt retries)? Document the chosen semantics in the migration docstring.
- 🟢 nit — `BEGIN TRANSACTION` / `COMMIT` work but consider `with self._conn:` (DuckDB's context-manager transaction syntax) which auto-rollbacks on exception — simpler, less code, same guarantee.

PR #1410, opened 2026-05-16
[RELEASE] fix(alerts): evaluator reads DuckDB when OTLP metrics_store is empty (closes #1404)

### LINKED ISSUE #1404: [Product P0 from PR #1399] Alerts evaluator reads OTLP-only metrics_store; ~99% of OSS installs can never fire a threshold rule on real spend
From the post-merge product review of PR #1399.

## TL;DR
The Alerts pipeline (a Cloud-Pro funnel feature) is structurally broken on every OSS install that doesn't have OTLP traffic flowing — which is essentially all of them by default. A user can configure a threshold rule, spend real money on Claude all day, and the rule will never fire because `daily_spent` stays at 0 forever.

## Root cause
`dashboard.py:_budget_monitor_loop` evaluates threshold rules against `_get_budget_status()`, which reads from `metrics_store["cost"]`. That ring buffer is fed ONLY by OTLP ingestion (`/v1/metrics`, `/v1/traces`). DuckDB has the real cost rows from session JSONLs and the HTTP interceptor; the evaluator never looks there.

Surfaced by the alerts harness (PR #1399) as `spend_visibility/real_spend_visible` drift on a vanilla 0.12.229 install: `ground='>=0.000307' actual=0.0`.

## Why this is a Cloud-Pro conversion P0
- Alerts Center is gated as Cloud-Pro per `project_alerts_pro_feature`.
- A user who clicks past the paywall, configures a \$5/day threshold rule, and hears crickets for a week churns — they got zero value from the paid feature.
- A user who DOES get one threshold alert in week one (\"you spent \$5 today on opus\") is probably a paid conversion. We are leaving every one of those on the table.
- This is the single highest-leverage bugfix for OSS→Cloud-Pro conversion in the queue.

## Fix
Mirror DuckDB cost rows into `metrics_store[\"cost\"]` on a tick (e.g. piggyback on the existing 60s `_budget_monitor_loop`):

1. At the top of each evaluator pass, query DuckDB for cost rows since `now() - 24h` that aren't already in the in-memory ring buffer.
2. `_d._add_metric(\"cost\", row)` for each, tagged with a source field so we don't double-count if OTLP ALSO ingests later.
3. Then run the existing eval pass.

Alternative: change `_get_budget_status()` to read directly from DuckDB and bypass `metrics_store` for the cost path entirely. Cleaner long-term, larger blast radius.

## Verification
Re-run the alerts harness — `spend_visibility/real_spend_visible` should flip to PASS, and a real openclaw turn followed by a low-threshold rule should fire within 60s.

```bash
CLAWMETRY_HARNESS_HOOKS=1 python3 scripts/accuracy_harness/alerts.py
```

vivekchand:
## Summary

Fixes #1404 — alert threshold rules silently never fire on the typical OSS install (~99% of users).

**Root cause:** `dashboard._get_budget_status` summed only the in-process `metrics_store[\"cost\"]` ring buffer, which is fed exclusively by the optional OTLP exporter. `pip install clawmetry` (without the `[otel]` extra) never wires OTLP, so that buffer stays empty forever — `daily_spent` is locked at 0 and threshold rules like "alert when daily spend > \$5" never trip on real OpenClaw spend.

**Fix:** Priority-ordered cost lookup:
1. OTLP `metrics_store[\"cost\"]` when it has a fresh entry (≤5 min old) covering today — keeps the fast path for OTLP-fed installs.
2. Else, DuckDB aggregate `SELECT SUM(cost_usd) ...` over the `events` table via the existing v3-aware `local_store.query_aggregates`. Covers all four envelope shapes (`message`, `assistant`, `subagent:assistant`, `model.completed`) for free.

`_get_budget_status` now returns `cost_source ∈ {"otlp","duckdb"}` so the accuracy harness can verify which path fired.

## Before / After (local repro of harness's `assert_spend_visibility`)

Same setup both runs: one real-shape `model.completed` event ingested into DuckDB (cost=\$0.0042), `metrics_store["cost"]` cleared (no-OTLP install).

| Phase | Stage | `daily_spent` | Harness check | Threshold rule (>\$0.001) |
|---|---|---:|---|---|
| **BEFORE** (main @ 0c648c6) | `spend_visibility` | \$0.0000 | DRIFT (0 < \$0.0021) | NEVER FIRES |
| **AFTER** (this PR) | `spend_visibility` | \$0.0042 | PASS (0.0042 ≥ \$0.0021) | FIRES |

Harness execution count BEFORE: persistent 1/4 PASS (only `create` passes, the listed status in #1404). AFTER, all downstream stages should pass — `spend_visibility` PASS → real_spend visible to evaluator → `fire` row landed → `dispatch` webhook delivered. Full live harness run requires a running dashboard + openclaw CLI which aren't available in worktree mode; the regression is reproduced and verified via the unit-level pre/post simulation pasted above.

## Code path

- `dashboard.py:641` — `_otel_cost_is_fresh()` + `_duckdb_cost_since()` helpers, plus the priority-ordered fallback inside `_get_budget_status()`.
- `dashboard.py:end-of-status-dict` — new `cost_source` field for diagnostics.
- `tests/test_alerts_evaluator_duckdb.py` — 5 regression tests, all pass in <1s.

## Test plan

- [x] `pytest tests/test_alerts_evaluator_duckdb.py` — 5/5 PASS
- [x] `pytest tests/test_per_agent_budget.py tests/test_alert_evaluator_unit.py tests/test_alert_rules_local_store.py tests/test_evaluate_alerts_integration.py` — all related-area tests still green (one pre-existing failure on `test_api_alerts_rules_fast_path_serves_local_store` is unrelated; reproduces on `main`).
- [x] `make lint` — Python syntax OK
- [x] Local BEFORE/AFTER simulation of harness's `assert_spend_visibility` (shown above)
- [ ] Live alerts harness on prod after release auto-publishes 0.12.232

## Constraints honoured

- 106-LOC `dashboard.py` diff + 227-LOC test file = **333 LOC total** (cap: <400)
- Schema unchanged
- Dispatch path unchanged
- No new dependencies
- Read-only fallback — DuckDB is queried, never written
- `cost_source` field is additive — existing consumers ignore unknown keys

## Does the feature work now?

**Yes.** On no-OTLP installs (the typical OSS user) the evaluator now sees real OpenClaw spend from DuckDB. Threshold rules fire on real data after 0.12.232 publishes.

### DIFF
--- a/dashboard.py
+++ b/dashboard.py
@@ -638,8 +638,81 @@ def _dispatch_configured_webhooks(alert_type, payload):
         _send_webhook_alert(discord_url, payload, payload_type="discord")
 
 
+# ── DuckDB cost fallback (issue #1404) ────────────────────────────────
+# OTLP is the *cheap* aggregation path: ``metrics_store["cost"]`` is a
+# pre-summed in-process buffer the evaluator reads in O(N) without touching
+# DuckDB. But the typical OSS install never wires an OTLP exporter (it's
+# explicitly optional — ``pip install clawmetry[otel]`` is opt-in), so on
+# the vast majority of nodes that buffer stays empty and threshold rules
+# silently never fire on real spend.
+#
+# This helper recomputes daily/weekly/monthly spend by aggregating the
+# billable-turn rows the sync daemon already persists into DuckDB. We only
+# invoke it when OTLP is empty/stale (see ``_otel_cost_is_fresh``) so
+# OTLP-fed installs keep their fast path.
+_OTLP_FRESH_WINDOW_SEC = 300  # 5 min — matches AGENT_DOWN threshold
+
+
+def _otel_cost_is_fresh(since_ts: float) -> bool:
+    """True iff ``metrics_store['cost']`` has at least one entry timestamped
+    within the last ``_OTLP_FRESH_WINDOW_SEC`` seconds (i.e. an OTLP exporter
+    is actively pushing cost data). When False, the evaluator falls back to
+    DuckDB so threshold rules can still fire on real OpenClaw spend.
+
+    ``since_ts`` is the earliest period start we care about (e.g. today_start
+    for daily rules). Even one fresh OTLP row covering the period means the
+    in-memory buffer is the source of truth — DuckDB fallback is unnecessary.
+    """
+    cutoff = time.time() - _OTLP_FRESH_WINDOW_SEC
+    with _metrics_lock:
+        for entry in metrics_store["cost"]:
+            ts = entry.get("timestamp", 0)
+            if ts >= cutoff and ts >= since_ts:
+                return True
+    return False
+
+
+def _duckdb_cost_since(since_iso: str) -> float:
+    """Sum billable-turn USD over the ``events`` table since ``since_iso``.
+
+    Reuses the v3-aware helpers from ``clawmetry.local_store`` so every
+    OpenClaw envelope shape (Anthropic-SDK ``message``, v3 ``assistant``,
+    ``subagent:assistant``, slim ``model.completed``) is covered. Never
+    raises — returns 0.0 on any error so the evaluator still gets a number.
+
+    Issue #1404: without this, ~99% of OSS installs see a zero spend metric
+    inside the alert evaluator and no real-spend rule ever fires.
+    """
+    try:
+        from clawmetry import local_store
+        store = local_store.get_store()
+        # ``query_aggregates`` is a per-day SUM(cost_usd) rollup keyed by
+        # ``substr(ts, 1, 10)`` — exactly the shape we need to roll back up
+        # into daily / weekly / monthly totals. Cheap on DuckDB (columnar
+        # SUM over a single column with a ts range scan).
+        rows = store.query_aggregates(since=since_iso)
+    except Exception:
+        return 0.0
+    total = 0.0
+    for r in rows or []:
+        try:
+            total += float(r.get("cost_usd") or 0)
+        except (TypeError, ValueError):
+            continue
+    return total
+
+
 def _get_budget_status():
-    """Calculate current spending vs budget limits."""
+    """Calculate current spending vs budget limits.
+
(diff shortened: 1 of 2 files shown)


vivekchand:
## UX review

🟢 **nit (backend-only)** — Diff is pure backend: `dashboard.py` evaluator + new test file. No UI surface touched, no template/CSS/JS changes. `cost_source` is wired into the response dict (`dashboard.py:799`) but no UI consumes it yet — consider adding a one-line diagnostic badge in the Alerts tab ("daily_spent computed via DuckDB fallback") so users in the no-OTLP common case can see the path is healthy. Otherwise no UX surface to scrutinize.

vivekchand:
## Product review

🔴 **P0 — this is THE conversion fix; ship a comms moment around it.** ~99% of OSS installs (no `[otel]` extra) had `daily_spent=0` forever, meaning every "alert when spend > $X" rule silently never fired. Users who set up an alert and never got pinged would have concluded the product is broken — and many of them were one paid alert away from converting. This PR doesn't just fix a bug; it unbreaks the single feature most likely to drive Cloud-Pro conversions. **Action:** changelog entry + dashboard banner for users with rules but $0 historical fires ("your alert rule has never fired — recent fix means it will now"). Don't let this land silently.

🟡 **P1 — `cost_source` is the seed of a Pro upsell surface.** The new `cost_source` field (`otlp` vs `duckdb`) is the cheapest possible signal of "this user has a no-OTLP install" — the exact cohort that should see "Want richer telemetry? Cloud-Pro auto-routes spend through OTel + retains 90 days" upsell. **Action:** wire a one-time conversion CTA on Alerts tab when `cost_source=="duckdb"` AND the user has 0 historical alert fires.

🟢 **nit — surface the "last alert fired" timestamp on the Alerts tab.** With this fix, alerts will start firing for users who've had silent rules for weeks. A "last fired: 5 min ago" pill turns the silent fix into a visible win. Cheap, drives engagement.

PR #1398, opened 2026-05-16
[RELEASE] fix(usage): widen /api/usage event_type to match real OpenClaw v3 (#1394)

vivekchand:
Closes #1394.

## Summary
- The Tokens tab was returning ``input/output/cache_read/cache_write = 0`` across every window on real OpenClaw v3 installs because ``_try_local_store_usage`` derived its per-day chart from ``query_aggregates`` (raw ``token_count`` / ``cost_usd`` columns) and stamped 0 into every split. PR #1386 widened 4 query methods but missed ``/api/usage`` itself.
- Fix walks each ``assistant`` / ``subagent:assistant`` / ``model.completed`` event blob via a new ``_extract_usage_splits`` helper that handles both Anthropic-SDK keys (``cache_read_input_tokens``, ``cache_creation_input_tokens``) and OpenClaw v3 camelCase (``cacheRead``, ``cacheWrite``).
- Sibling dedup: real installs emit BOTH an ``assistant`` (full envelope, has cache splits) AND a ``model.completed`` (slim, no cache) for every LLM turn, ~150 ms apart. New ``query_daily_usage_splits`` collapses sibling pairs via a (session_id, ±1 s) window so input/output don't double.

## Predicates widened (before → after)
- ``query_daily_usage_splits`` (new) filters on ``_BILLABLE_TURN_EVENT_TYPES = (message, assistant, subagent:assistant, model.completed)``.
- Distinct from ``_ASSISTANT_EVENT_TYPES`` (used by fallback detector, excludes subagents intentionally per #1385).

## Helper changes
- New ``_extract_usage_splits(data) -> {input_tokens, output_tokens, cache_read_tokens, cache_write_tokens}`` walks 4 envelope shapes (priority-ordered, first non-zero ``input`` wins).
- New ``_extract_usage_cost(data) -> float`` for the same multi-shape walk over ``cost.total``.
- ``_extract_input_tokens`` refactored to delegate to ``_extract_usage_splits``.

## Live harness verification (real OpenClaw v3)

```
BEFORE 0.12.230:
  summary: 0 pass / 20 drift / 20 total checks
  /api/usage today input        18 → 0    DRIFT
  /api/usage today output       21 → 0    DRIFT
  /api/usage today cache_read   85791 → 0 DRIFT
  /api/usage today cache_write  213 → 0   DRIFT
  (same drift across week/month/all_14d)

AFTER 0.12.230:
  summary: 20 pass / 0 drift / 20 total checks
[...]

Issue #1419, opened 2026-05-16
[Product P0 from PR #1410] Ship comms moment: alerts evaluator now sees real spend on no-OTLP installs


vivekchand:
From the post-merge product review of PR #1410.

PR #1410 unbreaks alert threshold rules for ~99% of OSS installs (no `[otel]` extra had `daily_spent=0` forever; no "alert when spend > $X" rule ever fired on real spend).

This is THE single feature most likely to drive Cloud-Pro conversions. Don't let it land silently.

**Action items:**
1. Add changelog entry calling out the user-visible effect ("alert rules now fire on real OpenClaw spend, not just OTLP-fed installs").
2. Dashboard banner / Alerts tab notice for users who have rules configured but $0 historical fires: "Your alert rules previously didn't fire on real spend — recent fix means they will now."
3. Wire the new `cost_source` field (`dashboard.py:799`, values `otlp` vs `duckdb`) into the Alerts tab as a one-time conversion CTA: when `cost_source=="duckdb"` AND `historical_fires==0`, show "Want richer telemetry + 90d retention? Upgrade to Cloud-Pro."
4. Add "last fired: 5 min ago" pill to Alerts tab — converts the silent fix into a visible win.

**Conversion impact:** users who set up an alert and never got pinged would have concluded the product is broken. Many of them were one paid alert away from converting.

PR #1417, opened 2026-05-16
feat(insights): weekly LLM-over-DuckDB insights digest (Slack/email/Telegram)

vivekchand:
## Summary
First piece of the "AI wiki over your agent data" moat. 10 hand-authored SELECT templates run on the local DuckDB; results synthesised by Claude Sonnet into a Monday digest; dispatched via the existing channel adapters (Slack/email/Telegram) — Cloud-Pro per `project_alerts_pro_feature.md`.

**Gated by `CLAWMETRY_INSIGHTS=1`** for v1 soak. NOT `[RELEASE]`-tagged. After 1-week soak we flip Pro users on by default.

### What ships
- **`clawmetry/insights.py`** — 10 canned SQL templates, `WeeklyDigestGenerator`, Sonnet synthesis (Anthropic `/v1/messages`), Anthropic key resolver (config → env → cloud-relayed), `deliver()` for telegram/slack/email/dashboard_only, Mon-9am cron thread (drift-tolerant). Templates pass `dives_sql_safety.validate_sql` at module import.
- **`routes/insights.py`** — `/api/insights/{preview,history,send-now,config}` + minimal `/insights` HTML page. Feature-gated 404s when flag off; API-key redacted on GET config.
- **`clawmetry/local_store.raw_select_safe()`** — single allowlisted entry-point for static SQL templates. SELECT/WITH only, `$name` DuckDB binds, 500-row cap. Wired into `routes/local_query._DAEMON_METHODS` so the dashboard process reaches it through the daemon proxy when the writer lock is held.
- **`tests/test_insights.py`** — 10 tests: SQL safety, end-to-end on empty DB, raw_select_safe rejects/accepts, config round-trip, route gating, scheduler math. All pass.

### Hallucination guardrail
Every narrative bullet is paired with the raw rows it summarised. `WeeklyDigest.insights[i].rows` is the source of truth — the narrative is best-effort. The UI renders both side-by-side (`<details>` block under each insight).

### Cost discipline
Templates are STATIC — no LLM-generated SQL in v1, so safety surface is flat. Only synthesis hits the API.
- Per-user weekly cost: ~$0.05–$0.60 (depends on data volume).
- This box's actual run: 590 synthesised tokens × Sonnet pricing → **$0.005/run**.

### Sample digest output (live DuckDB on this box, with stubbed synthesis)

```
ClawMetry Weekly Insights — Week of 2026-05-09

1,230 events across 6 sessions; $0.01 spent; 265,508 tokens.

[...]

Issue #1394, opened 2026-05-16
[accuracy-audit 2026-05-16] tokens drift: /api/usage today/cache_read ground=85791 actual=0 (delta=-85791; 20 drifts total)


vivekchand:
## Drift report — `/api/usage`

Auto-filed by `scripts/accuracy_harness/tokens.py`.

### Drifted (window, metric) pairs
| window | metric | ground | actual | delta | tolerance |
|---|---|---:|---:|---:|---:|
| today | input | 18 | 0 | -18 | ±1 |
| today | output | 21 | 0 | -21 | ±1 |
| today | cache_read | 85791 | 0 | -85791 | ±3 |
| today | cache_write | 213 | 0 | -213 | ±3 |
| today | total | 39 | 52 | +13 | ±1 |
| week | input | 18 | 0 | -18 | ±1 |
| week | output | 21 | 0 | -21 | ±1 |
| week | cache_read | 85791 | 0 | -85791 | ±3 |
| week | cache_write | 213 | 0 | -213 | ±3 |
| week | total | 39 | 52 | +13 | ±1 |
| month | input | 18 | 0 | -18 | ±1 |
| month | output | 21 | 0 | -21 | ±1 |
| month | cache_read | 85791 | 0 | -85791 | ±3 |
| month | cache_write | 213 | 0 | -213 | ±3 |
| month | total | 39 | 52 | +13 | ±1 |
| all_14d | input | 18 | 0 | -18 | ±1 |
| all_14d | output | 21 | 0 | -21 | ±1 |
| all_14d | cache_read | 85791 | 0 | -85791 | ±3 |
| all_14d | cache_write | 213 | 0 | -213 | ±3 |
| all_14d | total | 39 | 52 | +13 | ±1 |

### Ground-truth event log
```json
[
  {
    "tag": "ACCURACY_AUDIT_6ee15e69_msg1",
    "elapsed_s": 4.39,
    "input": 6,
    "output": 7,
    "cacheRead": 28526,
    "cacheWrite": 71,
    "sessionId": "01c6c2d6-44bd-4dbf-a554-ade04c5869a9",
    "model": "claude-opus-4-7"
  },
  {
    "tag": "ACCURACY_AUDIT_6ee15e69_msg2",
    "elapsed_s": 4.29,
    "input": 6,
    "output": 7,
    "cacheRead": 28597,
    "cacheWrite": 71,
[...]

PR #1424, opened 2026-05-16
[RELEASE] fix(ux): one-time Tokens-tab banner for cache-split data restoration (closes #1401)

### LINKED ISSUE #1401: [UX P0 from PR #1398] Silent Tokens-tab fix needs one-time 'now-showing-real-data' banner
From the post-merge UX review of PR #1398.

Every existing user on 0.12.229 and older has been staring at a Tokens tab full of zeros (input/output/cache_read/cache_write). PR #1398 lands the real numbers but the jump from $0 → actual spend on next dashboard open will look like a regression or a fabrication, not a fix.

Action: add a one-time dismissible banner on the Tokens tab for one release ("Token splits now include cache reads/writes — see CHANGELOG #1394") so the visual jump is legible rather than suspicious. Auto-dismiss after first dismiss or after one [RELEASE] cycle.

Severity: P0 — silently distrusted dashboards bleed Cloud-Pro conversions.

vivekchand:
Closes #1401

## Problem
PR #1398 silently fixed input/output/cache_read/cache_write tokens being zero on real OpenClaw v3 installs. Users who visited the Tokens tab before 0.12.230 saw $0 everywhere. After the fix, values jump from $0 → real spend — without context, this looks like a regression or fabrication rather than a fix. Silently-distrusted dashboards bleed Cloud-Pro conversions.

## Fix
One-time dismissible banner at the top of the Tokens tab:

> 📈 Token splits now include cache reads/writes — numbers are accurate from this version.

**Mechanics:**
- Inline `<script>` IIFE checks `localStorage.getItem('clawmetry_tokens_fix_banner_v1')` on page load
- If not dismissed, sets `display:flex` on the banner div
- Dismiss button sets the localStorage key and hides the div
- Fresh installs that never saw the zero state will also see it once — acceptable; it's a one-sentence informational notice, not alarming

**No server-side state, no API call, no dependency.** Banner fires at most once per browser profile.

## Files changed
- `dashboard.py` — 13 lines added inside `#page-usage`, before the Main Usage Stats section

## Test plan
- [ ] Navigate to Tokens tab → banner appears (if not previously dismissed)
- [ ] Click × → banner disappears, `clawmetry_tokens_fix_banner_v1` key set in localStorage
- [ ] Reload page → banner does NOT reappear
- [ ] Open new private/incognito tab → banner appears again
- [ ] Other tabs (Sessions, Brain, etc.) unaffected

---
_Generated by [Claude Code](https://claude.ai/code/session_017Ayqq7QzArFJLcw33tzNWy)_

### DIFF
--- a/dashboard.py
+++ b/dashboard.py
@@ -4274,7 +4274,19 @@ function clawmetryLogout(){
   
   <!-- Cost Warnings -->
   <div id="cost-warnings" style="display:none; margin-bottom: 16px;"></div>
-  
+
+  <!-- One-time banner: token splits now accurate (issue #1401) -->
+  <div id="tokens-fix-banner" style="display:none;margin-bottom:16px;padding:10px 14px;background:#1a3a2a;border:1px solid #2a5a3a;border-radius:8px;font-size:13px;color:#60ff80;display:flex;align-items:center;justify-content:space-between;">
+    <span>&#128200; Token splits now include cache reads/writes &mdash; numbers are accurate from this version. <a href="#" onclick="event.preventDefault();" style="color:#60ff80;text-decoration:underline;">See CHANGELOG #1394</a></span>
+    <button onclick="document.getElementById('tokens-fix-banner').style.display='none';localStorage.setItem('clawmetry_tokens_fix_banner_v1','1');" style="background:transparent;border:none;cursor:pointer;font-size:16px;color:#60ff80;padding:0 0 0 12px;">&times;</button>
+  </div>
+  <script>
+  (function(){
+    var b = document.getElementById('tokens-fix-banner');
+    if(b && !localStorage.getItem('clawmetry_tokens_fix_banner_v1')) b.style.display='flex';
+  })();
+  </script>
+
   <!-- Main Usage Stats -->
   <div class="grid">
     <div class="card">


vivekchand:
## UX review

🔴 **P0 — "See CHANGELOG #1394" link does NOTHING.** `<a href="#" onclick="event.preventDefault();">See CHANGELOG #1394</a>` (`dashboard.py:4280`). Users will click it expecting to see what changed; nothing happens. That's worse than no link — it teaches users not to trust links in the dashboard. Either link to `https://github.com/vivekchand/clawmetry/issues/1394` directly (target="_blank"), or strip the link entirely and let the message stand. Fake links in informational banners are the exact "feels like a debug build" pattern flagged on PR #1417's Send-to-channel button.

🟡 **P1 — banner color (#60ff80 + #1a3a2a) is the same green as success/CTAs elsewhere.** Informational banners should use a neutral/info palette (blue or amber), not the green that means "action succeeded." Reading the dashboard, this banner reads like a confirmation — not a notice. Consider `#1a2a3a` bg + `#60a5fa` text, matching the existing info-banner pattern in `#alert-banner`.

🟡 **P1 — banner inserted via inline `<script>` IIFE inside template** (`dashboard.py:4283-4288`). Works but doesn't compose with the rest of the dashboard's lifecycle. If a future tab adds a similar one-time banner, we'll have N IIFEs racing on page-load. Centralize via a tiny helper (`oneTimeBanner(elemId, lsKey)`) in `auth-bootstrap.js` or `app.js`.

🟢 **nit — banner copy "📈 Token splits now include cache reads/writes"** is engineer-speak. A non-technical operator (`feedback_simple_ui_for_nontechnical.md` audience) doesn't know what "cache reads/writes" or "splits" mean. Reframe: "Your token costs now include all model interactions, including cached requests. Numbers are accurate from this version."

🟢 **nit — `localStorage.clawmetry_tokens_fix_banner_v1`** key is good (versioned). Keep this convention; future "fix announcement" banners should follow `clawmetry_<topic>_banner_v<N>`.

vivekchand:
## Product review

🔴 **P0 — banner is the right move but reads like an apology.** Per `project_clawmetry_traction.md` we're at 120K+ installs; this banner ships to all of them. The current copy ("numbers are accurate from this version") implies "prior numbers were inaccurate" — that's an admission of a months-long bug to a userbase that may not have noticed. Reframe positively: "Token attribution upgraded — your dashboard now tracks cache reads/writes for the full picture." Same fact, no self-incrimination.

🟡 **P1 — missed conversion event.** Users who see this banner have actively used the Tokens tab, which means they care about cost. That's the highest-converting cohort. Add a tiny CTA: "See historical accuracy in your Cloud-Pro retention archive →" (links to /cloud upgrade). One sentence, no lift.

🟡 **P1 — fresh installs see this banner too** (per PR body: "acceptable; one-sentence informational notice"). Disagree — fresh installs have NEVER seen the buggy state. Showing them a confidence-eroding banner about a fix they don't need is pure noise. Gate the banner on: `first_install_at < 2026-05-15` (when 0.12.230 shipped). Cheap server-side check via `/api/version` history or install_at field. If the field doesn't exist, ship a one-line `POST /api/seen_tokens_banner` server-side acknowledgement instead of localStorage so we can also count engagement.

🟢 **nit — telemetry hook on click vs dismiss.** Wire a counter for "banner shown" / "banner dismissed" / "changelog link clicked" so we know whether this kind of in-product comms lands. Critical learning for #1419 (alerts comms) which uses the same pattern.

Issue #1420, opened 2026-05-16
[Product P0 from PR #1417] Insights digest leaks paid-tier value + Anthropic-key friction wall blocks first-touch


vivekchand:
From the post-merge product review of PR #1417.

Two coupled P0s on the Weekly Insights Digest:

### P0a — Pro upsell narrative missing; ships full feature to ALL users on `CLAWMETRY_INSIGHTS=1`

Per memory `project_alerts_pro_feature.md` digests should be Cloud-Pro-only. PR #1417 ships the full feature (LLM synthesis + Slack/Telegram/email dispatch + cron-scheduled delivery) gated only by an env var, with zero conversion narrative. When the soak ends and we flip to Pro-only, free users will lose access and feel rugged.

**Better tier split:**
- **Cloud-Free:** 1 weekly digest, dashboard view only, no dispatch, ClawMetry-subsidized synthesis (~$0.05/user fits inside CAC)
- **Cloud-Pro:** daily/scheduled cron + Slack/Telegram/email dispatch + cloud-relayed Anthropic key (no setup)

Every `/insights` view becomes a conversion event with: "Want this delivered to Slack every Monday at 9am? Upgrade to Pro."

### P0b — Anthropic API key requirement = pre-Pro friction wall

Forcing OSS users to wire `ANTHROPIC_API_KEY` (`clawmetry/insights.py` config) before they can preview the digest = ~95% drop-off on first-touch. The first "wow" moment must require zero setup.

**Fix:** cloud-relay synthesis through the existing daemon→cloud heartbeat-piggyback (memory `project_relay_transport_decision.md`) for free preview. User's own Anthropic key only required for unlimited Pro dispatches.

### Telemetry gap (P1, related)

No tracking of "who clicked Send to channel" — that's the single most important conversion signal in the PR. Wire a `/api/track` event (or admin-analytics entry per `reference_admin_analytics_page.md`) so Pro candidates emerge.

### Files to touch
- `clawmetry/insights.py` — add cloud-relay key resolver
- `routes/insights.py` — add tier-aware dispatch gating + upsell banner
- `clawmetry-cloud/routes/admin.py` — add insights conversion tracking

PR #1428, opened 2026-05-16
feat(insights): add Insights nav tab (gated by CLAWMETRY_INSIGHTS=1)

vivekchand:
## Summary
P1 follow-up from the post-merge UX review of PR #1417 — without a nav entry, the Weekly Insights Digest page at `/insights` is unreachable unless you already know the URL.

## Implementation
- New `<a class="nav-tab" id="insights-tab" href="/insights">` entry alongside the existing Memory/Security/NemoClaw tabs (both nav locations updated — OSS variant @ dashboard.py:3935 + cloud variant @ dashboard.py:10059).
- Hidden by default (`style="display:none;"`).
- IIFE on page load probes `/api/insights/config` — returns 404 (`feature_disabled`) when `CLAWMETRY_INSIGHTS=1` is unset (per `routes/insights.py:_gated()`), so the tab stays invisible until the user opts in. Mirrors the existing NemoClaw tab gate pattern (app.js:13186).

## Test plan
- [x] `python3 -c "import dashboard"` — clean
- [ ] Manual w/ `CLAWMETRY_INSIGHTS=1` — Insights tab appears in nav, click navigates to `/insights`
- [ ] Manual w/o the env var — tab stays hidden (no 404 console noise; `r.ok` check skips toggle)

## Out of scope
- The `/insights` page navigates AWAY from the dashboard SPA (full page load) rather than rendering inline. Tradeoff intentional — keeps blast radius small until soak ends. Future: render insights as an in-app tab section.

### DIFF
--- a/clawmetry/static/js/app.js
+++ b/clawmetry/static/js/app.js
@@ -13194,3 +13194,18 @@ function _ncEsc(s) {
   } catch(e) {}
 })();
 } } // end if(false) stub */
+
+// On page load: reveal Insights tab if the feature flag is on. Insights endpoints
+// 404 with `{"error":"feature_disabled"}` when CLAWMETRY_INSIGHTS=1 is unset, so
+// a fast HEAD-style probe of /api/insights/config keeps the tab hidden by default
+// (no surface for users who haven't opted in yet) without leaking a spurious 200.
+(function() {
+  try {
+    fetch('/api/insights/config').then(function(r) {
+      if (r.ok) {
+        var t = document.getElementById('insights-tab');
+        if (t) t.style.display = '';
+      }
+    }).catch(function(){});
+  } catch(e) {}
+})();
--- a/dashboard.py
+++ b/dashboard.py
@@ -3933,6 +3933,7 @@ function clawmetryLogout(){
     <div class="nav-tab" onclick="switchTab('memory')">Memory</div>
     <div class="nav-tab" onclick="switchTab('security')">Security</div>
     <div class="nav-tab" id="nemoclaw-tab" onclick="switchTab('nemoclaw')" style="display:none;">NemoClaw</div>
+    <a class="nav-tab" id="insights-tab" href="/insights" style="display:none;text-decoration:none;color:inherit;" title="Weekly LLM-over-DuckDB digest (set CLAWMETRY_INSIGHTS=1 to enable)">Insights</a>
     <!-- History tab hidden until mature -->
     <!-- <div class="nav-tab" onclick="switchTab('history')">History</div> -->
   </div>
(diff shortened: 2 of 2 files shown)


vivekchand:
## UX review

🟡 **P1 — Insights tab has zero discovery hint when feature flag is OFF.** When `CLAWMETRY_INSIGHTS=1` is unset, the tab simply doesn't render. Users with a configured cron, real DuckDB data, and the perfect cohort to value insights see literally nothing pointing them at the feature. Better pattern: render the tab in a "locked" state with a tooltip "Set CLAWMETRY_INSIGHTS=1 to unlock" — lets users discover the capability before they have to know to opt in.

🟡 **P1 — Insights tab navigates AWAY from the SPA via full page load** (`<a href="/insights">`), breaking continuity with every other nav-tab that uses `switchTab(name)`. PR body acknowledges this as intentional ("keeps blast radius small until soak ends"), but the inconsistent behavior is jarring — users expect the same in-page panel-swap they get on Brain/Memory/etc. Track for post-soak follow-up: render insights as an in-app tab section.

🟢 **nit — every page load probes `/api/insights/config`** (synchronous `fetch` in IIFE). For users without the flag, that's a wasted 404 round-trip on every dashboard load forever. Combine with the Nemoclaw probe into a single `/api/feature-flags` manifest that returns all gates in one call; cheaper, easier to audit.

🟢 **nit — no "new" badge or subtle visual on first reveal.** The tab pops in silently when the flag is detected. A one-time `★` or "NEW" pip would draw the eye to a feature the user just enabled.

vivekchand:
## Product review

🔴 **P0 — invisible-when-off pattern is wrong for a Pro upsell lever.** Per `project_alerts_pro_feature.md` insights digests are Cloud-Pro-only post-soak. This PR currently makes the tab disappear entirely when `CLAWMETRY_INSIGHTS=1` is unset — meaning **every free user, every Cloud-Free user, sees zero evidence that insights exist.** That's the worst possible Pro funnel: hide the value prop from the people you're trying to convert. Better: show a Pro-locked tab ("Insights 🔒") that opens an upsell modal — "Weekly LLM-narrated digest of your agent telemetry — included in Cloud-Pro." Same architecture, infinitely better conversion shape. This compounds the issue already filed in #1420.

🟡 **P1 — gating-by-env-var is a developer-affordance, not a product mechanism.** A non-technical operator (`feedback_kody_kendall_user.md`) can't `export CLAWMETRY_INSIGHTS=1` and restart the daemon. The gate should be a UI checkbox in Settings ("Enable Weekly Insights — beta") with no terminal involvement, per `feedback_simple_ui_for_nontechnical.md`. Env-var-only = engineering-only audience.

🟡 **P1 — no telemetry on "tab revealed" events.** This is a feature flag flip — knowing how many users have `CLAWMETRY_INSIGHTS=1` set is critical to deciding when soak ends and Pro flip-on begins. Wire a one-line `POST /api/track` (or admin-analytics entry per `reference_admin_analytics_page.md`) on tab-reveal.

🟢 **nit — IIFE pattern matches NemoClaw cleanly.** Architectural consistency is right; just add a single `/api/feature-flags` aggregator before the third feature ships and we'll have N+1 probes.

PR #1434, opened 2026-05-16
ci: add auth-bootstrap no-reload smoke guard

### LINKED ISSUE #1368: [Product P0 from PR #1363] Add fast lint guard for auth-bootstrap.js IIFE regressions
From the post-merge product review of PR #1363.

## Problem
PR #1358 (PR-C zero-click bootstrap) shipped a `location.reload()` inside the page-load IIFE. Every Playwright E2E test ERRORed at setup with "Execution context was destroyed". We didn't catch it on PR-C — it surfaced on PR #1360 (CLI banner, no JS changes), where it looked like #1360's fault. Five minutes of CI per innocent PR is unacceptable for a regression that's deterministic and fast to check.

## Proposed fix
Add a required CI lint check on `clawmetry/static/js/auth-bootstrap.js`:
1. `node --check` (already runs locally, not in CI gate)
2. Headless smoke: spin up the dashboard, hit `/`, assert `document.readyState === 'complete'` after 2s + `window.performance.navigation.type === TYPE_NAVIGATE` (i.e. did NOT reload). Fails in <30s vs the 5-minute E2E run.

## Acceptance
- New CI job runs on every PR that touches `auth-bootstrap.js` (or `*.js` under `clawmetry/static/`).
- Job fails fast on any IIFE that calls `location.reload()` directly. (Could be a regex check + the smoke test as defence in depth.)
- PR cycle time for JS changes drops back to <2 min for syntax-only PRs.

JaiCodes77:
Closes #1368

## What

Adds a fast CI guard that catches `location.reload()` regressions in the zero-click auth bootstrap flow. PR #1358 shipped a `location.reload()` inside the page-load IIFE which broke every Playwright E2E test — this prevents that class of bug from landing again.

## How

New workflow `.github/workflows/auth-bootstrap-guard.yml` that runs on PRs touching `clawmetry/static/js/*.js`. Two checks:

1. **Syntax check** — `node --check auth-bootstrap.js` (fast fail for parse errors)
2. **Headless smoke test** — starts the dashboard, loads `/` in Playwright Chromium, and asserts:
   - `document.readyState === 'complete'`
   - `performance.navigation.type === TYPE_NAVIGATE` (i.e. no reload happened)
   - Bootstrap token was stored in localStorage correctly

Fails in <30s vs the 5-minute E2E run that was catching this before.

## Note

The existing `ci.yml` already runs `node --check` on all `*.js` files, so the syntax step here is redundant by design — this workflow is self-contained so it stays useful even if the main CI job changes. The real value-add is the behavioral smoke test.

### DIFF
--- a/.github/workflows/auth-bootstrap-guard.yml
+++ b/.github/workflows/auth-bootstrap-guard.yml
@@ -1,0 +1,95 @@
+name: Auth Bootstrap Guard
+
+on:
+  pull_request:
+    paths:
+      - "clawmetry/static/js/*.js"
+      - ".github/workflows/auth-bootstrap-guard.yml"
+
+permissions:
+  contents: read
+
+jobs:
+  auth-bootstrap-guard:
+    name: Auth bootstrap no-reload smoke
+    runs-on: ubuntu-latest
+    timeout-minutes: 3
+
+    steps:
+      - uses: actions/checkout@v4
+
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.11"
+
+      - uses: actions/setup-node@v4
+        with:
+          node-version: "20"
+
+      - name: Check auth-bootstrap.js syntax
+        run: node --check clawmetry/static/js/auth-bootstrap.js
+
+      - name: Guard bootstrap IIFE does not reload
+        run: |
+          sed -n '1,58p' clawmetry/static/js/auth-bootstrap.js | grep -v '^[[:space:]]*//' | \
+            node -e "const fs=require('fs'); const s=fs.readFileSync(0,'utf8'); if (/location\\.reload\\s*\\(/.test(s) || /window\\.location\\.reload\\s*\\(/.test(s)) { console.error('auth bootstrap IIFE must not call location.reload()'); process.exit(1); }"
+
+      - name: Install runtime deps
+        run: |
+          pip install flask requests waitress cryptography
+          npm install playwright
+          npx playwright install --with-deps chromium
+
+      - name: Start dashboard
+        run: |
+          OPENCLAW_GATEWAY_TOKEN=ci-test-token python3 dashboard.py --host 127.0.0.1 --port 8900 --no-debug > /tmp/clawmetry.log 2>&1 &
+          for i in $(seq 1 30); do
+            curl -sf -H "Authorization: Bearer ci-test-token" http://127.0.0.1:8900/api/health && exit 0
+            sleep 1
+          done
+          cat /tmp/clawmetry.log
+          exit 1
+
+      - name: Smoke test zero-click bootstrap does not reload
+        run: |
+          node <<'NODE'
+          const { chromium } = require('playwright');
+
+          (async () => {
+            const browser = await chromium.launch();
+            const page = await browser.newPage();
+
+            await page.goto('http://127.0.0.1:8900/', {
+              waitUntil: 'load',
+              timeout: 10000,
+            });
+
+            await page.waitForTimeout(2000);
+
+            const result = await page.evaluate(() => ({
+              readyState: document.readyState,
+              navType: window.performance.navigation.type,
+              expectedNavType: window.performance.navigation.TYPE_NAVIGATE,
+              token: localStorage.getItem('clawmetry-token'),
+            }));
+
+            await browser.close();
+
+            if (result.readyState !== 'complete') {
+              throw new Error(`document.readyState was ${result.readyState}, expected complete`);
+            }
+
+            if (result.navType !== result.expectedNavType) {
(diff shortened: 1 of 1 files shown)


vivekchand:
@JaiCodes77 Thanks for picking this up and shipping fast! Read through the diff — solid work. Defense-in-depth (regex + behavioral smoke) matches the acceptance criteria exactly, 3-min timeout is the right shape, path-scoping keeps it from running on unrelated PRs.

A few specific notes (none blocking — happy to merge once you've had a chance to look):

1. **`performance.navigation.type` is deprecated** (Navigation Timing API Level 1). Still works everywhere Playwright runs so it's fine for now, but the modern equivalent is:
   ```js
   const entry = performance.getEntriesByType('navigation')[0];
   if (entry && entry.type === 'reload') { /* reload happened */ }
   ```
   Optional swap.

2. **Regex window** — `sed -n '1,58p'` checks the first 58 lines, but the IIFE could grow past that. Could you widen to the whole file or grep for the literal string instead?
   ```bash
   if grep -nE 'location\.reload\s*\(' clawmetry/static/js/auth-bootstrap.js; then
     echo 'auth-bootstrap must not call location.reload()'; exit 1
   fi
   ```
   Same defence, smaller surface to maintain.

3. **No newline at end of file** — small style nit, GitHub's diff shows `\ No newline at end of file` on the last line. Most editors fix this on save.

4. **Token-flow check is great** — asserting `localStorage.getItem('clawmetry-token') === 'ci-test-token'` validates the whole zero-click bootstrap path end-to-end, not just the absence-of-reload. That's a real win beyond the original issue.

CI should run shortly. Ping back when you've pushed updates (or if you'd rather merge as-is and iterate after — the regex / deprecated-API points are both true follow-ups, not blockers).

vivekchand:
## UX review

🟢 **nit (CI infra-only)** — Pure GitHub Actions workflow addition. No template/JS/CSS, no user-visible surface. Users feel this indirectly via fewer JS-bootstrap regressions slipping through (per the PR-#1358/#1363 incident this was filed to prevent).

🤖 Pre-merge review @ 9e95215. Pushing new commits will re-trigger review if diff >50 lines or >2h elapsed.

vivekchand:
## Product review

🟢 **good — fast-fail gate for the exact regression class.** PR #1358's `location.reload()` cost 5 minutes per innocent JS PR until the root cause was found. This brings the detection budget to <30s and runs only on JS PRs (no wasted CI on unrelated changes). That's the right shape.

🟢 **nit — `OPENCLAW_GATEWAY_TOKEN=ci-test-token` end-to-end check is a freebie.** The smoke test was written to assert no-reload, but it also incidentally validates that the whole zero-click bootstrap (detect token on loopback → store in localStorage → checkAuth) flows correctly. Worth calling out in the workflow logs so anyone debugging a token-flow bug in the future knows this gate exercises it.

🟢 **nit — make it a required check on `main`** in branch protection (per follow-up bullet in PR #867's actionlint pattern). Gate is advisory until then.

🤖 Pre-merge review @ 9e95215. Pushing new commits will re-trigger review if diff >50 lines or >2h elapsed.

JaiCodes77:
@vivekchand  requested changes are in place, please review when free.

vivekchand:
@JaiCodes77 Nicely done! All three suggestions landed cleanly:

- ✅ `performance.getEntriesByType('navigation')[0]?.type` — used optional chaining too, which handles the edge case where the entry isn't ready.
- ✅ `awk '1; /^\}\)\(\);$/ { exit }'` — clever, extracts up to the first IIFE-closing line so the regex check scales with the file.
- ✅ trailing newline.

Looks ready to merge from my side. Waiting on @vivekchand to flip the button since this is your PR; I won't merge it on your behalf. CI on the new SHA should run shortly.

vivekchand:
## UX review

🟢 **nit (CI infra-only, unchanged from prior review)** — Refinement push (`46f2482`) is 10 lines, all in the new `.github/workflows/auth-bootstrap-guard.yml`. Still pure infra; no template/JS/CSS user-facing surface.

🤖 Pre-merge review @ 46f2482. Pushing new commits will re-trigger review if diff >50 lines or >2h elapsed.

vivekchand:
## Product review

🟢 **good — addresses all P1 nits from prior review.** Modern Navigation Timing API, awk-based scope extraction (no fixed line-count), trailing newline. The defense-in-depth (regex + behavioral smoke) shape is unchanged and still the right call for the regression class this targets.

🟢 **nit (carryover from prior review, still applies)** — Make this a required check on `main` in branch protection settings; otherwise the gate is advisory and a determined merger can still push past a failing run.

🤖 Pre-merge review @ 46f2482. Pushing new commits will re-trigger review if diff >50 lines or >2h elapsed.

Issue #1431, opened 2026-05-16
[Product P0 from PR #1428] Insights tab invisible-when-off destroys Pro funnel


vivekchand:
From the post-merge product review of PR #1428.

PR #1428 makes the Insights tab disappear entirely when `CLAWMETRY_INSIGHTS=1` is unset. Per `project_alerts_pro_feature.md` insights digests are Cloud-Pro-only post-soak. So **every free user, every Cloud-Free user, sees zero evidence that insights exist** — that's the worst possible Pro funnel: hide the value prop from the people you're trying to convert.

**Fix:** show a Pro-locked tab ("Insights 🔒" or "Insights — Pro") when the flag is unset. Click opens an upsell modal:

> **Weekly Insights Digest**
> LLM-narrated summary of your agent telemetry, delivered every Monday.
> Cost drivers, anomalies, week-over-week deltas — auto-dispatched to Slack/email/Telegram.
>
> **Included in Cloud-Pro.** [Start 7-day free trial]

Same architecture (existing `/insights` route stays as-is), infinitely better conversion shape.

This compounds with the same issue already filed in **#1420** (insights tier leak + Anthropic-key wall) — both should land together as the "Insights Pro funnel" sprint:
- #1420 — fix tier split (free preview vs Pro dispatch) + cloud-relay synthesis
- #1428 follow-up (this issue) — add locked-tab + paywall modal

### Related P1s from same review
- Replace env-var gating with a Settings checkbox (env-var-only = engineering-only audience per `feedback_simple_ui_for_nontechnical.md`).
- Wire `POST /api/track` on "tab revealed" so we can measure soak adoption before Pro flip-on.
- Combine probes (`/api/insights/config` + `/api/nemoclaw/status` + future ones) into a single `/api/feature-flags` manifest before the third feature ships.

PR #1435, opened 2026-05-16
[RELEASE] fix(e2e): allowlist /api/insights/config 404 in cloud-contract spec

vivekchand:
## Summary
Unblocks cloud deploy of 0.12.234 — failed at 09:22 UTC on \`cloud-contract\` spec's \"zero unexpected JS errors\" check.

[RELEASE]-tagged because cloud's deploy gate pulls the contract spec from OSS main; the fix needs to land in a published OSS version before cloud's next deploy attempt sees it.

## Root cause
The new Insights nav-tab IIFE (PR #1428) probes \`/api/insights/config\`. The endpoint returns 404 on cloud because \`CLAWMETRY_INSIGHTS=1\` is unset (feature is OSS-only soak per PR #1417). The JS \`if (r.ok)\` correctly hides the tab on 404, but the browser auto-logs the failed fetch as \`console.error\` which trips the cloud-contract gate.

## Fix
Allowlist matches the existing pattern for \`/api/diagnostics\` 410 and \`/api/config-diagnostics\` 404 — both \"cloud-disabled returns 4xx but JS handles it gracefully\" cases.

## Test plan
- [x] \`node --check tests/e2e/cloud-contract.mjs\` — clean
- [ ] CI: cloud-contract job (no live run on this PR — verified in prod cloud deploy gate after merge)
- [ ] After release: re-run cloud deploy of 0.12.234 should pass

## Long-term
Issue #1431 tracks the proper fix: change \`routes/insights.py\` \`_gated()\` to return 200 \`{enabled:false}\` instead of 404. That removes the allowlist need entirely and matches the \"Pro-locked vs invisible\" UX direction. Until that lands, this allowlist is the floor.

### DIFF
--- a/tests/e2e/cloud-contract.mjs
+++ b/tests/e2e/cloud-contract.mjs
@@ -82,7 +82,15 @@ function isHarmlessConsoleError(e) {
     // shape: new OSS endpoint, returns 404 on cloud. Filtering
     // them here matches the existing /api/skills 410 pattern.
     (/\/api\/diagnostics/.test(e) && /\b410\b/.test(e)) ||
-    (/\/api\/config-diagnostics/.test(e) && /\b404\b/.test(e))
+    (/\/api\/config-diagnostics/.test(e) && /\b404\b/.test(e)) ||
+    // /api/insights/config returns 404 when CLAWMETRY_INSIGHTS=1 is unset
+    // (the common case on cloud — feature is OSS-only soak per PR #1417).
+    // The IIFE probe in app.js intentionally swallows the failure via
+    // `if (r.ok)` to keep the Insights tab hidden, but the browser still
+    // logs the 404 as console.error which we can't suppress from JS.
+    // Issue #1431 will fix this properly by returning 200 {enabled:false}
+    // instead of 404; until then, allowlist matches the existing pattern.
+    (/\/api\/insights\/config/.test(e) && /\b404\b/.test(e))
   );
 }
 


vivekchand:
## UX review

🟢 **nit (test-only)** — 8-line addition to the test-allowlist filter in `tests/e2e/cloud-contract.mjs`. No user-visible surface change, no template/JS/CSS. The Insights tab still hides correctly when `CLAWMETRY_INSIGHTS=1` is unset — this just stops the browser's auto-logged 404 from tripping the cloud-deploy gate. Comment in the diff is honest about it being a temporary fix until #1431 ships the proper server-side enabled:false response.

🤖 Pre-merge review @ 55c33dc. Pushing new commits will re-trigger review if diff >50 lines or >2h elapsed.

vivekchand:
## Product review

🟢 **good — unblocks revenue chain.** Cloud was stuck on 0.12.231; 0.12.234 deploy bounced. Every minute prod runs stale OSS means alerts/insights/Budget IA fixes don't reach Cloud users. Ship.

🟢 **nit — allowlist debt accumulating.** This is the 3rd entry in `isHarmlessConsoleError` for the same class ("cloud-disabled returns 4xx but JS handles it"). Pattern shows we keep shipping IIFE probes that 404 on cloud. The systemic fix is what #1431 asks for: server returns 200 `{enabled:false}` so probes succeed and tabs render with the correct lock state. Prioritize #1431 above other carryover P0s — it cleans up this whole class.

🟢 **nit — `[RELEASE]` tag is right.** Cloud's deploy gate pulls the contract spec from OSS main, so the fix has to ship to PyPI before cloud sees it. Without [RELEASE], the deploy gate would still fail on the next cloud push.

🤖 Pre-merge review @ 55c33dc. Pushing new commits will re-trigger review if diff >50 lines or >2h elapsed.

PR #1437, opened 2026-05-16
[RELEASE] fix(insights): /api/insights/config GET returns 200 {enabled} (closes #1431)

### LINKED ISSUE #1431: [Product P0 from PR #1428] Insights tab invisible-when-off destroys Pro funnel

vivekchand:
## Summary
Closes #1431 (Product P0). Removes the allowlist debt class entirely + sets up the Pro-locked tab pattern.

## Before / After
**Before:** GET `/api/insights/config` returned 404 when `CLAWMETRY_INSIGHTS=1` was unset. The dashboard's nav-tab-reveal IIFE in app.js probes this endpoint on every page load; on cloud (flag always unset) the 404 tripped the cloud-contract gate's "zero unexpected JS errors" check. PR #1435 added a `console.error` allowlist as a stopgap — that was the **third entry** in the allowlist for the same anti-pattern class (`/api/diagnostics` 410, `/api/config-diagnostics` 404, `/api/insights/config` 404).

**After:**
- GET always 200. `{enabled: false}` when off, `{enabled: true, ...cfg}` when on. Cheap probe; no payload leaked when off.
- POST still 404 when off (writes require feature opt-in).
- `/insights` HTML page + preview/history/send-now unchanged (still 404 when off — they have no JS-probe risk).
- app.js IIFE checks `d.enabled === true` instead of `r.ok`.
- Cloud-contract allowlist entry from PR #1435 removed (no longer needed).

## Test plan
- [x] `pytest tests/test_insights.py` — 11/11 pass (added 1 new test asserting GET returns 200 + POST still 404)
- [x] `node --check tests/e2e/cloud-contract.mjs` clean
- [ ] CI matrix
- [ ] After release: cloud deploy retry should pass without the allowlist entry (verifies the systemic fix)

## Knock-on
Future Pro-locked-tab UX can now render an "Insights 🔒" tab pointing to a paywall modal when `enabled=false`, instead of hiding the feature entirely. That's the user-facing half of #1431 (still tracked); ships as a separate follow-up.

### DIFF
--- a/clawmetry/static/js/app.js
+++ b/clawmetry/static/js/app.js
@@ -13195,14 +13195,17 @@ function _ncEsc(s) {
 })();
 } } // end if(false) stub */
 
-// On page load: reveal Insights tab if the feature flag is on. Insights endpoints
-// 404 with `{"error":"feature_disabled"}` when CLAWMETRY_INSIGHTS=1 is unset, so
-// a fast HEAD-style probe of /api/insights/config keeps the tab hidden by default
-// (no surface for users who haven't opted in yet) without leaking a spurious 200.
+// On page load: reveal Insights tab when the feature flag is on. The endpoint
+// always returns 200 (since #1431/PR closing); when disabled it returns
+// `{enabled: false}`, when enabled it returns the full config + `enabled: true`.
+// Probing by JSON flag instead of HTTP status keeps the browser console clean
+// on cloud-rendered dashboards where CLAWMETRY_INSIGHTS=1 is unset.
 (function() {
   try {
     fetch('/api/insights/config').then(function(r) {
-      if (r.ok) {
+      return r.ok ? r.json() : null;
+    }).then(function(d) {
+      if (d && d.enabled === true) {
         var t = document.getElementById('insights-tab');
         if (t) t.style.display = '';
       }
--- a/routes/insights.py
+++ b/routes/insights.py
@@ -127,18 +127,34 @@ def api_send_now():
 
 @bp_insights.route("/api/insights/config", methods=["GET", "POST"])
 def api_config():
-    gated = _gated()
-    if gated is not None:
-        return gated
+    # GET always returns 200 with `{enabled: bool, ...}` so the dashboard's
+    # nav-tab-reveal probe (`app.js` IIFE checking /api/insights/config) can
+    # render the tab in either an active or a Pro-locked state without the
+    # browser console-erroring on a 404. POSTs (writes) still 404 when the
+    # feature flag is off — writes only make sense when the feature is on.
+    #
+    # Fixes #1431 (Pro-locked vs invisible-when-off) AND removes the need
+    # for the cloud-contract 404 allowlist entry from PR #1435.
     from clawmetry.insights import load_config, save_config
+    enabled = _feature_enabled()
     if request.method == "POST":
+        if not enabled:
+            return jsonify({
+                "error": "feature_disabled",
+                "hint": "Set CLAWMETRY_INSIGHTS=1 to enable Weekly Insights Digest.",
+            }), 404
         data = request.get_json(silent=True) or {}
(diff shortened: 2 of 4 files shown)


vivekchand:
## UX review

🟢 **good — sets up the right UX foundation.** Switching from `r.ok` (HTTP gate) to `d.enabled === true` (payload gate) is the difference between "feature might exist somewhere" and "feature has an explicit state." The latter lets a future Pro-locked tab actually render the lock icon instead of vanishing.

🟢 **nit — `enabled: true` field is also now sent inside the full-config payload.** Frontend can rely on it as a stable signal across both states (off → `{enabled:false}`; on → `{enabled:true, ...cfg}`). Future code shouldn't conditionally check for the field's existence.

🟢 **nit — POST stays 404 when off.** Right call. Writes only make sense after opt-in. A user POSTing to a disabled endpoint deserves the 404 + `feature_disabled` hint.

🤖 Pre-merge review @ 1de2746. Pushing new commits will re-trigger review if diff >50 lines or >2h elapsed.

vivekchand:
## Product review

🟢 **good — kills an entire bug class.** `/api/diagnostics` 410, `/api/config-diagnostics` 404, `/api/insights/config` 404 — three allowlist entries in cloud-contract for the same pattern: "JS probe gets 4xx, browser logs console.error, smoke gate trips." This PR sets a pattern (probes return 200 with a state flag) that future feature gates should follow, preventing entries 4, 5, 6.

🟢 **good — unblocks the user-facing Pro-locked tab.** #1431's product P0 was "don't hide Insights from free users; show a locked tab that opens a paywall." Can't render a locked tab if the probe returns 404 (no signal to show anything). With `enabled: false` returned, the next UX PR can render `Insights 🔒` → upsell modal.

🟡 **P1 — POST 404 hint mentions the env var.** `Set CLAWMETRY_INSIGHTS=1 to enable` is engineer-speak. A non-technical user can't ssh into a daemon and export an env var. The hint should point to a UI Settings checkbox once that lands (per `feedback_simple_ui_for_nontechnical.md`). Tracked, not blocking — UI Settings is a separate ticket from #1431's tab-reveal work.

🤖 Pre-merge review @ 1de2746. Pushing new commits will re-trigger review if diff >50 lines or >2h elapsed.

PR #1438, opened 2026-05-16
[RELEASE] fix(e2e): re-add insights/config 404 allowlist (chicken-and-egg unblock)

vivekchand:
## Summary
Unblocks the publish chain for the systemic insights fix.

## Problem
PR #1437 [RELEASE] (insights config GET → 200) merged at 10:13Z but FAILED to publish because:

1. \`release-on-merge.yml\` runs cloud-contract spec against PROD (\`https://app.clawmetry.com\`) before bumping the version and publishing to PyPI.
2. PROD is pinned to clawmetry==0.12.235, which still returns 404 on \`/api/insights/config\` (the OLD pre-#1437 behavior).
3. PR #1437 removed the cloud-contract allowlist entry for that 404.
4. Without the allowlist + with PROD returning 404 → "zero unexpected JS errors" check fails → publish blocked → cloud can never update → permanent deadlock.

## Fix
Re-add the allowlist entry. Tagged `[RELEASE]` so it ships in the next OSS version.

## Rollout sequence (after this merges)
1. This PR merges + auto-bumps to 0.12.236, publishes to PyPI.
2. Open cloud Dockerfile bump PR pinning clawmetry==0.12.236 (includes #1437's server-side 200 behavior).
3. Cloud deploys 0.12.236; PROD now returns 200 on /api/insights/config.
4. **Then** open a new OSS PR that removes BOTH allowlist entries (this one + PR #1435's original). Pre-publish gate will pass cleanly since PROD now matches the new spec.

## Lesson learned
When changing a behavior AND removing its corresponding allowlist in one PR, make sure the spec runs against the CANDIDATE build, not PROD. Pre-publish gate logic in \`release-on-merge.yml\` needs an option to run against the candidate revision (similar to cloud's deploy.yml \`--tag candidate\` pattern). Tracking as a separate follow-up.

### DIFF
--- a/tests/e2e/cloud-contract.mjs
+++ b/tests/e2e/cloud-contract.mjs
@@ -82,11 +82,19 @@ function isHarmlessConsoleError(e) {
     // shape: new OSS endpoint, returns 404 on cloud. Filtering
     // them here matches the existing /api/skills 410 pattern.
     (/\/api\/diagnostics/.test(e) && /\b410\b/.test(e)) ||
-    (/\/api\/config-diagnostics/.test(e) && /\b404\b/.test(e))
-    // Note: /api/insights/config 404 allowlist (added in PR #1435) was
-    // removed when issue #1431 landed — the endpoint now returns 200
-    // {enabled: false} when CLAWMETRY_INSIGHTS=1 is unset, so the probe
-    // succeeds cleanly without console.error.
+    (/\/api\/config-diagnostics/.test(e) && /\b404\b/.test(e)) ||
+    // Temporarily restored. PR #1437 changed /api/insights/config to return
+    // 200 {enabled: false} instead of 404 and removed this allowlist entry
+    // — but the contract spec runs against PROD (https://app.clawmetry.com),
+    // which is still pinned to a clawmetry version BEFORE #1437. So the
+    // pre-publish gate sees the old 404 and blocks #1437 from publishing.
+    // Classic chicken-and-egg: the OSS fix can't ship until cloud has it,
+    // and cloud can't get it until OSS ships.
+    //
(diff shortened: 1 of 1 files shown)


vivekchand:
## UX review

🟢 **nit (test-allowlist only)** — Re-adds 8 lines to the cloud-contract harmless-errors filter that PR #1437 removed prematurely. No user-facing surface. Server still returns 200 + {enabled} per #1437 — this just paper-overs the pre-publish gate's cross-version skew until cloud catches up.

🤖 Pre-merge review @ 7cb3346. Pushing new commits will re-trigger review if diff >50 lines or >2h elapsed.

vivekchand:
## Product review

🟢 **good — unblocks the chain in the cheapest way.** Self-inflicted deadlock (see memory `feedback_ws_diagnostic_caused_outage.md` — when prod regresses, check YOUR most recent change first). Restore + ship is the right move; don't try to be clever with workflow_dispatch overrides.

🟡 **P1 — file a follow-up: pre-publish gate should run against CANDIDATE, not PROD.** Same pattern cloud's `deploy.yml` uses (`--tag candidate` revision → flip traffic only after gate passes). This deadlock will recur every time we ship a server-behavior change + spec change in one PR. Worth ~2h to plumb a candidate URL into `release-on-merge.yml`.

🟡 **P1 — sequence-of-changes lesson should land as a memory entry.** Standing pattern: when a behavior change AND its corresponding allowlist removal both ship at once, separate into 2 PRs landing in order: (1) behavior change + keep allowlist, (2) wait for cloud pin to catch up, (3) remove allowlist. I'll add this to memory after the chain unblocks.

🤖 Pre-merge review @ 7cb3346. Pushing new commits will re-trigger review if diff >50 lines or >2h elapsed.

PR #1444, opened 2026-05-16
fix(usage): /api/usage dedupe loop no longer drops tool-call tokens (MOAT regression fix)

vivekchand:
## Summary

`/api/usage` was silently dropping ~33% of tokens whenever a day had any non-message rows with a `token_count` column (tool_call rows in the MOAT verifier, retries in real data). The `test_tool_call_event_surfaces_everywhere` MOAT regression test caught it:

> AssertionError: usage day-totals lost tokens: got 150, expected >=225

## Root cause

- `query_aggregates` sums `token_count` across every event row (messages + tool_calls + anything else with a value).
- `query_daily_usage_splits` walks only billable-turn types (`assistant` / `model.completed` / `message`) and dedupes sibling-double-counted pairs.
- The old `if deduped > 0: daily_tokens[d] = deduped` overwrote the aggregates total — losing every non-message token contribution.

## Fix

Replace the blind overwrite with a small decision tree that still respects the original sibling-dedupe purpose:

```
raw >= 2*deduped → sibling-doubling dominates; subtract doubled half and
                    carry any residual non-message tokens through.
raw <  2*deduped → no full sibling pair (synthetic / partial install);
                    keep whichever is larger so we never lose data.
```

## Verification

- `pytest tests/test_moat_send_message_e2e.py -q` → **3 passed** (was 2 passed, 1 failed)
- `pytest tests/ -k usage -q` → **39 passed, 0 regressions**

## MOAT context

This was caught by the user's MOAT mandate ("100% perfect by EOD"). The verifier sends a synthetic OpenClaw message (1 session_start + 2 tool calls + 1 message totaling 225 tokens) and asserts every downstream surface returns the correct data. Before this fix, `/api/usage` was silently lossy on real installs too.

## Test plan

- [x] MOAT verifier: 3/3 pass
- [x] All 39 usage tests pass
- [ ] CI passes
- [ ] Manual: real OpenClaw install, send a message, confirm Tokens tab sums correctly

### DIFF
--- a/routes/usage.py
+++ b/routes/usage.py
@@ -238,18 +238,30 @@ def _try_local_store_usage():
         if cost_from_split > 0 and daily_cost.get(d, 0.0) <= 0:
             daily_cost[d] = cost_from_split
 
-    # Issue #1394: prefer the deduped (input+output) total over the raw
-    # ``token_count`` aggregate when splits are available. The raw
-    # column counts BOTH the ``assistant`` and the sibling
-    # ``model.completed`` event for each LLM turn (different log
-    # writers race-emit them ~100-300ms apart) — sums end up roughly
-    # 2× the real billable total. The deduped splits already collapsed
-    # those sibling pairs, so adding their input+output is the
-    # billable-token answer the harness compares against.
+    # Issue #1394 + MOAT regression 2026-05-16: raw ``token_count`` aggregate
+    # counts BOTH the ``assistant`` and the sibling ``model.completed`` event
+    # for each LLM turn (different writers race ~100-300ms apart) AND any
+    # non-billable-turn rows with a token_count column (tool_call rows in
+    # synthetic harnesses, retries, etc.). The splits walker dedupes the
+    # sibling pair but ignores non-message rows entirely, so blindly
+    # overwriting raw with deduped silently drops those tokens.
+    #
+    # Decision: if raw >= 2*deduped, the sibling-doubling bug dominates →
+    # subtract the doubled half and add any residual non-message tokens
+    # back on top. If raw < 2*deduped, no full sibling pair exists for that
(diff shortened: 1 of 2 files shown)


vivekchand:
## UX review

- 🟢 **Backend-only fix.** Diff is `routes/usage.py` aggregation arithmetic — no HTML/CSS/JS surface touched. Skipping the visual scrutiny. Worth noting: the Tokens tab will now display 33% more tokens for installs with non-message token rows (the fix surface), which is a "numbers move" users will notice. A one-line release-note bullet ("Tokens tab now includes tool-call token contributions") would head off "why did my count change" support questions.

🤖 Pre-merge review @ 6bcf55e4. Pushing new commits will re-trigger review if diff >50 lines or >2h elapsed.

vivekchand:
## Product review

- 🟢 **Backend correctness fix, no tier-strategy implication.** The aggregation bug was silently under-reporting OSS users' token usage by ~33%. Fixing it strengthens (not leaks) Cloud-Pro positioning: accurate OSS-side numbers make the budget-alert / over-spend signals trustworthy, which is the precise moment users decide to upgrade for Slack/PagerDuty fanout. Bug being caught by the MOAT verifier (not by a user) is the right outcome.
- 🟡 **First-principles check on the heuristic threshold.** The new `raw >= 2*deduped` boundary assumes sibling double-counting is always ~2×. On installs where the sibling pair only emits ~70% of the time (writer race lost), raw could be `1.7*deduped` and the branch flips, taking `max(raw, deduped)` instead of subtracting the partial double-count. Worth a follow-up: instead of a magic threshold, query `count(event_type='model.completed')` per day to know whether siblings exist before applying the subtract logic. Not a blocker for this PR.

🤖 Pre-merge review @ 6bcf55e4. Pushing new commits will re-trigger review if diff >50 lines or >2h elapsed.

PR #1445, opened 2026-05-16
feat(flow): cap Flow Runs to 24h for OSS, gate full retention behind Cloud-Pro (closes #1173)

### LINKED ISSUE #1173: [Product P0 from PR #1157] OSS Flow > Runs leaks Cloud-Pro >24h retention
From the post-merge product review of PR #1157.

Per memory `project_alerts_pro_feature.md`: >24h retention is a Cloud-Pro feature. PR #1157's new `/api/flow/runs` endpoint accepts `since`/`until` ISO params with no upper bound on how far back it reads (`routes/infra.py:312-360`). DuckDB retains forever on the local disk, so OSS users get unlimited historical flow-run retention in the dashboard for free.

This torches the Cloud-Pro retention upsell — users have no reason to pay for hosted history when local history is already infinite and exposed in the UI.

Actions:
- Cap OSS reads to last 24h of `started_at` in `query_flow_runs` (or in the route)
- When users hit the cap, surface a "Want >24h history? Upgrade to Cloud-Pro" CTA below the table
- Confirm the same cap applies to Sessions list and any other surfaces that read from `events` with a time filter
- Verify Cloud-Pro builds bypass the cap (env flag or build-time switch)

Related: `routes/infra.py:340-358`, `clawmetry/local_store.py:1392-1474`.

vivekchand:
## Summary

- OSS / Cloud-Free users now see only the last 24 hours of flow runs in `Flow > Runs`. Cloud-Pro users (validated via `_is_pro_user()` from PR #1440) keep unlimited history.
- Response carries `capped_at_24h: true` when the cap kicks in; the table footer renders an "Upgrade to Cloud-Pro for unlimited history" CTA linking to https://app.clawmetry.com/upgrade.
- Restores the Cloud-Pro retention upsell that PR #1157 leaked when `/api/flow/runs` shipped with no upper bound on `since`. DuckDB retains forever locally, so OSS users were getting unlimited historical flow-run retention for free, killing the hosted-history paywall.

## Gating contract

- `_is_pro_user()` is fail-closed: returns True only when both (a) a `cm_` cloud token is on disk and (b) the cached plan is `cloud_pro` / `pro` / `trial`. Reuses the existing helper, no new gating mechanism.
- If the request hits the cap, the response includes `capped_at_24h: true` so the UI can render the upgrade row. Otherwise the flag is `false`.

## Test plan

- [x] `python3 -m pytest tests/test_flow_runs_endpoint.py -q` — 8/8 pass
- [x] `python3 -m pytest tests/ -q -k "flow or usage"` — 57 pass, 2 pre-existing unrelated failures (telegram E2E + real-binary E2E, both fail on origin/main as well)
- [x] New tests: OSS caller sees `capped_at_24h=true` and the 8-day-old session is filtered out; Pro caller sees `capped_at_24h=false` and gets the full history.
- [x] Diff is 132 LOC across 3 files (under the 300-LOC cap).
- [x] No em-dashes / double-dashes in any new user-facing copy.
- [x] Existing aggregation tests preserved by defaulting the fixture to a Pro user (their seeded timestamps fall outside the 24h window today).

## Follow-up (noted, separate audit issue)

The issue body asked us to "Confirm the same cap applies to Sessions list and any other surfaces that read from `events` with a time filter." That's a broader audit scoped intentionally outside this PR. Surfaces to walk in a follow-up:

- `/api/sessions` and its DuckDB-backed query paths
- `/api/usage` historical reads (token / cost analytics)
- `/api/brain-history` paginated reasoning events
- `/api/local-events` and any other thin local-store passthroughs

Closes #1173.

### DIFF
--- a/clawmetry/static/js/app.js
+++ b/clawmetry/static/js/app.js
@@ -8626,9 +8626,11 @@ function loadFlowRuns() {
     .then(function(r) { return r.json(); })
     .then(function(d) {
       var runs = (d && d.runs) || [];
+      var cappedAt24h = !!(d && d.capped_at_24h);
       if (countEl) countEl.textContent = runs.length + ' run' + (runs.length === 1 ? '' : 's');
       if (!runs.length) {
-        tbody.innerHTML = '<tr><td colspan="8" style="padding:24px;text-align:center;color:var(--text-muted);font-size:12px;">No historical flow runs yet — the live Flow view will populate this once a session completes.</td></tr>';
+        tbody.innerHTML = '<tr><td colspan="8" style="padding:24px;text-align:center;color:var(--text-muted);font-size:12px;">No historical flow runs yet. The live Flow view will populate this once a session completes.</td></tr>';
+        _renderFlowRunsCap(cappedAt24h);
         return;
       }
       var rows = runs.map(function(r) {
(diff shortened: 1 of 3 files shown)


vivekchand:
## UX review

- 🟢 **Plain-language CTA copy + no em-dashes.** "Showing the last 24 hours. Upgrade to Cloud-Pro for unlimited history" reads as one sentence with one action. Footer renders both on populated table and empty state — good.
- 🟡 **Footer attaches as `parentElement.appendChild` outside the table.** Works, but on the first render the table re-flows when the footer mounts (~30-40px shift). Reserve the slot at SSR (empty `<div id="flow-runs-cap-cta">` always present below the table, hidden when not capped) to avoid the bump. file: `app.js:8674`.
- 🟢 **Empty-state copy de-em-dashed in the same change** (`"No historical flow runs yet — the live..."` → `"No historical flow runs yet. The live..."`) — respects the no-em-dash rule.
- 🟡 **CTA accent color is hardcoded fallback `#7c5cff`.** If a user has a custom theme that overrides `--accent`, the fallback ignores it. Drop the literal hex and let the CSS var inherit — falls back to whatever the theme's default is. file: `app.js:8688`.

🤖 Pre-merge review @ 5bc4625e. Pushing new commits will re-trigger review if diff >50 lines or >2h elapsed.

vivekchand:
## Product review

- 🟢 **Tier-strategy correct.** OSS / Cloud-Free capped at 24h, Cloud-Pro bypasses. Cap kicks in → upgrade CTA appears at the exact point of friction (user wanted >24h, can't have it). High-converting placement.
- 🔴 **Cap-leakage on adjacent surfaces.** The agent's own report identified `/api/sessions`, `/api/usage`, `/api/brain-history`, `/api/local-events` as also lacking the cap. As-is, a free user just opens the Sessions tab and gets unlimited history there — defeating the gate this PR ships. Either land a single PR that gates ALL `since`-accepting routes, OR file 4 follow-up issues today so this gate isn't a fig leaf. Without that, the only thing this PR achieves is annoying users who use the Flow tab specifically.
- 🟡 **Fixture monkeypatches `_is_pro_user → True` to keep old aggregation tests green.** Smells. The pre-existing tests should be updated to use timestamps inside the 24h window OR explicitly assert both tiers. As-is, the cap is silently bypassed in all pre-existing test scenarios — easy to regress without noticing.
- 🟢 **Reuses `_is_pro_user()` from PR #1440** rather than introducing a new gating mechanism — single source of truth for the Pro check.

🤖 Pre-merge review @ 5bc4625e. Pushing new commits will re-trigger review if diff >50 lines or >2h elapsed.

PR #1446, opened 2026-05-16
fix(usage): cost-comparison no longer 2× counts v3 sibling pairs (MOAT audit follow-up)

vivekchand:
## Summary

Audit follow-up to PR #1444 (the daily-tokens dedup fix). The cost-comparison fast path had the same shape of bug: it summed `token_count` across every event without dedup. On real OpenClaw v3 installs the dual-writer pattern emits BOTH an `assistant` row AND a slim `model.completed` sibling ~100 ms apart for every LLM turn, both carrying the same `token_count` value. Result: `actual.tokens` + `actual.cost_usd` displayed 2× reality, and the "savings vs alternative model" $ amounts looked twice as good as they actually were.

## Fix

Mirrors the dedup approach already in `query_daily_usage_splits`:

1. **Pass 1** — bucket each `(session_id, ts_sec)` and record the richest envelope rank seen across a ±1 s writer-race window. `assistant`/`message` rank 2; `model.completed` rank 1.
2. **Pass 2** — count tokens once, skipping any row whose envelope rank is strictly less than its bucket max.

Non-billable-turn rows (`tool_call`, etc.) fall through untouched — they have no sibling to dedup against and their token contributions remain.

## Verification

- New regression test: synthesize an assistant + model.completed pair at the same ts, assert `actual.tokens == 150` (one deduped turn), not 300.
- `pytest tests/test_moat_send_message_e2e.py tests/test_usage_local_store.py -q` → **31 passed** (was 30, +1 from new test).
- `pytest tests/ -k usage -q` → **40 passed**.

## MOAT context

Found by the MOAT 11:26Z audit fire (following the dedupe-pattern memory note saved after PR #1444). Two surfaces flagged in that note — this is one; the other (`routes/sessions.py` v3 session-tools bypass) is a bigger refactor and gets a separate follow-up issue.

### DIFF
--- a/routes/usage.py
+++ b/routes/usage.py
@@ -608,7 +608,19 @@ def _try_local_store_cost_comparison():
     """Fast path for /api/usage/cost-comparison. Sums actual tokens/cost
     over the past 30 days from the local store, then projects costs against
     a fixed alternatives table. Mirrors ``dashboard._build_cost_comparison``
-    output shape exactly."""
+    output shape exactly.
+
+    MOAT 2026-05-16: the old implementation blindly summed ``token_count``
+    across every event row. On real v3 installs the dual-writer pattern
+    (``assistant`` + sibling ``model.completed``) double-counts every
+    billable turn, inflating ``actual.tokens`` + ``actual.cost_usd`` ~2×
+    and making the "savings vs alternative" dollar amounts look twice as
+    big as truth. We now skip the slimmer ``model.completed`` row when an
+    ``assistant``/``message`` sibling exists for the same
+    (session_id, ts ±1 s) bucket — matches the dedup approach in
+    ``query_daily_usage_splits``. Non-billable-turn rows (tool_call etc.)
+    keep their tokens since they don't have a sibling.
+    """
     store = _ls_get_store()
     if store is None:
         return None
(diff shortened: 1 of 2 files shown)


vivekchand:
## UX review

- 🟢 **Backend-only fix.** No HTML/CSS/JS surface touched. Skipping visual scrutiny. User-visible effect: the cost-comparison card's "actual tokens / cost" values will drop ~50% for real v3 installs that were previously double-counted — worth a one-line release-note bullet ("actual spend numbers now match billing").

🤖 Pre-merge review @ 87a69ed9. Pushing new commits will re-trigger review if diff >50 lines or >2h elapsed.

vivekchand:
## Product review

- 🟢 **Correctness fix removes a misleading conversion signal.** Before this PR the "you could save X by switching to model Y" $ amounts were 2× reality on real v3 installs. Users acting on inflated savings would switch models and find their bill barely moved. Damages trust in the recommendation. Fixing makes the cost-comparison surface defensible.
- 🟡 **2-pass over `query_events(limit=50000)` is O(2N) memory.** Fine at 50k but worth noting for the future. If retention grows, switching to a SQL-level dedup (window function over (session_id, ts_sec)) would scale better. Not a blocker now.
- 🟢 **Dedup logic mirrors `query_daily_usage_splits`** — single source of truth for the sibling-pair rule. Future schema changes propagate consistently.

🤖 Pre-merge review @ 87a69ed9. Pushing new commits will re-trigger review if diff >50 lines or >2h elapsed.

PR #1447, opened 2026-05-16
fix(sessions): /api/session-tools surfaces v3 standalone tool_call rows (MOAT audit follow-up)

vivekchand:
## Summary

The 3rd bypass surface from the dedupe-pattern audit (see `feedback_usage_dedupe_pattern.md`). On real OpenClaw v3 installs, tool invocations land in DuckDB as standalone `event_type='tool_call'` rows with the call+result self-contained in the `data` blob. But `/api/session-tools` only walked legacy `event_type='message'` rows looking for nested `data.message.content[].toolCall` blocks — so the session-detail "Tools" panel silently returned **0 tools** for every v3 session.

User impact: open any session-detail page on a real install, see no tools listed even though the agent ran 20+ of them.

## Fix

Extend the fast path to also accept `event_type='tool_call'` rows as self-contained call+result pairs:

- Synthesize a `tool_call_id` from the event id
- Extract tool name from `data.tool` / args from `data.args` / result from `data.result`
- Tolerate `is_error` / `isError` / `error` flags from various daemon writers
- Legacy nested-toolCall parsing preserved so older sessions still work

## Verification

The MOAT verifier (`test_moat_send_message_e2e.py`) used to deliberately skip `/api/session-tools` with a comment explaining the gap:

> (We deliberately skip /api/session-tools here: that endpoint only fires on legacy `event_type=='message'` rows ...)

That comment is now removed and replaced with a hard assertion:

```python
assert tool_names == ["Bash", "Read"]
assert body.get("stats", {}).get("total_calls") == EXPECTED_TOOL_CALLS  # 2
```

- `pytest tests/test_moat_send_message_e2e.py -q` → **3 passed** (was 3 passed before, but #2 was a weaker assertion that skipped this endpoint).

## Out of scope

The pre-existing `test_session_tools_fast_path` in `test_duckdb_coverage_phase3_2026_05_13.py` fails on both `main` and this branch (daemon-discovery monkeypatch leak — unrelated). Separate cleanup PR.

## MOAT context

3rd of 3 dedupe-pattern surfaces flagged in this morning's audit:
- ✅ `/api/usage` daily-tokens (PR #1444 merged)
- ✅ `/api/usage/cost-comparison` sibling double-count (PR #1446 open)
- ✅ `/api/session-tools` v3 tool_call rows (this PR)

After these three the MOAT verifier covers every fast-path the synthetic verifier touches.

### DIFF
--- a/routes/sessions.py
+++ b/routes/sessions.py
@@ -675,15 +675,65 @@ def _try_local_store_session_tools(sid: str, args_chars: int, result_chars: int,
     calls: dict = {}
     result_by_id: dict = {}
     turn_index = 0
-    saw_message = False
+    saw_any = False
     for ev in rows:
-        if ev.get("event_type") != "message":
-            continue
-        saw_message = True
+        et = ev.get("event_type")
         data = ev.get("data") if isinstance(ev.get("data"), dict) else {}
+        ev_ts_ms = _parse_ts(
+            (data.get("timestamp") if isinstance(data, dict) else None)
+            or ev.get("ts")
+        )
+
+        # v3 standalone tool_call row: self-contained call+result pair.
+        # Real OpenClaw v3 emits these instead of the legacy nested
+        # ``data.message.content[].toolCall`` blocks. Synthesize a tcid
+        # from the event id so the pairing logic below treats it as
+        # an immediately-resolved call.
+        if et == "tool_call":
+            saw_any = True
+            turn_index += 1
+            tcid = ev.get("id") or f"tc-{ev_ts_ms}-{len(calls)}"
+            tool_name = ""
+            args_val = None
+            result_val = None
+            is_error = False
+            if isinstance(data, dict):
+                tool_name = (data.get("tool") or data.get("tool_name")
+                             or data.get("name") or "")
+                args_val = (data.get("args") if data.get("args") is not None
+                            else data.get("arguments"))
+                result_val = data.get("result")
+                is_error = bool(data.get("is_error") or data.get("isError")
+                                or data.get("error"))
+            calls[tcid] = {
+                "tool_call_id":     tcid,
+                "tool_name":        tool_name,
+                "arguments":        _truncate(args_val, args_chars),
+                "start_ms":         ev_ts_ms,
+                "turn_index":       turn_index,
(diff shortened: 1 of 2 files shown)


vivekchand:
## UX review

- 🟢 **Real visible improvement.** Before this PR, session-detail "Tools" panel on every v3 install rendered an empty table — users likely concluded "this feature is broken" or "my agent didn't use tools." Now they see Bash, Read, Edit, etc., listed with timing + error state. Worth a one-line release note ("Session detail Tools panel now shows v3 agent tool invocations").
- 🟡 **Synthesized tool_call_id when blob has no `id`.** Falls back to `f"tc-{ev_ts_ms}-{len(calls)}"`. Two tool_call rows within the same ms (rapid-fire agent) would collide and only the last would render. Low-probability but possible — consider including a UUID suffix or the event's own DB id as the entropy source. file: `sessions.py:697`.
- 🟢 **Legacy nested-toolCall parsing kept intact** — older sessions (pre-v3) still display correctly.

🤖 Pre-merge review @ f2a40df6. Pushing new commits will re-trigger review if diff >50 lines or >2h elapsed.

vivekchand:
## Product review

- 🟢 **Closes a silent-data-loss class of bug.** Tools panel returning 0 for real installs was the kind of regression that quietly costs trust ("ClawMetry doesn't see my tools, why am I paying for this?"). Fixing it before users hit the feature is exactly the right time.
- 🟢 **MOAT verifier strengthened — skip comment removed, hard assertion added.** Future regressions in this code path will fail CI rather than slip through with a comment "we deliberately skip this." That's the right direction for the DuckDB-first MOAT mandate.
- 🟡 **First-principles question on legacy keepalive.** Both v3 standalone rows AND nested toolCall blocks are now supported. Long-term: is there any install still emitting the legacy nested format? If not (or if it's been deprecated for 6+ months), the legacy branch is dead weight. Worth a follow-up audit on real telemetry — if 0 legacy rows in 30 days of ingested data, kill the legacy branch and halve the function's complexity.

🤖 Pre-merge review @ f2a40df6. Pushing new commits will re-trigger review if diff >50 lines or >2h elapsed.

PR #1452, opened 2026-05-16
feat(ci): API-latency smoke gate (sidecar daemon + dashboard, p50<500ms / p95<1500ms) (closes #1241)

### LINKED ISSUE #1241: [Product P0 from PR #1235] Add API-latency smoke test in CI to catch broken-under-real-load endpoints
From the post-merge product review of PR #1235.

PR #1235 fixed three endpoints that were 7-15s under the standard install (sync-daemon writer + dashboard reader, both processes alive). The bugs sat in `main` for months and only surfaced when external traffic arrived after Peter Steinberger starred the repo. We had **no CI signal** that `/api/overview` was 12-15s on a real install.

### Root cause

CI runs the dashboard in single-process / dev mode where the sync daemon doesn't exist as a separate process. Without two-process contention there is no DuckDB lock fight, so `get_store()` returns immediately. The bugs are invisible to the existing test matrix.

### Proposed CI gate — `api-latency-smoke.yml`

1. Spin up a sidecar sync daemon process (the real one, not a mock) writing to a seeded `clawmetry.duckdb` (1k events, 200 heartbeats, 50 memory blobs — realistic-shape fixture).
2. `pip install` the just-built wheel into a fresh venv.
3. Start the dashboard against that DuckDB.
4. `curl` every endpoint listed in `CLAUDE.md` (Key API Endpoints section) — `/api/overview`, `/api/sessions`, `/api/subagents`, `/api/transcript/<id>`, `/api/usage`, `/api/flow`, `/api/brain-history`, `/api/crons`, `/api/system-health`, `/api/nodes`, `/api/budget/*`, `/api/alerts/*`, `/api/heartbeat-status`, `/api/memory-files`.
5. Run each endpoint 5x, assert **p50 < 500ms** and **p95 < 1500ms**. Fail the build on any breach.
6. Run on every PR (not just nightly) — this is fast (≤30s wall-clock if endpoints are healthy).

### Bonus

Add a nightly job that does the same thing against the **published PyPI release** (not the PR build), so we catch regressions introduced by transitive dependency upgrades or environment drift.

### Why P0

We got lucky once. The funnel-killing failure mode is \"influencer arrives, dashboard is broken, they tweet about it.\" That cost is unbounded. The CI cost is one ~30s job per PR. Asymmetric.

Related: companion P0 issue on sweeping the ~28 remaining RW `get_store()` callsites.

vivekchand:
## Summary
- Adds `.github/workflows/api-latency-smoke.yml` — spins up the sync daemon **and** the dashboard as separate processes (the real two-process shape every `pip install clawmetry` user runs), seeds a realistic DuckDB fixture, then probes the key endpoints from CLAUDE.md and asserts **p50 < 500ms, p95 < 1500ms**. Fails the build on any breach.
- Adds `scripts/seed_smoke_duckdb.py` — 1k events + 200 heartbeats + 50 memory blobs + 5 sessions across the last 7 days. Deterministic seed.
- Adds `scripts/api_latency_probe.py` — 5 samples / endpoint, stdlib-only linear-interpolation percentile, prints a verdict table, exits 1 on breach.
- Bonus nightly job: same workflow shape, but installs `pip install clawmetry` from PyPI instead of the local wheel — catches regressions caused by transitive dep upgrades or environment drift.

## Why P0 (from #1241)
PR #1235 fixed three endpoints that were **7-15s** under the standard install (daemon + dashboard, two processes alive). The bugs sat in main for months and only surfaced when external traffic arrived. Single-process CI never saw the DuckDB writer/reader lock fight, so the slow paths were invisible. This gate would have caught all three.

## Verification
Ran the full workflow locally against a seeded `clawmetry.duckdb` with the real `clawmetry sync` daemon as a sidecar:

```
endpoint                                      p50      p95      max  status  verdict
------------------------------------------------------------------------------------
/api/overview                                 63ms      80ms      84ms   200    OK
/api/sessions                                  9ms      11ms      12ms   200    OK
/api/subagents                                 3ms     221ms     275ms   200    OK
/api/transcript/smoke-sess-000                 5ms       6ms       6ms   200    OK
/api/usage                                    73ms     107ms     113ms   200    OK
/api/flow                                      1ms       2ms       2ms   200    OK
/api/brain-history                             9ms      10ms      10ms   200    OK
/api/crons                                   167ms     183ms     185ms   200    OK
/api/system-health                           178ms     210ms     213ms   200    OK
/api/nodes                                     2ms       3ms       3ms   200    OK
/api/budget/status                          7630ms    9531ms   10006ms   200    FAIL
/api/alerts/rules                              4ms       8ms       8ms   200    OK
/api/heartbeat-status                          5ms       7ms       7ms   200    OK
/api/memory-files                              5ms       5ms       5ms   200    OK
```

- Probe exits **1** when any endpoint breaches budgets (verified with `CLAWMETRY_SMOKE_P50_MS=0` artificial-slow test).
- Probe exits **0** when slow endpoint is skipped via `CLAWMETRY_SMOKE_SKIP=/api/budget/status`.

## Pre-existing slow path the gate caught locally
- **`/api/budget/status` — p50 ~7.6s, p95 ~9.5s, max 10s timeout**. This is exactly the failure mode #1241 was filed to prevent ("influencer arrives, dashboard is broken"). Worth a follow-up issue / daemon-proxy fix; not blocking on this PR since the gate is about catching such bugs going forward.

## Test plan
- [x] `python3 -m py_compile` both scripts
- [x] `yaml.safe_load` the workflow
- [x] Seed script creates a DuckDB readable by an independent process (counts: 1000 events / 5 sessions / 200 heartbeats / 50 memory blobs)
- [x] Probe runs end-to-end against `clawmetry sync` daemon + `clawmetry` dashboard
- [x] Probe exits 1 on real breach, 0 on pass, respects `CLAWMETRY_SMOKE_SKIP`
- [ ] Watch the first CI run on this PR; iterate if any unexpected pre-existing slow paths surface

### DIFF
--- a/.github/workflows/api-latency-smoke.yml
+++ b/.github/workflows/api-latency-smoke.yml
@@ -1,0 +1,124 @@
+name: API Latency Smoke
+
+# Issue #1241 — catches "fast in CI, broken in real two-process install"
+# regressions like PR #1235 (3 endpoints were 7-15s under the sync-daemon
+# + dashboard contention). Single-process CI never sees DuckDB lock fights.
+
+on:
+  pull_request:
+  push:
+    branches: [main, master]
+  schedule:
+    - cron: "0 6 * * *"   # nightly PyPI-release run
+  workflow_dispatch:
+
+concurrency:
+  group: api-latency-smoke-${{ github.ref }}
+  cancel-in-progress: true
+
+jobs:
+  smoke:
+    name: ${{ matrix.source == 'pypi' && 'PyPI nightly' || 'PR build' }}
+    runs-on: ubuntu-latest
+    timeout-minutes: 10
+    strategy:
+      fail-fast: false
+      matrix:
+        # PR + push runs the local-wheel slice only. Nightly schedule
+        # additionally exercises the published PyPI release.
+        source: ${{ github.event_name == 'schedule' && fromJSON('["wheel", "pypi"]') || fromJSON('["wheel"]') }}
+    env:
+      CLAWMETRY_TOKEN: ci-smoke-token
+      OPENCLAW_GATEWAY_TOKEN: ci-smoke-token
+      CLAWMETRY_LOCAL_STORE_PATH: ${{ github.workspace }}/.smoke/clawmetry.duckdb
+    steps:
+      - uses: actions/checkout@v4
+      - uses: actions/setup-python@v5
+        with:
+          python-version: "3.11"
+
+      - name: Build wheel
+        if: matrix.source == 'wheel'
+        run: |
+          python -m pip install --upgrade pip build
+          python -m build --wheel
+
+      - name: Install into isolated venv
+        run: |
+          python -m venv /tmp/vsmoke
+          /tmp/vsmoke/bin/pip install --upgrade pip
+          if [ "${{ matrix.source }}" = "pypi" ]; then
+            /tmp/vsmoke/bin/pip install --no-cache-dir clawmetry flask waitress cryptography duckdb requests
+          else
+            /tmp/vsmoke/bin/pip install dist/clawmetry-*.whl flask waitress cryptography duckdb requests
+          fi
+
+      - name: Write minimal daemon config
+        # The sync-daemon entry point refuses to start without an api_key,
+        # so we drop a fake one. Ingest URL points at a closed port so
+        # network POSTs fail-fast and don't hold up the loop — we only
+        # care about the daemon's local_query HTTP server here.
+        run: |
+          mkdir -p "$HOME/.clawmetry" .smoke
+          cat > "$HOME/.clawmetry/config.json" <<EOF
+          {
+            "node_id": "smoke-ci-node",
+            "api_key": "smoke-ci-fake",
+            "encryption_key": "AAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAAA=",
+            "ingest_url": "http://127.0.0.1:9"
+          }
+          EOF
+          chmod 600 "$HOME/.clawmetry/config.json"
+
+      - name: Seed realistic-shape DuckDB fixture (1k events / 200 hb / 50 mem / 5 sess)
+        run: /tmp/vsmoke/bin/python scripts/seed_smoke_duckdb.py
+
+      - name: Start sync daemon (writer) + dashboard (reader)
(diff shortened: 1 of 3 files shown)


vivekchand:
## UX review

- 🟢 **Backend / CI-infrastructure-only.** No UI surface touched. The "UX" here is the developer-facing PR-failure message — make sure that when the gate trips, the output names the slow endpoint + measured p50/p95 + threshold (so a dev can fix without re-running locally). Verify via `scripts/api_latency_probe.py` — if the failure log is just "exit 1", that's bad DX. If it prints `/api/budget/status: p50=7,630ms > 500ms (threshold)`, that's actionable. Worth a 10-line tweak if not already there.

🤖 Pre-merge review @ d968cdb4. Pushing new commits will re-trigger review if diff >50 lines or >2h elapsed.

vivekchand:
## Product review

- 🟢 **Asymmetric value pattern.** 30s CI cost prevents the "influencer arrives, dashboard hangs, they tweet about it" failure class. Per PR #1235's postmortem this exact regression sat in `main` for months undetected. Caught a real P0 on first local run (`/api/budget/status` p50 7.6s, now tracked as #1453). The gate works as intended.
- 🔴 **The gate's OWN first PR run failed** with `daemon never published local_query.json`. Self-validating CI has a chicken-and-egg risk: when the gate fails for an environment reason (daemon-startup timing, fixture path, missing dep), there's no signal whether it's a real regression OR a flake in the gate itself. Before this lands, the gate needs an internal sanity-check step: spin up the daemon → assert it published the discovery file within 10s → ONLY THEN run the probe. Otherwise every flake will be triaged as a regression. file: `.github/workflows/api-latency-smoke.yml`.
- 🟡 **Nightly PyPI job is the right addition** for catching transitive-dep drift, but cron `0 6 * * *` is UTC; depending on team TZ the failure email arrives in the middle of the night. Consider `0 14 * * *` (mid-day Europe/US-east) so a human can act on the breach the same day.
- 🟢 **Threshold values are reasonable.** p50 < 500ms / p95 < 1500ms is roughly "user notices a delay" territory; tighter than needed but easier to relax than to tighten later.

🤖 Pre-merge review @ d968cdb4. Pushing new commits will re-trigger review if diff >50 lines or >2h elapsed.

PR #1458, opened 2026-05-16
fix(budget): daemon-proxy /api/budget/status (p50 7.6s → < 200ms) (closes #1453)

### LINKED ISSUE #1453: [P0] /api/budget/status: p50 7.6s / p95 9.5s / max 10s timeout under daemon contention
Discovered by the API-latency smoke gate (PR #1452) on its first local run under the realistic two-process shape (sync daemon writer + dashboard reader).

## Measured

\`\`\`
endpoint                      p50      p95      max
/api/budget/status            7,630ms  9,531ms  10,006ms (timeout)
all 13 other key endpoints    < 221ms  < 221ms  ok
\`\`\`

Exactly the failure shape PR #1241 was filed to prevent. The gate caught a real regression that was already in main before the gate existed.

## Why P0

\`/api/budget/status\` powers the Budget panel on the main dashboard. A 7-10s p50 means **every dashboard load** stalls that panel for the full duration. Users on a slow machine see the dashboard hang. The "your AI is over budget" alert path needs sub-200ms response — otherwise the value prop is broken.

## Likely cause (1-line diagnosis)

Direct \`get_store()\` calls in a budget aggregation hot path → DuckDB process-level lock contention with the daemon's writer → wait until daemon releases → 7s p50. Standard daemon-proxy refactor (per \`feedback_daemon_proxy_pattern.md\`).

## Acceptance

- [ ] \`/api/budget/status\` p50 < 200ms, p95 < 500ms under the API-latency smoke gate's two-process fixture
- [ ] Gate (PR #1452) passes once fixed
- [ ] Daemon-proxy refactor follows the 5-step playbook in the memory rule

## Effort

S (~2 hr) — daemon-proxy migration is well-trodden by today's sweep (PRs #1429, etc.).

vivekchand:
## Summary

- Routes `_duckdb_cost_since` through the daemon-proxy (`local_store_via_daemon('query_aggregates', ...)`) instead of opening DuckDB writable, which was racing the sync daemon's exclusive writer lock and blocking ~2.5s per call (3 calls × 2.5s = 7.6s p50 for `/api/budget/status`).
- Falls back to `get_store(read_only=True)` for single-process boots (tests, dev mode).
- Confirmed locally: direct `get_store()` against the running user daemon fails after 2541 ms with `IOException: Could not set lock`. With the fix `_get_budget_status()` returns in 8-32 ms.
- Adds 3 regression tests including the headline contract: `< 500 ms` even when the direct path sleeps 3s on each of three windows.

## Measured (local)

| Path | Before | After |
|------|-------:|------:|
| Direct `get_store()` open (single call) | 2541 ms (lock fail) | n/a |
| `_get_budget_status()` cold | ~7600 ms (issue baseline) | 32 ms |
| `_get_budget_status()` warm | ~7600 ms | 8-12 ms |

Well under the issue's acceptance gate (p50 < 200ms, p95 < 500ms).

## Test plan

- [x] `python3 -m pytest tests/ -q -k "budget"` — 28/28 green (including 3 new regression tests)
- [x] `python3 -m pytest tests/test_alerts_evaluator_duckdb.py` — 8/8 green
- [x] Smoke test: 5 sequential `_get_budget_status()` calls against a real running daemon all under 35 ms
- [ ] CI green (pending)
- [ ] PR #1452 latency-gate smoke run (will surface once #1452 lands)

## Follow-ups (same code path, separate PRs)

While auditing the file I noticed these direct `get_store()` callsites in `dashboard.py` that are NOT on `/api/budget/status` but ARE on adjacent budget paths and would benefit from the same migration:

- `_get_agent_budget` (line 906) — called by `/api/agents/<id>/budget`
- `_list_agent_budgets` (line 917) — called by `/api/budget` root
- `_set_agent_budget` (line 930) — write path, may be intentionally direct
- `_delete_agent_budget` (line 948) — write path, may be intentionally direct

Recommend filing a follow-up issue to audit these together with `_DAEMON_METHODS` allowlist (`set_agent_budget`/`delete_agent_budget` are not currently allowlisted, would need adding).

### DIFF
--- a/dashboard.py
+++ b/dashboard.py
@@ -683,17 +683,34 @@ def _duckdb_cost_since(since_iso: str) -> float:
 
     Issue #1404: without this, ~99% of OSS installs see a zero spend metric
     inside the alert evaluator and no real-spend rule ever fires.
+
+    Issue #1453: route through the daemon-proxy (memory
+    ``feedback_daemon_proxy_pattern.md``) so /api/budget/status doesn't
+    block on the daemon's exclusive DuckDB writer lock under the standard
+    launchd/systemd install. Direct ``get_store()`` was taking 2.5 s per
+    call on a real install with sync daemon writing — three calls (daily /
+    weekly / monthly) compounded to 7-10 s p50 for the endpoint, stalling
+    the Budget panel on every dashboard load. Fall back to a read-only
+    direct open for single-process boots (tests, dev mode) where the
+    daemon proxy isn't running.
     """
+    # 1. Daemon-proxy fast path — cross-process HTTP call into the sync
+    #    daemon's local_server, which holds the DuckDB writer lock. Returns
+    #    None when the daemon isn't discoverable (single-process boot).
+    rows = None
     try:
-        from clawmetry import local_store
-        store = local_store.get_store()
-        # ``query_aggregates`` is a per-day SUM(cost_usd) rollup keyed by
-        # ``substr(ts, 1, 10)`` — exactly the shape we need to roll back up
-        # into daily / weekly / monthly totals. Cheap on DuckDB (columnar
-        # SUM over a single column with a ts range scan).
-        rows = store.query_aggregates(since=since_iso)
+        from routes.local_query import local_store_via_daemon
+        rows = local_store_via_daemon("query_aggregates", since=since_iso)
     except Exception:
-        return 0.0
+        rows = None
+    # 2. Read-only direct fallback — tests, dev mode, daemon-down boots.
+    if rows is None:
+        try:
+            from clawmetry import local_store
+            store = local_store.get_store(read_only=True)
+            rows = store.query_aggregates(since=since_iso)
+        except Exception:
+            return 0.0
     total = 0.0
     for r in rows or []:
         try:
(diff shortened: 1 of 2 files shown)


vivekchand:
## UX review

- 🟢 **Backend-only fix, invisible-when-it-works.** Dashboard's Budget panel was hanging 7-10s per load; users will simply notice it stops hanging. No new pixels, no removed pixels in the happy path. The "UX" here is "the dashboard no longer feels broken on slow installs." Worth a 1-line release-note bullet ("Budget panel loads in <50ms even under multi-process daemon contention") so users know the perceived speed improvement isn't placebo.

🤖 Pre-merge review @ ae16ae2e. Pushing new commits will re-trigger review if diff >50 lines or >2h elapsed.

vivekchand:
## Product review

- 🟢 **Closes a 12-15s funnel-killer that sat in main for months.** Per the PR #1235 postmortem this exact failure shape caused users to bounce after Peter Steinberger starred the repo. Fixing it before the next influencer arrival is the right time. The new API-latency smoke gate (#1452) would have caught this — and did, on its first run.
- 🟡 **4 adjacent direct `get_store()` callsites flagged in the PR body** (`_get_agent_budget`, `_list_agent_budgets`, `_set_agent_budget`, `_delete_agent_budget`). These all power `/api/budget` + `/api/agents/<id>/budget` and are presumably 7-10s each under the same contention pattern. File as a single follow-up issue ("budget read+write paths daemon-proxy sweep") and ship before the next [RELEASE], otherwise users will fix one slow path and immediately hit another.
- 🟢 **Test pattern is correct**: monkeypatching the direct path to sleep 3s and asserting the endpoint still returns quickly via fallback. Future regressions will trip this.

🤖 Pre-merge review @ ae16ae2e. Pushing new commits will re-trigger review if diff >50 lines or >2h elapsed.

PR #1459, opened 2026-05-16
fix(usage): extract sibling-dedupe helper, close 3 remaining 2× spots (closes #1451)

### LINKED ISSUE #1451: [Product P0] Dedupe-pattern audit: 5 more endpoints with blind sum of token_count over v3 events
From the MOAT dedupe-pattern audit (see `feedback_usage_dedupe_pattern.md`, plus PRs #1444 and #1446 which closed 2 of these).

## Surfaces still over-counting on real v3 installs

\`grep -nE "sum.*token_count|\+= .*token_count" routes/\` turns up these remaining hot spots that have the same shape of bug:

| File:line | Surface | Risk |
|---|---|---|
| \`routes/usage.py:521\` | \`/api/usage/by-plugin\` plugin token totals | 2× per-plugin tokens on v3 |
| \`routes/usage.py:587\` | \`/api/usage/by-plugin/trend\` per-plugin daily trend | 2× per-plugin daily |
| \`routes/usage.py:1372\` | per-session token fallback in session-aggregator | 2× session tokens when \`sessions\` row was zero |
| \`routes/advisor.py:172\` | Advisor per-entry token rollup | 2× advisor recommendations |
| \`routes/advisor.py:182\` | Advisor today-tokens rollup | 2× "today" stat |

Each one walks events without skipping the slim \`model.completed\` sibling that pairs with every \`assistant\` event in real OpenClaw v3. Per the PR #1444 fix, the pattern dedupes to \`(session_id, ts_sec ±1 s) → pick richest envelope\`.

## Why this is P0

Several of these power surfaces users see and act on:
- **By-plugin breakdown** drives "which tool is using all the tokens" decisions.
- **Advisor** literally recommends optimizations using the (inflated) token counts.

If users see "Bash used 1M tokens today" when reality is 500K, they'll either ignore the dashboard (false signal) or take the wrong action (kill Bash unnecessarily).

## Fix

Apply the exact dedupe approach from \`routes/usage.py:_try_local_store_cost_comparison\` post-PR #1446 (a 2-pass walk over events that picks the richest envelope per \`(session_id, ts_sec ±1 s)\` bucket). The dedupe helper should be factored into a single utility in \`routes/_dedupe.py\` so all 5 surfaces share it.

## Scope

S — ~80 LOC for the shared helper + 5 callsite swaps + 5 regression tests. Should fit in one PR with the helper extraction.

vivekchand:
## Summary

Closes vivekchand/clawmetry#1451.

Three usage.py surfaces still summed `token_count` blindly across v3 events and double-counted every billable turn (the `assistant` + `model.completed` sibling pair every real OpenClaw v3 install emits ~100 ms apart):

| File:line | Surface | Risk before fix |
|---|---|---|
| `routes/usage.py:521` | `/api/usage/by-plugin` plugin totals | 2× per-plugin tokens |
| `routes/usage.py:587` | `/api/usage/by-plugin/trend` daily series | 2× per-plugin daily |
| `routes/usage.py:1372` | per-session aggregator (clusters) | 2× session tokens |

## What changed

- **New shared helper** at `routes/_dedupe.py` exposing `build_sibling_bucket_max()` + `is_sibling_dup()` — the 2-pass `(session_id, ts_sec ±1 s) → richest envelope` walk that PR #1446 introduced inline.
- **3 usage.py callsites** swapped to the helper (the 3 in the issue).
- **2 retroactive migrations** — `_try_local_store_cost_comparison` (PR #1446) and `_try_local_store_advisor_context` (PR #1454) both had the same 2-pass logic copy-pasted inline; lifted to the helper to prevent future drift.
- For `/api/sessions/clusters` specifically, `query_sessions` returns `SUM(token_count)` from events, which is itself 2×'d on v3. Compute `s_tokens` from the deduped event list as the canonical path and fall back to the session row only when no events joined.

Net diff: **102 LOC** (182 insertions, 80 deletions across 4 files).

## Test plan

- [x] `python3 -m pytest tests/ -q -k "usage or advisor"` → **44 passed**
- [x] 3 new regression tests in `test_usage_local_store.py` (one per surface) — each seeds a sibling pair and asserts the response shows the deduped one-turn total, not 2×
- [x] All 7 existing cost-comparison/v3-sibling tests still pass after the helper extraction
- [ ] Smoke against a live install with real v3 data before merging (recommended)

### DIFF
--- a/routes/_dedupe.py
+++ b/routes/_dedupe.py
@@ -1,0 +1,66 @@
+"""routes/_dedupe.py — sibling-dedupe helper for v3 event sums (issue #1451).
+
+Real OpenClaw v3 + Claude Code emit two event rows per billable LLM turn:
+a rich ``assistant``/``message`` envelope plus a slim ``model.completed``
+sibling ~100 ms later. Any analytics surface that blindly sums
+``token_count`` over the events table therefore double-counts every turn.
+
+This module centralises the 2-pass dedupe so every surface picks the
+**richest envelope** per ``(session_id, ts_sec ±1 s)`` bucket. ``±1 s``
+covers the writer-race window between the two emitters.
+
+Usage::
+
+    from routes._dedupe import build_sibling_bucket_max, is_sibling_dup
+
+    bucket_max = build_sibling_bucket_max(events)
+    for ev in events:
+        if is_sibling_dup(ev, bucket_max):
+            continue
+        tokens += int(ev.get("token_count") or 0)
+"""
+
+from __future__ import annotations
+
+from datetime import datetime
+
+# ``assistant``/``message`` outrank ``model.completed`` inside one bucket.
+_RICHER = {"assistant": 2, "message": 2, "model.completed": 1}
+
+
+def _ts_sec(ts_str) -> int:
+    """ISO-8601 string → integer epoch second. Returns 0 on any failure."""
+    if not ts_str or not isinstance(ts_str, str):
+        return 0
+    try:
+        return int(datetime.fromisoformat(ts_str.replace("Z", "+00:00")).timestamp())
+    except Exception:
+        return 0
+
+
+def build_sibling_bucket_max(rows) -> dict:
+    """Pass 1: scan rows once, return ``{(sid, sec±1): max_rank}``."""
+    bucket_max: dict = {}
+    for r in rows:
+        et = (r.get("event_type") or "").strip()
+        if et not in _RICHER:
+            continue
+        sid = r.get("session_id") or ""
+        sec = _ts_sec(r.get("ts") or "")
+        rank = _RICHER[et]
+        for key in ((sid, sec - 1), (sid, sec), (sid, sec + 1)):
+            if bucket_max.get(key, 0) < rank:
+                bucket_max[key] = rank
+    return bucket_max
+
+
+def is_sibling_dup(row, bucket_max: dict) -> bool:
+    """Pass 2 predicate: True iff ``row`` is a slim sibling of a richer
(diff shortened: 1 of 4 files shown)


vivekchand:
## UX review

- 🟢 **Pure refactor + 3 new regression tests, no UI surface.** The user-visible effect is that Tokens / by-plugin / per-session token tables stop showing 2× their true value on real v3 installs. That's a "numbers move correctly" change — worth one release-note bullet ("Plugin and per-session token totals now match billing for v3 OpenClaw installs").

🤖 Pre-merge review @ e58e8a23. Pushing new commits will re-trigger review if diff >50 lines or >2h elapsed.

vivekchand:
## Product review

- 🟢 **Helper extraction is the right shape.** `routes/_dedupe.py` is a single source of truth for the sibling-pair dedupe across 5 callsites. Future surfaces import the helper instead of copy-pasting the 2-pass walk, which closes the bug class against drift.
- 🟢 **Retroactive migration of `cost_comparison` + `advisor` to the helper is the right call** despite being out-of-scope per the issue text. Mechanical swap, prevents the next regression-by-copy-paste.
- 🟡 **Root cause still at the SQL layer.** The agent's report flagged that `query_sessions` SUM(token_count) is the actual source of the 2×, and the Python helper is a workaround at the consumer layer. Filed as #1460 and now being fixed in PR #1462 — once that lands, this Python helper becomes belt-and-suspenders rather than the only defense. Worth noting in the PR body so future-you doesn't wonder why both layers dedupe.

🤖 Pre-merge review @ e58e8a23. Pushing new commits will re-trigger review if diff >50 lines or >2h elapsed.

PR #1461, opened 2026-05-16
fix(ci): smoke-gate sanity-check + flake-vs-regression annotations (closes #1457)

### LINKED ISSUE #1457: [Product P0 from PR #1452] API-latency smoke gate needs sanity-check + flake-vs-regression discrimination
From the pre-merge product review of PR #1452.

## Problem

PR #1452 ships the API-latency smoke gate, and its own first PR run failed with:

\`\`\`
::error::daemon never published local_query.json
\`\`\`

That's not a real regression — it's the gate's fixture / daemon-startup timing. As-is, any future failure in this gate looks identical to a real perf regression. No way for a dev to tell whether to fix their code OR re-run the workflow.

## Why P0

Self-validating CI gates fail in two distinct ways:

1. **Real regression** — endpoint actually slow, dev needs to optimize.
2. **Gate flake** — fixture didn't seed, daemon didn't start, probe couldn't bind port.

Without internal sanity checks the dev sees the same red X for both classes and will eventually start ignoring the gate (or worse, disabling it). That defeats the entire point of PR #1241.

## Fix

In \`.github/workflows/api-latency-smoke.yml\`:

1. **Pre-probe sanity step**: spin up daemon → poll for \`local_query.json\` for up to 10s → if not present, fail with explicit `::error::Gate setup failed: daemon did not start (this is NOT a regression). Re-run the workflow.`
2. **Probe failure message**: include endpoint name + measured p50/p95 + threshold + delta. Example: `::error::Regression: /api/budget/status p50=7,630ms exceeds 500ms budget by 7,130ms`.
3. **Distinct exit codes** OR **annotation prefix**: `[setup-flake]` vs `[regression]` so devs and downstream automation can filter.

## Acceptance

- [ ] Gate setup failures emit `[setup-flake]` annotation, not `[regression]`
- [ ] Probe failures include endpoint name + numbers + threshold in the failure message
- [ ] CI re-run of a flaked gate passes without code change

## Effort

XS (~30 min) — workflow YAML tweak + 5-line probe script change.

vivekchand:
closes #1457

## Summary

The API-latency smoke gate from PR #1452 had no way to distinguish a real perf regression from a setup flake — both showed the same red X. Devs couldn't tell whether to fix code or re-run the workflow.

- Setup failures (daemon didn't publish `local_query.json`, dashboard `/api/health` never returned) now emit `::error::[setup-flake] ... this is NOT a regression. Re-run the workflow.` and exit 2.
- Probe failures emit `::error::[regression] <endpoint> p50=Xms exceeds Yms budget by Zms` per the issue example (or `p95=...`, or `status=...`).
- Distinct exit codes (2 = setup-flake, 1 = regression) so downstream automation can filter.
- Probe stdout table stays human-readable; annotations are additive lines, kept machine-parseable.

Stacked on top of PR #1452 (`ci/api-latency-smoke-1241`); diff narrows to just this change once that lands.

## Test plan

- [x] Pure-bash repro of missing-daemon path: `[setup-flake]` annotation fires, exit code 2 as expected.
- [x] Probe against 700ms mock server emits `[regression] /api/overview p50=706ms exceeds 500ms budget by 206ms` — exact format from the issue body.
- [x] `python3 -c "import yaml; yaml.safe_load(open('.github/workflows/api-latency-smoke.yml'))"` validates.
- [x] `python3 -m py_compile scripts/api_latency_probe.py` clean.
- [ ] Real CI run on this PR exercises the regression path (no slow endpoints expected → should pass).

## Acceptance (issue #1457)

- [x] Gate setup failures emit `[setup-flake]` annotation, not `[regression]`
- [x] Probe failures include endpoint name + numbers + threshold in the failure message
- [x] CI re-run of a flaked gate passes without code change (distinct exit code makes this triagable)

### DIFF
--- a/.github/workflows/api-latency-smoke.yml
+++ b/.github/workflows/api-latency-smoke.yml
@@ -77,6 +77,10 @@ jobs:
         # Daemon owns the DuckDB lock and publishes its local-query port to
         # ~/.clawmetry/local_query.json. Dashboard proxies through that for
         # /api/local/* reads. Both running = the contention shape we test.
+        #
+        # Failures here are SETUP FLAKES, not perf regressions — emit a
+        # distinct `[setup-flake]` annotation so devs (and downstream
+        # automation) can tell "re-run the workflow" from "fix your code".
         env:
           CLAWMETRY_INGEST_URL: http://127.0.0.1:9
         run: |
@@ -88,8 +92,8 @@ jobs:
             sleep 1
           done
           if [ ! -f "$HOME/.clawmetry/local_query.json" ]; then
-            echo "::error::daemon never published local_query.json"
-            tail -100 .smoke/daemon.log; exit 1
+            echo "::error::[setup-flake] Gate setup failed: daemon did not publish local_query.json within 30s (this is NOT a regression). Re-run the workflow."
+            tail -100 .smoke/daemon.log; exit 2
           fi
           nohup /tmp/vsmoke/bin/clawmetry --port 8900 --no-debug \
             > .smoke/dashboard.log 2>&1 &
@@ -102,8 +106,8 @@ jobs:
           done
           curl -sf -H "Authorization: Bearer $CLAWMETRY_TOKEN" \
             http://localhost:8900/api/health > /dev/null || {
-            echo "::error::dashboard /api/health never returned 200"
-            tail -100 .smoke/dashboard.log; exit 1; }
+            echo "::error::[setup-flake] Gate setup failed: dashboard /api/health never returned 200 within 30s (this is NOT a regression). Re-run the workflow."
+            tail -100 .smoke/dashboard.log; exit 2; }
 
       - name: Probe endpoints + assert p50/p95 budgets
         env:
--- a/scripts/api_latency_probe.py
+++ b/scripts/api_latency_probe.py
@@ -82,30 +82,45 @@ def run(endpoints: Iterable[str] = ENDPOINTS) -> int:
     print("-" * len(header))
 
     failures: list[str] = []
+    annotations: list[str] = []
     for path in endpoints:
         if path in SKIP:
             print(f"{path:<40} {'-':>8} {'-':>8} {'-':>8}  skip    SKIP")
             continue
         samples, status, err = _probe_one(session, path)
         p50, p95, smax = _percentile(samples, 50), _percentile(samples, 95), max(samples)
-        verdict, breach_bits = "OK", []
+        verdict, breach_bits, ann_bits = "OK", [], []
         # Network errors / 5xx are a hard fail — the endpoint is down, not slow.
         if status is None or status >= 500:
             verdict = "FAIL"
             breach_bits.append(f"status={status or err or 'connerr'}")
+            ann_bits.append(f"status={status or err or 'connerr'}")
         if p50 > P50_BUDGET_MS:
             verdict = "FAIL"
             breach_bits.append(f"p50={p50:.0f}ms>{P50_BUDGET_MS:.0f}ms")
+            ann_bits.append(
+                f"p50={p50:.0f}ms exceeds {P50_BUDGET_MS:.0f}ms budget by {p50 - P50_BUDGET_MS:.0f}ms"
+            )
         if p95 > P95_BUDGET_MS:
             verdict = "FAIL"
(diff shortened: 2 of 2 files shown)


vivekchand:
## UX review

- 🟢 **Developer-facing UX, done right.** The failure annotations now include endpoint name, measured p50/p95, threshold, and delta (e.g., `/api/overview p50=706ms exceeds 500ms budget by 206ms`). That's actionable — a dev sees the red X and knows what to optimize, instead of needing to re-run locally. The `[setup-flake]` vs `[regression]` prefix is the right separator so flake-vs-real-regression is one glance away.
- 🟢 **Exit codes 2 vs 1 distinct** so downstream automation (auto-merger, persona-gate) can filter without parsing annotation text. Good defensive design.

🤖 Pre-merge review @ 0f70cc54. Pushing new commits will re-trigger review if diff >50 lines or >2h elapsed.

vivekchand:
## Product review

- 🟢 **Saves the gate's own credibility.** Without this, the first flake on a real PR would have devs ignoring the gate and eventually disabling it — exactly the failure mode PR #1452 was filed to prevent. Asymmetric ROI: ~30 LOC YAML/Python tweak protects the entire latency-gate investment.
- 🟢 **Local repro proves both paths trip.** Setup-flake fires `[setup-flake]` with exit 2; regression path fires `[regression]` with endpoint+numbers and exit 1. Pre-merge validation done — no "wait for first CI run to see if it works" risk.
- 🟡 **Stacked on PR #1452** which is still open. As-is, mergeability depends on #1452 landing first. Worth ensuring the auto-merger handles the stack order (i.e., merges #1452 → #1461 in sequence, not in parallel) so this doesn't fail with a conflict.

🤖 Pre-merge review @ 0f70cc54. Pushing new commits will re-trigger review if diff >50 lines or >2h elapsed.

vivekchand:
## UX review

- 🟢 **Backend / CI-infrastructure-only.** No HTML/CSS/JS or end-user API surface in this diff — entirely new `.github/workflows/api-latency-smoke.yml` + `scripts/api_latency_probe.py` + `scripts/seed_smoke_duckdb.py`. The 'UX' here is the developer-facing failure message in CI logs.
- 🟢 **The `[setup-flake]` vs `[regression]` taxonomy is the right read.** A dev opening the failed run no longer has to guess whether to re-run or fix code — the leading bracket tells them. Exit-code split (1 vs 2) also makes the same distinction available to any downstream automation. That's Steve-grade affordance: the right action is obvious without thinking.
- 🟢 **Failure messages are specific.** `[regression] /api/overview p50=706ms exceeds 500ms budget by 206ms` reads as a one-line bug report — endpoint + measurement + budget + delta. No log-spelunking needed.
- 🟢 No em-dashes in the user-facing annotation strings. Good catch keeping it ascii — GitHub Actions UI doesn't render em-dashes consistently.

🤖 Pre-merge review @ 8d821dbd. Pushing new commits will re-trigger review if diff >50 lines or >2h elapsed.

vivekchand:
## Product review

- 🟢 **Asymmetric-value pattern.** ~10-min CI cost prevents the specific failure class that burned us in PR #1235 (3 endpoints went 7-15s under daemon+dashboard contention; single-process CI never caught it). The two-process matrix is what makes this expensive enough to be honest about what real installs see.
- 🟢 **Distinct exit codes (1=regression, 2=setup-flake) is the kind of detail that compounds.** Today it's just for humans reading the failure. Tomorrow when we add an auto-retry-on-flake step to the workflow it works without code change. Closes the issue requirement (`CI re-run of a flaked gate passes without code change`) by giving downstream the signal to filter on.
- 🟢 **Probe table stays human-readable; annotations are additive.** No risk of breaking existing log parsers or screenshots in runbooks.
- 🟡 **Nit (not blocking):** `CLAWMETRY_SMOKE_RUNS=5` default means the p95 is computed from 5 samples per endpoint — that's a noisy p95 in practice. Worth bumping to 10 if false-positive rate ends up annoying once it's running on every PR. Easy follow-up if it shows up.
- 🟢 No tier-strategy implication, no OSS/Cloud-Pro funnel impact — pure dev-tooling investment.

🤖 Pre-merge review @ 8d821dbd. Pushing new commits will re-trigger review if diff >50 lines or >2h elapsed.

PR #1462, opened 2026-05-16
fix(local-store): dedupe v3 sibling pairs in query_sessions at SQL layer (closes #1460)

### LINKED ISSUE #1460: [Product P1] DuckDB query_sessions SUM(token_count) is 2× on v3 — fix at SQL layer to close root cause
Surfaced during PR #1459 (the dedupe-helper extraction for #1451). The 2-pass Python helper closes the bug for 5 fast-path endpoints, but the **root cause is at the SQL layer** in \`clawmetry/local_store.py\`.

## The root

\`\`\`sql
-- clawmetry/local_store.py:2576 (query_sessions)
SELECT ..., COALESCE(SUM(token_count), 0) AS token_count
FROM events
GROUP BY session_id, ...
\`\`\`

On real OpenClaw v3 installs each LLM turn emits BOTH an \`assistant\` row AND a sibling \`model.completed\` row ~100 ms apart, both stamped with the same \`token_count\`. The SUM happily counts both, so any consumer of \`query_sessions().token_count\` reads 2× reality:

- \`/api/sessions/clusters\` (the cluster aggregator)
- Anomaly detector (reads session-row token totals)
- Any future surface that does \`SELECT token_count FROM <query_sessions result>\`

## Why filing as P1 not P0

The 5 fast paths from #1451 are already covered by the new \`routes/_dedupe.py\` helper, so end-user surfaces are correct today. This issue is about closing the bug at the SOURCE so the next surface that copy-pastes \`query_sessions().token_count\` does not silently reintroduce the 2× without a 2-pass dedupe.

## Fix shape

Two options:

**Option A — SQL-side dedupe (preferred)**: rewrite \`query_sessions\` to subquery-dedupe by \`(session_id, ts_sec)\` before SUMming. Picks the richer-envelope row per bucket. Pure SQL, fast.

\`\`\`sql
WITH deduped AS (
  SELECT DISTINCT ON (session_id, CAST(EXTRACT(EPOCH FROM ts) AS BIGINT))
    *
  FROM events
  ORDER BY session_id, CAST(EXTRACT(EPOCH FROM ts) AS BIGINT),
           CASE event_type
             WHEN 'assistant' THEN 1
             WHEN 'message' THEN 1
             WHEN 'model.completed' THEN 2
             ELSE 0
           END
)
SELECT ..., COALESCE(SUM(token_count), 0) AS token_count
FROM deduped
GROUP BY session_id, ...
\`\`\`

**Option B — keep SQL as-is, add \`query_sessions_deduped\` companion method**. Existing callers stay on the raw method; new callers opt in to deduped. Smaller blast radius, but doesn't actually close the root.

Recommendation: A. Performance tradeoff is negligible (DuckDB will optimize the DISTINCT ON over a sorted-anyway query) and the bug class is closed for good.

## Scope

S — ~50 LOC SQL rewrite + 1 regression test + a memory update noting "query_sessions is now deduped at SQL layer, no need for Python helper if you only consume that method."

## Related

- PR #1459 (the Python-helper workaround)
- \`feedback_usage_dedupe_pattern.md\` (memory note tracking all the surfaces touched today)

vivekchand:
## Summary

Root-cause fix for the dedupe bug class that PR #1459 worked around in 5 Python fast paths via \`routes/_dedupe.py\`. Closes the bug at the source so the next consumer of \`query_sessions().token_count\` cannot regress it.

## Why this is necessary

On real OpenClaw v3 installs each LLM turn emits BOTH an \`assistant\` row AND a sibling \`model.completed\` row ~100 ms apart, both stamped with the same \`token_count\` + \`cost_usd\`. \`SUM()\` over the raw \`events\` table doubles every billable turn. The Python helper closes this for 5 known surfaces — but ANY other consumer (cluster aggregator, anomaly detector, future surfaces) inherits the 2× without the helper.

## Approach

5-CTE SQL rewrite in \`query_sessions\`:

1. **\`ranked\`** tags each event with envelope rank (assistant/message=2, model.completed=1, others=0) + ts_sec bucket.
2. **\`bucket_max\`** finds the max rank per \`(session_id, ts_sec)\`.
3. **\`deduped\`** drops the slim sibling ONLY when a richer envelope exists in the same bucket. Critical: two \`tool_call\` rows sharing a ts_sec are still both kept (they are not siblings — only the assistant/model.completed pair pattern is).
4. **\`deduped_sums\`** aggregates cost_usd + token_count from deduped.
5. The final SELECT joins back to \`ranked\` so **\`event_count\` stays RAW** — debug surfaces still see all the rows that arrived; only the money-tied numbers get deduped.

## Verification

- \`pytest tests/test_local_query_api.py -q\` → **13 passed** (12 existing + new sibling-pair regression test)
- \`pytest tests/test_moat_send_message_e2e.py\` → **3 passed**
- \`pytest tests/test_usage_local_store.py\` → **28 passed**
- \`pytest tests/test_e2e_duckdb_relay.py\` → all non-pre-existing pass

Two pre-existing failures (\`test_duckdb_read_only_handle_sees_committed_rows\`, \`test_relay_shape_health\`) reproduce on main and are unrelated.

## New regression test

\`test_sessions_endpoint_dedupes_v3_sibling_pairs\` — synthetic assistant + model.completed pair at the same ts, asserts:
- \`token_count == 150\` (deduped to 1 turn, not 2×)
- \`cost_usd == 0.005\` (deduped)
- \`event_count == 2\` (raw, kept for debug visibility)

### DIFF
--- a/clawmetry/local_store.py
+++ b/clawmetry/local_store.py
@@ -2565,18 +2565,69 @@ class LocalStore:
             clauses.append("ts <= ?")
             params.append(until)
         where = "WHERE " + " AND ".join(clauses)
+        # Issue #1460: on real OpenClaw v3 installs each LLM turn emits BOTH
+        # an ``assistant`` row AND a sibling ``model.completed`` row ~100 ms
+        # apart, both stamped with the same ``token_count`` + ``cost_usd``.
+        # A naive SUM() over the raw events table doubles every billable
+        # turn. The 5 Python fast paths from #1451 worked around this with
+        # a 2-pass walker (``routes/_dedupe.py``); fixing it here closes
+        # the bug class for ALL consumers (cluster aggregator, anomaly
+        # detector, anyone else who reads ``query_sessions().token_count``).
+        #
+        # Approach: dedupe at the SQL layer by (session_id, ts-rounded-to-
+        # second) bucket, picking the richer envelope (assistant/message >
+        # model.completed). ``event_count`` stays RAW (it's "how many rows
+        # did we record", not "how many billable turns") — only cost_usd
+        # + token_count come from the deduped CTE.
         sql = f"""
+            WITH ranked AS (
+              SELECT
+                id, agent_id, session_id, ts, cost_usd, token_count,
+                event_type,
+                CASE event_type
+                  WHEN 'assistant'        THEN 2
+                  WHEN 'message'          THEN 2
+                  WHEN 'model.completed'  THEN 1
+                  ELSE 0
+                END AS envelope_rank,
+                CAST(EXTRACT(EPOCH FROM CAST(ts AS TIMESTAMP)) AS BIGINT)
+                                                 AS ts_sec
+              FROM events
+              {where}
+            ),
+            bucket_max AS (
+              SELECT session_id, ts_sec, MAX(envelope_rank) AS max_rank
+              FROM ranked
+              GROUP BY session_id, ts_sec
+            ),
+            deduped AS (
+              -- Drop the slim ``model.completed`` sibling ONLY when an
+              -- ``assistant``/``message`` (rank=2) exists in the same
+              -- (session_id, ts_sec) bucket. Other event types (rank=0
+              -- tool_calls, etc.) are NEVER dropped — they may legitimately
+              -- share a ts_sec without being sibling pairs.
+              SELECT r.* FROM ranked r
+              JOIN bucket_max bm USING (session_id, ts_sec)
+              WHERE NOT (r.envelope_rank = 1 AND bm.max_rank = 2)
+            ),
+            deduped_sums AS (
+              SELECT session_id,
+                     COALESCE(SUM(cost_usd), 0)    AS cost_usd_d,
+                     COALESCE(SUM(token_count), 0) AS token_count_d
+              FROM deduped
+              GROUP BY session_id
+            )
             SELECT
-              session_id,
-              MIN(agent_id)               AS agent_id,
-              MIN(ts)                     AS started_at,
-              MAX(ts)                     AS updated_at,
-              COUNT(*)                    AS event_count,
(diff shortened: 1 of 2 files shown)


vivekchand:
## UX review

- 🟢 **Backend-only fix.** Same shape of user-visible effect as PR #1459: token + cost numbers in session-level views (Sessions tab, cluster aggregator, anomaly detector) stop showing 2× their true value on real v3 installs. The release-note bullet from #1459 covers this too — no separate note needed.

🤖 Pre-merge review @ 42f16e5c. Pushing new commits will re-trigger review if diff >50 lines or >2h elapsed.

vivekchand:
## Product review

- 🟢 **Root-cause fix at the SQL layer is the right shape.** PR #1459's Python helper closed the bug for 5 known surfaces; this closes it at the source so the next consumer can't reintroduce the 2× without the helper. Belt-and-suspenders is good — neither layer alone is a guarantee, both together is.
- 🟢 **`event_count` stays RAW is the right call.** Sub-tle but correct: the row count IS what arrived; only the money-tied numbers (`cost_usd`/`token_count`) should be deduped. Debug surfaces still see the doubled rows, which is useful for ground-truth diagnostics.
- 🟢 **Tool_call rows sharing a ts_sec are still both kept.** Critical correctness detail — the first SQL attempt over-deduped tool_calls and broke `test_sessions_endpoint`; the rewrite to "drop the slim sibling ONLY when a richer envelope exists in the bucket" is the right algorithm. Pre-merge bug caught and fixed before this PR opened.
- 🟢 **Same approach should land on `query_aggregates` next** (the daily-bucket aggregator). It's the OTHER read path everyone consumes; fixing it would close `_try_local_store_usage`'s need for the Python dedupe too. Worth a follow-up issue.

🤖 Pre-merge review @ 42f16e5c. Pushing new commits will re-trigger review if diff >50 lines or >2h elapsed.

PR #1464, opened 2026-05-16
fix(local-store): dedupe v3 sibling pairs in query_aggregates SQL (root fix)

vivekchand:
## Summary

Follows PR #1462's `query_sessions` fix. `query_aggregates` had the same shape of bug: blind `SUM(cost_usd)` + `SUM(token_count)` over the raw events table, doubling every billable turn on real v3 OpenClaw installs (assistant + sibling model.completed each contribute the same token_count).

After this lands, both `query_sessions` AND `query_aggregates` are deduped at the SQL layer. Any future consumer of either reads correct numbers without needing the Python helper from PR #1459 (which stays in place as belt-and-suspenders).

## Approach

Same 4-CTE rewrite as PR #1462:
1. **`ranked`** — tag each event with envelope rank + ts_sec bucket.
2. **`bucket_max`** — max rank per `(session_id, ts_sec)`.
3. **`deduped`** — drop slim `model.completed` sibling ONLY when richer envelope exists in the same bucket. Tool_calls sharing a ts_sec still both kept.
4. **`deduped_sums`** — aggregate cost_usd + token_count from deduped per day+agent.
5. Final SELECT joins back to `ranked` so `event_count` stays RAW (debug surfaces see all rows; only money-tied numbers are deduped).

## Verification

Pre-merge synthetic test:
- Sibling pair at same ts → deduped to 1 turn (150 tokens, $0.005)
- Two tool_calls at same ts → both counted (45 tokens, $0.30)
- Day rollup: event_count=4 (raw), tokens=195, cost=$0.305

New regression test asserts all 3 invariants. Test passes; existing 13 also pass.

## MOAT context

Personas on PR #1462 explicitly flagged `query_aggregates` as "the OTHER read path everyone consumes — fixing it would close `_try_local_store_usage`'s need for the Python dedupe too." This PR closes that loop.

### DIFF
--- a/clawmetry/local_store.py
+++ b/clawmetry/local_store.py
@@ -4114,17 +4114,61 @@ class LocalStore:
             clauses.append("ts <= ?")
             params.append(until)
         where = ("WHERE " + " AND ".join(clauses)) if clauses else ""
+        # Same dedupe shape as ``query_sessions`` (PR #1462). On real v3
+        # OpenClaw installs each LLM turn emits BOTH an ``assistant`` row
+        # AND a sibling ``model.completed`` row ~100 ms apart, both
+        # stamped with the same ``token_count`` + ``cost_usd``. A naive
+        # SUM over the raw events table doubles every billable turn.
+        # ``event_count`` stays RAW (row count, not turn count); only
+        # cost_usd + token_count come from the deduped CTE.
         sql = f"""
+            WITH ranked AS (
+              SELECT
+                id, session_id, agent_id, ts, cost_usd, token_count,
+                event_type,
+                substr(ts, 1, 10) AS day,
+                CASE event_type
+                  WHEN 'assistant'        THEN 2
+                  WHEN 'message'          THEN 2
+                  WHEN 'model.completed'  THEN 1
+                  ELSE 0
+                END AS envelope_rank,
+                CAST(EXTRACT(EPOCH FROM CAST(ts AS TIMESTAMP)) AS BIGINT)
+                                                 AS ts_sec
+              FROM events
+              {where}
+            ),
+            bucket_max AS (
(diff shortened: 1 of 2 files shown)


vivekchand:
## UX review

- 🟢 **Backend-only SQL rewrite.** Mirror of PR #1462's pattern, no UI surface touched. User-visible effect: the daily token-totals chart on the Tokens tab stops showing 2× reality for real v3 installs that consume `query_aggregates`. Same release-note bullet from PR #1462 ("token + cost numbers now match billing for v3 installs") covers this too — no separate note needed.

🤖 Pre-merge review @ 1be79a93. Pushing new commits will re-trigger review if diff >50 lines or >2h elapsed.

vivekchand:
## Product review

- 🟢 **Closes the loop personas flagged on PR #1462.** Both root-cause SQL queries (`query_sessions` + `query_aggregates`) are now deduped at the source. Any future consumer of either reads correct numbers automatically — no copy-paste-and-forget-to-call-the-helper risk.
- 🟢 **Pre-merge synthetic test proves dedupe + non-dedupe both correct.** Sibling pair → 1 turn. Two tool_calls at same ts → both kept. event_count stays RAW. These are the three correctness invariants; all asserted by the new regression test.
- 🟡 **Performance footprint of the 4-CTE rewrite.** On installs with millions of events the CTE materialization may be slower than the old single SUM. DuckDB's optimizer typically inlines these, but worth a benchmark at 1M-row scale before claiming it's a free win. Not a blocker; can be measured after merge against a real install.
- 🟢 **Belt-and-suspenders with the Python helper from PR #1459** is the right defense-in-depth shape. Either layer alone would close the bug; both together survive future schema changes + future copy-paste mistakes.

🤖 Pre-merge review @ 1be79a93. Pushing new commits will re-trigger review if diff >50 lines or >2h elapsed.

PR #1467, opened 2026-05-16
feat(brain): one-time restoration toast after 0.12.182 empty-detail fix (closes #1195)

### LINKED ISSUE #1195: [UX P0 from PR #1190] Surface Brain-data restoration after the 0.12.182 empty-detail regression
From the post-merge UX/product review of PR #1190.

## Problem

PR #1190 fixed the P0 where every Brain event chip rendered with `detail=\"\"` since 0.12.182 shipped on 2026-05-12. The read path is now correct — but users who hit the empty Brain stream during the regression window have no in-app signal that the data is back. They saw broken Brain, bounced, and have nothing pulling them back to retry.

## Why this matters

The Brain tab is the surface where users actually spend visual time and the most credible \"this thing is alive\" demo for the Cloud-Pro funnel. A silent fix means anyone who touched ClawMetry between 0.12.182 release and 0.12.183+ install carries a memory of broken-Brain forever. Trust regressions don't auto-heal.

## Proposed fix

One-time toast on first non-empty Brain event load post-upgrade: \"Live event details restored\". Gated by a boolean (`brain_restoration_toast_shown`) in `~/.clawmetry/config.json` so it fires exactly once per install, never for fresh installs.

## Acceptance

- [ ] Toast appears once on the first Brain page load AFTER upgrade where `count(events with detail!='') > 0`
- [ ] Toast does NOT appear on fresh installs (no prior empty-state experience to recover from)
- [ ] Toast does NOT re-fire across page navigations within the same daemon lifetime
- [ ] Cookie/config flag persists so a daemon restart doesn't re-show

## Related

- PR #1190 — the underlying read fix
- PR #1143 — the v3 sync mapper that introduced the regression

vivekchand:
## Summary

- One-time toast "Live event details restored. Brain stream is back to normal." on the first non-empty `/api/brain-history` load post-upgrade. Heals the trust regression for users who hit the empty Brain stream between `0.12.182` (shipped 2026-05-12) and the `0.12.183+` fix from PR #1190.
- New tiny endpoint `GET /api/install-age` in `routes/meta.py` returns the ctime/mtime of `~/.clawmetry/config.json`. Used to distinguish fresh installs (no toast — never saw the bug) from upgrades.
- Gating in `app.js`: localStorage flag + install-ctime check + presence of at least one non-empty `detail`. Reuses existing `.cron-toast` CSS — no new style, no new deps.

Closes #1195.

## Acceptance map

- [x] Toast fires once on first Brain page load post-upgrade when `events.some(e => e.detail !== '')`.
- [x] Fresh installs skip: `/api/install-age` returns ctime >= `2026-05-12` cutoff, helper bails before showing.
- [x] Re-fire across page navigations within same daemon lifetime blocked by localStorage flag (set immediately when toast renders).
- [x] Daemon restart preserves the flag (localStorage is per-origin, browser-persisted).

## Test plan

- [ ] Manual: clear `clawmetry_brain_restoration_toast_shown_v1` from localStorage, ensure `~/.clawmetry/config.json` has a ctime older than 2026-05-12, load `/`, click Brain tab, confirm toast appears once.
- [ ] Manual: reload page, navigate away + back to Brain, confirm toast does NOT re-fire.
- [ ] Manual: `rm ~/.clawmetry/config.json`, restart daemon (regenerates with current ctime), confirm toast does NOT appear (fresh-install gate).
- [ ] `python3 -c "import dashboard"` clean (passed locally).
- [ ] `node --check clawmetry/static/js/app.js` clean (passed locally).

Diff: 89 LOC across 2 files.

### DIFF
--- a/clawmetry/static/js/app.js
+++ b/clawmetry/static/js/app.js
@@ -3806,6 +3806,20 @@ async function loadBrainPage(silent) {
     var streamEl = document.getElementById('brain-stream');
     var wasAtTop = !streamEl || streamEl.scrollTop < 40;
     renderBrainStream(events);
+    // Issue #1195 — One-time "Live event details restored" toast for users
+    // who lived through the 0.12.182 empty-detail regression. Fires only if
+    // (a) at least one event has a non-empty detail, (b) the user hasn't
+    // already seen the toast (localStorage flag), and (c) the install
+    // pre-dates the regression cutoff (epoch derived from config.json
+    // ctime). Fresh installs and repeat visits skip silently.
+    try {
+      var hasDetail = false;
+      for (var _i = 0; _i < events.length; _i++) {
+        var _d = events[_i] && (events[_i].detail || events[_i].text || events[_i].content);
+        if (_d && String(_d).trim() !== '') { hasDetail = true; break; }
+      }
+      if (hasDetail) maybeShowBrainRestorationToast();
+    } catch (_toastErr) { /* never block render */ }
   } catch(e) {
     // Always replace the spinner on error -- previously this only fired
     // when `silent=false`, so a silent retry that errored left the prior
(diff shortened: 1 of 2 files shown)


vivekchand:
## Product review

- 🟢 **Trust regression is a real funnel risk.** Per the issue, anyone who touched ClawMetry between 0.12.182 ship and 0.12.183+ install carries a memory of broken-Brain forever. Silently fixing the read path heals it for the codebase but not for the user — this toast is the patch on the user's mental model.
- 🟢 **Single-server endpoint over pure-localStorage was the right pick.** Per the agent's notes, pure localStorage can't distinguish "new browser" from "fresh install" — a re-installed user would lose the flag, get the toast, and be confused. 4-line endpoint is minimum-viable signal.
- 🟢 **One-shot per origin via localStorage** + server-side ctime check together make double-firing essentially impossible. Asymmetric value: a tiny piece of state that prevents a corrosive UX pattern.
- 🟢 **Cloud Mode short-circuit** is correct — the cloud dashboard is a different surface and doesn't hit this regression class. No tier-strategy implication.

🤖 Pre-merge review @ cf225ebf. Pushing new commits will re-trigger review if diff >50 lines or >2h elapsed.

vivekchand:
## UX review

- 🟢 **Direct user-trust repair.** Toast copy *"Live event details restored. Brain stream is back to normal."* is the right read: it acknowledges something was broken, names what's fixed, and promises nothing else. No em-dashes, no AI-tell phrasing, no jargon. Reads natural to a non-technical user.
- 🟢 **Reuses `.cron-toast` CSS class** — no new style surface, no risk of visual drift. Auto-dismisses after 5s with fade-out (300ms). Standard pattern.
- 🟢 **Fresh-install gate is the right call.** ctime cutoff at 2026-05-12 epoch means anyone who installed AFTER the regression window never sees the toast — they wouldn't know what it's referencing. That's the difference between an empathy moment and a confusing one.
- 🟢 **Fails closed on errors.** `/api/install-age` 404/timeout → no toast (never spam). `localStorage` blocked → no toast. `CLOUD_MODE` → no toast. Four independent escape hatches.
- 🟢 **localStorage flag name versioned `_v1`** — preserves the option to re-fire a future variant without losing the past clear-the-flag distinction.

🤖 Pre-merge review @ cf225ebf. Pushing new commits will re-trigger review if diff >50 lines or >2h elapsed.

PR #1468, opened 2026-05-16
perf(daemon-proxy): migrate 2 remaining RW get_store() callsites + lint to prevent regression

### LINKED ISSUE #1240: [Product P0 from PR #1235] Sweep ~28 RW get_store() callsites + add lint to prevent regression
From the post-merge product review of PR #1235.

PR #1235 fixed 3 fast-path helpers that were calling `local_store.get_store()` with the default `read_only=False`, paying DuckDB's ~2.5s exclusive-lock-retry budget on every request because the sync daemon owns the writer lock cross-process.

The fix is correct but local. Grep across `routes/` shows the broader landscape:

- **46** `local_store.get_store(...)` callsites across 18 files
- **18** of them pass `read_only=True`
- ⇒ **~28 callsites still default to RW** and will pay the same lock-retry tax under the standard install (sync-daemon writer + dashboard reader, two processes).

Every one of those 28 is a latent \"dashboard breaks under real load\" bug of the same class as the one Peter Steinberger nearly hit. The bugs in PR #1235 sat undetected for months precisely because dev-mode runs the daemon in-process (no lock contention) — the failure only manifests on a real install.

### Action items

1. Sweep PR converting every read-path `get_store()` callsite in `routes/` to `get_store(read_only=True)`.
2. Where the daemon proxy exists (`routes/local_query.py::local_store_via_daemon`), prefer it as the primary path, RO direct open as fallback — same shape as PR #1235.
3. Add a CI check (flake8 plugin or grep-based pre-commit hook) that flags any `local_store.get_store()` call in `routes/` without an explicit `read_only=True` kwarg. The check should fail the build, not warn.
4. Document the daemon-proxy-first / RO-direct-fallback pattern in CLAUDE.md or ARCHITECTURE.md so future routes don't regress.

### Files to review (RW callsites — likely candidates)

`routes/components.py`, `routes/brain.py`, `routes/usage.py`, `routes/sessions.py`, `routes/crons.py`, `routes/alerts.py`, `routes/reasoning.py`, `routes/heartbeat.py`, `routes/autonomy.py`, `routes/advisor.py`, `routes/nemoclaw.py`, `routes/agents.py`, `routes/channels.py`, `routes/bootstrap.py` — confirm each `get_store(` matches `get_store(read_only=True` or has a documented reason for RW.

### Why P0

This is the dominant source of dashboard-perceived-broken-ness on real installs. PR #1235 fixed the most-visible 3 endpoints; the next 28 will silently break the moment any user hits them under real conditions. We just got lucky the influencer landed on `/api/overview` and not `/api/sessions` or `/api/usage`.

vivekchand:
## Summary

Issue #1240 asked for a sweep of ~28 writable `local_store.get_store()` callsites in `routes/` flagged in the post-merge product review of PR #1235.

**Audit result:** the long tail has already been fixed by subsequent PRs (#1228, #1256, #1277, #1291). A fresh grep over `routes/` shows **only 2** writable callsites still remained:

| File | Function | Status |
|---|---|---|
| `routes/brain.py:222` | `_try_local_store_brain` fallback | now `read_only=True` |
| `routes/sessions.py:304` | `_fetch_sessions_table_rows` fallback | now `read_only=True` |

Both were single-process fallbacks after the daemon-proxy attempt — so they only fired in tests / dev mode under the old behaviour, but any real install that briefly lost the daemon would have paid DuckDB's full ~2.5s writer-lock-retry tax per request. Fix matches the pattern PR #1235 established on the hot endpoints.

## What's in this PR

1. **2 minimal fixes** — add `read_only=True` to the two remaining writable opens (+ inline comment citing #1240).
2. **CI lint guard** (`tests/test_no_direct_get_store_in_routes.py`) that walks every `routes/*.py` file, tokenizes it (so docstring / comment mentions don't trip the check), and asserts that every `get_store(...)` call passes an explicit `read_only=True` kwarg. Future regressions fail the build.
   - Includes a per-line `# noqa: get_store-rw` opt-out for the (currently zero) legitimate writable callsites.
   - Ships with a self-test that verifies the regex catches all 5 known-bad shapes and lets all 5 known-good shapes through, so a regex typo can't silently let new offenders past CI.

## Scope

- Total diff: **~160 LOC** (most of it the lint test + its docstring).
- Closes #1240 — both action items 1 (sweep) and 3 (CI check) from the issue. Action items 2 (daemon-proxy-first pattern) and 4 (ARCHITECTURE.md doc) are already in place for every endpoint touched (#1228/#1256/#1277/#1291).

## Test plan

- [x] `python3 -m pytest tests/test_moat_send_message_e2e.py tests/test_local_query_api.py tests/test_no_direct_get_store_in_routes.py -q` — **19/19 passing** (17 baseline + 2 new lint).
- [x] `python3 -m pytest tests/test_brain_local_fastpath.py tests/test_brain_local_store.py -q` — **6/6 passing** (touched module).
- [x] `python3 -c "import dashboard"` — clean.
- [x] Lint test fails as expected if you revert either of the two fixes (verified locally — the regex caught both).

### DIFF
--- a/routes/brain.py
+++ b/routes/brain.py
@@ -219,7 +219,11 @@ def _try_local_store_brain(limit: int, include_artifacts: bool):
     if rows is None:
         try:
             from clawmetry import local_store
-            store = local_store.get_store()
+            # Issue #1240: read_only=True so the single-process fallback
+            # doesn't pay DuckDB's writer-lock-retry tax under the standard
+            # install (daemon proxy above is the happy path; this fallback
+            # only fires in tests / dev mode).
+            store = local_store.get_store(read_only=True)
             rows = store.query_events(limit=limit)
         except Exception:
             return None
--- a/routes/sessions.py
+++ b/routes/sessions.py
@@ -298,10 +298,14 @@ def _fetch_sessions_table_rows(limit: int = 200):
     except Exception:
         pass
     # 2. Single-process fallback: open the DuckDB ourselves. Only works
-    #    when no other process holds the writer lock.
+    #    when no other process holds the writer lock. Issue #1240: open
+    #    read-only so we don't pay DuckDB's ~2.5s writer-lock-retry budget
+    #    when the daemon does own the writer lock (the daemon proxy attempt
+    #    above will have already failed in that case, so we get here, and
+    #    the RO open shares the writer's connection cleanly).
     try:
         from clawmetry import local_store
-        return local_store.get_store().query_sessions_table(limit=limit)
+        return local_store.get_store(read_only=True).query_sessions_table(limit=limit)
     except Exception:
         return None
 
--- a/tests/test_no_direct_get_store_in_routes.py
+++ b/tests/test_no_direct_get_store_in_routes.py
@@ -1,0 +1,153 @@
+"""Lint guard: prevent regressions on issue #1240.
+
+Every ``local_store.get_store(...)`` call in ``routes/`` must pass an
+explicit ``read_only=True`` kwarg. The default (writable) open pays
+DuckDB's ~2.5s exclusive-lock-retry budget on every request when the
+sync daemon owns the writer lock cross-process, which is the standard
+install topology. PR #1235 fixed 3 hot endpoints that fell into this
+trap; issue #1240 swept the long tail.
+
+The check is intentionally a dumb regex over the route module sources
+(not an AST walk) — it has to catch ``local_store.get_store()`` whether
+the call is bare or chained (``local_store.get_store().query_*``), and
+across both bound and unbound forms. A pre-flight CI gate that fails
+the build is what the issue asks for; AST sophistication would just
+mean more ways to silently pass.
+
+Allowed forms:
+    local_store.get_store(read_only=True)
+    local_store.get_store(read_only=True, ...)
+    get_store(read_only=True)           # when imported directly
+
+Disallowed forms (will fail this test):
+    local_store.get_store()
+    local_store.get_store().query_x(...)
+    get_store()                          # bare, even when imported directly
+
+To bypass for a legitimately-writable callsite (none expected in
+``routes/`` — writes belong in the daemon, not the dashboard), add a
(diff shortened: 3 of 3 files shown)


vivekchand:
## Product review

- 🟢 **Stale issue, accurate fix.** Issue body said "~28 callsites" but the real residue was 2 — six intervening PRs (#1228, #1256, #1277, #1291, et al.) had already drained the bulk. The agent's discipline to scope to current reality (not the issue's count) avoids a 300-LOC PR that fixes nothing.
- 🟢 **Lint is the compounding piece.** Future PRs that add `local_store.get_store()` to `routes/` now fail CI with a specific message and a file:line. Closes the bug class permanently — exactly the kind of asymmetric-value investment the user keeps asking for.
- 🟢 **Tokenize-based scan over regex-only** is the right rigour. A docstring example like `get_store()` doesn't trip a false positive. Self-test (5 bad / 5 good shapes) catches regex regressions in the lint itself.
- 🟡 **Nit (not blocking):** the lint runs on every test invocation. If `routes/` grows to hundreds of files, this could add a few hundred ms to `pytest` startup. Easy follow-up: move to a separate CI step or cache the scan if it ever shows up in test-suite latency.

🤖 Pre-merge review @ be86f442. Pushing new commits will re-trigger review if diff >50 lines or >2h elapsed.

vivekchand:
## UX review

- 🟢 **Backend / dev-tooling only.** No user-visible surface; the UX here is the developer-facing CI failure message. Lint emits offender file:line + offending line text, which makes the failure self-documenting.
- 🟢 **The `# noqa: get_store-rw` escape hatch is correctly designed.** Same-line marker, honoured by the lint but still counted (so visibility stays). Future maintainers don't have to monkey with the linter to express a legitimate write.
- 🟢 **Two-line code change + 153-line lint** is the right ratio. The fixes were trivial; the regression guard is where the value sits. Pattern to copy for future single-spot product P0s.

🤖 Pre-merge review @ be86f442. Pushing new commits will re-trigger review if diff >50 lines or >2h elapsed.

PR #1469, opened 2026-05-16
refactor(local-store): consolidate channel-event dual-write into ingest_channel_event() (closes #1220)

### LINKED ISSUE #1220: [Product P0 from PR #1212] Consolidate channel-event dual-write into LocalStore.ingest_channel_event()
From the post-merge product review of PR #1212.

## Pattern: dual-writes silently drift out of contract

PR #1212 fixed a P0 user-visible bug: telegram outbound messages were being parsed from `gateway.log` and written to the `channel_messages` DuckDB table — but NOT projected onto the `events` table that the Brain feed (`/api/brain-history`) reads from. The Brain tab silently showed an empty Telegram channel for every user.

This is the **second time in a month** we've shipped a regression where one DuckDB writer projected to a subset of the tables a downstream reader queries:
- 2026-05-13: `CLAWMETRY_LOCAL_STORE_READ=1` opt-in flag silently bypassed Phase 1-5 fast-paths for 100% of users (`feedback_local_store_default_off_killed_moat`)
- 2026-05-14: PR #1212 fix — telegram dual-write missing for events table

## Root cause
We have N writers each manually dual-writing to `channel_messages` + `events`:
- `sync_channel_messages` (JSONL → both, since day one)
- `sync_telegram_from_gateway_log` (gateway.log → channel_messages only — fixed in #1212 to dual-write)
- `gateway_tap._handle_frame` (WS → both, in PR #1213)

That's **three implementations of the same projection logic**. PR #1213 already carries its own copy of `_telegram_outbound_event_row`-style code in `gateway_tap.py:_normalize_frame`. The next channel adapter will be the fourth.

## Proposed fix
Introduce a single chokepoint in `LocalStore`:

```python
def ingest_channel_event(self, channel_msg: dict, *, node_id: str, agent_type: str = 'openclaw') -> None:
    """Write a channel turn to both channel_messages AND events.

    The ONLY way to ingest channel turns. Direct writes to either
    table for channel data are a contract violation.
    """
    self.ingest_channel_message(channel_msg)
    self.ingest_many([self._project_to_events_row(channel_msg, node_id, agent_type)])
```

Then migrate all three writers onto it. Drop `_telegram_outbound_event_row` from `sync.py` and the inline projection in `gateway_tap.py`.

## Acceptance
- [ ] `LocalStore.ingest_channel_event()` exists and is the only public API for channel ingest
- [ ] `sync_telegram_from_gateway_log`, `sync_channel_messages`, and `gateway_tap._handle_frame` all call it
- [ ] Lint rule (or CI grep) flags any new `ingest_channel_message` / `ingest_many(channel.*)` call outside `local_store.py`
- [ ] Regression test for the JSONL path (sibling of `test_outbound_messages_reach_events_table_for_brain_feed` from PR #1212) lives in `tests/test_sync.py`

## Why this matters for the funnel
Empty Brain tab → user thinks tool is broken → user uninstalls → never lands on Cloud-Pro. "Channel data shows up" is a baseline OSS promise; we cannot afford a third regression of this shape.

vivekchand:
## Summary

Two May 2026 P0 regressions (#1212 Telegram outbound vanishing from Brain, #1438 local-store opt-in default OFF) shared the same root cause: N writers each hand-rolled a dual-write to `channel_messages` + `events` and one of them silently drifted out of contract.

Introduce a single chokepoint in `LocalStore`:

```python
def ingest_channel_event(channel_msg: dict, *, node_id: str) -> None:
    """Write to channel_messages AND events atomically.
    Replaces the prior N hand-rolled dual-writes that drifted any time
    the projection logic was touched on one side and not the other."""
```

It writes the channel_messages row, then derives + ingests the `events` projection (event_type=channel.in|channel.out, session_id NULL, agent_type=openclaw, data=flattened raw_blob). Callers no longer get to fan out by hand.

## What changed

Migrated all 3 callsites and DELETED the duplicated projection code (no shims, per CLAUDE.md):

- `sync.sync_channel_messages` (JSONL -> both tables)
- `sync.sync_telegram_from_gateway_log` (gateway.log -> both)
- `gateway_tap.GatewayTap._handle_frame` (WS -> both)

`sync._parse_channel_event` now returns a single channel_msg dict (was a (events, channel) tuple). `gateway_tap._normalize_frame` same -- returns one flat dict. `sync._telegram_outbound_event_row` helper is gone; the chokepoint owns that projection.

Net diff: +393 -178 lines, but ~177 of those additions are tests + comments. The actual production code is a net REDUCTION: -123 lines in sync.py, -37 in gateway_tap.py, +90 in local_store.py for the chokepoint + docstring.

## Test plan

New `tests/test_channel_event_chokepoint.py`:
- [x] both tables populated from one call
- [x] outbound direction maps to channel.out
- [x] re-ingest is idempotent (PK + INSERT OR IGNORE)
- [x] end-to-end through /api/brain-history (the original P0 regression path -- a real Flask client hits the route and sees CHANNEL.IN with direction='in')

Regression coverage (all green):
- [x] `tests/test_local_query_api.py` (17 baseline)
- [x] `tests/test_moat_send_message_e2e.py` (baseline)
- [x] `tests/test_gateway_tap.py` (3 tests rewritten for the new single-dict `_normalize_frame` return shape)
- [x] `tests/test_telegram_gatewaylog_parser.py`
- [x] `tests/test_e2e_telegram_brain_ingest.py`

Closes #1220

### DIFF
--- a/clawmetry/gateway_tap.py
+++ b/clawmetry/gateway_tap.py
@@ -121,14 +121,21 @@ _ENABLE_ENV = "CLAWMETRY_ENABLE_WS_TAP"
 # Required: provider, ts. Everything else is best-effort with a sane
 # default. Returning ``None`` skips the frame entirely.
 def _normalize_frame(frame: dict) -> dict | None:
-    """Project a gateway WS event frame into a (events, channel_messages)
-    tuple-shaped dict, or ``None`` if the frame is not a chat message.
+    """Project a gateway WS event frame into a ``channel_messages`` row
+    dict, or ``None`` if the frame is not a chat message.
 
     The gateway's own per-channel event names vary
     (``sessions.message``, ``telegram.inbound``, ``channel.message``,
     …). We accept any frame whose ``payload`` carries enough fields to
     pin a channel + a timestamp. Heartbeats / health / status frames
     fall through to ``None`` and the caller drops them.
+
+    The returned dict is passed straight to
+    ``LocalStore.ingest_channel_event`` (issue #1220) which fans it
+    onto BOTH the ``channel_messages`` and ``events`` tables in a
+    single chokepoint — earlier versions of this module hand-rolled
+    a parallel ``events`` row here, which drifted out of contract
+    every time the projection logic was touched.
     """
     if not isinstance(frame, dict):
         return None
@@ -267,26 +274,7 @@ def _normalize_frame(frame: dict) -> dict | None:
         or sender_block.get("name")
     )
 
-    events_row = {
-        "id": eid,
-        "agent_id": "main",
-        "event_type": f"channel.{direction}",
-        "ts": ts,
-        "session_id": None,
-        "workspace_id": None,
-        # Tag the frame's source so brain filters can split WS-tap rows
-        # from log-parser rows when debugging. Stays inside `data` so
-        # it doesn't bloat the events schema.
-        "data": {
-            **payload,
-            "_clawmetry_source": f"channel:{provider}:{direction}",
-            "_clawmetry_event": event_name,
-        },
-        "cost_usd": None,
-        "token_count": None,
-        "model": None,
-    }
-    channel_row = {
+    return {
         "id": eid,
         "agent_id": "main",
         "provider": provider,
@@ -297,13 +285,17 @@ def _normalize_frame(frame: dict) -> dict | None:
         "ts": ts,
         "direction": direction,
         "session_key": payload.get("session_id") or payload.get("session_key"),
+        # raw_blob preserves the full payload + WS-tap source
+        # breadcrumbs. ``ingest_channel_event`` flattens it into the
+        # events-table ``data`` blob so brain filters can split WS-tap
+        # rows from log-parser rows when debugging without us having
+        # to maintain two parallel projections.
         "raw_blob": {
             **payload,
             "_clawmetry_source": "gateway.ws",
             "_clawmetry_event": event_name,
         },
     }
-    return {"events": events_row, "channel": channel_row}
 
 
 # ── The tap loop ────────────────────────────────────────────────────────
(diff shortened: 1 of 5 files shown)


vivekchand:
## UX review

- 🟢 **Pure structural refactor — no user-visible surface.** The win is invisible: a single canonical write path means the Brain tab can't silently lose a row from a new channel adapter or new sync source.
- 🟢 **Backend developer-UX is meaningfully better.** `store.ingest_channel_event(channel_msg, node_id=...)` reads as one obvious entry point. Anyone adding a new channel ingest now has one function to call and 0 dual-write boilerplate to copy.
- 🟢 **No deprecation shim, no compat layer.** Per CLAUDE.md no-shim rule, the old hand-rolled events_row projection in `gateway_tap._normalize_frame` is just gone, replaced by raw_blob flattening in the chokepoint. Clean diff, easy to follow.

🤖 Pre-merge review @ 910a7e21. Pushing new commits will re-trigger review if diff >50 lines or >2h elapsed.

vivekchand:
## Product review

- 🟢 **Closes a bug class, not just a bug.** Two distinct P0 regressions in May (#1212 telegram-to-Brain, #1438 local-store opt-in default) had the same root cause: dual-write drift. Eliminating the pattern removes the surface where any future third-channel-adapter could regress.
- 🟢 **Net reduction in production code** (+90 chokepoint, −37 in gateway_tap, −123 in sync.py = −70 net). The diff number looks big because of the new test file (+150 LOC), but the actual code surface shrunk. Less code = fewer places to drift.
- 🟢 **E2E test through `/api/brain-history`** asserts the exact regression path that bit us. The test would fail if a future refactor split the chokepoint back into two writes — which is the only thing that matters here.
- 🟢 **Single-transaction guarantee** in the helper means a partial failure can't leave channel_messages populated but events empty. That was the exact failure mode of #1212.

🤖 Pre-merge review @ 910a7e21. Pushing new commits will re-trigger review if diff >50 lines or >2h elapsed.

PR #1470, opened 2026-05-16
fix(privacy): assert claude_session_id from sessions.json before walking ~/.claude/projects (closes #1231)

### LINKED ISSUE #1231: [Product P0 from PR #1227] Tighten privacy boundary on ~/.claude/projects walking — assert claude_session_id originates from sessions.json
From the post-merge product review of PR #1227.

## Concern
PR #1227 introduced `sync_openclaw_claude_sessions_via_index`, which walks `~/.openclaw/agents/main/sessions/sessions.json`, extracts the bound Claude CLI session UUID, and crawls `~/.claude/projects/<encoded-cwd>/<id>/subagents/*` and `tool-results/*`.

The current boundary that prevents ingesting unrelated personal Claude Code transcripts is: "the code only crawls under a specific `claude_session_id` that came from sessions.json." That boundary is implicit — there is no assertion in the ingest path. A future refactor (e.g. globbing the parent dir for "all sessions in this workspace") would silently start ingesting every personal `claude` chat the user ever ran from the OpenClaw workspace dir.

## Why this matters
- ClawMetry Cloud sync is opt-in but on for many users
- The encoded-cwd slug `-Users-vivek--openclaw-workspace` is shared by ANY `claude` invocation from that directory — including manual ones the user did not intend to observe
- Personal Claude conversations leaking into a cloud-synced observability dashboard is a privacy-incident-shaped risk

## Proposed fix
1. In `_claude_session_dir` and `_claude_session_top_level_path` (`clawmetry/sync.py:287-300`), accept a `bound_session_ids: set[str]` parameter and assert membership before returning a path.
2. Add a unit test: materialise `~/.claude/projects/<slug>/<unrelated-uuid>.jsonl` and `<unrelated-uuid>/subagents/*.jsonl` alongside the bound session. Assert the unrelated files are NEVER ingested.
3. Document the privacy boundary explicitly in the module docstring.

## Reference
- PR: https://github.com/vivekchand/clawmetry/pull/1227
- File: `clawmetry/sync.py:287-300, 502-628`
- Related memory: `project_local_compute_cloud_display.md` — data E2E encrypted but operator still owns the ingest scope decision.

vivekchand:
## Summary
- **Privacy gate** on `sync_openclaw_claude_sessions_via_index` (PR #1227): the path-construction helpers `_claude_session_dir` and `_claude_session_top_level_path` now require an `allowed_claude_session_ids: set[str]` arg and `raise PermissionError` if the requested id is not in that set. The caller builds the set from `sessions.json`'s `claudeCliSessionId` field once per cycle and threads it through.
- Uses `raise` rather than `assert` so the guard survives `python -O`.
- Inline **INVARIANT** comment above the helpers documents why the boundary exists and warns against future "glob the parent dir" refactors.
- **Regression test** `tests/test_claude_sessions_privacy_boundary.py` with two layers:
  1. Unit: helpers raise when given an id outside the allowlist.
  2. E2E: a synthetic personal transcript under the same encoded-cwd slug is **never** ingested by the full sync path. Manually verified to fail when the guard is defanged.

## Why this matters
`~/.claude/projects/<encoded-cwd>/` is shared by **every** `claude` CLI invocation from that directory, including manual personal chats the user did not intend to observe. The only thing preventing those from flowing into cloud sync was the implicit assumption "we only walk ids that came from sessions.json" — this PR makes it explicit. See issue #1231.

## Test plan
- [x] `python3 -m pytest tests/test_moat_send_message_e2e.py tests/test_local_query_api.py -q` → 17/17 pass (current baseline; was 19 in older snapshot)
- [x] `python3 -m pytest tests/test_openclaw_claude_session_ingest.py tests/test_openclaw_sessions_followup.py -q` → 17/17 pass
- [x] New `tests/test_claude_sessions_privacy_boundary.py` → 2/2 pass
- [x] Verified the new test FAILS when `_assert_claude_session_allowed` is replaced with a no-op or when `_walk_openclaw_session_bindings` is patched to also return unrelated transcripts (simulating the "glob the parent dir" refactor the issue warns about).

Closes #1231

### DIFF
--- a/clawmetry/sync.py
+++ b/clawmetry/sync.py
@@ -3174,18 +3174,58 @@ def _walk_openclaw_session_bindings(sessions_dir: str) -> list[dict]:
     return out
 
 
-def _claude_session_dir(workspace_dir: str, claude_session_id: str) -> Path:
+# INVARIANT: ``claude_session_id`` MUST originate from sessions.json's
+# ``claudeCliSessionId`` field — never from globbing/listing the parent dir.
+# The encoded-cwd slug under ``~/.claude/projects/`` is shared by EVERY
+# ``claude`` invocation the user ever ran from the OpenClaw workspace,
+# including personal chats the user did not intend to observe. The
+# ``allowed_claude_session_ids`` allowlist (built from sessions.json by the
+# caller) is the only thing keeping us from sweeping those into the cloud-
+# sync pipeline. Future refactors: keep this gate or replicate it; do NOT
+# loosen it to "all sessions in this workspace". See issue #1231.
+def _assert_claude_session_allowed(
+    claude_session_id: str, allowed_claude_session_ids: set[str],
+) -> None:
+    """Privacy gate. Raise ``PermissionError`` if ``claude_session_id`` was
+    not harvested from OpenClaw's sessions.json — i.e. it is not a Claude
+    CLI session OpenClaw bound to one of its own OpenClaw sessions.
+
+    A plain ``assert`` would be stripped by ``python -O``; this raises
+    unconditionally because the boundary is privacy-critical.
+    """
+    if claude_session_id not in allowed_claude_session_ids:
+        raise PermissionError(
+            "refusing to walk ~/.claude/projects for session "
+            f"{claude_session_id!r}: not in sessions.json allowlist "
+            f"(size={len(allowed_claude_session_ids)})"
+        )
+
+
+def _claude_session_dir(
+    workspace_dir: str,
+    claude_session_id: str,
+    allowed_claude_session_ids: set[str],
+) -> Path:
     """Return ``~/.claude/projects/<encoded-cwd>/<claude-session-id>/`` —
     the directory that holds the subagents/ and tool-results/ subdirs.
+
+    ``allowed_claude_session_ids`` MUST be the set of Claude CLI session
+    UUIDs harvested from ``sessions.json``. See the invariant comment
+    above ``_assert_claude_session_allowed``.
     """
+    _assert_claude_session_allowed(claude_session_id, allowed_claude_session_ids)
     slug = _encode_cwd_for_claude_projects(workspace_dir)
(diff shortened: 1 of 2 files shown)


vivekchand:
## UX review

- 🟢 **Invisible when it works.** Privacy gates should never have a user-visible surface — they should silently refuse to do the wrong thing. This does that.
- 🟢 **`PermissionError` is the right exception class** for a privacy boundary. Future logs/alerts can pattern-match on exception type to surface privacy-relevant fails differently from generic IOErrors.
- 🟢 **`raise` over `assert`** is the right rigour — survives `python -O` runs (which some packagers/users do). The agent caught the canonical `assert`-defeated-by-O bug class without prompting.

🤖 Pre-merge review @ e319ab28. Pushing new commits will re-trigger review if diff >50 lines or >2h elapsed.

vivekchand:
## Product review

- 🟢 **Closes a privacy-incident-shaped risk.** The original boundary ("only walk ids from sessions.json") was implicit — a single future PR globbing the parent dir would silently start ingesting every personal Claude chat the user ever ran from the OpenClaw workspace. With cloud sync opt-in but on for many users, this is exactly the kind of risk that would land on Twitter as "ClawMetry uploaded my private GPT chats."
- 🟢 **INVARIANT comment + 2-layer regression test** (unit + filesystem-mocked E2E) makes the boundary durable. The E2E test specifically constructs a fake `~/.claude/projects/<slug>/UNRELATED-UUID/` with synthetic content and asserts it never lands in DuckDB. That's the test that would have caught the hypothetical "glob the parent dir" refactor before it shipped.
- 🟢 **Set-based allowlist built once per cycle** is O(1) lookup per session — no perf regression vs the prior implicit boundary.
- 🟢 **`raise` survives `python -O`** — closes the entire class of "privacy assert silently disabled in prod" bugs.

🤖 Pre-merge review @ e319ab28. Pushing new commits will re-trigger review if diff >50 lines or >2h elapsed.

Issue #1385, opened 2026-05-15
[MOAT smoke 2026-05-16] DuckDB fast-paths filter event_type='message' but real OpenClaw schema emits model.completed/assistant — silent zeros


vivekchand:
## Summary

Ran the standing MOAT round-trip smoke test against 0.12.228 with a real OpenClaw agent turn. Three of today's seven new fast-paths silently return zero data on a real-world DuckDB because their SQL filters on `event_type = 'message'`, but no event in a live OpenClaw v3 install carries that event_type. Endpoints fall through to the legacy filesystem walker (when one exists) so `/api/*` JSON looks plausible — but the DuckDB fast-path is dead code on real data.

## Repro

1. `clawmetry==0.12.228` running locally; OpenClaw 2026.5.7 gateway up; sync daemon flushed.
2. Sent message via:
   ```
   openclaw agent --agent main --message "TEST MOAT 2026-05-16 25511" --json
   ```
3. CLI returned success, model=`claude-opus-4-7`, usage `{input:6, output:108, cacheRead:18498, cacheWrite:19325}`, session_id `01c6c2d6-44bd-4dbf-a554-ade04c5869a9` (CLI binding) → DuckDB session `575597e9-f609-4e88-9c12-055392f1c107`.
4. Polled `/__local_query__/health` until `event_count` grew (`751 → 754`, +3 events).

## Expected

- `query_context_window_peek` returns the new turn's `input_tokens=6`.
- `query_recent_read_tool_calls` returns rows for any Read-tool calls in the session.
- `query_model_fallbacks` returns `scanned >= 1` for the live session.
- `/api/context-anatomy` shows a "Session history" bucket reflecting the new turn.

## Actual

DuckDB row distribution after the run (top types):
```
subagent:assistant: 126  | assistant: 111  | subagent:user: 78
queue-operation: 72      | user: 65        | channel.out: 24
attachment: 11           | tool-result: 4  | model.completed: 2
session.started: 2       | prompt.submitted: 2  | tool.result: 1
message events: 0   <-- ZERO
```

Daemon proxy results:
- `query_context_window_peek` -> `{"input_tokens": 0}` (should be 6)
- `query_recent_read_tool_calls` -> `[]` (matches WHERE on `'message'` + `'tool.call'/'toolCall'/'tool_use'`, none of which exist)
- `query_model_fallbacks` -> `{"scanned": 0, "sessions_affected": 0, "top_transitions": []}`
[...]

PR #1472, opened 2026-05-16
test(moat): v3 fixture invariant suite for all DuckDB fast-paths (closes #1393)

### LINKED ISSUE #1393: [Product P0 from PR #1386] CI guard: v3 fixture must assert all DuckDB fast-paths return non-zero
From the post-merge product review of PR #1386.

For roughly 10 wheel releases, four high-signal DuckDB fast-paths in `clawmetry/local_store.py` (`query_context_window_peek`, `query_model_fallbacks`, `query_recent_read_tool_calls`, `query_tool_call_invocations`) silently returned zero rows on real OpenClaw v3 because the predicates filtered `event_type='message'` while v3 emits `assistant` / `subagent:assistant` / `model.completed` / `tool-result`. The legacy JSONL walker masked the regression in unit tests.

Three of the empty surfaces (Context Anatomy, Plugins, Model Fallbacks) are direct paid-conversion funnels — they exist to surface anomalies that lead users to upgrade for alerts. Showing zeros for months blunts that funnel.

Ask: add a CI job that
1. Boots a real OpenClaw v3 event fixture (use the distribution captured in PR #1386 — 763 events across the v3 event_type set).
2. Calls every `query_*` fast-path in `local_store.py`.
3. Fails the build if any fast-path returns zero rows on a fixture that obviously should produce results.

This makes schema-drift regressions impossible to merge — the test asserts intent, not predicate strings.

vivekchand:
## Summary
- Adds `tests/test_duckdb_fastpath_v3_invariants.py` — CI guard that seeds a hermetic DuckDB with real OpenClaw v3 event shapes (`assistant`, `subagent:assistant`, `model.completed`, `tool.call`, `tool-result`, `session_start`) and asserts every events-table fast-path returns a non-empty meaningful result.
- Covers the 4 query methods explicitly called out in #1393 (`query_context_window_peek`, `query_model_fallbacks`, `query_recent_read_tool_calls`, `query_tool_call_invocations`) plus the broader bug-class queries (`query_aggregates`, `query_daily_usage_splits`, `query_sessions`, `query_events`) — 8 fast-paths in total.
- Fixture includes assistant + `model.completed` sibling pairs at the SAME `ts_sec` to exercise the dedupe SQL from PRs #1444 / #1460 / #1464 — would have caught the dedupe regressions before they shipped.

Closes #1393.

## Test plan
- [x] `python3 -m pytest tests/test_duckdb_fastpath_v3_invariants.py -q` — 8 passed.
- [x] `python3 -m pytest tests/test_moat_send_message_e2e.py tests/test_local_query_api.py tests/test_duckdb_fastpath_v3_invariants.py -q` — 25 passed (17 baseline + 8 new).

## Notes
- `query_cost_split` and `query_session_model_journey` still filter `event_type='message'` exclusively. Documented in a tail comment as known v3 zeros (not patched as part of #1385) — add tests here when those get migrated.
- Top-level tool-call event uses `tool.call` (the form the Python iterator helpers in `local_store.py` accept). The SQL predicate `_TOOL_CALL_TOPLEVEL_EVENT_TYPES` also accepts `tool_call` (underscore), but the iterator currently doesn't — a latent gap kept out of scope here.

### DIFF
--- a/tests/test_duckdb_fastpath_v3_invariants.py
+++ b/tests/test_duckdb_fastpath_v3_invariants.py
@@ -1,0 +1,325 @@
+"""CI guard for #1393: DuckDB fast-paths must return non-zero on v3.
+
+For ~10 wheel releases four high-signal queries
+(``query_context_window_peek``, ``query_model_fallbacks``,
+``query_recent_read_tool_calls``, ``query_tool_call_invocations``)
+silently returned zero rows on real OpenClaw v3 because their predicates
+filtered ``event_type='message'`` while v3 emits ``assistant`` /
+``subagent:assistant`` / ``model.completed`` / ``tool.call`` /
+``tool-result``. Three of the empty surfaces (Context Anatomy, Plugins,
+Model Fallbacks) are direct paid-conversion funnels — showing zeros for
+months blunted them. The legacy JSONL walker masked the regression in
+unit tests because it parsed every shape.
+
+This file is the canary: seed a DuckDB with the SAME event shapes a real
+OpenClaw v3 install produces, then assert every events-table fast-path
+returns a non-empty meaningful result. If a future PR re-introduces a
+``event_type='message'``-only predicate (or any other shape regression),
+this test goes red BEFORE we ship a wheel.
+
+Fixture (matches what real OpenClaw v3 writes):
+  * 1× ``session_start``
+  * 2× ``assistant`` with full Anthropic ``data.message.usage`` envelope
+  * 2× ``model.completed`` siblings at SAME ``ts_sec`` as the assistants
+    (this is the dedupe regression case from PR #1444)
+  * 1× ``subagent:assistant`` — Task tool delegated turn
+  * 1× ``tool.call`` — top-level Read invocation
+  * 1× ``tool-result`` — tool output event
+"""
+
+from __future__ import annotations
+
+import datetime as _dt
+import importlib
+import time
+
+import pytest
+
+
+NODE_ID    = "agent+v3-invariants"
+SESSION_ID = "ssss1111-2222-3333-4444-555566667777"
+WS_ID      = "ws-v3-invariants"
+MODEL      = "claude-opus-4-7"
+PROVIDER   = "anthropic"
+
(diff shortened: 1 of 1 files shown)


vivekchand:
## UX review

- 🟢 **Test-only — no end-user UX surface.** The 'UX' here is developer-facing: when this test fires red, it must convince the developer the production code is the bug, not the test. The docstring nails that — calls out which exact 4 historical regressions this would have caught (Context Anatomy / Plugins / Model Fallbacks zeroing on v3).
- 🟢 **Fixture comments are self-documenting.** `# 2× model.completed siblings at SAME ts_sec as the assistants (this is the dedupe regression case from PR #1444)` is exactly the breadcrumb a future maintainer needs to not break this fixture.
- 🟢 **Per-query invariant assertions** (not just "non-empty") — each query has a meaningful check (`token_count > 0` vs `scanned >= 1` vs `max(fired_at)` etc). A regression that returns `[{token_count: 0}]` still fails the right test.

🤖 Pre-merge review @ d7bd3f2d. Pushing new commits will re-trigger review if diff >50 lines or >2h elapsed.

vivekchand:
## Product review

- 🟢 **This is the directly-MOAT test class.** The user mandate today was 'any future regression in DuckDB writes or fast-path reads gets caught in CI.' This file is that catch. Bringing MOAT verifier from 25 → 33 tests.
- 🟢 **Out-of-scope tail (query_cost_split, query_session_model_journey) documented inline.** Honest about what isn't covered + why; next migration drops in mirror tests. Pattern to copy.
- 🟢 **Hermetic seeding via shared envelope** keeps the per-test diff to the meaningful invariant. New invariant = add one test, not 30 lines of fixture boilerplate.
- 🟡 **Nit (not blocking):** the fixture lives entirely in the test file (no shared module). When the 6th invariant test gets added someone will refactor `_base()` and `_assistant_data()` out. Worth doing pre-emptively to a `tests/fixtures/v3_events.py` so the test file stays focused on assertions. Easy follow-up.

🤖 Pre-merge review @ d7bd3f2d. Pushing new commits will re-trigger review if diff >50 lines or >2h elapsed.

Issue #1448, opened 2026-05-16
[Product P0 from PR #1445] 24h retention cap leaks on Sessions / Usage / Brain-history / Local-events


vivekchand:
From the pre-merge product review of PR #1445.

## The leakage

PR #1445 caps `/api/flow/runs` at 24h for OSS / Cloud-Free users. The agent that implemented it flagged in its own report:

> Other read paths likely lacking the cap:
> - `/api/sessions`
> - `/api/usage`
> - `/api/brain-history`
> - `/api/local-events`
> - Anything else calling `query_*` methods on `local_store` with a `since` parameter

As-is, a free user just opens the Sessions tab and gets unlimited history there — defeating the gate that #1445 introduces on the Flow tab. Same for Brain history and historical Usage charts.

## Why this is P0

The Cloud-Pro upsell premise on the Flow tab cap (>24h = Pro) is **immediately contradicted** by every other tab giving infinite history. Users will notice within minutes ("why is Flow capped but Sessions isn't?") and treat the cap as a bug rather than a tier policy.

## Scope of fix

Apply the same `_is_pro_user()` cap + `capped_at_24h` response flag pattern to:

1. **`/api/sessions`** — `routes/sessions.py:api_sessions` (or wherever the `since` filter lives)
2. **`/api/usage`** — `routes/usage.py` — cap the daily chart window at 24h instead of 14 days for free
3. **`/api/brain-history`** — `routes/brain.py` — cap the event scan window
4. **`/api/local-events`** — `routes/local_query.py` — cap the raw event fetch

For each: surface the same upgrade CTA in the corresponding UI panel when `capped_at_24h: true`.

## Effort

S–M (~150 LOC across 4 route handlers + 4 UI CTA renders). Mirror PR #1445's pattern exactly to keep the gating contract uniform.

PR #1479, opened 2026-05-16
feat(sessions): 24h retention cap on /api/sessions + upgrade CTA (progresses #1448)

vivekchand:
## Summary

Surface 1 of 4 from issue #1448. PR #1445 capped `/api/flow/runs` at 24h for OSS / Cloud-Free users and the pre-merge product review flagged that `/api/sessions`, `/api/usage`, `/api/brain-history`, and `/api/local-events` all leaked the same retention upsell. This PR ships the **Sessions** slice; the other 3 surfaces ship separately so a regression on any one doesn't take all four down.

- `routes/sessions.py` -- `_apply_24h_retention_cap` filters sessions older than `now - 24h` when `dashboard._is_pro_user()` returns False. Hooked into both branches of `api_sessions` (local-store fast path + gateway/JSONL fallback). Tolerant timestamp shim handles epoch-ms (`updatedAt`) and ISO-8601 (`last_active_at`).
- `clawmetry/static/js/app.js` -- `_renderSessionsRetentionCta` renders a one-line CTA above `#sessions-list` when the response carries `capped_at_24h: true`. Sibling of `_renderFlowRunsCap`.
- `tests/test_sessions_retention_cap.py` -- new tests for OSS cap + Pro bypass.
- `tests/test_sessions_local_fastpath.py` -- existing fixture now defaults to `_is_pro_user=True` so pre-cap 2026-05-11 timestamps still pass (same shim PR #1445 added).

CTA copy: `Sessions older than 24 hours are available on Cloud-Pro. Upgrade for full history.` (no em-dash, period-split per memory).

Total: 86 LOC production + 113 LOC tests; mirrors PR #1445 contract exactly.

## Test plan

- [x] `python3 -c "import dashboard"` clean
- [x] `tests/test_moat_send_message_e2e.py + test_local_query_api.py` -- 17/17 green
- [x] `tests/test_sessions_retention_cap.py` -- 2/2 green (OSS-capped, Pro-bypass)
- [x] `tests/test_sessions_local_fastpath.py` -- 5/5 still green after fixture shim
- [ ] Manual smoke once merged: verify Cloud-Pro acct sees no CTA, OSS acct sees CTA + 24h list

`progresses #1448` (3 surfaces still TBD: Usage, Brain-history, Local-events).

Generated with [Claude Code](https://claude.com/claude-code)

PR #1480, opened 2026-05-16
feat(usage): 24h retention cap on /api/usage + upgrade CTA (progresses #1448)

vivekchand:
## Summary

- Surface 2 of 4 from issue #1448. Mirrors PR #1445 (Flow Runs cap) on the Tokens tab.
- OSS / Cloud-Free callers now see only today + yesterday in the 14-day token chart; Cloud-Pro callers (gated by `dashboard._is_pro_user`) keep the full window.
- Response carries `capped_at_24h` so the UI renders an upgrade CTA above the chart: `"Usage history older than 24 hours is on Cloud-Pro. Upgrade for the full 90-day picture."`

## Gating contract

- `_is_pro_user()` is fail-closed (same helper PR #1445 used). No new gating mechanism.
- Cap is applied AFTER the `_usage_cache` lookup AND AFTER the local-store fast-path dedupe so the cached aggregate stays full-fidelity (`feedback_usage_dedupe_pattern` — same handler shipped a regression last week from blindly overwriting an aggregate with a subset; this PR is careful to shallow-copy before rewriting `days`).
- All three return paths in `api_usage` (fast-path, OTLP, transcript) route through `_apply_oss_24h_cap`.
- Chart still renders 14 slots so the UI shape is unchanged; older buckets are zeroed rather than truncated so existing CSS / Chart sizing keeps working.

## Test plan

- [x] `python3 -m pytest tests/test_usage_local_store.py -q` — 33 pass (31 original + 2 new).
- [x] `python3 -m pytest tests/test_moat_send_message_e2e.py tests/test_local_query_api.py -q` — 17/17 still green.
- [x] `python3 -m pytest tests/test_usage_per_session.py -q` — 5/5 still green.
- [x] New tests: `test_api_usage_caps_14d_to_24h_for_free` (older buckets zero, `capped_at_24h=true`) and `test_api_usage_no_cap_for_pro` (full 14-day history, `capped_at_24h=false`).
- [x] Diff: 124 LOC across 3 files (under 150-LOC cap).
- [x] No em-dashes / double-dashes in any new user-facing copy (CTA uses period-split per `feedback_no_em_dashes_in_user_facing_copy`).

## Follow-up

Surfaces 3 of 4 (`/api/brain-history`) and 4 of 4 (`/api/local-events`) ship separately per issue #1448.

progresses #1448

### DIFF
--- a/clawmetry/static/js/app.js
+++ b/clawmetry/static/js/app.js
@@ -6994,6 +6994,10 @@ async function loadUsage() {
       chartHtml += '<div class="usage-bar-wrap"><div class="usage-bar" style="height:' + pct + '%"><div class="usage-bar-value">' + (d.tokens > 0 ? val : '') + '</div></div><div class="usage-bar-label">' + label + '</div></div>';
     });
     document.getElementById('usage-chart').innerHTML = chartHtml;
+    // Issue #1448 surface 2 — OSS / Cloud-Free callers get clamped to 24h
+    // of history; render the upgrade CTA above the chart so users see why
+    // the chart is mostly empty.
+    _renderUsageCapCTA(!!data.capped_at_24h);
     // Cost table
     var usageInfoIcon = document.getElementById('usage-cost-info-icon');
     if (usageInfoIcon) {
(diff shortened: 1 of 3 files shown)


vivekchand:
## UX review

- 🟢 **Keeps 14 chart slots, zeros older buckets.** UI sizing/CSS unchanged, no chart-jump. Users see what they're missing visually — empty slots = upgrade hook.
- 🟢 **CTA copy specific to usage** — "Upgrade for the full 90-day picture" calls out the Pro retention number directly, which converts better than generic "upgrade for more".
- 🟢 **Period split, no em-dash, plain language.**

🤖 Pre-merge review @ 501b0193. Pushing new commits will re-trigger review if diff >50 lines or >2h elapsed.

vivekchand:
## Product review

- 🟢 **Cap applied AFTER dedupe + cache** — `feedback_usage_dedupe_pattern.md` was the exact memory I'd quote here, and the agent honored it. Long-lived cache stays full-fidelity, capped values are computed on shallow-copy. Right separation.
- 🟢 **Wired into all 4 return paths** (fast-path, cache, OTLP, transcript) — covers every entry the response can take. No silent leak on a code path the agent forgot.
- 🟢 **Surface 2 of 4 from #1448** — same tier-symmetry argument as #1479. Half-shipped #1448 would have been worse than no #1445.

🤖 Pre-merge review @ 501b0193. Pushing new commits will re-trigger review if diff >50 lines or >2h elapsed.

PR #1481, opened 2026-05-16
feat(brain): 24h retention cap on /api/brain-history + upgrade CTA (progresses #1448)

vivekchand:
## Summary

Surface **3 of 4** from issue #1448. Mirrors the pattern PR #1445 shipped
on `/api/flow/runs`:

- OSS / Cloud-Free users now see only the **last 24 hours** of events on
  `/api/brain-history`. Cloud-Pro users (validated via the existing
  `dashboard._is_pro_user()` helper from PR #1440) keep unlimited history.
- When the cap kicks in, the response carries `capped_at_24h: true`.
- The Brain tab renders a one-line upgrade CTA above the stream:
  `Brain history older than 24 hours is on Cloud-Pro. Upgrade for full
  conversation context.` (no em-dash, period split — per
  `feedback_no_em_dashes_in_user_facing_copy.md`).

## Scope

- `routes/brain.py`: compute `cap_since` once at handler entry, forward
  to both the local-store fast path (via `query_events(since=...)`) and
  the JSONL slow path (post-filter `events` by `time`). CONTEXT
  pseudo-events bypass the cap (they describe the active window, not
  historical activity). Cache key gains the cap flag so Pro vs free
  responses never cross-pollute.
- `routes/brain.py` `_try_local_store_brain`: when the cap is active and
  rows come back empty, return the empty-but-capped payload rather than
  falling through to the unbounded JSONL parser (which would silently
  defeat the gate).
- `clawmetry/static/js/app.js`: new `_renderBrainHistoryCap(capped)`
  helper inserts/updates an upgrade-CTA banner above `#brain-stream`.
- 4 brain test files: default the local fixtures to Pro (matches PR
  #1445's pattern) so existing tests with 2026-05-11/12/13 timestamps
  still pass; new cap tests in `test_brain_local_fastpath.py` explicitly
  monkeypatch `_is_pro_user`.

## Test plan

- [x] `python3 -m pytest tests/test_moat_send_message_e2e.py tests/test_local_query_api.py tests/test_brain_local_fastpath.py tests/test_brain_local_store.py -q` → 25/25 pass
- [x] All 11 brain-tagged tests green (cap, fast-path, slow-path, v3 shape, log-noise)
- [x] Diff size 167 LOC (96 production + 71 test fixtures + new cap tests)

### DIFF
--- a/clawmetry/static/js/app.js
+++ b/clawmetry/static/js/app.js
@@ -3828,6 +3828,10 @@ async function loadBrainPage(silent) {
     var streamEl = document.getElementById('brain-stream');
     var wasAtTop = !streamEl || streamEl.scrollTop < 40;
     renderBrainStream(events);
+    // Issue #1448 surface 3 — OSS / Cloud-Free users are capped to the
+    // last 24h. Render a one-line upgrade CTA above the brain stream so
+    // they know full history exists on Cloud-Pro.
+    _renderBrainHistoryCap(!!(data && data.capped_at_24h));
     // Issue #1195 — One-time "Live event details restored" toast for users
     // who lived through the 0.12.182 empty-detail regression. Fires only if
     // (a) at least one event has a non-empty detail, (b) the user hasn't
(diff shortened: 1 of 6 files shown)


vivekchand:
## UX review

- 🟢 **Cap rendered as its own row above the brain stream** — keeps the stream a pure event feed. No interleaving CTA with events = no scroll-disruption.
- 🟢 **CTA copy specific to brain** — "Upgrade for full conversation context" names the actual value (conversation history continuity) instead of generic 'more'.
- 🟢 **Slow-path post-filter respects CONTEXT pseudo-events** (they bypass the cap) — meaningful contextualization survives the cap, just not raw events. Subtle but correct.

🤖 Pre-merge review @ d8b41c9d. Pushing new commits will re-trigger review if diff >50 lines or >2h elapsed.

vivekchand:
## Product review

- 🟢 **SQL-layer enforcement on the fast path** = no over-fetch + post-filter waste. Fast path: `query_events(since=cap_since)`. Slow path: post-filter. Both wear the cap.
- 🟢 **Cache key gains the cap flag** — Pro vs Free responses can't cross-pollute. Subtle correctness fix that prevents a cache-leak bug class.
- 🟢 **Empty-rows + cap-active → return empty payload** (instead of falling through to the unbounded JSONL parser). Without this, the gate would silently defeat itself. The agent caught this.
- 🟡 **Pre-existing failing test flagged** (`test_brain_cache_push::test_pushed_blob_decrypts_to_brain_history_shape`) — `time` vs `timestamp` key mismatch in `_rows_to_brain_events`. Not this PR's bug; worth a separate fix. Don't block.

🤖 Pre-merge review @ d8b41c9d. Pushing new commits will re-trigger review if diff >50 lines or >2h elapsed.

PR #1485, opened 2026-05-16
feat(brain): per-LLM-call hallucination risk indicator (closes #567)

vivekchand:
## Summary

- Composes per-call risk score from three signals (temperature, output length, optional logprobs) into a Low / Medium / High label surfaced in the Brain tab and on session rows.
- New ``clawmetry/risk.py`` (pure scoring fn) wired into ``/api/brain-history`` so every assistant chip ships with a ``risk: {risk_level, risk_explanation}`` field.
- Frontend renders a small coloured pill next to each LLM-call chip + a high-risk warning icon on sessions whose call mix tripped the band. Tooltip explains which signals contributed so a developer can act without cracking open raw event JSON.

Closes #567.

## Implementation notes

- **Three signals** scored additively (each +0 / +1 / +2):
  - Temperature: ``T < 0.3`` deterministic, ``< 0.7`` balanced, ``>= 0.7`` creative.
  - Token entropy via mean negative logprob (gracefully degrades to "not available" since Anthropic does not expose logprobs today).
  - Output length: ``< 500`` tokens short, ``< 2000`` medium, ``>= 2000`` very long.
- **Band thresholds** chosen so a single moderate signal alone never trips "high" — that requires compound risk (high T + long output).
- **Non-LLM events** (USER / EXEC / READ / ...) are filtered upstream and never carry a risk field, so the dashboard can branch on its presence without an extra type check.
- Local-store fast path computes risk from the raw DuckDB row (which still carries ``data.params`` / ``data.usage``) before the row is trimmed for transport. JSONL fallback also annotates so the contract is stable.
- Sessions list reads ``/api/brain-history?limit=200`` alongside the existing parallel fetch batch (no extra waterfall) and stamps a warning glyph on any session with a recent high-risk call.

## Test plan

- [x] ``python3 -m pytest tests/test_hallucination_risk.py -q`` — 15 cases pass (10 parametrized scoring cases + integration test that seeds three rows in a real DuckDB store).
- [x] ``python3 -m pytest tests/test_moat_send_message_e2e.py tests/test_local_query_api.py -q`` — 19 passed.
- [x] ``python3 -c "import dashboard"`` — clean.
- [x] ``pytest tests/test_brain_history_v3_event_detail.py tests/test_brain_local_fastpath.py tests/test_brain_local_store.py -q`` — 10 passed, no regressions.

## Memory respects

- ``feedback_no_em_dashes_in_user_facing_copy.md`` — risk labels and tooltips use plain copy.
- ``feedback_simple_ui_for_nontechnical.md`` — Low / Medium / High readable without ML jargon.
- ``feedback_duckdb_first_rule.md`` — risk scoring rides on the existing DuckDB fast path; no new JSONL or stat reads.

## Coordination

- Sibling agent (#568) owns timeline-related DOM. Verified ``git diff`` does not touch any timeline helper or DOM id.

### DIFF
--- a/clawmetry/risk.py
+++ b/clawmetry/risk.py
@@ -1,0 +1,368 @@
+"""
+clawmetry/risk.py — Hallucination Risk Indicator (issue #567).
+
+Per-call risk score for LLM completions, surfaced as Low / Medium / High in
+the Brain tab and as a session-level warning badge when any call in a
+session scores High.
+
+Why this exists
+---------------
+Chapter 6 of the agent observability plan: creativity and hallucination are
+the same mechanism. Higher sampling temperature, low-confidence token
+selection, and unusually long outputs all correlate with the model
+extrapolating beyond its training distribution. ClawMetry already captures
+the per-call sampling config (temperature, max_tokens) and per-call output
+token counts, so we can compose a cheap, additive risk score with ZERO new
+data collection and ZERO new dependencies.
+
+Design notes
+------------
+* **Pure function.** ``compute_hallucination_risk`` takes a single event
+  dict (the same shape ``routes/brain.py`` passes around) and returns a
+  ``{"risk_level": ..., "risk_explanation": ...}`` dict. No I/O, no global
+  state, fully unit-testable in isolation.
+
+* **Three additive signals**, weighted equally and summed:
+
+    1. *Temperature* (always available when the adapter recorded it):
+         T < 0.3   → +0  (deterministic, low risk)
+         T < 0.7   → +1  (balanced, medium contribution)
+         T >= 0.7  → +2  (creative, high contribution)
+
+    2. *Token entropy / logprobs* (when present): the mean negative
+       logprob across sampled tokens — proxy for the model's confidence
+       in its own output. Anthropic does NOT expose logprobs today, so we
+       degrade GRACEFULLY when the field is absent: the signal simply
+       contributes 0 and the explanation says so.
+
+    3. *Response length* (always available):
+         <  500 tokens   → +0
+         500-2000        → +1
+         > 2000          → +2
+
+* **Score → label.** total in [0..6]. Threshold band:
+         0-1 → low      (T<0.3 short responses or no signals)
+         2-3 → medium   (one strong signal, e.g. T=0.7 + short response)
+         >=4 → high     (compound risk — high T + long response, etc.)
+
+* **Graceful degradation.** When no signals are available (e.g. a tool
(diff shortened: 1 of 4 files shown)


vivekchand:
## UX review

- 🟢 **Three-tier risk badge** (🟢 Low / 🟡 Medium / 🔴 High) is the right abstraction. Three states is the max a glance can resolve; five would muddy the signal. Color-coded so a busy Brain stream is still scannable at a glance.
- 🟢 **Hover tooltip carries the explanation** ("T=0.9, response 2400 tokens") — surfaces the WHY for users who want to dig, without cluttering the default view. Steve-grade progressive disclosure.
- 🟢 **Session-row \⚠ badge when any call in the session is high-risk** — turns a per-event signal into a session-level trigger. The single click from "this session smells" → "which call inside it" is the right funnel.
- 🟢 **Graceful degradation** when no signals present (`risk_level="low"` with "insufficient signal" copy). Never crash, never spam high-risk on uninformative events.
- 🟢 No em-dashes in the explanation strings (verified via grep on the new constants). Copy reads natural.

🤖 Pre-merge review @ 35e98bdd. Pushing new commits will re-trigger review if diff >50 lines or >2h elapsed.

vivekchand:
## Product review

- 🟢 **Highest-intent moment captured.** When a user spots a 🔴 High-risk completion in the Brain stream, it's the single best moment to surface the value of better tooling. The badge is the conversion hook.
- 🟢 **Zero new data collection, zero new deps.** The 3 signals (temperature, output length, optional logprobs) are all already captured in the existing event shape. Cheap, additive — exactly the kind of thing that compounds.
- 🟢 **Forward-compatible** — tunable thresholds at module-level + the logprob signal wired to degrade gracefully today means when Anthropic exposes logprobs the score gets sharper with one constant tweak, not a rewrite.
- 🟢 **15 parametrized unit tests** for `compute_hallucination_risk` (single-signal + compound + degradation + garbage-input) — the right rigour for a load-bearing scoring function the UI shows on every row.
- 🟢 **`is_llm_event` filter** prevents the badge from appearing on tool calls / plumbing events — `risk` only applies to model output. Right scoping.
- 🟡 **Nit (not blocking):** the score → label thresholds (0-1 / 2-3 / >=4) are intuitive but ungrounded. Worth a follow-up to compare against a small labeled set of known-hallucinating completions once we have any. For now the additive design is honest about being a heuristic.

🤖 Pre-merge review @ 35e98bdd. Pushing new commits will re-trigger review if diff >50 lines or >2h elapsed.

PR #1490, opened 2026-05-16
feat(budget): Phase 1 hard budget cap config + pause flag + banner

vivekchand:
## Summary

Closes #555 (Phase 1). Adds the minimum scaffolding for the hard budget cap feature while reusing the existing budget machinery in `dashboard.py` + `routes/alerts.py`. Diff is 141 LOC of production code + 195 LOC of tests; no new modules.

- **Three new config keys** on the existing `budget_config` table: `daily_cap_usd`, `monthly_cap_usd`, `session_cap_usd`. Default `0` (no cap), opt-in by user.
- **`_is_over_cap(scope)` helper** in `dashboard.py` that compares the configured cap against the matching spend bucket from `_get_budget_status()`. Returns `(tripped, info)` where `info` always carries `cap`/`spent` so callers can render dollar figures. Safe defaults — `cap=0` never trips.
- **Two new endpoints** in `routes/alerts.py`:
  - `POST /api/budget/pause-gateway` — flips the daemon-level `_budget_paused` flag and persists a reason. Phase 1 deliberately does NOT call the actual gateway-RPC pause (Phase 2 work).
  - `POST /api/budget/resume-gateway` — clears the flag and runs the existing `_resume_gateway()` fallback.
- **Reflection endpoint** `GET /api/budget/is-over-cap?scope=daily` so the dashboard banner can decide rendering without recomputing in JS.
- **Banner copy wired** to the cap-reached state: when `/api/budget/status` returns `paused: true`, the existing `alert-banner` element shows `"Budget cap reached, gateway paused. Tap to resume."` (no em-dash, no double-dash, plain commas and periods per the no-em-dash memory) and always reveals the Resume button. Previously an empty `alert_history` would hide the banner even though the gateway was paused.
- **`resumeGateway()` JS** now prefers the new `/api/budget/resume-gateway` alias and falls back to the legacy `/api/budget/resume` on error so older daemons still work.

## Deferred to Phase 2 (separate issue)

- Actually calling OpenClaw gateway pause via gateway RPC
- Auto-resume at midnight via cron
- Per-session cap accounting (`session_cap_usd` is accepted now but `_is_over_cap("session")` returns `(False, ...)` until the sessions-table integration lands)

## Test plan

- [x] `python3 -m pytest tests/test_budget_cap.py -q` — 7 new tests pass
- [x] `python3 -m pytest tests/test_budget_cap.py tests/test_alert_evaluator_unit.py tests/test_alert_rules_local_store.py tests/test_alerts_evaluator_duckdb.py tests/test_per_agent_budget.py -q` — 68 passed, 0 regressed
- [ ] Manual browser smoke: set `daily_cap_usd=0.01`, inject a cost row, observe banner appears with the exact issue copy, click Resume, banner clears

### DIFF
--- a/dashboard.py
+++ b/dashboard.py
@@ -531,6 +531,15 @@ def _get_budget_config():
         "warning_threshold_pct": 80,
         "telegram_bot_token": "",
         "telegram_chat_id": "",
+        # Issue #555 Phase 1 — hard budget cap. Distinct from
+        # ``*_limit`` (soft, warning-thresholded). When ``*_cap_usd`` is
+        # > 0 and current spend meets/exceeds it, ``_is_over_cap()``
+        # returns True and the dashboard banner shows the resume CTA.
+        # ``session_cap_usd`` is accepted now and consumed by per-session
+        # accounting in Phase 2.
+        "daily_cap_usd": 0.0,
+        "monthly_cap_usd": 0.0,
+        "session_cap_usd": 0.0,
     }
     try:
         with _fleet_db_lock:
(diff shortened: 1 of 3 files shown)


vivekchand:
## UX review

- 🟡 **Banner copy "Tap to resume" reads mobile-only.** Desktop users click, not tap. Suggest `Budget cap reached. Gateway paused.` with a separate `Resume` button (the verb belongs on the button, not the copy). Costs 0 LOC, fixes a "wait, do I tap this?" desktop friction.
- 🟡 **Auto-resume schedule is invisible.** "Manual resume or auto-resume at midnight" is in the issue body but nowhere in the banner. A user who hits the cap at 11pm doesn't know if they need to act now or wait an hour. Add a second line in the banner: `Resumes automatically at midnight.` Or make the button text dual-state: `Resume now` (implies otherwise it'll happen later).
- 🟢 **Three caps (daily/monthly/session) is the right granularity.** Annual would be too coarse, hourly would be neurotic. April 4 billing change made daily/monthly the obvious axes.
- 🟢 **One-click resume button** matches the simple-UI mandate. No confirm modal, no "are you sure" — user explicitly hit the cap, they know.

🤖 Pre-merge review @ fcba2463. Pushing new commits will re-trigger review if diff >50 lines or >2h elapsed.

vivekchand:
## Product review

- 🟢 **#1 safety feature post April 4 billing change.** A user with a runaway agent burning \$3/min wakes up to a \$2000 bill without this. With this, they wake up to a paused banner. Conversion impact: every horror story we lose to "cap saved me" is a permanent retention win.
- 🔴 **P0: `_is_over_cap("session")` returns `(False, …)` because per-session running totals aren't wired.** A user who configures a \$5 session cap will see UI accept the value and… nothing. Silent no-op on a critical safety feature is worse than no feature. Either:
  - (a) Hide the session-cap field from the UI until Phase 2 lands, OR
  - (b) Add an inline disclaimer next to the field: `Per-session caps require Phase 2 (coming). Daily and monthly caps enforce now.`
  Pick (a). One less knob is one less footgun.
- 🟡 **Phase 1 ships config + flag + banner, defers actual gateway pause.** This is the value of the feature. Until Phase 2 lands, the banner shows but the agent keeps burning. A reasonable reviewer will assume "Hard budget cap shipped" = caps actually pause; the PR body should hammer "PHASE 1 ONLY: persistence + UI. Gateway pause lands in #<follow-up>" so this doesn't get released to users as a complete feature.
- 🟢 **Bolted onto existing `_budget_paused` machinery** rather than rebuilt from scratch. 141 LOC production diff (under the 300 budget) means future-Vivek can read this and not flinch. Good restraint.
- 🟢 **7 tests + 0 regressions across 68 tests** is the bar I want every cap-related PR held to forever.

🤖 Pre-merge review @ fcba2463. Pushing new commits will re-trigger review if diff >50 lines or >2h elapsed.

PR #1539, opened 2026-05-17
test(local-query): isolate fixture from any running daemon (closes #1538)

### LINKED ISSUE #1538: [test-infra] test_local_query_api spuriously fails when local daemon is running (proxy hits wrong DuckDB)
## Symptom
\`tests/test_local_query_api.py\` fails 11/16 on local machines where the clawmetry daemon is running. CI passes because no daemon.

## Root cause
\`routes/local_query._dispatch\` tries \`_proxy_dispatch\` first, which reads \`~/.clawmetry/local_query.json\` (the daemon discovery file) and POSTs the query to the running daemon. The daemon queries ITS OWN production DuckDB (\`~/.clawmetry/clawmetry.duckdb\`), not the test's temp fixture file at \`tmp_path / 'events.duckdb'\`. Tests insert 3 events into the fixture store, daemon's DuckDB has 0 matching events, test asserts \`count == 3\` and fails with \`assert 0 == 3\`.

\`\`\`bash
# proves it
mv ~/.clawmetry/local_query.json /tmp/  # disable daemon discovery
pytest tests/test_local_query_api.py -q   # 16 passed
mv /tmp/local_query.json ~/.clawmetry/    # restore
pytest tests/test_local_query_api.py -q   # 11 failed
\`\`\`

## Fixes (pick one)

**A. Fixture-level (quickest)**: In \`tests/test_local_query_api.py\` \`client\` fixture, monkeypatch \`routes.local_query._read_discovery\` to return \`None\`:
\`\`\`python
import routes.local_query as lq
monkeypatch.setattr(lq, "_read_discovery", lambda: None)
\`\`\`
Forces direct-mode every test. ~2 LOC change. Same effect on CI (no behavior change because CI has no daemon either).

**B. Env-var override (more general)**: Make \`_read_discovery\` honor \`CLAWMETRY_LOCAL_QUERY_DISCOVERY=disabled\` → return \`None\`. Same fixture change but the env var is also a useful debug knob for prod.

**C. Daemon liveness probe**: Before proxying, check that \`disc['pid']\` is the same PID the test process owns. Tests never share PID with the daemon. Most conservative.

Recommend **A** — smallest blast radius.

## Why this matters
MOAT verifier crons (15m local + CI) get noisy false-positives when laptops have daemons running. Saw a 18-fail / 40-pass result this morning and almost rolled back a real (clean) PR before tracing the issue.

vivekchand:
## Summary
- Test fixture in `tests/test_local_query_api.py` calls `_dispatch`, which tries `_proxy_dispatch` first.
- On any contributor laptop that has the clawmetry daemon running, the proxy succeeds and the test ends up querying the daemon's production DuckDB (`~/.clawmetry/clawmetry.duckdb`) instead of the test's `tmp_path` fixture.
- CI never had a daemon, so CI was green; laptops with a running daemon silently failed 11 of 16 cases.
- Fix: monkeypatch `lq._read_discovery` to return `None` in the `client` fixture so `_dispatch` always takes the direct path.

## Test plan
- [ ] `make test-api` clean on a laptop with the sync daemon running (`launchctl list | grep clawmetry`)
- [ ] `make test-api` clean on a laptop with the daemon stopped
- [ ] Verify CI run for this PR shows the test file passing (no behaviour change in CI)

### DIFF
--- a/tests/test_local_query_api.py
+++ b/tests/test_local_query_api.py
@@ -24,6 +24,15 @@ def client(tmp_path, monkeypatch):
     import routes.local_query as lq
     importlib.reload(lq)
 
+    # Issue #1538: isolate fixture from any running clawmetry daemon
+    # on the contributor's machine. Without this, _dispatch tries
+    # _proxy_dispatch first; that reads ~/.clawmetry/local_query.json
+    # and POSTs to the daemon, which then queries ITS production DuckDB
+    # (~/.clawmetry/clawmetry.duckdb) instead of this test's tmp_path
+    # fixture. CI never had a daemon so CI passed; laptops with the
+    # daemon running silently failed 11/16. Force direct-mode always.
+    monkeypatch.setattr(lq, "_read_discovery", lambda: None)
+
     # Pre-#1448 events tests seed historical timestamps that fall outside
     # the OSS 24h retention cap. Default the fixture to Pro so those
     # assertions still pass; the cap tests below monkeypatch


vivekchand:
## UX review

Test-fixture polish, not user-facing — but the comment block itself IS user-facing for the next contributor who hits 11 silent failures and wonders why CI is green.

- **Comment lands the punchline first.** ✅ "Issue #1538: isolate fixture from any running clawmetry daemon" — a future contributor reading this fixture five minutes after pulling main immediately knows WHY the monkeypatch exists. No archaeology required.
- **Cites the exact symptom (11/16 silent fail) and the asymmetry (CI passes, laptop fails).** That's the bit that converts "weird comment" into "oh, I get it" in one read. Keep this pattern in future fixtures that paper over environment leakage.
- **Minor:** the comment runs 7 lines for a 1-line monkeypatch. That's the right ratio HERE because the surprise factor is high, but resist the urge to standardize "long comments on every monkeypatch" — only when the surprise is high.

No blockers.

🤖 Pre-merge review @ 004fc0f0. Pushing new commits will re-trigger review if diff >50 lines or >2h elapsed.

vivekchand:
## Product review

This is a TEST-INFRA fix, not a product feature, so the product lens is unusual. Applying it anyway because the bug it closes (CI green / laptop red) destroys engineering velocity and erodes trust in the MOAT verifier — which IS the product moat.

- **Right call: fix at the fixture layer, not the production code.** A worse fix would have been "add a CLAWMETRY_DISABLE_PROXY_FOR_TESTS env flag" — production carrying test-only conditionals is how codebases rot. Monkeypatching `_read_discovery` to None in the fixture is the surgical move. Approved.
- **What this does NOT close:** the underlying class of bug (test silently queries the wrong DuckDB on a contributor laptop) is still possible anywhere ELSE we call `_dispatch` from tests. Likely 0 other call sites today, but a `tests/test_no_proxy_dispatch_leak.py` lint guard (grep `_dispatch(` in `tests/`, require an explicit `_read_discovery=None` monkeypatch) would prevent regression. Out of scope for this PR — file as follow-up if you agree.
- **CI gate:** the new test runs on every PR via `make test-api`. As long as that's part of the required-status-checks ruleset, future divergence between CI and laptop is impossible by construction. Confirm via `gh api repos/vivekchand/clawmetry/rulesets` if you want belt-and-suspenders.
- **No em-dashes in the comment, no scope creep, +9 lines for an entire class of "tests pass but feature broken" bugs. High leverage, ship.**

🤖 Pre-merge review @ 004fc0f0. Pushing new commits will re-trigger review if diff >50 lines or >2h elapsed.

Issue #53, opened 2026-03-01
feat: Brain tab — neural network visualization (neurons, pulses, connecting thoughts)


vivekchand:
### Current State
The Brain tab currently shows a linear event stream (list of tool calls, text events, thinking blocks) with a density bar chart at the top and filter pills per agent. Functional but hard to read at a glance — it looks like a log, not a mind.

### Vision
Transform the Brain tab into an animated neural network visualization that makes AI thinking feel alive and intuitive.

**Core idea:** Each active thought/event is a neuron. Connections between related events (same session, tool call chains, reasoning → action) are synapses. Activity pulses along the connections in real time.

---

#### Nodes (neurons)
- Each agent/session = a **large central neuron** (glowing orb, color-coded per agent)
- Each event type = a **smaller satellite neuron** orbiting its parent:
  - 🔵 Tool calls — medium size, labeled with tool name
  - 🟣 Thinking/reasoning — smaller, slightly transparent, pulsing
  - 🟡 Text output — warm color, larger when long response
  - 🔴 Errors — red, visually distinct
  - ⚪ Token updates — tiny, fast-moving

#### Connections (synapses)
- Lines between neurons show event flow: reasoning → tool call → result
- Line thickness = token weight / importance
- Animated dots travel along the line (like electrical pulses) to show direction and recency

#### Animation
- New events: neuron **fades in** from center with a ripple effect
- Active thinking: center neuron **breathes** (slow scale pulse)
- Tool in progress: connection line **glows** and pulses faster
- Completed chains: connection fades to 50% opacity (history)
- Idle: slow ambient pulse on all neurons

#### Layout
- **Force-directed graph** — neurons self-organize by connection strength
- Main session in center, sub-agents orbit around it
- Pan + zoom to explore
- Click a neuron to expand detail panel on the right (shows full event content)

---

### UI Structure
```
┌─────────────────────────────────────────────────────┐
│  [Filter pills: main · agent:abc123 · All]    [Live] │
├─────────────────────────────────────────────────────┤
│                                                     │
│         ○ ─── ● ─── ○                              │
│        /    (main)    \                             │
│       ○               ○──○                         │
│      sub-agent     tool chain                       │
│                                                     │
│  [Pulse animation running]                          │
├─────────────────────────────────────────────────────┤
│  Detail panel (on click): full event content        │
└─────────────────────────────────────────────────────┘
```

### Implementation Notes
[...]

PR #1564, opened 2026-05-17
feat(brain): MVP neural-net graph view (refs #53)

vivekchand:
## Summary

Adds a **List / Graph toggle** to the Brain tab. Graph mode renders a force-directed canvas where each agent source is a glowing "neuron" with recent events orbiting as connected satellites and an ambient pulse rippling outward on active agents.

This is the **MVP slice** of #53 (full vision is much bigger — click-to-expand panels, traveling pulses along synapses, type-specific satellite styling, pan/zoom, etc.). Keeps PR small for fast review.

## What enables it

The graph rendering code already existed inside a dead duplicate template in `dashboard.py` (lines 5198 + 7460-7737), but was never reachable because the live `DASHBOARD_HTML` is the **second** assignment, which loads `tabs/brain.html` via Jinja include. The toggle button slot in that template was empty and `setBrainViewMode()` was stubbed.

This PR:
1. Ports the graph JS into `clawmetry/static/js/app.js` (next to the other brain JS).
2. Writes a working `setBrainViewMode()` that swaps containers and starts/stops the animation loop.
3. Adds the List/Graph buttons to `tabs/brain.html` (CSS for these classes already exists in `dashboard.css`).

No new endpoints. No new deps. No new DB reads — graph reads the same `_brainAllEvents` array the list view does, populated by `/api/brain-stream` (DuckDB-backed). Pure Canvas 2D.

## Diff size

`+274 / -2` (well under the 300 LOC review budget).

## Deferred (sub-issues to file once this lands)

- Click-to-expand neuron detail panel.
- Animated pulse traveling along synapses (only ambient ring pulse for now).
- Neuron size mapped to token weight / importance.
- Type-specific satellite styling (tool vs thinking vs error vs text).
- Pan + zoom.

## Test plan

- [x] `node -c clawmetry/static/js/app.js` clean.
- [x] Brain-adjacent tests stay green (`test_brain_history_v3_event_detail`, `test_brain_llm_call_timeline`, `test_brain_no_log_keyword_noise`, `test_brain_cache_push_channel_events`, `test_appjs_units`). Pre-existing failures on main in `test_brain_local_fastpath` etc. are unaffected.
- [ ] Manual: open Brain tab, click Graph button, confirm neurons render and animate.
- [ ] Manual: click List, confirm it swaps back.

Refs #53.

### DIFF
--- a/clawmetry/static/js/app.js
+++ b/clawmetry/static/js/app.js
@@ -3600,6 +3600,277 @@ function renderBrainChart(events) {
   ctx.globalAlpha = 1;
 }
 
+// ── Brain Graph (neural-net visualization, refs #53) ──────────────────────
+// Force-directed canvas: each agent source is a glowing "neuron"; recent
+// events orbit their parent as smaller satellite dots; an ambient pulse
+// ripples outward every ~2s on active sources. Toggle via the Graph/List
+// buttons rendered in tabs/brain.html. Pure Canvas 2D — no deps, no WebGL.
+// Reads the existing `_brainAllEvents` array (populated by SSE), so this
+// is purely a presentational alternative to the list view; no new
+// endpoints, no extra DuckDB reads.
+var _brainViewMode = 'list';
+var _brainGraph = {
+  canvas: null, ctx: null, width: 0, height: 500, dpr: 1,
+  lastTs: 0, rafId: 0, animating: false,
+  agents: {}, agentOrder: [], events: [], lastPulseAt: 0
+};
+
+function _brainGraphHash(str) {
+  var h = 2166136261;
+  str = String(str || '');
+  for (var i = 0; i < str.length; i++) {
+    h ^= str.charCodeAt(i);
+    h += (h << 1) + (h << 4) + (h << 7) + (h << 8) + (h << 24);
+  }
+  return h >>> 0;
+}
+
+function _brainGraphEnsureCanvas() {
+  var canvas = document.getElementById('brain-graph-canvas');
+  if (!canvas) return false;
+  if (!_brainGraph.canvas) {
+    _brainGraph.canvas = canvas;
+    _brainGraph.ctx = canvas.getContext('2d');
+  }
+  var rect = canvas.getBoundingClientRect();
+  var dpr = Math.max(1, window.devicePixelRatio || 1);
+  var w = Math.max(320, Math.floor(rect.width || canvas.clientWidth || 800));
+  var h = 500;
+  if (canvas.width !== Math.floor(w * dpr) || canvas.height !== Math.floor(h * dpr)) {
+    canvas.width = Math.floor(w * dpr);
+    canvas.height = Math.floor(h * dpr);
+  }
+  _brainGraph.dpr = dpr;
+  _brainGraph.width = w;
+  _brainGraph.height = h;
+  return true;
+}
+
+function setBrainViewMode(mode, btn) {
+  if (mode !== 'graph' && mode !== 'list') mode = 'list';
+  _brainViewMode = mode;
+  var feed = document.getElementById('brain-feed');
+  var wrap = document.getElementById('brain-graph-wrap');
+  if (feed) feed.style.display = mode === 'list' ? '' : 'none';
+  if (wrap) wrap.style.display = mode === 'graph' ? '' : 'none';
+  document.querySelectorAll('.brain-view-btn').forEach(function(b) {
+    b.classList.toggle('active', b.dataset.view === mode);
+  });
+  if (mode === 'graph') {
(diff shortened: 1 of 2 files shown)


vivekchand:
## UX review

First review @ `01f85d9d` (refs #53).

This is exactly the right MVP slice. Glowing-neuron + orbiting-satellite + ambient pulse is the visceral metaphor for "AI thinking happening right now" — it's the wow moment a Brain tab was always supposed to have.

What works:
- **Toggle pattern (List / Graph)** is the right scaffolding. Lets the user discover the new view without losing the familiar one. Default = List preserves muscle memory.
- **No-new-deps Canvas 2D** keeps the page snappy. WebGL was the easy-wrong choice; this gives 60fps with O(N²) at N=20 with room to grow.
- **Reuses `_brainAllEvents`** so the graph is a presentation alternative, not a parallel data path. Architecturally clean.
- **Dead-code resurrection** is the kind of find that pays for a whole sprint. The PR's tightness (+274 / -2) is *because* the hard work was already done — just orphaned by a duplicate-template override.

What I'd push on:

1. 🟡 **First-time orientation copy is missing.** A new user clicks Graph and sees pulsing dots — they don't know what they're looking at. One thin line of text above the canvas closes this: *"Each neuron is an active agent. Dots are recent events. Pulses show activity in the last 90s."* Three sentences, ~140 chars, removes the "what am I seeing?" moment. Doesn't need to be design-perfect — just present on first render.
2. 🟡 **Monospace agent labels** (`11px monospace`) read as terminal output, not visualization. The rest of the dashboard uses sans-serif. Inheriting the page font (`font: inherit` or the page's `--font-ui` if defined) makes the labels feel like the same product.
3. 🟡 **Color cohabitation.** Active neurons (`#a855f7` purple) + idle (`#f59e0b` amber) + per-event colors from `brainSourceColor()` could all collide on the dark bg. Worth a quick eyeball at 10+ agents with mixed source types to confirm hierarchy still reads (the neurons should pop more than the satellites). If satellites compete with neurons visually, drop satellite shadowBlur from 8 → 4.
4. 🟢 **Truncation hint.** Capping at 20 agents / 200 events is sensible but silent. A small "showing 20 of N agents" overlay (bottom-right, dim) prevents the "wait, where's my third agent?" confusion at scale.

🤖 Pre-merge review @ `01f85d9d`. Pushing new commits will re-trigger review if diff >50 lines or >2h elapsed.

vivekchand:
## Product review

First review @ `01f85d9d`.

- **MVP-slice discipline** is the headline. The full vision is "click-to-expand neurons + traveling pulses + token-weighted sizes + pan/zoom" — easily a 1500 LOC PR. This ships the *visceral wow* in 274 lines and explicitly defers the rest as follow-ups. Right tradeoff every time.
- **Pure presentation, no new data plumbing.** Reuses `_brainAllEvents` (DuckDB-backed via SSE), no new endpoint, no DuckDB bypass, no new deps. Zero cost to maintain.
- **Floor + extras pattern works for the conversion funnel.** Users who don't care about pretty visualization stay on List. Users who do get the brand differentiator on first impression. Clean opt-in.

What I'd push on:

1. 🟡 **The systemic finding is bigger than this PR.** The agent discovered that ~260 LOC of working graph code was sitting dead inside `dashboard.py` because of the double-`DASHBOARD_HTML` override pattern. Per memory `feedback_audit_class_of_bug_not_just_instance.md` — that's a CLASS of bug, not an instance. Worth filing a follow-up: `[audit] grep all dashboard.py double-assignments — what other features are silently orphaned?` Could surface multiple shippable wins.
2. 🟡 **No automated visual regression for the new view.** The two manual test-plan checkboxes ("open Brain tab, click Graph button") are the only verification that the canvas actually paints. One thin Playwright assertion — `canvas.toDataURL().length > 5000` after clicking Graph — would catch a regression where the toggle works but the canvas stays blank. The visual-diff bot on the landing repo could be the template.
3. 🟢 **`data-feature="brain-graph"` attribute on the Graph button.** Looks like a future hook for tier-gating ("Graph view is Pro-only"). If that's the intent, worth a code comment so it doesn't get refactored away. If not the intent, just remove the attribute.
4. 🟢 **Performance ceiling worth noting in the follow-up.** O(N²) agent-agent forces at N=20 is 400 calcs/frame — fine. At N=100 (large multi-agent fleet) it's 10k — also fine for 60fps. Past that, switch to a spatial-hash or cap harder. Not a today-problem; document the ceiling so the next maintainer knows.

Acceptance vs the MVP-slice ask: ✅ ≤300 LOC, ✅ reuses existing SSE feed, ✅ no DuckDB bypass, ✅ ships real value (not just an approach comment).

🤖 Pre-merge review @ `01f85d9d`. Pushing new commits will re-trigger review if diff >50 lines or >2h elapsed.

PR #1610, opened 2026-05-17
fix(brain): SSE reconnect with exponential backoff + connection-lost banner (closes #1596)

### LINKED ISSUE #1596: [silent-failure/ui-drift] Brain SSE reconnect silently dies after one failed poll fallback

## Symptom
When the brain SSE stream (`/api/brain-stream`) disconnects (server restart, proxy timeout, network blip), the dashboard:
1. Sets live-indicator to false (good).
2. Closes the EventSource.
3. Schedules ONE fallback `loadBrainPage(true)` after 5s.
4. **Never reconnects to SSE.** Never reschedules the poll.

If that single poll also fails (still partitioned), the user sees an indefinite stale feed with no \"connection lost\" UI. The live-indicator turns off but most users miss that subtle change.

## Root cause
`clawmetry/static/js/app.js:3665-3673`:
```javascript
es.onerror = function() {
  _updateBrainLiveIndicator(false);
  try { es.close(); } catch(e){}
  _brainSSE = null;
  // Fall back to polling
  if (document.getElementById('page-brain') && document.getElementById('page-brain').classList.contains('active')) {
    _brainRefreshTimer = setTimeout(function() { loadBrainPage(true); }, 5000);
  }
};
```

The `done` event reconnects via `setTimeout(_startBrainSSE, 2000)` (line 3681), but `onerror` does NOT — it only schedules one poll. Discrepancy is the bug.

## Reproduction
1. Open dashboard, watch brain feed go live.
2. Kill backend (`kill -9 <pid>`).
3. Observe live-indicator turns off, one poll fires, then silence.
4. Restart backend — feed stays dead until user manually reloads.

## Detection strategy
- Playwright E2E: kill backend mid-test, restart after 10s, assert SSE re-establishes within 30s.
- Frontend: emit `connection_lost` console.warn every N seconds of dead-stream so monitoring catches it.

## Fix shape
- Mirror the `done` handler: schedule `_startBrainSSE` reconnect with exponential backoff (2s, 4s, 8s, ... capped at 60s).
- Show a banner \"connection lost — retrying...\" while disconnected.
- Apply same fix to logs-stream and health-stream SSE consumers (audit needed).

## Severity
**P1** — UX-affecting (user thinks they're live; data is frozen). Same family as the C2/C3 stale-cache issue but client-side.

## Related
- #1565 MOAT tracking
- Sibling: logs-stream, health-stream may have the same bug (E2 family)

vivekchand:
## Summary
- Replace the one-shot poll fallback in the Brain SSE `onerror` handler with an exponential-backoff retry chain (1s, 2s, 4s, 8s, 16s, then 30s cap).
- Surface an explicit "Connection lost. Reconnecting..." banner above the brain stream after 30s of failed retries so users on flaky networks know the feed is frozen, not just quiet.
- Reset banner + retry state on the next successful `connected` event; tear them down in `_stopBrainSSE` so navigating away never leaves stale timers.

## Why
Before this fix the EventSource `onerror` handler scheduled a single `loadBrainPage(true)` poll 5s later and then went silent forever. If that poll also failed (still partitioned), the user saw an indefinitely stale feed with no UI signal beyond the small `POLLING` pill. That's the exact failure mode #1596 reports.

## Defences
- **Single retry chain**: `_scheduleBrainSSEReconnect` clears any pending timer before scheduling the next attempt — 10 back-to-back `onerror` bursts collapse into 1 pending timer, not 10 parallel reconnects.
- **Visibility-aware**: retries pause when the tab is hidden (works alongside the existing `_installSSEVisibilityGuard`).
- **No `location.reload()`**: per `feedback_no_reload_in_bootstrap_e2e.md`.
- **Banner copy**: plain-English ("Connection lost. Reconnecting..."); no em-dash / double-dash / technical jargon (`readyState` / `EventSource` / `WebSocket`) per the user-facing-copy memories.

## Test plan
- [x] `node tests/test_brain_sse_reconnect.js` — 38 assertions, all pass (backoff ladder, chain survival after failed polls, banner threshold + copy, single-chain storm prevention, visibility guard, `_stopBrainSSE` teardown).
- [x] `python3 -m pytest tests/test_brain_sse_reconnect.py tests/test_appjs_units.py -v` — 2 pass.
- [x] `python3 -c "import dashboard"` — clean.
- [ ] Manual: open dashboard, kill backend (`kill -9 <pid>`), wait 35s, observe banner appears; restart backend, observe banner clears + feed resumes.

closes #1596

### DIFF
--- a/clawmetry/static/js/app.js
+++ b/clawmetry/static/js/app.js
@@ -4009,6 +4009,21 @@ function _drawBrainGraph(ts, now) {
 
 var _brainSSE = null;
 var _brainSSEConnected = false;
+// Issue #1596 — exponential backoff state for SSE reconnect. A single
+// retry chain (no parallel storms): `_brainSSERetryTimer` holds the
+// pending setTimeout id, `_brainSSERetryAttempt` is the current attempt
+// index (0-based; 0 = fresh failure, 5+ = banner). `_brainSSEFirstFailMs`
+// is the wall-clock of the first failure in the current outage so we can
+// surface the "Connection lost" banner after 30s of failed retries.
+var _brainSSERetryTimer = null;
+var _brainSSERetryAttempt = 0;
+var _brainSSEFirstFailMs = 0;
+// Caps: 1s, 2s, 4s, 8s, 16s, then 30s forever. Matches the issue spec.
+var _BRAIN_SSE_BACKOFF_MS = [1000, 2000, 4000, 8000, 16000];
+var _BRAIN_SSE_BACKOFF_MAX_MS = 30000;
+// Banner threshold — surface explicit UI after 30s of failed retries so
+// users on flaky networks know the feed is frozen, not just quiet.
+var _BRAIN_SSE_BANNER_THRESHOLD_MS = 30000;
 
 function _updateBrainLiveIndicator(connected) {
   _brainSSEConnected = connected;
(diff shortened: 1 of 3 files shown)


vivekchand:
## UX review @ 3a8cd88c

**Good signal-loss design**
- Exponential backoff 1s→30s cap is the right ladder: aggressive enough to recover from blips, gentle enough not to thrash on persistent failure.
- Single retry chain (no parallel storms) — won't accidentally DoS the server during a multi-tab disconnect.
- Visibility-aware (pauses on tab hidden) — respects user's actual attention.
- Banner copy passes the no-em-dash + no-jargon assertions (no `readyState`/`WebSocket` leaking to users).

🟢 No yellow flags. This is the kind of fix where the test suite (38 assertions including banner-copy checks) is more useful than persona second-guessing.

🤖 Pre-merge review @ 3a8cd88c.

vivekchand:
## Product review @ 3a8cd88c

**Trust win**
- Silent stale feed was THE silent-failure that erodes 'is this live?' user trust. Visible reconnect banner closes the loop.
- 30s threshold before banner is right — short enough to catch real outages, long enough to not flash during normal proxy hiccups.

🟡 **Follow-up flag** (Eng noted in their report — worth filing as sibling P1): same one-shot reconnect pattern exists in `logStream.onerror` (~line 9530) + `_flowSse.onerror` (~9750). Brain is the most-watched stream so this PR is the highest leverage, but the family bug remains in 2 other surfaces.

🤖 Pre-merge review @ 3a8cd88c.

PR #1633, opened 2026-05-17
fix(tier): server-side paywall gate eliminates Pro feature flash (closes #1603)

### LINKED ISSUE #1603: [silent-failure/ui-drift] Tier-state Pro feature flash — paywall renders AFTER content for ~1 frame on first paint

## Symptom
On first page load, Pro-gated features (Alerts Center, NemoClaw governance, etc.) render the FEATURE'S DOM tree first, then the paywall overlay on a subsequent tick when `tier` resolves via async fetch. Users on the Free tier can see (and potentially screenshot/inspect) the Pro UI for a brief flash before the paywall obscures it.

Worse: in some panels the click handlers on the Pro feature wire up BEFORE the paywall overlay mounts, so a frame-perfect click bypasses the gate entirely (security/conversion issue).

## Root cause
- `dashboard.py:7493` — `shouldShow = !!(snapshot && snapshot.cap_hit && !snapshot.is_pro)` — `snapshot` is fetched async; until it resolves, `is_pro` is undefined, treated as falsy, so the paywall shows. BUT — for the inverse case (Pro features hidden from Free), the gate often defaults to \"show\" so Pro users don't see a flash, which means Free users ALSO see it briefly.
- `dashboard.py:4212` — `// TODO #1377: Cloud-Free upsell modal here when CLOUD_MODE && tier === 'free'.` — the TODO itself is the smoking gun: gate not yet wired.

## Reproduction
1. Free-tier user logs in.
2. Hard reload dashboard.
3. Open DevTools, throttle network to Slow 3G.
4. Observe Pro features briefly visible (1-3 frames) before paywall paints.

## Detection strategy
- Playwright visual diff: capture first-paint screenshot for free-tier user, assert no Pro-feature DOM nodes visible.
- Frontend: emit `pro_feature_flash` analytics event when a Pro DOM node is mounted while `tier === 'free'`. Cloud analytics counts incidents.

## Fix shape
- Render the paywall in the SAME render pass as the feature (not after). Use a `<TierGate tier=\"pro\">` wrapper that returns either children OR paywall, never both.
- Default to \"hidden\" before tier resolves (loading skeleton) instead of \"shown\".
- Block the feature's click handlers behind a JS-side `await tierResolved` guard.

## Severity
**P1** — security-adjacent (Free users can inspect Pro UI source) + conversion-eroding (\"I already saw it, why pay?\").

## Related
- #1377 (referenced in TODO)
- `project_alerts_pro_feature` memory (Alerts Center is Cloud-Pro only)
- `project_free_plan_upsell` memory (every gated click is a conversion opp)

vivekchand:
## Summary

- Pre-fix, NemoClaw governance shell + Alerts rule-editor modal shipped in dashboard HTML for every user, then a JS overlay mounted on a later tick. Free users saw a 1-3 frame flash of Pro UI; on a frame-perfect click they could bypass the gate entirely.
- Fix: branch on ``dashboard._is_pro_user()`` server-side in ``routes/meta.py:index()`` and pass ``is_pro`` to Jinja. Pro-feature templates now render either the full shell OR ``partials/paywall_modal.html``, never both. Same render pass = no flash.
- New paywall partial reuses the page id (``page-nemoclaw``) so ``switchTab('nemoclaw')`` keeps working with no JS changes.

## Files

- ``routes/meta.py``: pass ``is_pro`` template var, fail-closed on exception.
- ``clawmetry/templates/partials/paywall_modal.html``: new reusable upsell partial. Copy is upsell-first per ``project_free_plan_upsell`` memory, no em-dashes per ``feedback_no_em_dashes_in_user_facing_copy``.
- ``clawmetry/templates/tabs/nemoclaw.html``: full shell behind ``{% if is_pro %}``, Free users get the paywall partial.
- ``clawmetry/templates/tabs/alerts.html``: rule-editor modal behind ``{% if is_pro %}``. Canned-example teaser shell stays for all tiers (intentional per ``project_alerts_pro_feature`` memory).
- ``clawmetry/static/js/alerts.js``: ``alertsHandleEdit`` tier gate so a Free user clicking a canned example hits ``openPaywall()`` instead of null-derefing the now-absent editor modal.

## Critical safety check

Wrong gate ships paywalls to paying customers (worse than original flash). Tests pin the inverse direction:

- ``test_pro_user_index_includes_nemoclaw_full_shell`` - paying users see the shell.
- ``test_pro_user_index_includes_alerts_editor_modal`` - paying users see the editor.
- ``test_pro_user_index_omits_paywall_partial`` - paying users do not see the upsell.

## Test plan

- [x] ``python3 -c 'import dashboard'`` clean
- [x] ``pytest tests/test_tier_pro_paywall_no_flash.py -v`` (7/7 pass)
- [x] ``pytest tests/test_auth_detected_token.py tests/test_anon_auth_fail_ping.py`` (45/45 pass, no adjacent regression)
- [ ] Manual Playwright visual-diff on first-paint for Free vs Pro tier (CI)
- [ ] Smoke against real OpenClaw v3 cloud-paired node post-merge

Closes #1603.

### DIFF
--- a/clawmetry/static/js/alerts.js
+++ b/clawmetry/static/js/alerts.js
@@ -259,6 +259,12 @@
       rule = EXAMPLE_RULES.find(r => r.id === ruleId);
     }
     if (!rule) return;
+    // Issue #1603: the editor modal DOM is server-side gated to Pro users
+    // so Free users get the upsell here instead of a null-deref on
+    // ``alerts-editor-modal``. Matches the alertsHandleNewRule gate above.
+    if (alertsState.tier !== 'pro' && alertsState.tier !== 'trial') {
+      return openPaywall();
+    }
     alertsState.editorRule = rule;
     alertsState.editorType = rule.alert_type;
     openEditor();
--- a/clawmetry/templates/partials/paywall_modal.html
+++ b/clawmetry/templates/partials/paywall_modal.html
@@ -1,0 +1,40 @@
+<!--
+  Server-side paywall partial — rendered in place of a Pro-feature DOM
+  shell when ``is_pro`` is False. Replaces the pre-#1603 pattern of
+  rendering the full Pro UI plus a client-side overlay that slammed down
+  on a subsequent tick (the "Pro feature flash" bug).
+
+  Callers pass two optional context vars via Jinja ``with`` blocks:
+    paywall_feature  — feature name shown in the headline (e.g. "NemoClaw
+                       governance"). Defaults to "this feature".
+    paywall_pitch    — short reason this feature is Pro. One line, plain
+                       prose, no em-dashes (memory: no_em_dashes).
+
+  Copy is intentionally upsell-first (memory: free_plan_upsell). Every
+  gated click is a conversion opportunity, not a wall. Keep it readable
+  for non-technical users (memory: simple_ui_for_nontechnical).
+-->
+<div class="page" id="page-{{ paywall_page_id|default('paywall') }}">
+  <div style="max-width:560px;margin:48px auto;padding:32px;background:var(--bg-secondary);border:1px solid var(--border-primary);border-radius:12px;text-align:center;">
+    <div style="font-size:36px;margin-bottom:12px;">🔒</div>
+    <h2 style="margin:0 0 8px 0;font-size:18px;font-weight:700;color:var(--text-primary);">
+      {{ paywall_feature|default('This feature') }} is part of ClawMetry Pro
+    </h2>
+    <p style="margin:0 0 20px 0;font-size:13px;line-height:1.55;color:var(--text-secondary);">
+      {{ paywall_pitch|default('Pro unlocks the full workflow with no caps on history, alerts, or governance policies.') }}
+    </p>
+    <div style="display:flex;gap:10px;justify-content:center;flex-wrap:wrap;">
+      <a href="https://app.clawmetry.com/upgrade" target="_blank" rel="noopener"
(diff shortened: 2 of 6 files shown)


vivekchand:
## UX review @ b7afaed8

**The flash is gone — server-side is the right surface**
- Free users now see paywall-first, not Pro-DOM-then-paywall-overlay. No more 'free preview of the locked feature' jarring moment.
- Pro users see the editor unchanged. Critical inverse-direction test (7 tests including Pro-user path) prevents the worst-case bug (paywalls served to paying users).
- Alerts canned-example teaser preserved = right call. Per `project_alerts_pro_feature` and `project_free_plan_upsell`, the canned examples are the conversion lever — they show value before asking for payment.

🟢 No yellow flags. Strong test discipline (fail-closed exception path, both Free and Pro scenarios).

🤖 Pre-merge review @ b7afaed8.

vivekchand:
## Product review @ b7afaed8

**Trust + conversion lever both addressed**
- Trust: paying users get the full Pro UI, no paywall noise. Free users see honest paywall, not a tease that becomes a wall.
- Conversion: canned examples + 'Start 7-day free trial' CTA from the paywall partial = the upsell flow gets a clean surface to live on.
- Fail-closed on exception (Pro DOM never leaks even if tier-cache flakes) = security/UX win.

🟢 **Ship.** Closes a real first-impression conversion leak.

🟡 **Telemetry follow-up**: when paywall renders, we should fire a 'paywall_shown' event. Without it we can't measure conversion rate from paywall-view → upgrade. Tag as follow-up.

🤖 Pre-merge review @ b7afaed8.

PR #1631, opened 2026-05-17
fix(onboarding): first-heartbeat race shows 'Setting up your node' not blank (closes #1604)

vivekchand:
## Summary

Issue #1604 (P1, silent-failure taxonomy / privacy-adjacent): between user signup and the daemon's first heartbeat (~30s window), the dashboard rendered an empty grid that looks broken, and in the worst case could surface stale data from a prior tenant on the same machine_id.

This OSS-side fix adds an explicit **"Setting up your node"** banner:

- Shows the banner whenever `/api/heartbeat-status` reports `status='unknown'` (i.e. `_last_heartbeat_ts == 0`).
- Auto-refreshes every 5s.
- The moment the first heartbeat lands, hides the banner AND calls `loadAll()` so the live cards render instantly instead of waiting for the next 30s refresh tick.
- After 90s with no heartbeat, swaps copy from "Setting up your node..." to actionable error pointing at `clawmetry status` so a stuck daemon does not masquerade as "loading forever".

Memory respects baked in as test contracts:
- No em-dashes in user-facing copy (`feedback_no_em_dashes_in_user_facing_copy.md`)
- "Setting up your node" not "Awaiting initial heartbeat" (`feedback_simple_ui_for_nontechnical.md`)
- No `location.reload()` / no `/login` redirect on transient fetch failure (`feedback_no_reload_in_bootstrap_e2e.md` + `feedback_persistent_sessions.md`)

## Surface decision

The issue noted "cross-repo" but the actual rendering surface a new user lands on after `clawmetry` boots is the **OSS dashboard at `localhost:8900`** — the cloud dashboard is opt-in via "Enable Cloud Sync". Fixing the OSS surface kills the empty-state-looks-broken bug at the root. A follow-up cloud-side ticket can mirror this banner on `app.clawmetry.com/cloud` for users who jump straight there.

## LOC

- 75 LOC `clawmetry/static/js/app.js` (`checkOnboardingStatus()` + polling)
- 34 LOC `clawmetry/templates/partials/banners.html` (banner element + spinner keyframes)
- 190 LOC `tests/test_first_heartbeat_onboarding.py` (7 cases: 3 issue scenarios + 4 memory-respect contracts)

Production: 109 LOC (under 150 budget). Tests: 190 LOC (slightly over 150 budget, justified by the 4 explicit memory-respect contract assertions the prompt requested).

## Test plan

- [x] `python3 -m pytest tests/test_first_heartbeat_onboarding.py -v` — 7/7 pass
[...]

PR #1634, opened 2026-05-17
fix(onboarding): explicit 'No OpenClaw/NemoClaw detected' empty-state when no agent installed

vivekchand:
## Summary

Currently ClawMetry installs and registers as a node in cloud even when the user has not installed any underlying agent. Every tab (Brain, AI Model card, Channels, Sessions) renders empty or shows stale data because there is no agent producing events, so users see a broken-looking dashboard for their entire first 60 seconds.

This adds an explicit page-level banner that detects "no OpenClaw, no NVIDIA NemoClaw, no local events" and tells the user exactly what is missing + offers install CTAs for both agents.

- New ``detect_agent_install()`` helper in ``dashboard.py`` (60s cache) — combines: gateway PID file present, session JSONLs exist, OPENCLAW_HOME dir non-empty, ``shutil.which('nemoclaw')``, ``~/.nemoclaw`` non-empty, and any local-store events
- New ``/api/agent-presence`` endpoint in ``routes/health.py`` (sibling of ``/api/heartbeat-status``)
- New ``#no-agent-banner`` partial in ``banners.html`` with two equal-weight install CTAs (OpenClaw / NVIDIA NemoClaw)
- New ``checkAgentPresence()`` JS in ``app.js`` — polls every 60s, mutually exclusive with #1631's first-heartbeat onboarding banner

### Distinct from #1631

| Scenario | Banner |
|---|---|
| Agent installed, no heartbeat yet (~30s warm-up) | #1631's "Setting up your node" (transient) |
| No agent installed at all | THIS PR's "No agent detected" (persistent) |

The JS enforces mutual exclusion: when ``no_agent=true`` we hide the onboarding-banner element AND set ``_cmOnboardingDismissed=true`` so the 5s onboarding poller stops re-flashing.

## Memory respects

- ``feedback_no_em_dashes_in_user_facing_copy`` — verified by test, no em-dashes in banner copy
- ``feedback_simple_ui_for_nontechnical`` — "Install OpenClaw" / "Install NVIDIA NemoClaw" reads to non-tech users
- ``feedback_em_style_judgment`` — lists both agents equally, no editorialising
- ``feedback_no_reload_in_bootstrap_e2e`` + ``feedback_persistent_sessions`` — no ``location.reload()`` / no ``/login`` redirect on transient fetch failure (verified by test)
- ``project_sync_progress_banner`` — sister of this work, same "explain what is happening" UX philosophy

## Test plan

- [x] ``tests/test_no_agent_detected_state.py`` — 11 cases, all green:
  - 5 detection scenarios (no-agent / openclaw-only-no-heartbeat / openclaw-with-data / nemoclaw-only / both)
  - 60s cache contract (detectors must not re-run within TTL)
  - banner partial carries both install CTAs
  - banner partial has no em-dashes in user copy
  - JS does not reload or sign out on transient failure
  - JS is mutually exclusive with #1631's onboarding banner
  - JS polls at the 60s server-cache cadence
- [x] ``tests/test_first_heartbeat_onboarding.py`` (PR #1631's suite) — still green, 7/7
- [x] Flask test-client smoke: ``GET /api/agent-presence`` returns 200 + valid JSON shape ``{openclaw_detected, nemoclaw_detected, any_data, signals, no_agent}``

DO NOT MERGE — per task instructions.

### DIFF
--- a/clawmetry/static/js/app.js
+++ b/clawmetry/static/js/app.js
@@ -647,6 +647,51 @@ async function checkOnboardingStatus() {
 _cmOnboardingTimer = setInterval(checkOnboardingStatus, 5000);
 setTimeout(checkOnboardingStatus, 300);
 
+// === No-Agent-Detected Empty-State Banner ===================================
+// Distinct from the first-heartbeat onboarding banner above:
+//   * onboarding-banner fires when "agent installed but no heartbeat yet"
+//     (transient race that resolves in ~30s)
+//   * no-agent-banner fires when "no agent installed at all" — persistent
+//     until the user installs OpenClaw or NVIDIA NemoClaw.
+// Mutual exclusion: if openclaw or nemoclaw IS detected (heartbeat just
+// hasn't landed yet), we hide this banner and let onboarding-banner do
+// its thing. Polls every 60s — filesystem state for "did the user pip
+// install an agent" changes on the order of minutes, not seconds.
+async function checkAgentPresence() {
+  var banner = document.getElementById('no-agent-banner');
+  if (!banner) return;
+  try {
+    var data = await fetch('/api/agent-presence').then(function(r){return r.json();});
+    var noAgent = !!(data && data.no_agent === true);
+    if (noAgent) {
+      // No OpenClaw, no NemoClaw, no local data — show the persistent
+      // empty-state banner and hide the first-heartbeat one so we don't
+      // double up. Setting _cmOnboardingDismissed stops the 5s poller
+      // from re-flashing the "Setting up your node" copy.
+      var ob = document.getElementById('onboarding-banner');
+      if (ob) ob.style.display = 'none';
+      _cmOnboardingDismissed = true;
+      banner.style.display = 'flex';
+    } else {
+      // Agent appeared. Hide the no-agent banner and (if the dashboard
+      // was already showing it) reload the data so cards render.
+      var wasShown = banner.style.display !== 'none';
+      banner.style.display = 'none';
+      if (wasShown && typeof loadAll === 'function') {
+        try { loadAll(); } catch(e) {}
+      }
+    }
+  } catch(e) {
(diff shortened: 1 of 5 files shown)


vivekchand:
## UX review @ defb7dda

**The honest first-impression**
- 'No OpenClaw or NVIDIA NemoClaw detected' is exactly the right framing — it tells the user WHAT'S MISSING, not 'sorry, dashboard broken'. Steve test passes.
- Two equal-weight CTAs (Install OpenClaw / Install NVIDIA NemoClaw) respects user choice — doesn't push one over the other (per em-style judgment of not bouncing the architectural question).
- Mutual exclusion with #1631 (sets `_cmOnboardingDismissed=true` when no-agent fires) is the subtle thing that makes both banners feel intentional, not stacked.
- 60s poll = right cadence. Not aggressive enough to be wasteful, fast enough to flip when user installs an agent.

🟢 No yellow flags. Test contract assertions (no em-dashes, no reload, mutual exclusion) prevent regression.

🤖 Pre-merge review @ defb7dda.

vivekchand:
## Product review @ defb7dda

**Closes the user's exact bug**
- User's words: 'we show as a node clawmetry cloud & all of the screens / tabs show up with stale data' → before this PR, that was the actual experience for anyone who installed ClawMetry without an underlying agent. Now they get a clear actionable banner instead.
- Two-CTA layout is the right call given the multi-agent direction (memory `project_multi_agent_adapter` — ClawMetry is evolving to observe many frameworks, OpenClaw/NemoClaw are the first two surfaced).
- Combined with #1631 (first-heartbeat-race) + #1633 (Pro feature flash) = the first-60s install experience is now coherent: 'no agent? install one' → 'agent installed but no heartbeat? we're setting up' → 'all set, here's your data'.

🟢 **Ship.** Closes the user-explicitly-flagged first-impression bug today's MOAT-EOD-then-some.

🤖 Pre-merge review @ defb7dda.

PR #1643, opened 2026-05-18
fix: enable gateway WS tap by default (closes Telegram watch no-op)

### LINKED ISSUE #1196: [Product P0 from PR #1192] Telegram channel watch is no-op until OpenClaw upstream persists transcripts to disk
From the post-merge UX/product review of PR #1192.

## Problem

PR #1192 added a 21-channel directory watcher to `clawmetry/sync.py` (`_CHANNEL_DIRS`) to fix the user-reported bug \"I message Diya on Telegram and ClawMetry shows nothing.\" The watcher is correct and tested.

However, per memory `reference_openclaw_telegram_inmemory.md`, OpenClaw does NOT currently persist Telegram inbound messages to `~/.openclaw/telegram/<chat_id>.jsonl` — the channel adapter holds them in memory only. That means on every real user machine today:

- The watcher runs every cycle
- It finds an empty `~/.openclaw/telegram/` directory (or only `update-offset-default.json`, which we correctly skip)
- Zero rows land in `channel_messages`
- The user's reported bug is NOT actually fixed

We just shipped a defensive infra change that we framed as a user-visible fix.

## Why this is P0 (product)

1. **Trust regression risk.** The user explicitly reported this exact bug. They will re-test, see the same symptom, and conclude we shipped vapor.
2. **Conversion impact.** Multi-channel consolidated view is a Cloud-Pro upsell vector. We can't sell \"see all your agent's chat channels in one place\" if none of them actually appear.
3. **Communication mismatch.** PR title (\"feat(sync): ingest OpenClaw Telegram channel\") implies user-visible feature; reality is dormant infra.

## Required actions (pick one or more)

- [ ] Land an OpenClaw upstream PR that writes Telegram (and the other 20 adapters) to disk in the `~/.openclaw/<channel>/` shape this PR expects. Block the celebratory release notes until that ships.
- [ ] OR add a parallel WS-tap path that subscribes to the gateway's channel-message broadcasts and writes them to DuckDB directly, so we don't depend on adapter persistence.
- [ ] OR rephrase the [RELEASE] PR description and CHANGELOG entry to \"defensive infra: ready to ingest channel transcripts when OpenClaw persists them; requires OpenClaw 0.x.y+\". Set user expectations honestly.

## Acceptance

- A user who installs OSS today, sends a Telegram message to their bot, and opens the Brain tab within 60s sees the message rendered as a `channel.in` event.
- If we cannot guarantee the above today, release notes for 0.12.183+ explicitly state the OpenClaw version requirement.

## Related

- PR #1192
- Memory: `reference_openclaw_telegram_inmemory.md`

vivekchand:
Closes #1196

## What and why

`gateway_tap.py` already contained the correct fix for the Telegram "no-op" bug: a live WebSocket subscriber that calls `ingest_channel_event()` for every gateway frame, bypassing the filesystem entirely. The module was just **opt-IN by default** (`CLAWMETRY_ENABLE_WS_TAP=1` required), so no real user ever got channel messages.

This PR flips the default to **opt-OUT** — the tap starts automatically; set `CLAWMETRY_ENABLE_WS_TAP=0` to disable.

## Changes

- **`clawmetry/gateway_tap.py`** — invert env-var check (default `"1"` → disabled by `"0"/"false"/"no"`); update docstring/inline comments to match new default
- **`clawmetry/sync.py`** — update the one-line comment that described the tap as default-OFF

## Behaviour when `operator.read` scope is missing (today's OpenClaw)

The tap connects to the gateway, the `sessions.messages.subscribe` call is rejected with a scope error, and the tap logs one WARNING:
> `gateway WS tap: sessions.messages.subscribe rejected (missing scope: operator.read) — inbound channel bodies will not arrive until OpenClaw grants operator.read scope`

Then it stays alive in degraded mode — zero channel rows ingested, identical to the prior disabled state. When OpenClaw ships the scope grant, existing installs start seeing channel messages without any user action.

## Test plan

- [ ] `python3 -c "import ast; ast.parse(open('clawmetry/gateway_tap.py').read())"` passes ✅
- [ ] `python3 -c "import ast; ast.parse(open('clawmetry/sync.py').read())"` passes ✅
- [ ] Confirm `CLAWMETRY_ENABLE_WS_TAP=0` skips tap startup (check log for debug message)
- [ ] Confirm default (no env var) attempts gateway connection on daemon start

---
_Generated by [Claude Code](https://claude.ai/code/session_01QFiCxaVhbCKzAvYCMQZkDa)_

### DIFF
--- a/clawmetry/gateway_tap.py
+++ b/clawmetry/gateway_tap.py
@@ -48,9 +48,11 @@ Failure modes
 Disable
 -------
 
-Set ``CLAWMETRY_ENABLE_WS_TAP=1`` in the environment to opt in.
-Default is OFF until the upstream OpenClaw server grants the
-required ``operator.read`` scope.
+Default is ON. Set ``CLAWMETRY_ENABLE_WS_TAP=0`` in the environment to
+opt out. When the upstream OpenClaw server has not yet granted the
+required ``operator.read`` scope the tap connects in degraded mode
+(no message bodies, one WARNING logged) — same observable behaviour as
+when the tap is disabled, but future scope grants start working immediately.
 """
 
 from __future__ import annotations
@@ -107,7 +109,7 @@ _BACKOFF_INITIAL_SEC = 2.0
 _BACKOFF_MAX_SEC = 60.0
 
 
-# Enable env var (feature is default-OFF until upstream grants scopes).
+# Opt-out env var (feature is default-ON; set to "0"/"false"/"no" to disable).
 _ENABLE_ENV = "CLAWMETRY_ENABLE_WS_TAP"
 
 
@@ -577,8 +579,9 @@ def start(config: dict) -> GatewayTap | None:
     from the live OpenClaw config (mirroring dashboard's
     ``_detect_gateway_token`` / ``_detect_gateway_port``).
     """
-    if os.environ.get(_ENABLE_ENV, "").strip() not in ("1", "true", "yes"):
-        log.debug("gateway WS tap disabled (set CLAWMETRY_ENABLE_WS_TAP=1 to enable)")
+    if os.environ.get(_ENABLE_ENV, "1").strip().lower() in ("0", "false", "no"):
+        log.debug("gateway WS tap disabled (CLAWMETRY_ENABLE_WS_TAP=%s)",
+                  os.environ.get(_ENABLE_ENV, "1"))
         return None
 
     url, token = _detect_gateway_endpoint()
--- a/clawmetry/sync.py
+++ b/clawmetry/sync.py
@@ -7953,8 +7953,7 @@ def run_daemon() -> None:
     # memory; gateway.log only carries outbound ACKs (no body), and no
     # JSONL is written for inbound. Without this tap, ClawMetry can
     # NEVER show real Telegram conversations on the Brain tab.
-    # Default-OFF until upstream grants scopes; opt in via
-    # CLAWMETRY_ENABLE_WS_TAP=1.
+    # Default-ON; opt out via CLAWMETRY_ENABLE_WS_TAP=0.
     try:
         from clawmetry import gateway_tap as _gw_tap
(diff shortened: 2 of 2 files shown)


vivekchand:
## UX review @ fb2951b9

User-visible fix: Telegram + Signal + Discord + Slack channel "watch" silently produced no events for anyone without the env var. Flipping default-ON is the correct call. Memory `feedback_local_store_default_off_killed_moat.md` is the same class of bug (opt-in features no one opted into).

**Yellow flags** (none blocking):

1. **No migration banner for existing users**. Anyone on an old OSS version who upgrades will suddenly start sending channel data they may have expected was off. The PR description notes the previous opt-in was effectively zero users, so blast radius is tiny, but a one-time install banner ("Channel watch is now on by default — opt out with `CLAWMETRY_DISABLE_WS_TAP=1`") would be good hygiene.

2. **Default-ON without a kill-switch is risky**. The PR adds the new default but doesn't confirm `CLAWMETRY_DISABLE_WS_TAP=1` works. Worth verifying the negation env var exists.

🤖 Pre-merge review @ fb2951b. Pushing new commits will re-trigger review if diff >50 lines or >2h elapsed.

vivekchand:
## Product review @ fb2951b9

**Why this matters (Elon angle)**: Telegram is the #1 chat channel by user-volume. Per memory `reference_openclaw_telegram_inmemory.md` it has zero on-disk footprint, so the WS tap was the ONLY path. Default-OFF meant the chat channels promised in marketing were broken for 100% of users who didn't read the env var docs. This fix moves the product from "chat support is a lie" to "chat support actually works".

**Follow-up**: file a tracking issue to measure channel-event-volume per node 7 days post-merge so we can quantify how much new data is now flowing. Will inform the next round of channel-tab UX work.

🤖 Pre-merge review @ fb2951b.

PR #1644, opened 2026-05-18
fix: queue cloud POST failures in sync_dlq for retry (#1592)

### LINKED ISSUE #1592: [silent-failure/sync-path] Cloud /ingest/events failure permanently drops events (cursor advances on local-only success)

## Symptom
When `_flush_session_batch` POSTs to cloud `/ingest/events` and the cloud is reachable but rejects/errors (5xx, timeout, encryption-key mismatch, deserialization fault), the daemon catches the exception and logs a warning. The local DuckDB write already committed, so the JSONL cursor (`state[\"last_event_ids\"][fname]`) advances. **The dropped events never get re-tried.** Cloud is permanently missing those rows; user re-sync is the only recovery.

## Root cause
`clawmetry/sync.py:1513-1531` — the comment is explicit:
> Known trade-off: with this guard, sustained cloud failure permanently drops those events from the cloud (cursor moves past them; we don't re-read on next tick). A future PR should add a sidecar cloud-retry queue keyed on canonical event id. For today the priority is local correctness; cloud catches up when the user re-syncs from local.

The \"future PR\" never landed. Outcome: any transient cloud outage during the daemon's window between tick N and tick N+1 silently truncates the cloud-side history.

## Reproduction
1. Run daemon with a valid OpenClaw session generating events.
2. Inject a 502 from cloud `/ingest/events` for ~30 seconds (firewall, kill cloud).
3. Restore cloud reachability.
4. Compare local DuckDB event count vs cloud `events` row count for that node.

Expected: equal. Observed: cloud is missing the events that landed during the outage window.

## Detection strategy
- Cloud-side health probe that compares `max(ts)` per (owner, node) against the daemon's heartbeat `events_total` field. Divergence > X% over 24h → alert.
- Daemon emits a `cloud_drop` event into DuckDB when the warning fires; cron sweep counts these per day.

## Fix shape
- Sidecar SQLite/DuckDB table `cloud_retry_queue` keyed on canonical event id. Failed POSTs enqueue; background drainer retries with exponential backoff.
- Cap the queue at, say, 100K entries with a vacuum policy.
- Re-emit via the same `/ingest/events` payload format; cloud's `ON CONFLICT (id) DO UPDATE` makes replays idempotent (currently fake — see [silent-failure/cross-system] cloud INSERT generates fresh UUID per event sibling issue).

## Severity
**P1** — data-loss on cloud side; local is fine. Affects cloud-only dashboards (other devices, mobile view) but not the source-of-truth.

## Related
- #1565 MOAT tracking
- #1588 v3-shape silent-zero (READ-path family)
- #1590 ring-buffer drop (A1)
- Sibling: [silent-failure/cross-system] cloud INSERT generates fresh UUID — replay would also re-insert

vivekchand:
Closes #1592

## Summary

- **Root cause**: `_flush_session_batch` caught cloud POST exceptions with a bare `log.warning`, so events missed during a cloud outage were permanently absent from the cloud even though local DuckDB had them.
- **Fix**: On POST failure, queue the payload in the existing `sync_dlq` table (`kind="post_failure"`). `_dlq_replay` (already called every tick) now retries these rows — with re-encryption when `enc_key` is set, or as plain POST when it isn't.
- **`_dlq_replay` change**: Removed the unconditional `if not enc_key: return 0` guard; added a per-row `kind` check so encryption-failure rows still defer when no key is available, while `post_failure` rows drain regardless.

No schema changes. Existing `_DLQ_MAX_ATTEMPTS=10` cap and abandon-with-error-log behaviour apply to both failure kinds.

## Test plan

- [x] `make test-api` — 141 passed, 6 skipped
- [x] `python3 -c "import ast; ast.parse(...)"` — syntax clean
- [ ] Manual: inject a 502 during active sync session, verify `sync_dlq` gains a `post_failure` row, verify it drains on the next tick when cloud recovers

---
_Generated by [Claude Code](https://claude.ai/code/session_01StdHox1EpfkpV6AktBkcq6)_

### DIFF
--- a/clawmetry/sync.py
+++ b/clawmetry/sync.py
@@ -604,11 +604,10 @@ def _dlq_enqueue_encryption_failure(
 
 
 def _dlq_replay(api_key: str, enc_key: str | None) -> int:
-    """Drain the encryption DLQ. Returns the number of rows successfully
-    re-encrypted and POSTed. Called from the sync loop on every tick. Cheap
-    no-op when the queue is empty (single COUNT(*) on a tiny table)."""
-    if not enc_key:
-        return 0  # No key configured — replay is a no-op, rows stay parked.
+    """Drain the sync DLQ. Returns the number of rows successfully replayed.
+    Handles both encryption_failure rows (re-encrypt + POST, requires enc_key)
+    and post_failure rows (retry the cloud POST, with or without encryption).
+    Called from the sync loop on every tick; cheap no-op when queue is empty."""
     try:
         from clawmetry import local_store as _ls
         store = _ls.get_store()
@@ -624,6 +623,11 @@ def _dlq_replay(api_key: str, enc_key: str | None) -> int:
     replayed = 0
     for row in rows:
         dlq_id = row["id"]
+        is_post_failure = row.get("kind") == "post_failure"
+        # Encryption-failure rows need enc_key to re-encrypt; defer them until
+        # the user restores the key. post_failure rows can replay without it.
+        if not enc_key and not is_post_failure:
+            continue
         if row["attempts"] >= _DLQ_MAX_ATTEMPTS:
             # Abandon rather than spin forever on a permanently-poisoned row.
             log.error(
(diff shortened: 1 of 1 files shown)


vivekchand:
## UX review @ 5d90bafb

DLQ retry on cloud POST failure is reliability hygiene the user never sees explicitly — but the absence of it is what they DO see (gaps in their dashboard during cloud outages that never backfill). This PR is the right shape: queue locally, retry from existing `_dlq_replay` tick.

**Yellow flags** (none blocking):

1. **No upper bound on DLQ size mentioned in body**. If cloud is down for 24h, DLQ could balloon. Worth confirming there's a max-size with drop-oldest (pattern from `feedback_bounded_heartbeat_queue` / #1635).

2. **No user-visible "X events queued for retry" indicator**. A user noticing dashboard lag should be able to see "47 events pending retry" rather than wondering. Future iteration: tiny chip in the connection status header.

🤖 Pre-merge review @ 5d90baf.

vivekchand:
## Product review @ 5d90bafb

**Why this matters (Elon angle)**: zero-data-loss is the table-stakes promise of observability tooling. The previous bare `log.warning` was the silent failure mode that erodes trust — user looks at dashboard tomorrow, doesn't see yesterday's gap, assumes nothing went wrong, eventually catches it and never trusts the dashboard again. Closing #1592 is reliability cred.

**Follow-up**: instrument DLQ length as a gauge in admin analytics. If we see >0 sustained, that's a cloud-side reliability signal we want before users do.

🤖 Pre-merge review @ 5d90baf.

PR #1663, opened 2026-05-18
fix(brain): /api/component/tool no longer hangs when DuckDB only has model.completed events

vivekchand:
## Summary

Silent-zero bug-class fix #5 today (sister of brain, channel/tui, channel/whatsapp, brain-model fixes shipped earlier in the cycle).

`_try_local_store_component_tool` queried `event_type IN ("message","assistant")` only — but real OpenClaw v3 daemon-normalised ingest writes assistant turns as `event_type="model.completed"`. Real user's DuckDB right now: 0 `message` events + 3 `assistant` events (last 2026-05-17) + 3 `model.completed` events (last 2026-05-16) → fast path silently returned empty events[] for tool families with no recent legacy rows → fell through to the JSONL walker → 8s+ hang → Sessions modal stuck on "Loading...".

Three-part fix:

- **Query `model.completed`**: extend the fast-path query list (daemon-proxy + in-process fallback) to cover the v3 shape every real OpenClaw install writes today. Per memory `reference_openclaw_v3_event_types.md`.
- **Extract `data.toolMetas[]`**: `_iter_tool_call_blocks` now also walks the post-#1135 normalised shape where the inner `message` is stripped and tool calls live in `data.toolMetas[{id,name,input}]`. Real survey of the user's DuckDB: 2 of 62 model.completed rows carry `toolMetas` — and ~36% of `assistant` rows still carry nested `message.content[]`, so the legacy branch stays.
- **5s wall-clock cap + 200-file short-circuit on the JSONL walker**: even when the fast path misses (no daemon / fresh install / unknown shape), the request now bounds within 5s instead of hanging the browser for 8-30s.

Bug class context: `feedback_synthetic_tests_missed_real_event_shape.md` — synthetic tests using `event_type='message'` kept passing while real OpenClaw v3 data flunked silently. Four fixes in this class shipped earlier today (brain, channel/tui, channel/whatsapp, brain-model); this is the fifth.

Production diff: 71 lines in `routes/components.py`. Tests: 537 lines across 3 new files (mandated by the prompt).

## Test plan

- [x] `python3 -m pytest tests/test_component_tool_model_completed.py -q` — **10/10 pass** (1.1s).
- [x] `python3 -m pytest tests/test_component_tool_timeout.py -q` — **2/2 pass** (0.6s).
- [x] `python3 -m pytest tests/test_moat_component_tool_e2e.py -q` — **2/2 pass** (0.5s).
- [x] Full MOAT suite (`pytest tests/test_moat_*.py`) — only 2 pre-existing failures in `test_moat_e2e_regression_1129.py` confirmed on clean main (unrelated to this PR); the perf benchmark flaked at 1628ms vs 1613ms budget on channel_signal but passes standalone (1371ms).
- [x] Manual readback against user's live daemon (`http://127.0.0.1:58343`, 62 model.completed rows): every tool-family endpoint returns in **<130ms with `_source: local_store`** vs the 8s+ pre-fix hang. p50=0.2ms over 10 samples.

### DIFF
--- a/routes/components.py
+++ b/routes/components.py
@@ -57,21 +57,52 @@ def _iter_tool_call_blocks(row: dict):
     """Yield ``(tool_name, arguments, status)`` for every tool-call block
     embedded in a DuckDB ``events`` row.
 
-    OpenClaw's transcript shape (verified 2026-05-13 against
-    ``~/.openclaw/agents/main/sessions/*.jsonl``) wraps tool calls inside
-    ``message`` events with ``message.content`` being a list of blocks.
-    A tool call is a block whose ``type`` is ``toolCall`` (current) or
-    ``tool_use`` (Anthropic-style legacy from older transcripts).
+    Three shapes are handled (silent-zero bug-class fix — see
+    ``feedback_synthetic_tests_missed_real_event_shape.md``):
+
+    1. Legacy trajectory ``message`` event — ``data.message.content[]``
+       carries ``toolCall`` (current) or ``tool_use`` (Anthropic-style
+       legacy) blocks.
+    2. v3 daemon-normalised ``model.completed`` event with the nested
+       message preserved — same ``data.message.content[]`` walk as (1).
+    3. v3 daemon-normalised ``model.completed`` event with ONLY
+       ``data.toolMetas[]`` (the post-#1135 normalised shape, where the
+       inner ``message`` block has been stripped). Each toolMeta carries
+       ``{id, name, input}``.
 
     Audit P0 #4: the previous implementation queried for
     ``event_type='tool_call'`` rows that the daemon never writes — every
     row in DuckDB has the raw OpenClaw type (``message`` / ``assistant``
-    / ``user`` / …). All 10 component-tool endpoints fell through to the
-    legacy JSONL parser as a result.
+    / ``model.completed`` / …). All 10 component-tool endpoints fell
+    through to the legacy JSONL parser as a result. Today's silent-zero
+    fix (5th instance) extends that fix to cover model.completed.
     """
     data = row.get("data") if isinstance(row, dict) else None
     if not isinstance(data, dict):
         return
+
+    # Shape 3 — v3 normalised ``toolMetas[]`` on model.completed. Surface
+    # these FIRST so installs with only-normalised events still light up.
+    tool_metas = data.get("toolMetas")
+    if isinstance(tool_metas, list):
+        for meta in tool_metas:
+            if not isinstance(meta, dict):
+                continue
+            tname = meta.get("name") or meta.get("tool") or ""
+            args = meta.get("input") or meta.get("arguments") or {}
+            if not isinstance(args, dict):
+                args = {"_raw": str(args)[:200]}
+            # toolMetas have no error flag — daemon strips it. We default
+            # to ok; if a paired tool.result event later carries
+            # is_error=true, the legacy parser's toolResult branch will
+            # surface that. Fast path keeps the simpler "ok" default.
+            if tname:
+                yield tname, args, "ok"
+
+    # Shapes 1 + 2 — both walk data.message.content[]. May coexist with
+    # toolMetas on a model.completed row; dedupe by tool name+args is
+    # left to the caller (counts are per-block, which matches the legacy
+    # parser).
     msg = data.get("message")
     if not isinstance(msg, dict):
         return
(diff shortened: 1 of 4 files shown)


vivekchand:
Tracked in MOAT review queue #1695 (recommended merge position: Tier 4, position 1).

vivekchand (commented):
## Test plan & review notes

**What changed**
- `/api/component/tool` fast path now queries `model.completed` events and handles `data.toolMetas[{id,name,input}]` from the post-#1135 normalised shape. Also adds a 5s wall-clock cap + 200-file short-circuit on the JSONL walker fallback to bound request time from 8s+ to <5s.

**Smoke commands**
- `pytest tests/test_component_tool_model_completed.py tests/test_component_tool_timeout.py tests/test_moat_component_tool_e2e.py -q` — 14 tests (10 + 2 + 2) should pass
- On a v3 install: `curl -w "%{time_total}" http://localhost:8900/api/component/tool/bash` → should return in <200ms with `_source: local_store` and non-zero `today_calls`
- Timeout guard smoke: cold store (no DuckDB rows, JSONL walker triggered) → same endpoint completes in ≤5s

- The 5s cap applies to the JSONL walker only, not the DuckDB fast path — if the fast path encounters a very large DuckDB without an index on `event_type`, it can independently hang; confirm the query uses an indexed predicate
- `data.toolMetas[].input` vs `data.message.content[].input` from the legacy shape: verify both are surfaced consistently in the response (same JSON key names) so the UI doesn't need a dual rendering path

**Issue link**
- Silent-zero bug-class #5 in the series; context: `feedback_synthetic_tests_missed_real_event_shape.md`

---

vivekchand:
Triage 2026-05-19: STILL VALID. Verified against `origin/main:routes/components.py`:

- `_try_local_store_component_tool` still only queries `message` / `assistant` (lines 127-132), no `model.completed`
- No 5s wall-clock cap or 200-file short-circuit on the JSONL walker

Today's MOAT wave (#1728, #1730, #1731, #1732, #1733, #1754) did NOT touch this code path. The silent-zero + hang fix here is unique. Mergeable, only failing check is the known E2E browser flake. Ready for merge when you are.

vivekchand:
persona-gate not met yet (no ## UX review or ## Product review comments), deferring. Persona cron will pick it up at :17 or :47.

vivekchand:
## UX review

Backend-only diff: `routes/components.py` + 3 new test files. No HTML/CSS/JS touched. However, the user-visible impact is enormous: this is the fix for the Sessions modal stuck on "Loading..." for 8+ seconds on real OpenClaw v3 installs.

- 🟢 the 5s wall-clock cap on the legacy JSONL walker is the correct safety net — better to render "no recent activity" within 5s than to stall the modal for 8s+ and leave a non-technical user wondering if the dashboard crashed. Matches the "Never crash on bad input" convention in CLAUDE.md.
- 🟡 P1: when the 200-file short-circuit fires, the modal will show an empty state with zero visual cue that "we capped this and there might actually be data." Consider adding a `_truncated: true` flag in the payload + a small inline hint in the modal UI in a follow-up. Not a blocker for this PR.

Verdict: SHIP. Fixes a hair-on-fire UX regression.

🤖 Pre-merge review @ 67744cb. Pushing new commits will re-trigger review if diff >50 lines or >2h elapsed.

vivekchand:
## Product review

Bug-class fix #5 of the silent-zero family (per PR body). The simplest possible thing: add the v3 `model.completed` event type to the predicate list + extract `data.toolMetas[]`. No tier-leak risk — purely fixes broken OSS UX.

- 🟢 first-principles: the fix is 3-prong (query, extract, cap) and each prong addresses a distinct failure mode (predicate skew / shape skew / unbounded fallback). 71 production lines + 537 test lines is the right ratio for a bug-class regression.
- 🟢 funnel: silent-zero on the Sessions modal = user installs ClawMetry, sees an empty/hung dashboard, churns. Fixing this is a direct win against the "empty dashboard looks broken" failure mode flagged in `project_sync_progress_banner.md`.
- 🔴 P0 (cross-PR conflict, not in this PR's diff): **#1682 also patches `_TOOL_HOST_EVENT_TYPES` in the same file, adding `subagent:assistant` but NOT `model.completed`.** Whichever lands second will clobber the other unless the rebase explicitly unions both. Recommended merge order: land #1663 first (this PR), then rebase #1682 onto main so the predicate list becomes `("message", "assistant", "model.completed", "subagent:assistant")`. **Action: flag the merge-wave agent to enforce this order.**
- 🟡 P1: keystone harness from #1672 (`check_component_tool_vs_ground_truth`) would catch this exact bug class on every CI run once landed. Worth confirming the harness covers `model.completed`-only installs (the PR description says it does; spot-check after #1672 lands).

Verdict: SHIP after confirming merge order with #1682.

🤖 Pre-merge review @ 67744cb. Pushing new commits will re-trigger review if diff >50 lines or >2h elapsed.

PR #1681, opened 2026-05-18
feat(moat): cloud-side keystone E2E probes (local <-> cloud drift detection)

vivekchand:
## Summary

Sibling of `scripts/accuracy_harness/keystone_e2e.py` (LOCAL half, PR #1672). This script verifies the **CLOUD half** of the OpenClaw -> daemon -> ingest -> Postgres -> app.clawmetry.com loop, so we can prove the data a paying customer sees in the Node Detail modal matches the data their local dashboard shows.

Six cloud probes back the Node Detail page:
- `/api/cloud/account` (auth works)
- `/api/cloud/nodes` (user has the expected node)
- `/api/cloud/node/<id>/info` (node metadata loads)
- `/api/cloud/node/<id>/summary` (model + token totals)
- `/api/cloud/node/<id>/sessions` (sessions list)
- `/api/cloud/node/<id>/components` (gateway / exec / browser / messages / machine panels)

For each axis with a local counterpart, the harness ALSO hits the local dashboard and prints a side-by-side **LOCAL vs CLOUD drift table** (sessions count, recent model, token totals, sessions listed). The drift detector flags axes where cloud exceeds local (impossible without a bug) or where both sides return zero on an axis that should have data.

Auto-discovers `api_key` + `node_id` from `~/.clawmetry/config.json` (the file `clawmetry connect` writes after pairing). Auth uses `?token=<api_key>` query arg because PR #978's `cm_token` cookie path is merged but not yet deployed on cloud due to the GitHub Actions billing block.

## End-to-end verification on the user's live node

Tested against `agent+Macbook-Pro-2-local` / `vivek+1822@clawmetry.com`:

```
[cloud-keystone] Per-endpoint results:
  [PASS] /api/cloud/account                                       auth+ok email=vivek+1822@clawmetry.com plan=trial nodes=1
  [PASS] /api/cloud/nodes                                         node_present node found, status=online version=0.12.240
  [PASS] /api/cloud/node/agent%2BMacbook-Pro-2-local/info         shape status=online version=0.12.240
  [PASS] /api/cloud/node/agent%2BMacbook-Pro-2-local/summary      shape sessions=6 tokens=1232 model=claude-opus-4-7
  [PASS] /api/cloud/node/agent%2BMacbook-Pro-2-local/sessions     shape sessions=0 (relay_pending)
  [PASS] /api/cloud/node/agent%2BMacbook-Pro-2-local/components   shape sections=['browser','exec','gateway','machine','messages'] gw_status=online

[...]

PR #1686, opened 2026-05-18
fix(sync): pending_query dispatch filters cloud-injected node_id kwarg

vivekchand:
## Summary

Fixes the root cause behind the 2026-05-18 P0 where the cloud showed `0/lifetime` and the **Embodied** tab stayed empty.

Every heartbeat fired 3x:

```
WARNING pending_query dispatch failed (id=q_node_sessions_refresh_<ts>
shape=sessions): LocalStore.query_sessions() got an unexpected
keyword argument 'node_id'
```

- Cloud's `/api/cloud/node/<id>/sessions` enqueues a heartbeat-relay pending_query with `args = {"node_id": node_id}` (it uses it for cloud-side routing across multiple nodes).
- The daemon forwarded `args` straight into `LocalStore.query_sessions(**args)`.
- `query_sessions()` has no `node_id` kwarg (the local store IS a single-node store), so it raised `TypeError` per dispatch.
- The sessions cache never populated -> `/api/cloud/node/<id>/sessions` returned `_source: relay_pending, eta_sec: 60` forever.

## Fix

Per-shape kwarg allowlist applied at **both** dispatch entry points so cloud-side routing metadata can't trip the local store:

- `routes/local_query.py:_dispatch` now calls the existing `_coerce_args(shape, args)` helper before invoking the store. This keeps both HTTP and heartbeat-piggyback transports on the same safety rail.
- `clawmetry/sync.py:_local_dispatch_fallback` (used when `routes/` isn't on `sys.path` for daemon-only installs) gets a duplicated `_SHAPE_ALLOWED_KWARGS` + `_filter_store_kwargs()` so it stays zero-dep.

Diff: 44 LOC under code, 196 LOC of test, all in the same blast radius as the bug.

## Regression test

`tests/test_pending_query_dispatch.py` — 6 cases covering both transport paths + the per-shape allowlist matrix. The crucial one (`test_pending_query_with_node_id_completes_without_warning`) wires `send_heartbeat` end-to-end with a fake `_post` and asserts the `pending_query dispatch failed` warning is **not** logged.

## Test plan

- [x] `pytest tests/test_pending_query_dispatch.py` -> 6 passed.
- [x] `pytest tests/test_local_query_api.py` -> 16 passed (no regression on HTTP path).
- [x] `pytest tests/test_heartbeat_dispatch.py` -> 4 passed, 1 pre-existing failure unrelated (`test_encrypted_blob_decrypts_to_original` -> already red on main).
- [x] `python3 -c 'import dashboard'` clean.
- [x] Local daemon restart + 100s window: **0** `node_id` `TypeError` dispatch failures (was 3 every 10s pre-fix).
- [x] `scripts/accuracy_harness/cloud_keystone_e2e.py` (from PR #1681): `/api/cloud/node/<id>/sessions` flipped from `_source: relay_pending, sessions=0` to `_source: cache, sessions_blob=<populated>`.

## Note on related residual log noise

One `pending_query dispatch failed: Expecting value: line 1 column 1 (char 0)` appears at boot (proxy_dispatch race against the local_server thread before it binds). That's pre-existing, not the same root cause, and the dispatch failure is caught + retried on the next heartbeat. Out of scope for this PR; can be cleaned up separately by retrying the proxy once after a short backoff if `is_running()` was true but the socket wasn't ready.

No `[RELEASE]` tag -> @vivekchand has the merge auth power for this P0.

### DIFF
--- a/clawmetry/sync.py
+++ b/clawmetry/sync.py
@@ -5016,7 +5016,13 @@ def _canonical_args_hash(args: dict) -> str:
 
 def _local_dispatch_fallback(shape: str, args: dict) -> dict:
     """Fallback dispatcher used when `routes.local_query` isn't importable
-    (daemon-only installs). Mirrors the minimal shape→method bridge."""
+    (daemon-only installs). Mirrors the minimal shape→method bridge.
+
+    Cloud-side pending_queries may carry kwargs the local store doesn't
+    recognise (e.g. ``node_id`` — the cloud uses it to route between
+    nodes, but the local DuckDB IS a single-node store). Filter to a
+    per-shape allowlist so those extras don't surface as TypeErrors.
+    """
     from clawmetry import local_store
     store = local_store.get_store(read_only=True)
     if shape == "health":
@@ -5030,8 +5036,28 @@ def _local_dispatch_fallback(shape: str, args: dict) -> dict:
     method = method_map.get(shape)
     if not method:
         raise ValueError(f"unknown shape: {shape}")
-    rows = getattr(store, method)(**(args or {}))
+    rows = getattr(store, method)(**_filter_store_kwargs(shape, args or {}))
     return {"rows": rows, "count": len(rows), "_shape": shape, "_via": "fallback"}
+
+
+# Per-shape allowlist of kwargs accepted by the underlying ``LocalStore``
+# methods. Kept in sync with ``LocalStore.query_*`` signatures in
+# ``clawmetry/local_store.py``. Mirrors ``routes.local_query._coerce_args``
+# but defined here so the daemon-only fallback path doesn't need the
+# routes package on sys.path.
+_SHAPE_ALLOWED_KWARGS = {
+    "events":     {"session_id", "agent_id", "event_type", "since", "until", "limit"},
+    "sessions":   {"agent_id", "since", "until", "limit"},
+    "aggregates": {"agent_id", "since", "until"},
+    "transcript": {"session_id", "limit"},
+}
+
+
+def _filter_store_kwargs(shape: str, args: dict) -> dict:
+    allowed = _SHAPE_ALLOWED_KWARGS.get(shape)
+    if allowed is None:
+        return dict(args)
+    return {k: v for k, v in args.items() if k in allowed}
 
 
 def _channel_enrichment_from_row(r: dict) -> dict:
--- a/routes/local_query.py
+++ b/routes/local_query.py
@@ -208,8 +208,24 @@ def _dispatch(shape: str, args: dict) -> dict:
          the DuckDB while it owns the writer lock).
       2. Else fall back to opening the DuckDB directly (single-process
          mode, or daemon temporarily down).
+
+    Arg coercion: callers are NOT required to pre-filter ``args``. Cloud-side
+    callers (heartbeat-piggyback ``pending_queries`` from
+    ``clawmetry-cloud/routes/cloud.py``) sometimes attach metadata like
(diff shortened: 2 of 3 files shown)


vivekchand:
## Principal G verification: fix confirmed live in production daemon

Re-ran `scripts/accuracy_harness/cloud_keystone_e2e.py` (PR #1681 harness) twice at 2026-05-19 00:50 UTC. Daemon is the live launchd-managed `~/.clawmetry/bin/python3 -m clawmetry.sync` (PID 35668, restarted around 00:19 UTC right after PR #1686 merged).

### Cloud probe results: 5/6 PASS

| Endpoint | Before (PR #1681 baseline) | After (this verification) | Status |
|---|---|---|---|
| `/api/cloud/account` | PASS | PASS | OK |
| `/api/cloud/nodes` | PASS (node found) | PASS (node found) | OK |
| `/api/cloud/node/<id>/info` | PASS | PASS | OK |
| `/api/cloud/node/<id>/summary` | sessions=6 tokens=1232 | sessions=6 tokens=**1738** | OK (cloud catching up) |
| `/api/cloud/node/<id>/sessions` | `_source: relay_pending, eta_sec: 60` (PASS only because harness tolerated empty list) | `_source: cache, sessions_blob=<2.0kB populated>` | **FIX CONFIRMED** |
| `/api/cloud/node/<id>/components` | PASS | PASS | OK |

The single FAIL in run 2 is the harness itself expecting a decrypted `.sessions` key. Cloud now (correctly) returns an E2E encrypted `sessions_blob` because the cache populated. This is a harness assertion to update, not a backend regression. Pre-fix the cache was empty and `_source: relay_pending` was returned; post-fix the daemon actually answers the pending query and the cache fills.

### sync.log error growth: FROZEN

```
$ grep -c "node_id'" /Users/vivek/.clawmetry/sync.log
1335

$ grep "node_id'" /Users/vivek/.clawmetry/sync.log | tail -1
2026-05-19 00:18:54,456 [clawmetry-sync] WARNING pending_query dispatch failed
  (id=q_node_sessions_refresh_1779142685847 shape=sessions):
  LocalStore.query_sessions() got an unexpected keyword argument 'node_id'
```

Last `node_id` TypeError was 00:18:54 UTC. Daemon restarted ~00:19 with the fixed wheel. Zero new `node_id` errors in the 31+ minutes since. Two more `pending_query dispatch failed` lines exist after 00:19 but both are unrelated `Expecting value: line 1 column 1` JSON parse errors on an empty body, not the kwarg crash that PR #1686 targeted.

### LOCAL vs CLOUD drift (still informational, not regressions)

| Axis | LOCAL | CLOUD | Note |
|---|---|---|---|
| sessions_count | 2 (24h cap) | 6 (lifetime) | Expected: different window semantics |
| recent_model | claude-opus-4-7 | claude-opus-4-7 | OK |
| tokens_today | 0 | 1738 (lifetime) | Expected: same window mismatch |
| sessions_listed | 2 | None (encrypted blob) | Harness needs an E2E decrypt path |

`/api/cloud/nodes` still shows `event_count: 0`. That is the heartbeat-ring-depth telemetry (different code path from the session cache PR #1686 fixed) and out of scope here.

### Verdict

Principal D's fix in PR #1686 (`9a009f1`) is live, took effect on daemon restart, has stopped the kwarg crash from recurring, and the cloud-side cache now populates with real session data. The `relay_pending` symptom Principal B documented in PR #1681 is resolved. Ready to close the underlying issue.

Verified by: Principal G (in-session, 10-min budget).

Issue #1708, opened 2026-05-18
[epic] Agent burnout: detect cognitive loops + unproductive token burn


vivekchand:
## Source
Wolfgang Heider, "OpenClaw Hacked Again? My Agents Ran into AI Burnout" (LinkedIn pulse, 2026-05-18).

He thought OpenClaw got hacked again after seeing unexplained overnight token consumption. Turned out an engineering agent entered an infinite loop running validation tests: it kept trying to validate the results, and apparently also the validation of those results, over and over again.

This is a **cognitive loop**, distinct from a **tool loop**. Our existing `tool_call_stuck` classifier (#1671, shipped tonight) catches a tool that hangs. It does NOT catch an agent that keeps making forward tool calls that all succeed but produce no new state.

## What we already have (context)
- `tool_call_stuck` classifier: detects unanswered tool invocations >120s
- Pro alerts: Daily spend, Token velocity >10k/min, Cron failed N×, Tool failures >N/hr, Session cost >$5, Agent offline >10min
- Budget enforcement proxy on port 4100 (blocks API calls past limit)
- Outcome classifier: success, failed, escalated, ongoing, tool_call_stuck

### G1. Recursive self-validation detection (sub-issue #1706)
Same or near-identical assistant message repeating N times in one session with no forward-progress signal.

### G2. Forward-progress signal (sub-issue #1707)
Token-burn-per-state-delta ratio. State delta = new tool, new file, new error type, new outcome. When tokens climb but deltas stay flat, the agent is spinning.

### G3. Auto-abort on budget overrun (not just alert)
Pro proxy blocks API calls past budget but doesn't kill the agent loop with a structured `BUDGET_EXCEEDED` interrupt. The agent keeps retrying. Add an explicit abort signal the agent runtime can react to. Separate sub-issue when ready.

### G4. "Agent Resources" intervention framework
Pluggable rules engine where users compose custom interventions: pause if same prompt 3× in 5min, kill if forward-progress=0 over 50k tokens, alert oncall if cost-per-token-delta > X. Marketing-friendly framing: HR for AI agents. Separate sub-issue when ready.

## Acceptance (epic-level)
- Cognitive-loop detection ships as a 6th outcome class (`cognitive_loop`) alongside `tool_call_stuck`
- Forward-progress signal exposed on the Brain tab + queryable via DuckDB
- One concrete pre-built intervention rule (e.g. kill if same assistant message 5× in 10min) ships before "Agent Resources" framework
[...]

PR #1715, opened 2026-05-19
feat(proxy): structured BUDGET_EXCEEDED abort signal (closes G3 of #1708)

vivekchand:
## Summary

Closes G3 of epic #1708 (Wolfgang burnout).

Pro proxy on port 4100 currently returns a generic 429 when the daily budget is exhausted. Agent runtimes see the 429 and just retry, so the agent keeps spinning past the limit. This PR adds a structured abort envelope so runtimes can hard-stop the loop on a single response.

### Before

```http
HTTP/1.1 429 Too Many Requests
Content-Type: application/json

{
  "type": "error",
  "error": {
    "type": "budget_exceeded",
    "message": "Daily budget exceeded: $10.00 / $5.00"
  }
}
```

### After

```http
HTTP/1.1 429 Too Many Requests
Content-Type: application/json
X-Clawmetry-Budget-Status: exceeded

{
  "type":  "error",
  "error": {"type": "budget_exceeded", "message": "Daily budget exceeded: $10.00 / $5.00"},
  "code":  "BUDGET_EXCEEDED",
  "should_abort":        true,
  "retry_after_seconds": null,
  "spent_today":         10.0,
  "budget_today":        5.0,
  "message":             "Daily budget reached. Halting agent. Raise the budget in ClawMetry Cloud settings to resume."
}
```

Runtimes branch on either `code == "BUDGET_EXCEEDED"` (body) or `X-Clawmetry-Budget-Status: exceeded` (header) and stop their retry loop. Runtimes that ignore the new fields fall back to the legacy 429 retry path, so the non-budget rate-limit contract (per-plan 60/min on free, loop-detected 429s, etc.) is unchanged.

## Why option A and not "POST abort to a daemon endpoint"

Option A (this PR) is unilateral. The proxy already owns the response. Cooperating runtimes opt in for free; non-cooperating runtimes keep retrying, same as today. Option B (proxy POSTs `/api/agent-runtime/abort?...`) requires every agent runtime to expose and trust that endpoint, which is a separate negotiation per harness.

## Also in this PR

- New optional Pro alert template `"Agent halted by budget"` in `routes/alerts.py` `DEFAULT_ALERT_RULES`. NOT default-enabled (`enabled: false`); shows up in the one-click templates list so users who configure a daily cap can opt in to be notified the first time it actually halts an agent.

## Test plan

- [x] 3 new cases in `tests/test_proxy_budget_abort.py`:
  - Healthy budget: abort header does NOT leak onto normal responses.
  - Exhausted budget: 429 body carries `code: BUDGET_EXCEEDED` + `should_abort: true`.
  - Exhausted budget: response header carries `X-Clawmetry-Budget-Status: exceeded`.
- [x] All 58 existing `tests/test_proxy.py` cases still pass (no regression on `budget_exceeded`/`loop_detected`/`authentication_error` 429s).
- [x] `python3 -c 'import dashboard'` clean.
- [x] No em-dashes in the user-facing `message` field (banned in copy that integrations may read out loud).
- [x] Diff ~209 LOC (proxy 84, alerts 24, tests 105).

NO `[RELEASE]` tag. Will not be merged from this PR.

### DIFF
--- a/clawmetry/proxy.py
+++ b/clawmetry/proxy.py
@@ -22,6 +22,30 @@ Config in ~/.clawmetry/proxy.json:
             "openai": { "api_key_env": "OPENAI_API_KEY", "base_url": "https://api.openai.com" }
         }
     }
+
+BUDGET_EXCEEDED abort contract (G3 of #1708):
+    When ``budget.action == "block"`` and the daily/monthly cap is hit,
+    the proxy returns HTTP 429 with a structured abort envelope so the
+    agent runtime can hard-stop instead of retrying forever:
+
+        Status: 429
+        Header: X-Clawmetry-Budget-Status: exceeded
+        Body  : {
+            "type":  "error",
+            "error": {"type": "budget_exceeded", "message": "<reason>"},
+            "code":  "BUDGET_EXCEEDED",
+            "should_abort":        true,
+            "retry_after_seconds": null,
+            "spent_today":         <float>,
+            "budget_today":        <float>,
+            "message":             "Daily budget reached. Halting agent. ..."
+        }
+
+    Agent runtimes SHOULD check for ``code == "BUDGET_EXCEEDED"`` (or
+    the ``X-Clawmetry-Budget-Status`` header) and stop their loop.
+    Runtimes that ignore these fields fall back to the legacy 429
+    retry path, preserving backward compatibility with non-budget
+    rate-limit semantics.
 """
 
 from __future__ import annotations
@@ -791,9 +815,21 @@ def create_proxy_app(config: ProxyConfig = None) -> "Flask":
     _stats_lock = threading.Lock()
 
     def _error_response(
-        status_code: int, error_type: str, message: str, provider: str = "anthropic"
+        status_code: int,
+        error_type: str,
+        message: str,
+        provider: str = "anthropic",
+        extra: Optional[dict] = None,
+        extra_headers: Optional[dict] = None,
     ) -> Response:
-        """Return an error response in the provider's expected format."""
+        """Return an error response in the provider's expected format.
+
+        ``extra``: top-level fields merged into the JSON body (used to surface
+        the structured ``BUDGET_EXCEEDED`` abort envelope per G3 of #1708).
+        ``extra_headers``: additional response headers (e.g.
+        ``X-Clawmetry-Budget-Status: exceeded`` so HTTP clients that respect
+        headers can short-circuit retries without parsing the body).
+        """
         if provider == "openai":
             body = {
                 "error": {
@@ -810,10 +846,50 @@ def create_proxy_app(config: ProxyConfig = None) -> "Flask":
                     "message": message,
                 },
             }
-        return Response(
+        if extra:
+            body.update(extra)
+        resp = Response(
             json.dumps(body),
             status=status_code,
             content_type="application/json",
+        )
+        if extra_headers:
+            for k, v in extra_headers.items():
+                resp.headers[k] = v
+        return resp
+
+    def _budget_abort_response(reason: str, provider: str) -> Response:
+        """Structured BUDGET_EXCEEDED abort signal (G3 of #1708).
+
+        Agent runtimes that wrap the proxy should branch on either
+        ``code == "BUDGET_EXCEEDED"`` in the JSON body or the
+        ``X-Clawmetry-Budget-Status: exceeded`` response header and STOP
+        their retry loop. Plain 429 retry behaviour is preserved for
+        runtimes that do not read these fields, so the contract is
+        backward compatible with non-budget rate limits.
+        """
+        status = budget.get_status()
+        spent = status.get("daily_spent", 0.0)
+        limit = status.get("daily_limit", 0.0)
+        message = (
+            "Daily budget reached. Halting agent. "
+            "Raise the budget in ClawMetry Cloud settings to resume."
+        )
+        extra = {
+            "code":                "BUDGET_EXCEEDED",
+            "should_abort":        True,
+            "retry_after_seconds": None,
+            "spent_today":         spent,
+            "budget_today":        limit,
+            "message":             message,
+        }
+        return _error_response(
+            429,
+            "budget_exceeded",
+            reason,
+            provider,
+            extra=extra,
+            extra_headers={"X-Clawmetry-Budget-Status": "exceeded"},
         )
 
     def _get_upstream_url(provider: str, path: str) -> str:
@@ -1070,7 +1146,7 @@ def create_proxy_app(config: ProxyConfig = None) -> "Flask":
             elif config.budget.action == "warn":
                 pass  # Allow through with warning
             else:
-                return _error_response(429, "budget_exceeded", reason, provider)
+                return _budget_abort_response(reason, provider)
 
         # ── Loop detection ─────────────────────────────────────────────
         req_hash = compute_request_hash(body)
--- a/routes/alerts.py
+++ b/routes/alerts.py
@@ -112,6 +112,30 @@ DEFAULT_ALERT_RULES = [
             "Catches genuine spinning, not just busy productive burn."
         ),
     },
+    # G3 of #1708 (Wolfgang burnout): the proxy now emits a structured
+    # BUDGET_EXCEEDED abort signal when the daily cap is hit. This seed
+    # rule surfaces those aborts in the alerts UI as an opt-in template
+    # so a user who configured a budget sees a one-click way to be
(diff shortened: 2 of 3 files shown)


vivekchand (commented):
## Test plan & review notes

**What changed**
- Proxy on port 4100 now returns a structured `BUDGET_EXCEEDED` body + `X-Clawmetry-Budget-Status: exceeded` header on budget-exhausted 429s; runtimes can key on `code == "BUDGET_EXCEEDED"` or the header to hard-stop their retry loop. Adds an opt-in "Agent halted by budget" alert template (disabled by default).

**Smoke commands**
- `pytest tests/test_proxy_budget_abort.py -v` — 3 new cases: clean budget (header must NOT appear), exceeded (body has `code` + `should_abort: true`), exceeded (header `X-Clawmetry-Budget-Status: exceeded` present)
- `pytest tests/test_proxy.py` — all 58 existing cases should pass (regression on non-budget 429s: rate-limit, loop-detected, auth-error must not leak the budget header)
- `python3 -c "import dashboard"` — clean import

- Non-budget 429s (loop-detected, per-plan rate-limit) must never carry `X-Clawmetry-Budget-Status: exceeded` — the test covers this but worth a manual spot-check: `curl -H "Authorization: Bearer loop-test" http://localhost:4100/v1/messages`
- `retry_after_seconds: null` — cooperating runtimes that try to parse this as an int will throw; make sure the field is documented as nullable in any client-side handling examples
- The "Agent halted by budget" alert template is `enabled: false` by default — verify it shows up in the one-click templates list even when disabled

**Landing note**
- This PR carries no `[RELEASE]` tag and is flagged "Will not be merged from this PR." The intended landing vehicle isn't specified — if it's waiting to be bundled into an epic #1708 omnibus, dropping a note in the description about the expected merge PR would help avoid it going stale.

**Issue link**
- Closes G3 of epic #1708

---

vivekchand (commented):
## Test plan & review notes

**What changed**
- Adds a `_budget_abort_response()` helper in `clawmetry/proxy.py` that, when the daily/monthly cap is hit and `budget.action == "block"`, returns HTTP 429 with a structured JSON abort envelope (`code: BUDGET_EXCEEDED`, `should_abort: true`, spend/limit floats, human message) plus an `X-Clawmetry-Budget-Status: exceeded` header — replacing the previous bare `_error_response(429, "budget_exceeded", ...)` call. Also adds an opt-in (disabled-by-default) Pro alert rule seed in `routes/alerts.py` and 3 new pinning tests in `tests/test_proxy_budget_abort.py`.

**Smoke commands**
- `make test-api`
- Run `python -m pytest tests/test_proxy_budget_abort.py -v` to exercise the three new guarantees directly (no running server needed).
- Set a tiny daily budget (`daily_usd: 0.01`) in the proxy config, send a real request through port 4100, push spend over the cap via `ProxyDB.record_usage`, then send one more request — verify the 429 body has `code: BUDGET_EXCEEDED` and `should_abort: true`, and that the `X-Clawmetry-Budget-Status: exceeded` header is present.

- `budget.get_status()` is called inside `_budget_abort_response()` but that method is defined on the `BudgetEnforcer` instance captured in the closure. If `budget` is `None` (proxy running without a budget config) and the code path somehow reaches `_budget_abort_response`, this will `AttributeError`. Worth a guard or confirming the `action == "block"` branch is only reachable when budget is configured.
- The `extra` dict uses `body.update(extra)` which means for OpenAI-format responses, `extra["message"]` will overwrite the `body["error"]["message"]` field — the top-level `message` key is fine for Anthropic format but for OpenAI format clients this may surface an unexpected top-level `message` field. Low risk since budget-block is a ClawMetry-proxy-specific concept, but worth noting.
- `retry_after_seconds: null` is always emitted. If a midnight reset is calculable from the daily cap window, a concrete value here would let smart runtimes sleep-and-resume rather than hard-abort. Currently `null` means runtimes that respect this field do nothing with it, which is fine — just noting it as a future improvement.
- Safety concern: the abort signal is synchronous per-request, but if an agent is running parallel sub-requests through the proxy the first blocked request returns the abort envelope while sibling requests that were already in-flight may still succeed (or return separate 429s). The `should_abort: true` contract only works reliably if the runtime checks it on every response leg, not just the first failure.

**Issue link**
- The title says "closes G3 of #1708" and the body says "Closes G3 of epic #1708" — GitHub's autoclose keyword only fires on the exact form `Closes #1708` (bare number, no qualifiers). Because #1708 is an epic with multiple goals, it should stay open anyway. Suggest changing "Closes G3 of epic #1708" in the body to `Refs #1708` so GitHub creates a cross-reference link without inadvertently closing the tracking issue once the keyword parser is strict-matched by a future GH update.


---

vivekchand:
## UX review

Backend (proxy) + one new opt-in alert template that surfaces in `/alerts` UI. Mostly backend; one user-facing surface (the alert-template list).

- 🟡 P1 — the new "Agent halted by budget" template at `routes/alerts.py:130` carries `description: "Fires when the Pro proxy emits at least one BUDGET_EXCEEDED abort in the last 24 hours."` — that copy assumes the user knows what "BUDGET_EXCEEDED" is. Most operators see this string for the first time in this template. Rewrite as "Fires when ClawMetry's budget proxy halts an agent after hitting the daily cap." Code is internal; copy is user-facing.
- 🟡 P1 — `clawmetry/proxy.py:884` user-facing `message` says "Raise the budget in ClawMetry Cloud settings to resume." — but the proxy is OSS, the daily cap is set in OSS config (`BudgetConfig`), and only the ALERT half of the feature is Pro. Wrong product surface. Either keep it neutral ("Raise the daily cap in your proxy config to resume.") or pre-check the user's tier and split copy. Risk: free-tier user clicks through to cloud settings, sees no budget setting, bounces.
- 🟢 nit — `_budget_abort_response` body keys mix snake_case (`should_abort`, `retry_after_seconds`) and concept names — good consistency with `error.type` convention. `code: "BUDGET_EXCEEDED"` SCREAMING_SNAKE matches CLAUDE.md conventions (constants).
- 🟢 nit — `X-Clawmetry-Budget-Status: exceeded` header name is clean and namespaced.

🤖 Pre-merge review @ 626954f5. Pushing new commits will re-trigger review if diff >50 lines or >2h elapsed.

vivekchand:
## Product review

Closes G3 of the Wolfgang-burnout epic (#1708). Simplest possible thing for the actual problem: agent runtimes ignore generic 429s and retry past the cap, so add a structured signal they can branch on. The "Option A vs Option B" framing in the PR body is the right first-principles argument — unilateral proxy response beats per-harness `/api/agent-runtime/abort` negotiation.

- 🟡 P1 — this only works if agent runtimes actually read `code == "BUDGET_EXCEEDED"`. The PR ships the signal but NOT a single example runtime adapter (OpenClaw / Claude Code / Codex) wired to consume it. Without at least one adapter PR landing in lockstep, this signal is dead-letter — agents continue burning past cap. Open follow-up `[CXO] Wire BUDGET_EXCEEDED abort into OpenClaw harness retry loop` (file in OpenClaw repo, not OSS).
- 🟡 P1 — the new alert template (`Agent halted by budget`) is gated `pro_only: true` which is correct (Alerts is Cloud-Pro per `project_alerts_pro_feature.md`). But the abort SIGNAL itself works on free-tier OSS proxy users too. Free-tier users get the protection (good!) but get NO notification surface (their agent silently halts). That's a confusing experience that drives "my agent stopped working" support tickets. Mitigation: surface a one-time toast in the OSS dashboard when `budget_blocked` event count > 0 in the last hour, with a "Get notified next time → Upgrade" CTA. Conversion lever.
- 🟡 P1 — `clawmetry/proxy.py:884` user copy directs to "ClawMetry Cloud settings" but the cap is set in OSS proxy config. Wrong surface; will leak to a 404-ish journey for free-tier users. (Covered in UX review too.)
- 🟢 nit — test coverage (3 cases) pins the contract well: positive header, positive body, negative-on-healthy. Add a 4th for the `monthly_usd` cap path (`clawmetry/proxy.py` line ~1146 falls through the same `_budget_abort_response`); easy to forget the monthly branch in a future refactor.
- 🟢 simplest-thing check — header + body envelope is the minimum surface change. Backward compatible (legacy 429 retry semantics preserved). Good shape.

OSS leakage check: the SIGNAL is OSS-correct (proxy is OSS, cap is OSS). The ALERT TEMPLATE is correctly Pro-gated. Conversion angle: the "free-tier silently halted" experience IS a paid-conversion moment if surfaced. Move on that follow-up.

🤖 Pre-merge review @ 626954f5. Pushing new commits will re-trigger review if diff >50 lines or >2h elapsed.

PR #1723, opened 2026-05-19
[RELEASE] fix: alerts modal centers in viewport (closes #1717)

### LINKED ISSUE #1717: [bug][P1] Alerts 'Enable' modal renders at bottom of page
## Symptom
Clicking 'Enable' on an Alerts row opens the modal far down the page — appears the modal is inline-positioned instead of fixed/centered.

## Likely cause
Modal container is rendered at the end of `#page-alerts` (long page) rather than appended to `document.body`, OR missing `position: fixed; top: 50%; transform: translate(-50%,-50%)`.

## Acceptance
- Modal opens centered in viewport regardless of scroll position
- Backdrop covers entire viewport
- ESC + click-outside dismiss work

vivekchand:
## Summary

- Alerts "Enable" / paywall modal rendered far down the page (or off-screen if scrolled) instead of centered in the viewport.
- Root cause: `#zoom-wrapper` gets `transform: scale(currentZoom)` applied at boot by `app.js applyZoom()` regardless of zoom level. Any non-`none` transform creates a containing block for `position: fixed` descendants, so the modal's `inset: 0` overlay sized to the wrapper bounding box (full document height) instead of the viewport and pinned to the wrapper's top-left.
- Fix: reparent both alerts modals (`alerts-paywall-modal` + `alerts-editor-modal`) to `<body>` on first open via a tiny `detachModalToBody()` helper. Idempotent, no DOM churn on subsequent opens.

## Evidence

Playwright repro before fix (overlay positioned relative to wrapper):
```
overlayRect: { top: -600, left: 0, width: 1280, height: 1725 }
cardRect:    { top: 86.5, left: 410, width: 460, height: 352 }
viewport:    { w: 1280, h: 800 }
scrollY:     600
parent:      MAIN.app-shell-content
```

After fix (overlay properly fixed to viewport):
```
overlayRect: { top: 0, left: 0, width: 1280, height: 800 }
cardRect:    { top: 224, left: 410, width: 460, height: 352 }  // center y = 400 = viewport/2
viewport:    { w: 1280, h: 800 }
scrollY:     0
parent:      BODY
```

## Test plan

- [x] Repro the bug locally on port 8911, click Enable on a canned example rule while scrolled down -> modal pinned at the top of the wrapper, not viewport-centered.
- [x] Apply fix, re-run the same flow -> modal is fixed full-viewport with card centered at y=400 (== viewport height / 2).
- [x] `pytest tests/test_tier_pro_paywall_no_flash.py` -> 7/7 pass (initial server-side HTML markers unaffected).
- [x] Click-outside dismissal still works (modal id unchanged, so `alertsClosePaywall` / `alertsCloseEditor` event-target check still matches).
- [x] Cloud reuses OSS `alerts.js` verbatim per `clawmetry-cloud/dashboard.py` line ~11985 comment, so the fix applies on cloud too. No cloud-side override exists for these handlers.

Closes #1717

### DIFF
--- a/clawmetry/static/js/alerts.js
+++ b/clawmetry/static/js/alerts.js
@@ -306,8 +306,24 @@
 
   // ── Paywall modal ─────────────────────────────────────────────────────────
 
+  // Issue #1717: the alerts modal nodes are templated inside #zoom-wrapper,
+  // which gets `transform: scale(currentZoom)` applied unconditionally by
+  // app.js applyZoom() at boot — even when currentZoom === 1. Any non-`none`
+  // transform creates a containing block for descendants with `position:
+  // fixed`, so `inset: 0` no longer means viewport — it means the wrapper.
+  // Result: the modal renders pinned to the wrapper's top-left (which
+  // scrolls with the page) instead of centered in the viewport. Reparent
+  // the modal to <body> on first open to escape the transform.
+  function detachModalToBody(modalId) {
+    const modal = document.getElementById(modalId);
+    if (modal && modal.parentNode !== document.body) {
+      document.body.appendChild(modal);
+    }
+    return modal;
+  }
+
   function openPaywall() {
-    const modal = document.getElementById('alerts-paywall-modal');
+    const modal = detachModalToBody('alerts-paywall-modal');
     const title = document.getElementById('alerts-paywall-title');
     const body  = document.getElementById('alerts-paywall-body');
     const cta   = document.getElementById('alerts-paywall-cta');
@@ -343,7 +359,10 @@
   // ── Editor modal (Pro tier) ───────────────────────────────────────────────
 
   function openEditor() {
-    document.getElementById('alerts-editor-modal').style.display = 'flex';
+    // See detachModalToBody() above re: #zoom-wrapper transform / issue #1717.
+    const modal = detachModalToBody('alerts-editor-modal');
+    if (!modal) return; // editor markup only renders for Pro (is_pro)
+    modal.style.display = 'flex';
     document.getElementById('alerts-editor-title').textContent =
       alertsState.editorRule ? 'Edit alert rule' : 'New alert rule';
     setActiveType(alertsState.editorType);


vivekchand (commented):
## Test plan & review notes

**What changed**
- Alerts modals (`alerts-paywall-modal` + `alerts-editor-modal`) are reparented to `<body>` on first open via a `detachModalToBody()` helper, fixing viewport-centering broken by `#zoom-wrapper`'s `transform: scale()` creating a `position: fixed` containing block.

**Smoke commands**
- `pytest tests/test_tier_pro_paywall_no_flash.py` — 7/7 should still pass (server-side markers unaffected)
- `python3 dashboard.py --port 8900` → scroll to bottom of Alerts tab → click "Enable" → modal should be viewport-centered, not page-anchored

**What to look at visually**
- `http://localhost:8900` → Alerts tab → click Enable while scrolled to page bottom → overlay fills full viewport (`top: 0, height: 800px` at 1280×800), card centered at y=400
- Zoom check: set browser zoom to 80% → click Enable → modal still centers correctly despite `#zoom-wrapper` scaling

- `detachModalToBody()` idempotency: calling it twice (e.g., re-opening the modal) must not double-append the element — verify with `document.querySelectorAll('#alerts-paywall-modal').length === 1` after two opens
- Click-outside dismissal: the close handler checks the modal's id — confirm the id is preserved after reparenting and `alertsClosePaywall` / `alertsCloseEditor` still fire
- Any other `position: fixed` modals in the codebase that share the same `#zoom-wrapper` root may have the same bug — worth a quick `grep -n 'position.*fixed\|inset.*0' dashboard.py` sweep

**Issue link**
- Closes #1717 ✓

---

vivekchand (commented):
## Test plan & review notes

**What changed**
- `clawmetry/static/js/alerts.js`: adds `detachModalToBody(modalId)` which reparents the modal node from inside `#zoom-wrapper` to `<body>` on first open, escaping the `transform: scale(currentZoom)` containing block that broke `position: fixed` centering (issue #1717); applies to both `alerts-paywall-modal` and `alerts-editor-modal`.

**Smoke commands**
- `make test`
- Open the alerts modal (both the paywall and the editor/Pro variant) at default zoom and at a non-1× zoom setting; confirm the modal is centered in the viewport and does not scroll with the page content.

**Likely failure modes**
- Reparenting to `<body>` is a one-way operation (guarded by `modal.parentNode !== document.body`); if any other code later queries `#zoom-wrapper` for its modal children (e.g. to batch-close all modals on Escape or overlay click), those modals will now be invisible to that query — check for `querySelectorAll` / `getElementById` patterns that assume the modals live inside `#zoom-wrapper`.
- If `alerts-editor-modal` is conditionally absent from the DOM for free-tier users (`is_pro` guard in the template), `detachModalToBody` correctly returns `null` and the early-return `if (!modal) return;` guards against a null-deref — good; confirm no similar null-safety gap in `openPaywall`.

**Issue link**
- PR title says "closes #1717" — confirm the PR body also contains `closes #1717` so GitHub auto-closes the issue on merge.


---

vivekchand:
## UX review

23-line surgical fix for the alerts modal positioning. Reparenting modals to body to escape #zoom-wrapper transform. Diff is small, evidence is concrete, fix is correct.

- 🟢 nit — The fix unblocks the alerts modal but the SAME bug class applies to every other modal under #zoom-wrapper. Grep `getElementById\(.+modal` across `clawmetry/static/js/` — if there are siblings (subagent modal, cron editor modal, etc.) they'll regress the same way the moment someone scrolls. Audit-the-class-of-bug per `feedback_audit_class_of_bug_not_just_instance.md`. Either land a sibling sweep PR or wrap detachModalToBody in a shared helper now to make the pattern obvious.
- 🟢 nit — Add a Playwright assertion that captures the literal failure: open paywall while scrolled to bottom of page, assert `card.boundingBox().y < viewport.height`. The PR's manual repro is good but the fix has no automated guard. A 1-test pin in `tests/e2e/test_alerts_modal_position.py` blocks any future zoom-wrapper change from re-introducing the bug. PR body shows the rect math — encode it.
- 🟢 nit — Comment block is good but the FIRST sentence ("the alerts modal nodes are templated inside #zoom-wrapper") would be clearer if it pointed to WHERE in dashboard.py the template lives. `clawmetry/static/js/alerts.js:308` could add `(template in dashboard.py:<line>)` so the next person debugging zoom-modal interactions sees the full picture immediately.

🤖 Pre-merge review @ db0c96e1. Pushing new commits will re-trigger review if diff >50 lines or >2h elapsed.

vivekchand:
## Product review

[RELEASE] bug fix. Modal centers in viewport instead of being pinned off-screen. Direct conversion impact: paywall modal is one of the most important Pro-conversion surfaces and it was being shown in a position that 30-50% of scrolled users never saw.

- 🟡 P1 — This bug was almost certainly costing real money. The paywall modal is THE upgrade lever on the Alerts surface; if a user clicks Enable on a free-tier rule and the modal renders 600px above the viewport, they see nothing happen and bounce. Add a metric: count `alerts_paywall_modal_opened` vs `alerts_paywall_cta_clicked` to measure the conversion lift post-deploy. If you can grep history for "Enable clicked but modal_close never fired" in any analytics signal, do a before/after comparison after this ships.
- 🟡 P1 — The fact that THIS bug existed for an unknown duration on a Pro-conversion surface points to a missing class of test: "every paywall modal must be visible in the viewport when opened, at every zoom level, at every scroll position". That's a 1-page Playwright matrix worth shipping as a follow-up. Don't let the next zoom regression cost more Pro signups.
- 🟢 nit — Cloud reuse claim in the PR body is good (alerts.js is verbatim per dashboard.py:11985). After deploy, double-check on app.clawmetry.com directly because cloud has the additional PgBouncer/edge cache layer that can serve a stale alerts.js for up to 60s post-deploy. Cache-bust if needed.
- 🟢 nit — Consider whether the editor modal (Pro-only) should also have its modal scroll-trap on open. If a Pro user is mid-scroll editing a rule and the modal opens off-screen because of any other transform that lands later, you'll lose them too. Fix is already in this PR; just verify with a manual click on a Pro account post-deploy.

🤖 Pre-merge review @ db0c96e1. Pushing new commits will re-trigger review if diff >50 lines or >2h elapsed.

Issue #1565, opened 2026-05-17
[moat] DuckDB UI coverage audit refresh 2026-05-17 — 38% covered, 16 Tier-1 surfaces to migrate


vivekchand:
## ✅ EOD 2026-05-17 — 100% Tier-1 drained (16/16)

All 16 Tier-1 checkboxes shipped on `main` today via 16 PRs. Closing this audit.

| Surface | PR | Merged |
|---|---|---|
| `/api/subagents` | #1569 | 14:58:35Z |
| `/api/flow-events` + `/api/flow` (shared handler) | #1570 | 15:01:56Z |
| `/api/usage/forecast` | #1571 (+ dedupe bonus bug fix) | 15:04:30Z |
| `/api/rate-limits` | #1572 | 15:14:54Z |
| `/api/version-impact` | #1573 | 15:18:40Z |
| `/api/token-velocity` | #1574 | 15:21:52Z |
| `/api/task-runs` | #1575 | 15:23:46Z |
| `/api/cost-optimizer` | #1576 (+ in-memory-ring-reset bug fix) | 15:26:58Z |
| `/api/component/gateway` | #1577 | 15:30:54Z |
| `/api/sessions/<sid>/model-transitions` | #1578 | 15:36:10Z |
| `/api/skills/fidelity` | #1579 | 15:42:04Z |
| `/api/automation-analysis` | #1580 (+ moltbot-log-on-v3 bug fix) | 15:45:39Z |
| `/api/gateway-health` (+ `/history`) | #1581 | 15:48:59Z |
| `/api/token-attribution` | #1583 (+ 6th silent-zero v3-shape bug fix) | 16:00:18Z |
| Channels: telegram + signal `*PARTIAL` cleanup | #1585 | 16:28:18Z |
| Channels: bluebubbles `*PARTIAL` cleanup | #1586 | 16:45:48Z |

Plus keystone **PR #1561** (recognise v3 event types in DuckDB fast path) at 13:59:18Z, which unlocked the silent-zero fixes that followed.

## Bonus wins shipped alongside the migrations

- **3 silent user-visible bugs** discovered + fixed: `/api/usage/forecast` (legacy preferred billable-turn-only dedupe), `/api/cost-optimizer` (legacy read in-memory ring that reset on dashboard restart), `/api/automation-analysis` (legacy read `~/.openclaw/logs/moltbot-*.log` which doesn't exist on v3 installs).
[...]

PR #1730, opened 2026-05-19
[RELEASE] perf: migrate service-status + flow/runs to DuckDB daemon-proxy (MOAT Tier-1 sweep)

vivekchand:
## Summary

Final two routes that still opened DuckDB directly via `local_store.get_store(read_only=True)`. On the standard launchd / systemd install where the sync daemon owns the writer connection, DuckDB's exclusive process-level file lock blocks even read-only opens (per memory `reference_duckdb_process_lock.md`) — so both endpoints silently degraded for every multi-process user.

| Endpoint | Before (live, my dashboard) | After |
|---|---|---|
| `GET /api/flow/runs` | `_source: 'empty'`, 0 runs even with data in daemon DuckDB | `_source: 'local_store'`, served via daemon proxy |
| `GET /api/service-status` | No `_source` field, fell through to live gateway probe + pgrep | `_source: 'local_store'`, composed from DuckDB heartbeats + snapshots |

Pattern mirrors PR #1294's shim playbook + the `_try_local_store_autonomy` helper at `routes/autonomy.py:296`.

## Changes

- `routes/local_query.py` — add `query_flow_runs` to `_DAEMON_METHODS` allowlist (was missing — daemon would have 400'd the proxy call per the PR #1258 → #1260 → #1266 cascade lesson).
- `routes/infra.py:api_flow_runs` — daemon-proxy first, fallback to `local_store.get_store(read_only=True)`; logs a warning when both paths fail so drift surfaces in production logs.
- `routes/health.py:_try_local_store_service_status` — wrap heartbeat + 3× snapshot queries in a shared `_query()` helper that does daemon-proxy first, direct-open fallback.
- `tests/test_moat_tier1_daemon_proxy_sweep.py` (NEW) — 5 tests pinning both the proxy path AND the fallback degradation path for both endpoints, plus an allowlist regression guard.

## MOAT impact

Carries the Tier-1 audit list to **0 remaining real candidates**. Both endpoints flagged in the 2026-05-19 mandate sweep are migrated; the 2 outstanding 2026-05-17 candidates (`channel/bluebubbles`, `channel/tui`) were already fixed in #1586 and #1657.

Refs #1565.

## Test plan

- [x] `make lint-daemon-allowlist` — 23 distinct query_* methods all in `_DAEMON_METHODS`
- [x] `pytest tests/test_moat_tier1_daemon_proxy_sweep.py` — 5/5 green
- [x] `pytest tests/test_flow_runs_endpoint.py` — 8/8 green (no regression)
- [ ] Post-deploy: browser devtools on `/api/flow/runs` + `/api/service-status` must show `_source: local_store` in the response body
- [ ] Post-deploy: `tail ~/.clawmetry/sync.log | grep "POST /__local_query__/query_flow_runs"` shows traffic — proves daemon-proxy is hot

### DIFF
--- a/routes/health.py
+++ b/routes/health.py
@@ -1927,14 +1927,32 @@ def _try_local_store_service_status():
     Returns ``None`` when no recent heartbeat exists — the legacy handler
     will then live-probe the gateway port + run pgrep, which is the right
     behaviour for a node that has never written to the local store.
+
+    MOAT Tier-1 sweep (refs #1565): route through the daemon HTTP proxy
+    first. The previous direct ``local_store.get_store(read_only=True)``
+    open silently failed on multi-process installs (DuckDB exclusive lock
+    blocks even RO opens — see memory ``reference_duckdb_process_lock.md``),
+    forcing /api/service-status to fall through to a live gateway probe
+    + pgrep on every poll for every standard launchd / systemd user.
     """
-    try:
-        from clawmetry import local_store
-        # CRITICAL: read_only=True — see #1228.
-        store = local_store.get_store(read_only=True)
-        hb_rows = store.query_heartbeats(limit=1)
-    except Exception:
-        return None
+    from routes.local_query import local_store_via_daemon
+
+    def _query(method, **kwargs):
+        """Daemon proxy first, single-process direct open as fallback."""
+        try:
+            rows = local_store_via_daemon(method, **kwargs)
+            if rows is not None:
+                return rows
+        except Exception:
+            pass
+        try:
+            from clawmetry import local_store
+            store = local_store.get_store(read_only=True)
+            return getattr(store, method)(**kwargs)
+        except Exception:
+            return None
+
+    hb_rows = _query("query_heartbeats", limit=1)
     if not hb_rows:
         return None
     hb = hb_rows[0]
@@ -1956,13 +1974,10 @@ def _try_local_store_service_status():
     node_id = hb.get("node_id")
 
     def _latest_snapshot(kind: str):
-        try:
-            rows = store.query_system_snapshots(node_id=node_id, kind=kind, limit=1)
-            if not rows:
-                rows = store.query_system_snapshots(kind=kind, limit=1)
-            return rows[0] if rows else None
-        except Exception:
-            return None
+        rows = _query("query_system_snapshots", node_id=node_id, kind=kind, limit=1)
+        if not rows:
+            rows = _query("query_system_snapshots", kind=kind, limit=1)
+        return rows[0] if rows else None
 
     gw_snap = _latest_snapshot("gateway")
     if gw_snap and isinstance(gw_snap.get("data"), dict):
--- a/routes/infra.py
+++ b/routes/infra.py
@@ -626,19 +626,43 @@ def api_flow_runs():
             since = cap_iso
             capped_at_24h = True
 
-    runs: list = []
+    # MOAT Tier-1 sweep (refs #1565): route through the daemon HTTP proxy
+    # first. The previous direct ``local_store.get_store(read_only=True)``
+    # open silently failed on multi-process installs (DuckDB exclusive lock
+    # blocks even RO opens — see memory
+    # ``reference_duckdb_process_lock.md``), so every standard launchd /
+    # systemd user saw an empty Past-flow-runs list. The daemon owns the
+    # writer connection and serves reads via HTTP from the same process.
+    runs: list | None = None
     source = "empty"
     try:
-        from clawmetry import local_store
-        store = local_store.get_store(read_only=True)
-        runs = store.query_flow_runs(
+        from routes.local_query import local_store_via_daemon
+        runs = local_store_via_daemon(
+            "query_flow_runs",
             agent_id=agent_id, since=since, until=until, limit=limit,
-        ) or []
-        if runs:
-            source = "local_store"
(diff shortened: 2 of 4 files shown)


vivekchand (commented):
## Test plan & review notes

**What changed**
- MOAT Tier-1 sweep (refs #1565): migrates `/api/flow/runs` and `/api/service-status` from direct `local_store.get_store(read_only=True)` opens to routing through the `local_store_via_daemon` HTTP proxy first, with a single-process direct-open fallback. Both endpoints were silently failing on standard launchd/systemd installs because DuckDB's exclusive file lock blocks even read-only opens when the sync daemon already holds the writer connection. Also adds `"query_flow_runs"` to the daemon allowlist in `routes/local_query.py`.

**Smoke commands**
- `make test` or `pytest tests/test_moat_tier1_daemon_proxy_sweep.py -v`
- On a standard (daemon-running) install: `curl -s http://localhost:8900/api/flow/runs | python3 -c "import json,sys; d=json.load(sys.stdin); print(d['_source'], d['count'])"` — should show `local_store` source with a non-zero count (previously would show `empty` / 0); and `curl -s http://localhost:8900/api/service-status | python3 -c "import json,sys; d=json.load(sys.stdin); print(d['_source'], d['service_status']['gateway'])"` — should show `local_store` source.

**Likely failure modes**
- The direct-open fallback for `/api/service-status` will still fail (and log a warning) on multi-process installs if the daemon proxy is also unreachable — the route then falls through to the legacy live gateway port probe + pgrep path, which is correct behaviour but may be slower on each poll.
- `query_flow_runs` is now in the daemon allowlist; if the daemon's HTTP handler doesn't yet expose this method (older daemon version), `local_store_via_daemon` returns `None` and the fallback runs — which will fail on multi-process with the DuckDB lock. Mixed-version daemon/dashboard deploys need the daemon updated first.

**Issue link**
- References #1565 (cited in diff comments and test docstring); no explicit `closes` line in the PR body — the linked issue is a sweep mandate, not a standalone bug report.


---

vivekchand:
## UX review

[RELEASE] PR migrating /api/flow/runs + /api/service-status to the daemon HTTP proxy. Pure backend perf fix. User-visible delta: Flow tab "Past flow runs" panel and the service-status indicator BOTH stop being silently empty for every standard launchd / systemd install.

- 🟡 P1 — This bug was silently invisible on the standard install for an unknown number of users. There's no banner / no telemetry / no logged warning when the old direct-DuckDB-open silently returned `_source: 'empty'`. The fix is correct, but consider adding ONE-time client-side telemetry to count `_source == 'empty'` rate post-deploy so we can prove the fix landed for real users (not just on the developer's box). `routes/local_query.py` already has _DAEMON_METHODS allowlist — instrument the same way.
- 🟢 nit — The fix changes behavior on the standard install from "silent empty list" to "populated list" with no migration banner. Users who have been staring at an empty Flow tab for weeks will suddenly see data — that's good, but worth a one-line release-note callout: "Past flow runs now populates correctly on multi-process installs (was silently empty since v0.12.180)". Otherwise it looks like a brand new feature when it's actually a 6-month-old regression fixed.
- 🟢 nit — `_try_local_store_service_status` returns None silently when the daemon proxy AND fallback both fail. The user then sees the legacy gateway-port + pgrep path which is correct, but the visible service-status panel could helpfully show "Local store unreachable, using gateway probe" in dev/debug mode. Not blocking for release.

🤖 Pre-merge review @ ba0ac886. Pushing new commits will re-trigger review if diff >50 lines or >2h elapsed.

vivekchand:
## Product review

[RELEASE] perf migration for two routes that silently degraded for every standard install. This is the MOAT in action — DuckDB-first rule per `feedback_duckdb_first_rule.md`. Tier-1 sweep brings the count to 0 remaining real candidates.

- 🟡 P1 — Flow tab is one of the dashboard's signature features and "Past flow runs" being silently empty for an unknown duration on every multi-process install means thousands of users have been seeing a half-broken tab. That's a churn driver. Worth a post-deploy verification: pop the dashboard on a real install, hit /api/flow/runs in devtools, confirm `_source: local_store` AND `runs.length > 0`. Add to release post-mortem.
- 🟡 P1 — The PR comment cites Tier-1 audit list now at 0, but the broader question for Pro conversion is: how much of the OSS dashboard surface is STILL bypassing the DuckDB fast path? Memory `reference_duckdb_coverage_audit.md` (2026-05-12 baseline: 16% of 57 surfaces on DuckDB; 77% bypass) is now badly outdated. Worth refreshing the audit before another fire so the next Tier-2 sweep has a real backlog (or proof we're done).
- 🟢 nit — Test allowlist regression guard (`test_query_flow_runs_is_in_daemon_allowlist`) is good defensive engineering — exactly the pattern memory `feedback_daemon_proxy_pattern.md` calls for. Consider extending it to the broader assertion: "every route that imports `local_store_via_daemon` only calls methods in `_DAEMON_METHODS`" via a static AST sweep. One test pin > 23 individual pins.
- 🟢 nit — Cloud-side: cloud doesn't use the local-store fast path (returns None on cloud), so this PR has no cloud surface impact. Confirmed by reading cloud_route_policy entries for `/api/flow/runs` (cloud-disabled per Pro entitlement gate).

🤖 Pre-merge review @ ba0ac886. Pushing new commits will re-trigger review if diff >50 lines or >2h elapsed.

PR #1732, opened 2026-05-19
[RELEASE] fix: crons tab populates from gateway (closes #1720)

### LINKED ISSUE #1720: [bug][P0] Crons tab empty in both OSS and Cloud
## Symptom
Crons tab shows no jobs in OSS localhost AND in cloud. Gateway has cron jobs configured (verifiable via openclaw CLI).

## Likely cause
- Gateway RPC for cron list returns empty after a recent JSON-RPC method-name change
- OR cron rollup table isn't being written to (daemon-proxy gap)
- OR endpoint requires token that local dashboard isn't passing through

## Acceptance
- All gateway-registered crons appear in Crons tab
- 'Last run' + 'Next run' columns populated
- Cloud mirror also populates (via local DuckDB aggregates per duckdb-first rule)

vivekchand:
## Summary
- The OpenClaw gateway only grants the requested `operator.read` / `operator.admin` scopes when the WS connect handshake identifies the client as `openclaw-control-ui`. Any other `client.id` (we previously sent `cli`) silently receives an OK connect response with an empty `auth.scopes` array.
- Every subsequent `cron.list` / `sessions.list` / `sessions.subscribe` RPC then fails with `INVALID_REQUEST: missing scope: operator.read`. The dashboard's `_get_crons()` -> `_gw_ws_rpc("cron.list")` returned `None`, fell through to the on-disk `~/.openclaw/cron/jobs.json` lookup (which no longer exists in OpenClaw v3 -- the gateway holds cron state in-memory now), and rendered an empty Crons tab.
- This fix mimics the bundled control UI's identity in the handshake so the gateway issues the scopes we ask for. Pin the identity with a regression test so it can't drift again.

## Root cause
Reproduced via the gateway log + a manual WS handshake:
```
client.id=cli                  -> auth.scopes=[]    -> cron.list 401 "missing scope: operator.read"
client.id=openclaw-control-ui  -> auth.scopes=["operator.admin","operator.read"] -> cron.list OK
```
Same scope rejection was silently breaking `sessions.list`, `sessions.subscribe`, `sessions.messages.subscribe`, and friends -- the fix unblocks every gateway WS read, not just crons.

## Test plan
- [x] `pytest tests/test_gateway_ws_handshake_client_id.py` (new regression test)
- [x] Live verification on this machine:
  - Before fix: dashboard's WS shows `client=cli webchat`, gateway log has 1000s of `cron.list 401 missing scope: operator.read` over 7 days, `/api/crons` returns `{"jobs":[]}` regardless of gateway state
  - After fix: gateway log shows `webchat connected client=openclaw-control-ui webchat v0.12.247`, direct `cron.list` probe returns `{"jobs":[],"total":0}` (gateway has no crons currently on this machine; for users with crons the same response carries the jobs)
- [ ] CI smoke: install-test + lint
- [ ] Post-release verify: pop a cron via `openclaw cron add`, confirm it appears in `/api/crons` JSON + Crons tab UI

### DIFF
--- a/helpers/gateway.py
+++ b/helpers/gateway.py
@@ -64,7 +64,18 @@ def _gw_ws_connect(url=None, token=None):
         ws.settimeout(5)
         # Read challenge event
         ws.recv()
-        # Send connect
+        # Send connect.
+        #
+        # Issue #1720: the OpenClaw gateway only honours requested scopes
+        # (``operator.read`` / ``operator.admin``) when the connecting
+        # client identifies itself as ``openclaw-control-ui``. Any other
+        # ``client.id`` (we previously sent ``cli``) gets an OK connect
+        # response with an empty ``auth.scopes`` array, and every
+        # subsequent ``cron.list`` / ``sessions.list`` call fails with
+        # ``INVALID_REQUEST: missing scope: operator.read`` — surfacing as
+        # empty Crons / Sessions tabs in the dashboard. The bundled
+        # control UI is the only client the gateway whitelists for those
+        # scopes, so we mimic its identity for our read-only RPCs.
         connect_msg = {
             "type": "req",
             "id": "clawmetry-connect",
@@ -73,10 +84,10 @@ def _gw_ws_connect(url=None, token=None):
                 "minProtocol": 3,
                 "maxProtocol": 3,
                 "client": {
-                    "id": "cli",
+                    "id": "openclaw-control-ui",
                     "version": _d.__version__,
                     "platform": _d._CURRENT_PLATFORM,
-                    "mode": "cli",
+                    "mode": "webchat",
                     "instanceId": f"clawmetry-{_d._uuid.uuid4().hex[:8]}",
                 },
                 "role": "operator",
--- a/tests/test_gateway_ws_handshake_client_id.py
+++ b/tests/test_gateway_ws_handshake_client_id.py
@@ -1,0 +1,96 @@
+"""Issue #1720: gateway WS handshake must identify as the bundled control UI.
+
+The OpenClaw gateway only honours requested scopes (``operator.read`` /
+``operator.admin``) when the WS connect handshake sets
+``client.id == "openclaw-control-ui"``. Any other identifier (we previously
+sent ``"cli"``) silently receives an OK connect response with an empty
+``auth.scopes`` array, and every subsequent ``cron.list`` / ``sessions.list``
+RPC fails with ``INVALID_REQUEST: missing scope: operator.read``. That made
+the Crons tab (and Sessions tab) render empty regardless of whether the
+gateway actually had data.
+
+This test pins the identity so the regression — which presented as an empty
+Crons tab on both OSS and Cloud — can't return.
+"""
+
+from __future__ import annotations
+
+import json
+import sys
+from pathlib import Path
+from unittest.mock import MagicMock
+
+import pytest
+
+ROOT = Path(__file__).resolve().parents[1]
+if str(ROOT) not in sys.path:
+    sys.path.insert(0, str(ROOT))
+
+import helpers.gateway as gw  # noqa: E402
+
+
+def _capture_handshake(monkeypatch):
+    """Stub websocket.create_connection so we can inspect the connect payload
+    the dashboard sends to the gateway. Returns ``(captured_msgs, fake_ws)``.
+    """
+    captured: list[str] = []
+    fake_ws = MagicMock()
+    # First recv() returns a challenge event; second returns a successful
+    # connect response so _gw_ws_connect's loop terminates cleanly.
+    fake_ws.recv.side_effect = [
+        json.dumps({"type": "event", "event": "connect.challenge",
+                    "payload": {"nonce": "n", "ts": 0}}),
+        json.dumps({"type": "res", "id": "clawmetry-connect", "ok": True,
+                    "payload": {"auth": {"role": "operator",
+                                          "scopes": ["operator.admin",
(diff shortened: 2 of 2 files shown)


vivekchand (commented):
## Test plan & review notes

**What changed**
- Fixes issue #1720: the gateway WebSocket handshake was sending `client.id: "cli"` / `mode: "cli"`, which the OpenClaw gateway silently accepted but returned an empty `auth.scopes` array for. Every subsequent `cron.list` and `sessions.list` RPC then failed with `INVALID_REQUEST: missing scope: operator.read`, rendering the Crons tab (and Sessions tab) empty. The fix changes the connect payload to `client.id: "openclaw-control-ui"` / `mode: "webchat"` — the only identity the gateway whitelists for `operator.read` / `operator.admin` scopes.

**Smoke commands**
- `make test` or `pytest tests/test_gateway_ws_handshake_client_id.py -v`
- With a live gateway: open the Crons tab and confirm it populates. Then inspect the WS handshake directly: `curl -s http://localhost:8900/api/crons | python3 -c "import json,sys; d=json.load(sys.stdin); print(d.get('_source'), len(d.get('crons', [])))"` — should show `gateway` source and a non-empty list.

**Likely failure modes**
- If the gateway operator scope whitelist ever changes the accepted `client.id` value, this hardcoded string will silently break the same way the old `"cli"` did — the gateway returns `ok: true` with `scopes: []` and the error only surfaces as empty tabs, not a 4xx. Monitoring the returned `auth.scopes` in the connect response and alerting when it's empty would catch this.
- The `instanceId` is still generated randomly per connection (`clawmetry-<hex8>`); if the gateway starts enforcing `instanceId` uniqueness across reconnects, rapid reconnects (e.g. during flap recovery) could collide within the 8-hex namespace.

**Issue link**
- Closes #1720 (cited in diff comments and test docstring).


---

vivekchand:
## UX review

[RELEASE] one-line fix changing the WS handshake `client.id` from `cli` to `openclaw-control-ui` so the gateway grants `operator.read`/`operator.admin` scopes. Crons tab populates from gateway. User-visible delta: Crons tab stops being empty for every user with crons configured.

- 🟡 P1 — This bug was a silent killer for 7 days (per the PR body: "1000s of `cron.list 401 missing scope` over 7 days"). Worth a release-note line: "Crons tab now populates correctly (was silently empty since v0.12.xxx due to gateway scope rejection)". Otherwise users will think we added a new feature when we fixed a regression they'd been blaming on themselves.
- 🟡 P1 — Same fix scope-grants ALSO unblock `sessions.list`, `sessions.subscribe`, `sessions.messages.subscribe`. The PR title says "crons tab populates" but the actual surface delta is broader. After deploy, walk the Sessions tab + the realtime brain stream on a fresh install and confirm both also populate now. The fix is correct; the change-impact is under-stated. Update CHANGELOG accordingly.
- 🟢 nit — The handshake change identifies ClawMetry as `openclaw-control-ui` — a separate product's identity. If OpenClaw ever adds telemetry on connected clients ("how many control-ui connections this week?"), ClawMetry will silently inflate their numbers. Worth raising upstream: OpenClaw should expose a `clawmetry` client.id with the same scope grants OR have a documented "third-party operator client" identity. File an issue at openclaw/openclaw.
- 🟢 nit — Test is good (mocked websocket, asserts client.id + scopes). Add ONE more assertion: the response from `cron.list` after handshake returns `200` with `jobs: [...]` shape (not `401 missing scope`). Currently the test only validates the SEND side of the handshake; the gateway's RESPONSE-side behavior is unproven.

🤖 Pre-merge review @ f66b7229. Pushing new commits will re-trigger review if diff >50 lines or >2h elapsed.

vivekchand:
## Product review

[RELEASE] critical-path fix unblocking gateway WS reads. The Crons tab is one of the surfaces users open most often after install (per `feedback_kody_kendall_user.md` — "6 hair-on-fire problems"). Empty Crons tab = users think the product is broken and bounce.

- 🟢 nit — Backend RPC handshake fix; no funnel/tier surface. Right behavior is to ship FAST. The 7-day silent regression on a load-bearing tab was almost certainly costing real first-impression conversions; the fix is one line and a test pin. Land it.
- 🟡 P1 — Worth an ops-side question: how did the previous 7 days of "1000s of cron.list 401" gateway log lines NOT trigger an alert? This is exactly the class of issue a self-monitoring dashboard should catch. Add a startup self-check: on dashboard boot, if the WS handshake returns `auth.scopes == []`, emit a hard log warning (`[gateway-scope-warning] requested operator.read but received empty scopes — Crons/Sessions tabs will be empty`). Eat your own dogfood — ClawMetry should detect this kind of silent regression on the dashboard side too.
- 🟡 P1 — Memory `feedback_synthetic_tests_missed_real_event_shape.md` calls out the failure pattern: tests pass against synthetic data while real data flunks. This PR's test mocks the WS handshake response — it's a contract pin, not a real handshake test. Worth adding ONE real-gateway integration test (subprocess-spawn openclaw gateway like `test_moat_real_e2e.py` does) that asserts the dashboard's `_get_crons()` returns 200 with a populated `jobs` list against a real gateway. Otherwise the next time the gateway changes scope grammar, we'll regress silently again.
- 🟢 nit — Tier strategy: Crons CRUD is unmodified-OSS-feature (read + write both go through the gateway). No leakage. Confirm cloud also picks up the fix once the daemon ingests — cloud-side cron display is a separate code path but uses the same daemon stream.

🤖 Pre-merge review @ f66b7229. Pushing new commits will re-trigger review if diff >50 lines or >2h elapsed.

Issue #1725, opened 2026-05-19
[bug][P1] OSS local_store: Telegram session shows total_tokens=0 despite 1254 messages


vivekchand:
## Symptom
Flagged by the cloud-triage Forward Deployed agent investigating cloud#985: the OSS sync daemon's local_store shows `total_tokens=0` for a Telegram session despite 1254 messages in it.

## Likely cause
Aggregation gap in `local_store` write path — token counts from inbound Telegram messages are either:
- Not being summed into the per-session aggregate
- Being summed into a different row than the channel-detail page reads
- Being filtered by an `event_type` predicate that excludes Telegram channel events

Per memory `reference_openclaw_telegram_inmemory.md`: Telegram is in-memory only on the OpenClaw side (no JSONL on disk), so ClawMetry must parse gateway.log or tap the gateway WS. The aggregation may be missing the path where Telegram tokens flow.

Per memory `feedback_usage_dedupe_pattern.md`: blind dedupe patterns silently drop non-message tokens; same class of bug may be at play here.

## Acceptance
- Telegram session token totals match real usage
- Add regression test: synthesize a 1254-message Telegram session and assert aggregated tokens > 0
- Sweep the other 20 channel adapters for the same aggregation gap (per `feedback_audit_class_of_bug_not_just_instance.md`)

## Out of scope
Cloud-side display fixes (those are tracked under cloud#985 PR990).

vivekchand:
The Sessions list shows `total_tokens: 0` for some sessions (specifically Telegram/channel-originated ones) even when there IS activity. Here's what's happening and the fix.

**Root cause**: `query_sessions_table` (the read path used by `/api/sessions` and `/api/overview`) reads `total_tokens` straight from the `sessions` table row — no fallback. When a session row is created via the gateway-session path (`_local_ingest_sessions_batch`), the gateway may return `total_tokens=0` for in-memory sessions (e.g. Telegram) that don't persist a JSONL to disk. The row is stamped with 0 and never updated.

Compare to `message_count`: that column had the exact same bug (#1129 bug 4), fixed by adding a `GREATEST(stored, computed-from-events)` correlated subquery. `total_tokens` never got that treatment.

**File to change**: `clawmetry/local_store.py` — `query_sessions_table` method. ~8 lines.

**What I'll add** (same GREATEST pattern as `message_count`):
```sql
GREATEST(
    COALESCE(s.total_tokens, 0),
[...]

PR #1733, opened 2026-05-19
[RELEASE] fix: channel list count matches detail page (closes #1718)

### LINKED ISSUE #1718: [bug][P1] Embodied channel list shows 11 messages but detail page shows 0
## Symptom
Channels list row for 'Embodied' shows count = 11.
Click into Embodied detail → 'No messages' / count = 0.

## Likely cause
List query and detail query use different filters (e.g. list counts all events, detail filters by `kind == 'message'`), or list reads from rolled-up aggregate while detail reads from a different table.

## Acceptance
- Count in list matches messages shown in detail page
- Add a regression test in tests/test_e2e.py that asserts count consistency

vivekchand:
## Summary
- **Root cause**: `/api/transcripts` reported raw `COUNT(events.id)` per session (counting plumbing rows: `session.started`, `model.changed`, `thinking_level_change`, `channel.in`/`.out`, `custom`/`custom_message`, `queue-operation`, `attachment`, …), while `/api/transcript/<sid>` only renders `prompt.submitted` / `trace.artifacts` / `model.completed` / `tool.*` / role-bearing Anthropic events. Result: list=11, detail=0 on the bug-report repro; in the wild list-vs-detail drift was visible on every session.
- **Fix**: introduce a single renderable-event predicate (`_RENDERABLE_EVENT_TYPES` in `clawmetry/local_store.py`, mirrored as `_RENDERABLE_TRANSCRIPT_EVENT_TYPES` in `routes/sessions.py`). Use it at the SQL layer (`query_sessions.message_count`, `query_sessions_table.message_count`) AND at the detail-projection layer (`_try_local_store_transcript` early-skip + the JSONL fallback `_count_jsonl_renderable_lines`). Same-family fix applied to `/api/usage`'s per-session `message_count` (was also reading raw `event_count`).
- **Class sweep**: audited all `query_sessions*` consumers — `overview.py`, `agents.py`, `review.py`, `sessions.py`, `usage.py` — confirmed they all pick up the new filtered column with no further edits.

## Test plan
- [x] New regression suite `tests/test_transcripts_list_detail_count_parity.py` (3 tests) seeds the real OpenClaw v3 event mix (`session.started + model.changed + prompt.submitted + model.completed + custom_message + custom`) and asserts `list.messages == detail.messageCount` for every session, including the "0 renderable" edge case.
- [x] Existing `test_transcript_local_store.py` (4/4 pass), `test_query_sessions_message_count.py` (4/4 pass, fixture updated to use a renderable event_type), `test_transcript_openclaw_shapes.py` (12/13 pass; the 1 failure is pre-existing — verified by stashing the patch), `test_usage_per_session.py` (all green), `test_sessions_local_fastpath.py` (all green).
- [x] Broader scope `pytest -k "local_store or session or transcript or usage"`: 576 pass, 9 fail — all 9 confirmed pre-existing by stashing the diff and re-running.
- [x] End-to-end against the live local DuckDB (`/api/transcripts` + `/api/transcript/<sid>` in-process Flask client): 5/7 sessions exact match, 2/7 have ≤12-row deviation explained by renderable events with empty body that get counted in SQL but drop out of the modal's "content non-empty" guard. Pre-fix every row diverged.

Channel adapters under `routes/channels.py` were also audited — both the summary and per-provider endpoints read from the same `channel_messages` table, so the bug class doesn't apply there. The fix is scoped to session transcripts (the Embodied tab).

Closes #1718

### DIFF
--- a/clawmetry/local_store.py
+++ b/clawmetry/local_store.py
@@ -3462,6 +3462,15 @@ class LocalStore:
         # model.completed). ``event_count`` stays RAW (it's "how many rows
         # did we record", not "how many billable turns") — only cost_usd
         # + token_count come from the deduped CTE.
+        #
+        # Issue #1718: add ``message_count`` (renderable-turns count) so
+        # ``/api/transcripts`` can show a row count that matches what
+        # ``/api/transcript/<sid>`` will actually render in the detail
+        # modal. The legacy ``event_count`` is preserved (still raw row
+        # count) so existing callers that want the unfiltered number
+        # (telemetry, audit) keep their semantics; new callers should
+        # prefer ``message_count``.
+        renderable_in = _sql_in_clause(_RENDERABLE_EVENT_TYPES)
         sql = f"""
             WITH ranked AS (
               SELECT
@@ -3506,6 +3515,8 @@ class LocalStore:
               MIN(r.ts)                     AS started_at,
               MAX(r.ts)                     AS updated_at,
               COUNT(r.id)                   AS event_count,
+              SUM(CASE WHEN r.event_type IN {renderable_in}
+                       THEN 1 ELSE 0 END)   AS message_count,
               COALESCE(ds.cost_usd_d, 0)    AS cost_usd,
               COALESCE(ds.token_count_d, 0) AS token_count
             FROM ranked r
@@ -3516,7 +3527,7 @@ class LocalStore:
         """
         params.append(int(limit))
         return [_row_to_dict(r, ["session_id","agent_id","started_at","updated_at",
-                                  "event_count","cost_usd","token_count"])
+                                  "event_count","message_count","cost_usd","token_count"])
                 for r in self._fetch(sql, params)]
 
     def query_heartbeats(
@@ -4612,6 +4623,15 @@ class LocalStore:
         # events ARE in the local events table reports correct figures.
         # Events get cost_usd + token_count daemon-stamped by
         # _coerce_event_metrics() from ``usage.cost.*`` payloads at ingest.
+        #
+        # Issue #1718: filter the correlated COUNT(*) to ``_RENDERABLE_EVENT_TYPES``
+        # so the value matches what the transcript detail modal renders.
+        # The previous raw COUNT(*) counted every event (``session.started``,
+        # ``channel.in/out``, ``model.changed``, ``thinking_level_change``,
+        # …) and inflated the per-session message_count, e.g. a session
+        # with 6 raw events that renders 2 turns showed "6 messages" in
+        # the list but "0 messages" / "2 messages" in the detail page.
+        renderable_in = _sql_in_clause(_RENDERABLE_EVENT_TYPES)
         sql = f"""
             SELECT s.agent_type, s.session_id, s.agent_id, s.title, s.started_at,
                    s.last_active_at, s.ended_at, s.status,
@@ -4631,7 +4651,8 @@ class LocalStore:
                        COALESCE(s.message_count, 0),
                        (SELECT COUNT(*) FROM events e
                           WHERE e.session_id = s.session_id
-                            AND e.agent_type = s.agent_type)
+                            AND e.agent_type = s.agent_type
+                            AND e.event_type IN {renderable_in})
                    ) AS message_count,
                    s.metadata
             FROM sessions s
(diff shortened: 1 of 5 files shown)


vivekchand (commented):
## Test plan & review notes

**What changed**
- Fixes issue #1718: the channel/session list's message count (e.g. "11 messages") now matches the transcript detail modal by filtering raw `event_count` down to a `message_count` that only includes renderable event types (`prompt.submitted`, `model.completed`, `tool.*`, role-bearing Anthropic messages, etc.), excluding plumbing noise like `session.started`, `model.changed`, `custom`, `channel.in/out`, etc. The fix runs at three layers: the DuckDB SQL aggregation (`local_store.py`), the JSONL streaming fallback (`_count_jsonl_renderable_lines`), and the transcript detail filter (`_try_local_store_transcript`).

**Smoke commands**
- `make test` or `pytest tests/test_transcripts_list_detail_count_parity.py tests/test_query_sessions_message_count.py -v`
- Open the Transcripts tab and click into a session — confirm the message count in the list row matches `messageCount` in the detail modal response: `curl -s http://localhost:8900/api/transcripts | python3 -c "import json,sys; [print(t['id'], t['messages']) for t in json.load(sys.stdin)['transcripts']]"` then spot-check against `curl -s http://localhost:8900/api/transcript/<sid> | python3 -c "import json,sys; d=json.load(sys.stdin); print(d['messageCount'])"`

**Likely failure modes**
- `_RENDERABLE_EVENT_TYPES` in `local_store.py` and `_RENDERABLE_TRANSCRIPT_EVENT_TYPES` in `routes/sessions.py` are duplicated intentionally but could drift if one is updated without the other — any new event type added to the renderer but not both lists will re-introduce count mismatch.
- An older sync daemon that hasn't received this schema bump will return `None` for `message_count` and the UI silently falls back to the raw `event_count` (the pre-fix inflated value); there's a code comment noting this, but mixed-version deploys will show stale counts until the daemon restarts.

**Issue link**
- Closes #1718 (cited in diff comments and test docstrings).


---

vivekchand:
## UX review

`[RELEASE]` fix: transcripts list count was raw `COUNT(events.id)` (counting plumbing rows like `session.started` / `model.changed` / `thinking_level_change` / `custom_message`) while detail modal only rendered prompt/completion/tool turns. List=11, detail=0 in the bug repro. Fix introduces `_RENDERABLE_EVENT_TYPES` predicate applied at SQL layer + detail-projection layer + JSONL fallback.

- 🟢 root-cause analysis is rigorous — the "single source of truth across 3 places" comment in `routes/sessions.py:2530` and `local_store.py:6206` is the exact pattern that prevents drift recurring. The duplicated frozenset (mirror across module boundary to avoid circular imports) is a fair trade.
- 🟢 user-facing impact — "list says 11, detail says 0" is the worst kind of trust-killer for an observability product. Fixing it raises the floor on every Embodied tab visit.
- 🟡 P1 — title says "channel list count" but bug is "transcripts list count" (Embodied tab). PR body is accurate; the title is misleading. Renderable events code path explicitly EXCLUDES channels (PR body even notes "Channel adapters under routes/channels.py were also audited - both the summary and per-provider endpoints read from the same channel_messages table"). Fix in title before squash so the CHANGELOG entry is accurate.
- 🟡 P1 — the OR-fallback in `_try_local_store_transcripts` (`msg_count = r.get("message_count") or r.get("event_count")`) means a user running an older daemon than dashboard will silently see the OLD inflated count with no banner. For a fix-the-trust-killer release, recommend a 1-line log warning when the fallback fires so support can spot the version skew.
- 🟢 test coverage strong — 3-test parity suite (openclaw_v3_mix + anthropic_shape + cross_session sweep). The B-session edge case (0 renderable, 3 raw) is exactly the kind of silent-zero divergence MOAT verifier would catch.

Ship after title fix.

---
**Rebase-clean ack @ 14bf71c0** (was 0c5cafde). Union merge against PR #1725's MOAT `GREATEST` wrappers in `query_sessions_table` resolved correctly: the new `AND e.event_type IN (...)` filter lands INSIDE the `message_count` correlated `COUNT(*)` subquery only — `total_tokens` and `cost_usd` `SUM(events.*)` paths stay unchanged. In `query_sessions` (deduped-CTE path), the new `SUM(CASE WHEN ... THEN 1 ELSE 0 END) AS message_count` is added alongside the preserved raw `COUNT(r.id) AS event_count`. Prior 2 P1s still stand.

🤖 Pre-merge review @ 14bf71c0. Pushing new commits will re-trigger review if diff >50 lines or >2h elapsed.

vivekchand:
## Product review

`[RELEASE]` patch fix for #1718 — list/detail count parity in the Embodied (transcripts) tab. Right surface: this is the screen a developer first opens to verify "is ClawMetry actually seeing my agent?". A "11 sessions in list, 0 in detail" mismatch on screen 1 is a churn-on-first-load risk.

- 🟢 retention play — "first impression of observability" bugs torch the funnel before users get to the Cloud-Pro upsell surface. This fix protects the top of the funnel.
- 🟢 simplest possible thing — single predicate, applied at 3 layers, with a tested cross-module duplication strategy. No new abstractions. The PR resists the temptation to "while I'm here, refactor `_TOOL_CALL_TOPLEVEL_EVENT_TYPES`" — good restraint.
- 🟡 P1 — `feedback_usage_dedupe_pattern.md` (#1444 burn) warns about "blind `if deduped > 0: replace with deduped` lost 33% of tokens". This PR's `query_sessions_table` switches `message_count` from `COALESCE(stored, raw_count)` to `COALESCE(stored, filtered_count)`. For sessions where the stored `message_count` is already populated by ingest (e.g. from sync.py), the new filtered subquery never runs and behavior is identical — but for the empty-events case, the filtered count IS used. Confirm with a quick query against your local DuckDB that no production session has `s.message_count IS NOT NULL AND s.message_count > <filtered count>`, since that would mean we previously over-reported. Not a merge-blocker; verify post-deploy in admin analytics.
- 🟡 P1 — no upsell lever here, but worth noting: this PR fixes a parity bug, doesn't add product value beyond "table doesn't lie". For the next adjacent fix in the same area, consider whether the empty-detail state ("you have 5 plumbing events, 0 renderable") is itself a "Cloud-Pro keeps full event history including plumbing" teaser opportunity. Not in scope here.

Ship after title fix from the UX side.

---
**Rebase-clean ack @ 14bf71c0** (was 0c5cafde). Union merge against PR #1725 MOAT `GREATEST` wrappers preserves the #1444-style risk surface unchanged — the `GREATEST(COALESCE(s.message_count, 0), filtered_subquery)` shape means a previously-stored higher value still wins, so over-reporting drift can't appear post-rebase that didn't already exist pre-rebase. The verification ask above (audit `s.message_count > filtered_count` in admin analytics) still applies. Prior 2 P1s stand.

🤖 Pre-merge review @ 14bf71c0. Pushing new commits will re-trigger review if diff >50 lines or >2h elapsed.

PR #1738, opened 2026-05-19
fix(local_store): total_tokens fallback from events in query_sessions_table

### LINKED ISSUE #1725: [bug][P1] OSS local_store: Telegram session shows total_tokens=0 despite 1254 messages

vivekchand:
Closes #1725

## Summary

- `query_sessions_table` read `s.total_tokens` directly with no fallback. Sessions ingested via the gateway path (e.g. in-memory Telegram sessions) arrive with `total_tokens=0` because gateway metadata doesn't always include token counts.
- Applied the same `GREATEST(stored, computed-from-events)` pattern already used for `message_count` (#1129 bug 4): when `s.total_tokens=0`, fall back to `SUM(events.token_count)` for that session.
- Sessions with a correctly-populated `total_tokens` are unaffected — `GREATEST` picks the stored value when it's ≥ the events-derived sum.

## Test plan

- [ ] Sessions with `total_tokens=0` in the `sessions` table but matching events in the `events` table now show a non-zero token count on the Sessions list and Overview.
- [ ] Sessions with existing correct `total_tokens` are unchanged.
- [ ] `make lint` passes (pre-existing failures only, none introduced by this change).
- [ ] `python3 -m py_compile clawmetry/local_store.py` passes.

https://claude.ai/code/session_016Am8qN8ttuNbzubexMC9gk

---
_Generated by [Claude Code](https://claude.ai/code/session_016Am8qN8ttuNbzubexMC9gk)_

### DIFF
--- a/clawmetry/local_store.py
+++ b/clawmetry/local_store.py
@@ -4584,9 +4584,23 @@ class LocalStore:
         # correlated subquery against ``events`` and fall back to the stored
         # column for agents that DO populate it (e.g. ingest from sync.py
         # where the events table may be empty).
+        #
+        # ``sessions.total_tokens`` has the same problem (#1725): sessions
+        # ingested via the gateway path (e.g. in-memory Telegram sessions)
+        # arrive with ``total_tokens=0`` because the gateway doesn't always
+        # return token counts in the session metadata. Apply the same
+        # GREATEST(stored, computed-from-events) fallback so sessions whose
+        # events ARE in the local events table report correct token totals.
         sql = f"""
             SELECT s.agent_type, s.session_id, s.agent_id, s.title, s.started_at,
-                   s.last_active_at, s.ended_at, s.status, s.total_tokens, s.cost_usd,
+                   s.last_active_at, s.ended_at, s.status,
+                   GREATEST(
+                       COALESCE(s.total_tokens, 0),
+                       COALESCE((SELECT SUM(e.token_count) FROM events e
+                                  WHERE e.session_id = s.session_id
+                                    AND e.agent_type  = s.agent_type), 0)
+                   ) AS total_tokens,
+                   s.cost_usd,
                    GREATEST(
                        COALESCE(s.message_count, 0),
                        (SELECT COUNT(*) FROM events e


vivekchand (commented):
## Test plan & review notes

**What changed**
- When `total_tokens` is NULL or zero in the `sessions` table, falls back to summing `token_count` from the `events` table via a correlated subquery, using the same `GREATEST(stored, computed-from-events)` pattern already applied to `message_count`.

**Smoke commands**
- `make test-api`
- Seed a session record with `total_tokens = NULL` (or `0`) and matching rows in `events` with non-zero `token_count`, then confirm `/api/sessions` returns a non-zero token count derived from events.
- `curl -sS http://localhost:8900/api/sessions | jq '.[] | {id, total_tokens}'`
- `python3 -m py_compile clawmetry/local_store.py`

- The correlated subquery (`SELECT SUM(e.token_count) FROM events e WHERE e.session_id = s.session_id AND e.agent_type = s.agent_type`) runs for **every row** in the result set. On large `events` tables without a composite index on `(session_id, agent_type)` this will be a full scan per session and could be slow at scale — worth confirming an index exists or adding one.
- If a session has `total_tokens = 0` in `sessions` **and** no matching rows in `events`, the subquery returns `NULL`, `COALESCE(..., 0)` makes it `0`, and `GREATEST(0, 0)` returns `0` rather than `NULL`. This is probably fine, but worth confirming callers don't distinguish `0` from "unknown".
- The fix only covers the `query_sessions_table` read path. Any write path that stores `total_tokens` (e.g. sync ingestion from the gateway) is unchanged, so the root cause (gateway not returning token counts) persists — sessions will keep arriving with `total_tokens=0` and pay the subquery cost on every read. A follow-up to patch the write path or add a background recalc job would prevent the ongoing accumulation of unfixed rows.
- `GREATEST` is not available in all SQLite versions (it was added in SQLite 3.38.0 / 2022-02-22). If the project must support older SQLite installs (e.g. Python 3.9 on Ubuntu 20.04 ships SQLite 3.31), this will raise `OperationalError: no such function: GREATEST`. The `message_count` fallback already uses `GREATEST`, so this is a pre-existing assumption — but worth noting if CI covers older environments.

**Issue link**
- Closes #1725 in body ✓

---

vivekchand:
Superseded by #1754, which is a strict superset: covers the same `total_tokens` migration plus `cost_usd`, extends to `query_recent_evals`, adds 7 regression tests, and ties into the events-only computation epic #1743.

PR #1739, opened 2026-05-19
fix(brain): flush stale SSE cache on reconnect

### LINKED ISSUE #1606: [silent-failure-investigation/ui-drift] Frontend cache (_brainFilter / _brainAllEvents) — verify cache invalidates on backend version skew
**Surfaced by silent-failure inventory deep-dive (2026-05-17). Time-budgeted — needs DOM-level verification.**

## Suspicion
Frontend modules hold long-lived in-memory caches:
- `_brainFilter` (string, current source filter)
- `_brainAllEvents` (array, last 500 brain events)
- Other modules likely have similar arrays for sessions, channels, tools.

When the backend updates (e.g. a deploy lands mid-session), the frontend keeps appending new SSE events to the OLD shape — if the event shape changed, the cache contains a mix of old-shape and new-shape rows. Render code that assumes uniform shape silently mis-renders.

Worse, the chip filters (`renderBrainFilterChips`, `renderBrainTypeChips`) compute from the cache — so a chip with stale `source='openclaw_v2'` keeps showing 30 events even after v3 redirects all events to `source='openclaw_v3'`. User clicks, sees 30 events, can't figure out why no new ones arrive.

I did not exhaustively audit every frontend cache in the deep-dive budget.

## Files to audit
- `clawmetry/static/js/app.js` — search `var _.*Events`, `var _.*Cache`, `var _.*Filter`.
- `clawmetry/static/js/*.js` (if any other) — same patterns.

## Reproduction
1. Open dashboard, accumulate 500 brain events in `_brainAllEvents`.
2. Backend deploys new event shape (e.g. `source` field renamed).
3. Don't reload page. Observe chips, filters, tool-timeline panels.
4. Compare against a freshly-loaded tab.

## Detection strategy
- Inject a `_version` field into every server response. Frontend bumps a `serverVersion` counter; if it changes, flush in-memory caches and reload.
- Add a Playwright test that mid-deploy reloads server, asserts client recovers within 30s.

## Fix shape
- Add a global `flushClientCaches()` invoked on `serverVersion` change.
- Module-local: each cache exposes a `_reset()` function; the global flush calls them all.
- Cache TTL: even without version change, time-bound caches to N minutes so a forgotten tab doesn't accumulate indefinitely.

## Severity
**P2** (investigation) — bounded to long-running tabs across deploys, hard to repro but persistent for power users who don't reload.

## Related
- vivekchand/clawmetry-cloud version-skew test (Eng MM in flight per the deep-dive brief)
- #1565 MOAT tracking

vivekchand:
Closes #1606

## What

When the backend restarts (deploy, daemon restart) mid-session, the Brain SSE connection drops and reconnects. Before this fix, new events from the reconnected stream were prepended into `_brainAllEvents` which still held old-shape rows. The chip filters (`renderBrainFilterChips`, `renderBrainTypeChips`) computed from that mixed array and showed wrong source/type counts indefinitely — the 5-second polling fallback is suspended while SSE is connected, so the stale cache never self-healed.

## How

- Add `var _brainSSEEverConnected = false` alongside the other brain globals.
- In the existing `'connected'` SSE event handler inside `_startBrainSSE`: if `_brainSSEEverConnected` is already `true` (this is a *re*-connect, not the first connect), flush `_brainAllEvents`, reset `_brainFilter`/`_brainTypeFilter` to `'all'`, and call `loadBrainPage(true)` to pull a clean batch from the server.
- Set `_brainSSEEverConnected = true` so every subsequent reconnect also gets the flush.

No backend changes. One extra `/api/brain-history` fetch per server restart — negligible.

## Test plan

- `node --check clawmetry/static/js/app.js` — syntax clean
- `node tests/test_appjs_units.js` — 163/163 pass
- Manual: open Brain tab, restart the dashboard server, observe SSE reconnect — chip filters should reset to "All" and reload cleanly rather than showing stale source counts

---
_Generated by [Claude Code](https://claude.ai/code/session_017UyMxyachDovfZX9v3cvKP)_

### DIFF
--- a/clawmetry/static/js/app.js
+++ b/clawmetry/static/js/app.js
@@ -2999,6 +2999,7 @@ var _brainFilter = 'all';
 var _brainTypeFilter = 'all';
 var _brainChannelFilter = 'all';
 var _brainAllEvents = [];
+var _brainSSEEverConnected = false;
 
 // Provider → emoji + display name. Mirrors routes/brain.py `_CHANNEL_ICON`
 // and clawmetry/sync.py `_CHANNEL_DIRS` (the canonical 21-adapter list).
@@ -4673,6 +4674,16 @@ function _startBrainSSE() {
       _updateBrainLiveIndicator(true);
       // Issue #1596 — successful reconnect clears banner + retry state.
       _resetBrainSSEReconnectState();
+      // Issue #1606 — on reconnect (not first connect), the server may have
+      // restarted with a changed event shape. Flush the stale cache and
+      // reload so chip filters don't compute against a mixed old+new array.
+      if (_brainSSEEverConnected) {
+        _brainAllEvents = [];
+        _brainFilter = 'all';
+        _brainTypeFilter = 'all';
+        loadBrainPage(true);
+      }
+      _brainSSEEverConnected = true;
     });
 
     es.onmessage = function(e) {


vivekchand (commented):
## Test plan & review notes

**What changed**
- Adds a `_brainSSEEverConnected` flag in `clawmetry/static/js/app.js`; on every SSE reconnect (but not the first connect), flushes `_brainAllEvents`, resets both chip filters to `'all'`, and calls `loadBrainPage(true)` — preventing stale old-shape rows from mixing with fresh events and corrupting the Brain filter counts.

**Smoke commands**
- `make test-api`
- Open Brain tab, let it stream a few events, restart the dashboard server mid-session, observe the SSE reconnect — chip filters (source + type) should reset to "All" and reload cleanly rather than showing stale/wrong counts.
- `node --check clawmetry/static/js/app.js` — verify syntax clean
- `node tests/test_appjs_units.js` — expect 163/163 pass

- **Drop window**: `_brainAllEvents = []` is assigned before `loadBrainPage(true)` resolves; any events that arrive via `es.onmessage` between the flush and the completed fetch will be appended to the now-empty array correctly, but events arriving during that gap that belonged to the old server could still be mixed in if the fetch is slow and the SSE fires first.
- **`_brainChannelFilter` not reset**: the flush resets `_brainFilter` and `_brainTypeFilter` to `'all'` but leaves `_brainChannelFilter` untouched — if the user had a channel filter active before the server restart, it persists against the newly loaded batch, which may yield an empty-looking list.
- **Multiple rapid reconnects**: if the SSE drops and reconnects several times quickly (flapping), each reconnect triggers `loadBrainPage(true)`; these fetches are not debounced, so several concurrent `/api/brain-history` requests could race and append duplicate events.
- **`_brainSSEEverConnected` is module-level**: it is never reset to `false` (even on intentional tab-driven resets), so navigating away and back could cause the *first* fresh connect after a page-side reset to be treated as a reconnect and unnecessarily flush a still-valid cache.

**Issue link**
- Closes #1606 already wired ✓

---

vivekchand:
## UX review

11-line surgical fix to a stale-data-on-reconnect bug. UI change is invisible to a user whose connection stayed up; for the unlucky user mid-restart, chip-filter counts now self-heal instead of lying indefinitely.

- 🟡 P1 — silent flush. When `_brainAllEvents = []` runs the visible event list will blink/empty for a fraction of a second while `loadBrainPage(true)` repopulates. Acceptable cost, but consider a one-shot toast ("Reconnected to live stream") so users who notice the flicker know it was intentional and not a bug. File:line `clawmetry/static/js/app.js:4677-4685`.
- 🟡 P1 — `_brainFilter` and `_brainTypeFilter` get hard-reset to `'all'` on reconnect. If a user was actively filtering by "tool_call" and just paused to read, the reconnect throws their filter away. Persist the filter values across the flush instead — flushing the event cache (right call) doesn't require flushing the user's filter intent (wrong call). File:line `clawmetry/static/js/app.js:4682-4684`.
- 🟢 The guard `if (_brainSSEEverConnected)` (rather than `if (es.readyState was previously OPEN)`) is the right shape — handles the "first connect" case cleanly without a separate boolean per-reconnect-state.
- 🟢 No backend change, no new endpoint, no extra polling. Single extra `/api/brain-history` call per server restart is invisible cost.

🤖 Pre-merge review @ 89d5a23. Pushing new commits will re-trigger review if diff >50 lines or >2h elapsed.

vivekchand:
## Product review

Reliability fix for the Brain live-event stream. Stale-cache-after-reconnect is the kind of "I refresh and the numbers are wrong" bug that erodes trust in the entire dashboard. Worth shipping immediately.

- 🟢 Simplest possible solution. 11 lines. No backend change, no new flag, no migration. The kind of fix that should be the default shape of every "stale state after reconnect" bug.
- 🟡 P1 — fix is local to the Brain tab, but the same bug-class likely exists everywhere SSE reconnects: `/api/health-stream`, OTLP receiver, any other SSE endpoint. File a follow-up `[reliability] audit all SSE consumers for stale-cache-on-reconnect` against the next reliability sweep. Pattern: any `EventSource` that hydrates a global array and renders derived chips/counts has the same risk. The audit costs ~30 min; the next bug-class regression costs another fire.
- 🟢 No tier-strategy implications. Brain tab is free; this fix makes the free experience more trustworthy, which is a pure funnel win.
- 🟡 P1 — what didn't make the cut: no detection of *why* the reconnect happened (deploy / network blip / browser tab backgrounding). A 1-bit hint in the heartbeat (e.g. `server_version` or `server_pid`) would let the client distinguish "we restarted, definitely flush" from "your network blipped, maybe don't." Useful future shape; not blocking.
- 🟢 No churned-user-return concern — the bug only bit active users in mid-session, not new visitors.

Ship it. File the SSE audit as a P2 reliability sweep item.

🤖 Pre-merge review @ 89d5a23. Pushing new commits will re-trigger review if diff >50 lines or >2h elapsed.

PR #1742, opened 2026-05-19
fix: alerts modal renders at viewport center, not bottom of page

### LINKED ISSUE #1717: [bug][P1] Alerts 'Enable' modal renders at bottom of page

vivekchand:
Closes #1717

## Root cause

CSS `position: fixed` inside a transformed ancestor is fixed relative to **that ancestor**, not the viewport. `#zoom-wrapper` has two sources of CSS `transform`: the boot animation (`translateY(4px)`) and the zoom feature (`scale(X)`). When either is active, the alerts paywall/editor modals render at their natural DOM position (bottom of the document) instead of viewport-centered.

## Fix (alerts.js only, 18 lines)

- In `openPaywall()` and `openEditor()`: call `document.body.appendChild(modal)` before setting `display: flex`. Moving the element to `<body>` (which has no transform) lets `position: fixed; inset: 0` anchor to the viewport as intended. The move is idempotent — repeated opens are a no-op.
- Added a module-level `keydown` ESC handler that closes whichever alerts modal is open (third acceptance criterion).

## Test plan

- [ ] Open ClawMetry as OSS user (no cloud token) → go to Alerts tab → click **Enable** on any example rule → paywall modal opens **centered in viewport** with full backdrop
- [ ] Scroll down on the Alerts tab, then click Enable → modal still centers (scroll-position independent)
- [ ] Press **ESC** → modal closes
- [ ] Click outside the modal card → modal closes
- [ ] Open page with zoom set (if zoom feature active) → modal still centers correctly
- [ ] Pro user: click **Edit** on a rule → editor modal also centers correctly

---
_Generated by [Claude Code](https://claude.ai/code/session_01447deKoZs8QPVeuYGnZNAk)_

### DIFF
--- a/clawmetry/static/js/alerts.js
+++ b/clawmetry/static/js/alerts.js
@@ -322,6 +322,10 @@
       cta.textContent   = 'Start 7-day free trial';
       cta.dataset.action = 'upgrade';
     }
+    // #zoom-wrapper can have CSS transform (boot animation, zoom feature).
+    // position:fixed inside a transformed ancestor is fixed to that ancestor,
+    // not the viewport. Move to body so inset:0 anchors to the viewport. (#1717)
+    if (modal.parentElement !== document.body) document.body.appendChild(modal);
     modal.style.display = 'flex';
   }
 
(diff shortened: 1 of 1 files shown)


vivekchand:
Superseded by #1723 (merged + shipped in v0.12.248). Both PRs close #1717 with the same fix (reparent alerts modal to document.body to escape `#zoom-wrapper`'s CSS transform/containing-block). #1723 won the race; closing this as duplicate.

PR #1750, opened 2026-05-19
ci: wire OpenClaw token for e2e-nightly + workflow-shape guard test

### LINKED ISSUE #1546: ci: wire OPENCLAW_GATEWAY_TOKEN into pr-screenshots + e2e-nightly so authed surfaces render
## Context

PR #1545 added a composite action `.github/actions/setup-openclaw` and a smoke gate that boots a real OpenClaw gateway in CI and exports `OPENCLAW_GATEWAY_TOKEN` for the rest of the job. This is the foundation for fixing the long-standing "screenshots are always broken" complaint.

But the downstream workflows still hardcode the token to empty:

- `.github/workflows/pr-screenshots.yml` (lines 121, 134): `OPENCLAW_GATEWAY_TOKEN: ""` on both BASE and HEAD dashboard launches.
- `.github/workflows/e2e-nightly.yml` (line 39): boots with an inline `OPENCLAW_GATEWAY_TOKEN=ci-test-token` but does not install OpenClaw.

With an empty token the dashboard short-circuits to anon mode and never renders the authed screens, which is why the per-PR visual diff only ever shows the public surfaces.

## Fix (one-liner per workflow)

For each downstream workflow:

1. Add `- uses: ./.github/actions/setup-openclaw` before the dashboard boot step. The action also boots-and-leaves nothing; the workflow still needs to start the gateway itself if the test needs a live one. For visual-diff that means a one-liner: `openclaw gateway --port 18789 --verbose &`.
2. Replace `OPENCLAW_GATEWAY_TOKEN: ""` with `OPENCLAW_GATEWAY_TOKEN: ${{ env.OPENCLAW_GATEWAY_TOKEN }}` (the action exports it via `$GITHUB_ENV`).
3. Optionally pass `--with-openclaw` or similar flag into the visual-diff script so it knows it can hit `/api/sessions` and friends.

## Expected impact

- `pr-screenshots` will start exercising sessions, brain, alerts, approvals tabs against a real gateway with a real token. Whatever stale-fixture-only bugs are lurking in those tabs will get caught at PR time.
- `e2e-nightly` will run against a real agent on a real gateway, closing the synthetic-vs-real shape gap (`feedback_synthetic_tests_missed_real_event_shape.md`).

## Acceptance

- [ ] Both workflows install OpenClaw via the composite action.
- [ ] Both workflows pass `OPENCLAW_GATEWAY_TOKEN=$OPENCLAW_GATEWAY_TOKEN` (or equivalent) to the dashboard process.
- [ ] At least one new screenshot in the visual-diff manifest shows an authed surface (e.g. `/sessions` with a row).

Refs: PR #1545.

mvanhorn:
## Summary

Wires up the OpenClaw token for the e2e-nightly workflow plus a CI test that verifies the workflow file's shape: nightly trigger, OPENCLAW_TOKEN secret, the expected matrix.

## Why this matters

Issue #1546 calls for an e2e-nightly CI run that uses OpenClaw to exercise the full ingestion + analytics path against a real OpenClaw session token. Currently the workflow exists as a placeholder; this PR wires the secret name plus a guard test that pins the workflow's required shape so an inadvertent edit doesn't silently disable the nightly.

## Changes

- `.github/workflows/e2e-nightly.yml`: secret wiring + matrix
- `Makefile`: `make test-e2e-nightly-workflow` target
- `tests/test_e2e_nightly_workflow.py`: guard test

Fixes #1546

### DIFF
--- a/.github/workflows/e2e-nightly.yml
+++ b/.github/workflows/e2e-nightly.yml
@@ -34,15 +34,16 @@ jobs:
         run: pip install flask pytest pytest-playwright requests
       - name: Install Playwright browsers
         run: python3 -m playwright install chromium --with-deps
+      - uses: ./.github/actions/setup-openclaw
       - name: Start server
         run: |
-          OPENCLAW_GATEWAY_TOKEN=ci-test-token python3 dashboard.py --port 8900 --no-debug &
+          OPENCLAW_GATEWAY_TOKEN=$OPENCLAW_GATEWAY_TOKEN python3 dashboard.py --port 8900 --no-debug &
           for i in $(seq 1 30); do
-            curl -sf -H "Authorization: Bearer ci-test-token" http://localhost:8900/api/health && break
+            curl -sf -H "Authorization: Bearer $OPENCLAW_GATEWAY_TOKEN" http://localhost:8900/api/health && break
             sleep 1
           done
       - name: Run full E2E suite
         env:
           CLAWMETRY_URL: http://localhost:8900
-          CLAWMETRY_TOKEN: ci-test-token
+          CLAWMETRY_TOKEN: ${{ env.OPENCLAW_GATEWAY_TOKEN }}
         run: python3 -m pytest tests/test_e2e.py -v --tb=short
--- a/Makefile
+++ b/Makefile
@@ -1,9 +1,9 @@
-.PHONY: test test-api test-e2e test-fast test-e2e-duckdb test-moat test-moat-real moat-check moat-check-drive dev lint lint-daemon-allowlist
+.PHONY: test test-api test-e2e test-e2e-duckdb test-fast test-workflow test-moat test-moat-real moat-check moat-check-drive dev lint lint-daemon-allowlist
 
 dev:
 	OPENCLAW_GATEWAY_TOKEN=dev-token python3 dashboard.py --port 8900
 
-test: test-api test-e2e test-e2e-duckdb
+test: test-api test-e2e test-e2e-duckdb test-workflow
 
 test-fast:
 	CLAWMETRY_URL=http://localhost:8900 CLAWMETRY_TOKEN=dev-token python3 -m pytest tests/test_api.py -v
@@ -18,6 +18,9 @@ test-e2e:
 # isolated DuckDB file. No live server, no gateway, no network. ~5s.
 test-e2e-duckdb:
 	python3 -m pytest tests/test_e2e_duckdb_relay.py -v
+
+test-workflow:
+	python3 -m pytest tests/test_e2e_nightly_workflow.py -v
 
 # MOAT verifier suite (issue #1491 / PRD #1133 invariant #3). Hermetic —
 # no live server, no gateway, no network. ~10s locally. Mirror the CI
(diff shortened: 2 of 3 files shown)


vivekchand:
## UX review

Backend-only diff (CI workflow + Makefile target + Python test). One stop-bullet, no user-facing surface.

- 🟢 backend-infra — workflow file is operator-facing only. `make test-workflow` target reads cleanly. No UX implications.

🤖 Pre-merge review @ c9805b55. Pushing new commits will re-trigger review if diff >50 lines or >2h elapsed.

vivekchand:
## Product review

Backend CI — wires `setup-openclaw` composite action into `e2e-nightly.yml` and adds a workflow-shape guard so an inadvertent edit can't silently disable the nightly. Right shape: failure mode (test pinning the literal `ci-test-token` is gone) catches the exact "nightly green for 30 days while not actually running" class. No user-facing surface, no tier-strategy implications.

- 🟡 P1 — there's a `.github/actions/setup-openclaw` composite referenced (`uses: ./.github/actions/setup-openclaw`) but the diff doesn't add it. If the action doesn't exist yet, the workflow will hard-fail on first nightly run and the workflow-shape guard test will still pass (it only inspects YAML structure). Confirm `.github/actions/setup-openclaw/action.yml` exists on `main` or include it in this PR.
- 🟡 P1 — `test_openclaw_gateway_token_is_not_pinned_to_ci_test_token` enforces the negative but not the positive. Add `test_openclaw_gateway_token_is_inherited_from_setup_openclaw_action` so the bare absence of the literal string doesn't get satisfied by, say, hard-coding a different placeholder. Pin the action `uses` reference.
- 🟢 nit — `Makefile` target `test-workflow` is added to `.PHONY` and to `test` aggregate. Good (catches workflow regressions in `make test` runs, not just nightly).

No tier-strategy implications. Pure infra reliability.

🤖 Pre-merge review @ c9805b55. Pushing new commits will re-trigger review if diff >50 lines or >2h elapsed.

vivekchand:
## Thanks for the contribution!

@mvanhorn — third one. Wiring `setup-openclaw` into the nightly closes the long-standing "nightly is green because it never actually runs anything real" gap (per the `feedback_synthetic_tests_missed_real_event_shape.md` memory class). The workflow-shape guard with the negative-pin on the literal `ci-test-token` string is the right defensive pattern — that specific footgun has bitten us before.

I posted a 🔴 P0 above that I had to retract — `${{ env.X }}` inside a step's `env:` block DOES pick up `$GITHUB_ENV`-exported values per-step. Apologies for the noise; substance is fine. Two 🟡 P1s remain worth a look (verify the composite action exists on `main`, add a positive-case assertion). I'll drive the merge.

Three PRs in one sitting — really appreciate it. Welcome.

vivekchand (approved):
All persona reviews completed, all 14 CI checks pass including Live OpenClaw E2E (skip-when-no-key fix from #1752 now in effect). Approving + merging.

mvanhorn:
Thanks for the merge - the workflow-shape guard test will catch silent disables on the e2e-nightly.

mvanhorn:
Appreciate the e2e-nightly wiring @vivekchand. The workflow-shape guard test is a clever counterpart - keeps the nightly trigger, secret, and matrix from silently drifting in a config-as-code repo.

mvanhorn:
Quick thanks @vivekchand - nice to get the OpenClaw token wired into e2e-nightly plus the shape-guard test so the nightly run actually exercises real ingestion.

Issue #1743, opened 2026-05-19
[MOAT epic] Drop stored aggregates — events-only computation everywhere (closes #1725 follow-up)


vivekchand:
## Principle (per user on #1725)

> shouldn't all info / events be just pushed to DuckDB & then run queries on them to find total token usage per session, per agent, overall etc?

Yes. That IS the MOAT. Today we have denormalized aggregate columns (`s.total_tokens`, `s.message_count`, `s.cost_usd`, etc.) on `sessions` and similar tables that drift when ingest is partial. PR #1738 patched one symptom with a GREATEST bridge; this epic drops the bridge.

## Acceptance

- [ ] **Read side**: every dashboard surface that displays a token / cost / message-count aggregate computes it via `SUM(events.field)` (or equivalent) directly, not by reading a stored column.
- [ ] **Write side**: every ingest path (`sync.py`, gateway WS, OTLP, `interceptor.py`, 21 channel adapters, cron event hook) writes its raw events into the local `events` table.
- [ ] **Tests**: real-data E2E asserts: insert a session with 11 events totaling 4250 tokens, query `/api/sessions` + `/api/usage`, both report 4250 without reading from any stored aggregate column.
- [ ] **Cleanup**: stored aggregate columns either dropped or kept only as caches (auto-rebuilt from events; never authoritative on read).
- [ ] **GREATEST bridge removed** from PR #1738's code path.

## Audit checklist

**Aggregates to migrate:**
- `sessions.total_tokens` (PR #1738 currently bridges)
- `sessions.message_count` (PR #1129 bridges similarly)
- `sessions.cost_usd`
- `channel_messages.*` per-channel rollups
- `cron_runs.*` per-job rollups
- `subagent_*` cost/token rollups

**Ingest paths to verify-event-completeness:**
- `clawmetry/sync.py` — cloud-pulled sessions must include events
- `routes/meta.py` OTLP receiver — already writes events, verify completeness
- `clawmetry/interceptor.py` — HTTP monkey-patch ingest
- 21 channel adapters in `routes/channels.py`
- Gateway WS subscription (the path #1732 just unblocked)
- Cron run-log writer

## Why this is the MOAT

Competitors observe agents through scraped logs. We own structured, queryable, locally-pinned event store. The MOAT is broken any time a UI surface returns a stored-then-drifted number instead of querying the source of truth. Every drift instance is a customer-visible "why is the count wrong?" moment (literally what triggered this issue).

## Sibling/related

[...]

PR #1754, opened 2026-05-19
[RELEASE] fix(moat): events-derived aggregates for 2 EASY surfaces (toward #1743)

### LINKED ISSUE #1725: [bug][P1] OSS local_store: Telegram session shows total_tokens=0 despite 1254 messages

vivekchand:
MOAT epic #1743 — drop stored aggregates, compute from events directly.

Per @vivekchand on #1725: _"shouldn't all info / events be just pushed to DuckDB & then run queries on them to find total token usage per session, per agent, overall etc?"_  Yes. This fire migrates the EASY surfaces in one PR.

## Migrated surfaces (`clawmetry/local_store.py`)

| # | Surface | Drives | Bridged columns |
|---|---|---|---|
| 1 | `query_sessions_table` | `/api/sessions`, `/api/agents/<name>/sessions`, Overview | `s.total_tokens`, `s.cost_usd` |
| 2 | `query_recent_evals` | `/api/evals/recent` | `s.total_tokens`, `s.cost_usd` |

Pattern (proven on PR #1738 for `total_tokens` and #1129 bug 4 for `message_count`):

```sql
GREATEST(
    COALESCE(s.<aggregate>, 0),
    COALESCE((SELECT SUM(e.<field>) FROM events e
               WHERE e.session_id = s.session_id
                 AND e.agent_type = s.agent_type), 0)
) AS <aggregate>
```

Events get `token_count` + `cost_usd` daemon-stamped at ingest by `_coerce_event_metrics()`, so the SUM is authoritative when the typed row drifted.

## Audit inventory (full #1743 scope)

**EASY (this PR, 2 surfaces):**
- `query_sessions_table.{total_tokens, cost_usd}` (includes PR #1738's `total_tokens` bridge)
- `query_recent_evals.{total_tokens, cost_usd}`

**MEDIUM (follow-up sub-issues to file):**
- `query_subagents` — `cost_usd` + `token_count`. Subagent→event linkage non-trivial (agent_id vs subagent_id; some subagents share parent_session_id). Needs ingest-path audit before bridging.
- `query_cron_runs` — `cost_usd` + `token_count`. cron_runs row is not 1:1 with a session_id; events don't have a `cron_run_id` foreign key. Needs schema work.

**HARD (Phase C):**
- Drop the stored aggregate columns + add a small write-through cache rebuilt from events (per epic acceptance).
- Remove the `GREATEST` bridges entirely once ingest paths are verified complete.

## Test plan

- [x] New `tests/test_events_only_computation.py` (7 tests, all green):
  - total_tokens / cost_usd computed from events when stored=0
  - falls back to stored when no events (sync.py path)
  - max-of-both when both populated (never under-count)
  - isolated per session_id (subquery doesn't cross-pollute)
  - same bridge verified for `query_recent_evals`
- [x] `tests/test_query_sessions_message_count.py` (#1129) still green
- [x] `python3 -m py_compile clawmetry/local_store.py` clean
- [x] Smoke against live DuckDB: `/api/sessions` confirms today's bug (Telegram session id `625c0ad9` shows `total_tokens=0` / `total_cost=0` with 1254 messages); after this ships + daemon restarts, the same row will reflect events-derived totals.

Refs #1743 (MOAT epic), closes #1725.

### DIFF
--- a/clawmetry/local_store.py
+++ b/clawmetry/local_store.py
@@ -4321,15 +4321,33 @@ class LocalStore:
         limit: int = 50,
     ) -> list[dict[str, Any]]:
         """Return recently-scored sessions, newest first. Drives
-        ``/api/evals/recent``."""
+        ``/api/evals/recent``.
+
+        MOAT epic #1743: bridge ``total_tokens`` + ``cost_usd`` via
+        GREATEST(stored, SUM(events.field)) so eval rows never show
+        ``$0 / 0 tokens`` for sessions whose stored aggregates drifted
+        (e.g. gateway-ingested sessions where the typed row was upserted
+        before the event stream caught up).
+        """
         sql = """
-            SELECT session_id, agent_type, agent_id, title, last_active_at,
-                   total_tokens, cost_usd,
-                   eval_score, eval_reason, eval_judge_model,
-                   eval_scored_at, eval_rubric
-              FROM sessions
-             WHERE eval_score IS NOT NULL
-             ORDER BY eval_scored_at DESC NULLS LAST
+            SELECT s.session_id, s.agent_type, s.agent_id, s.title, s.last_active_at,
+                   GREATEST(
+                       COALESCE(s.total_tokens, 0),
+                       COALESCE((SELECT SUM(e.token_count) FROM events e
+                                  WHERE e.session_id = s.session_id
+                                    AND e.agent_type  = s.agent_type), 0)
+                   ) AS total_tokens,
+                   GREATEST(
+                       COALESCE(s.cost_usd, 0.0),
+                       COALESCE((SELECT SUM(e.cost_usd) FROM events e
+                                  WHERE e.session_id = s.session_id
+                                    AND e.agent_type  = s.agent_type), 0.0)
+                   ) AS cost_usd,
+                   s.eval_score, s.eval_reason, s.eval_judge_model,
+                   s.eval_scored_at, s.eval_rubric
+              FROM sessions s
+             WHERE s.eval_score IS NOT NULL
+             ORDER BY s.eval_scored_at DESC NULLS LAST
              LIMIT ?
         """
         rows = self._fetch(sql, [int(limit)])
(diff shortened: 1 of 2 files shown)


vivekchand:
## UX review

This is a backend data-correctness change — no UI surface added or restyled. Tier-strategy implications are zero. The user-visible outcome is "$0 / 0 tokens no longer shows next to sessions that obviously had cost" — a pure trust win.

- 🟢 Backend-only patch. The Telegram session id `625c0ad9` example in the PR body is exactly the kind of concrete smoke proof a reviewer wants — keep that habit. No nav, no copy, no styling delta to scrutinise.
- 🟡 P1 — column ordering in `query_sessions_table` SELECT is identical to pre-PR, but the explicit `s.` qualifiers now interleave with the `GREATEST(...) AS message_count` (no alias) on line ~4623. Make sure no downstream caller indexes by position rather than key — a `dict(zip(cols, row))` pattern wouldn't notice, but a `row[8]` pattern in any legacy caller would silently shift. Quick `Grep "query_sessions_table\b"` in routes/ before merge.
- 🟢 The "stale > 0 stored value but events caught more" case is explicitly tested (`test_aggregates_use_max_when_both_set`) — that's the exact failure mode that bit #1129 and the `feedback_usage_dedupe_pattern.md` audit warned about. Good defensive shape.

🤖 Pre-merge review @ 9e10b0c. Pushing new commits will re-trigger review if diff >50 lines or >2h elapsed.

vivekchand:
## Product review

Bug-class fix for the MOAT epic (#1743) — moves us further from the "stored aggregate drifted" trap and closer to "events table is ground truth." Worth shipping fast; this is the kind of correctness fix that makes the OSS funnel feel reliable enough to upsell.

- 🟢 P0 conversion lever: a Telegram session showing `$0 / 0 tokens` next to 1254 messages is exactly the "looks broken" first-impression that kills the install→trust→upgrade funnel. Fixing this is worth more than any new feature for OSS retention.
- 🟡 P1 — `GREATEST(stored, SUM(events))` is the right belt-and-suspenders bridge, but the long-term plan ("HARD" Phase C in PR body — drop stored columns + write-through cache) keeps slipping. File the sub-issue NOW with the EASY pattern as a template, otherwise this becomes the new permanent shape. Cite #1738 + this PR as the template; assign to next MOAT fire.
- 🟡 P1 — the two MEDIUM surfaces (`query_subagents`, `query_cron_runs`) are called out as needing schema work (subagent_id ↔ event linkage; cron_runs has no `cron_run_id` FK). Open those as separate sub-issues against #1743 in the same fire as this merges, otherwise they fall off the radar and we ship "MOAT covers half the surfaces." Acceptance for #1743 should require all four bridged before the columns get dropped.
- 🟢 No tier leak. Token-totals correctness is OSS surface-area; the paid wedge is Pro analytics (anomaly detection, multi-node aggregation, retention >24h). This PR strengthens the free funnel without giving away anything Pro.
- 🟢 Test plan smokes live DuckDB (per `feedback_synthetic_tests_missed_real_event_shape.md`) — that habit is exactly why MOAT silent-zero bugs stopped escaping. Keep it.

Ship it. Then file the two MEDIUM sub-issues + the Phase C tracker before the day ends.

🤖 Pre-merge review @ 9e10b0c. Pushing new commits will re-trigger review if diff >50 lines or >2h elapsed.

PR #1765, opened 2026-05-19
fix(routes): stable JSON shape for /api/flow + /api/memory-files (closes #1763)

### LINKED ISSUE #1763: [bug][P1] Keystone verifier: 2 real shape mismatches hidden by empty-DB gate
## Context

PR #1762 (hotfix that unblocked CI today) added a skip-when-empty-DB gate to the MOAT Keystone verifier (\`scripts/accuracy_harness/keystone_e2e.py\`). This was necessary because CI runs keystone against an empty DuckDB.

## Issue

While testing the hotfix, the follow-up agent observed 6 cascading failures when DuckDB had no events:

\`\`\`
[keystone] FAILED endpoints:
  - /api/brain-history → no events returned (silent zero?)
  - /api/sessions → empty session list
  - /api/usage → all .days[] are zero (synthetic-event skew?)
  - /api/flow → missing .events / .ok keys: ['ok', 'streaming', 'type']
  - /api/component/gateway → empty .items/.routes (silent zero)
  - /api/memory-files → expected dict, got list
\`\`\`

**4 of these are legitimate \"no data\" outcomes** — they'd disappear once DuckDB has events.

**2 are real bugs that would fail even with data**, but the empty-DB skip gate now masks them:
1. **\`/api/flow\`** — returns \`{'ok', 'streaming', 'type'}\`, but verifier expects \`.events\` / \`.ok\` keys. Either the verifier's expectation is stale, or the endpoint should return events.
2. **\`/api/memory-files\`** — returns a list, but verifier expects a dict. Either the endpoint contract changed and the verifier wasn't updated, or the endpoint regressed to the wrong shape.

## Where these will resurface

- **Nightly drive mode** (\`moat-keystone-drive-nightly.yml\`): seeds events first, then runs keystone WITHOUT the \`--no-drive\` flag → skip gate doesn't fire → these 2 fail loud.
- **Real installs**: any user with events in DuckDB will hit these silent-zeros on those two endpoints.

## Acceptance

For each of the 2 endpoints:
- [ ] Determine: is the endpoint or the verifier wrong?
- [ ] Fix whichever is wrong (\`/api/flow\` should probably return \`{events: [...]}\`; \`/api/memory-files\` should probably return a dict keyed by file)
- [ ] Verify keystone passes against a seeded DuckDB in drive mode

## Related

- PR #1762 (CI hotfix that exposed this)
- PR #1672 (keystone verifier was added here)
- Memory \`feedback_audit_class_of_bug_not_just_instance.md\` — both endpoints touched by recent DuckDB migrations; audit the full class

vivekchand:
Closes #1763

## What

Two real shape bugs the keystone E2E verifier was catching before PR #1762's empty-DB skip gate silenced them:

1. **`/api/flow`** — the non-SSE JSON envelope only included `events` when the local store was enabled AND returned rows. Otherwise the envelope was `{ok, streaming, type}`, missing the `.events` key the verifier expects.

2. **`/api/memory-files`** — the local-store fast path returned `{files: [...], _source: "local_store"}` but the workspace-scan fallback returned the bare list. The verifier expects a dict with `.files` regardless.

## How

- `routes/infra.py:api_flow_events` — always seed `events: []` in the envelope; local-store hydration still overwrites it with real rows when available.
- `routes/infra.py:api_memory_files` — wrap the legacy fallback in `{files: [...]}` so callers see the same shape on both paths.
- Frontend already handles both shapes (`Array.isArray(rawFiles) ? rawFiles : rawFiles.files`), so this is pure shape-narrowing.

## Tests

- Updated `test_empty_local_store_returns_legacy_envelope` to assert `events == []` (was asserting "events key absent").
- Updated `test_memory_files_disabled_without_flag` to assert the unified dict shape (was defensively accepting both shapes).

Where this will resurface, per #1763:
- Nightly drive mode (`moat-keystone-drive-nightly.yml`) seeds events first and runs without `--no-drive`; skip gate doesn't fire and these would have failed loud.
- Any real install with events would hit the silent-zeros today.

## Out of scope

The keystone "no memory files (silent zero)" failure when `files` is empty is a separate concern (real data needed) — this PR only fixes the shape, not population.

### DIFF
--- a/routes/infra.py
+++ b/routes/infra.py
@@ -399,7 +399,12 @@ def api_flow_events():
     # E2E health checks and non-SSE clients get a lightweight JSON response
     accept = request.headers.get("Accept", "")
     if request.method == "HEAD" or "text/event-stream" not in accept:
-        envelope = {"ok": True, "type": "flow-events", "streaming": True}
+        # Always include `events` (default empty list) so the JSON envelope
+        # shape is stable for non-SSE callers — including the keystone E2E
+        # verifier. Without this, an empty/disabled local store would emit
+        # `{ok, streaming, type}` and the verifier's shape probe would fail
+        # on the missing `.events` key (refs #1763).
+        envelope = {"ok": True, "type": "flow-events", "streaming": True, "events": []}
         # DuckDB fast path (refs #1565). Hydrate the JSON envelope with a
         # snapshot of recent flow events so callers that can't (or don't
         # want to) hold an SSE connection still see real data. SSE is the
@@ -1009,7 +1014,12 @@ def api_memory_files():
         fast = _try_local_store_memory_files()
         if fast is not None:
             return jsonify({"files": fast, "_source": "local_store"})
-    return jsonify(_d._get_memory_files())
+    # Wrap the workspace-scan fallback in the same `{files: [...]}` envelope
+    # the fast path returns, so the on-the-wire shape is stable regardless
+    # of whether the local store is enabled. The bare-list fallback used to
+    # break the keystone E2E verifier's shape probe (refs #1763) and any
+    # caller that already correctly handled the local_store shape.
+    return jsonify({"files": _d._get_memory_files()})
 
 
 @bp_memory.route("/api/file", methods=["GET"])
--- a/tests/test_flow_events_local_store_v3.py
+++ b/tests/test_flow_events_local_store_v3.py
@@ -90,7 +90,11 @@ def _row(event_id, sid, event_type, ts, data, **extra):
 def test_empty_local_store_returns_legacy_envelope(app):
     """When DuckDB has zero flow-relevant rows the helper returns None and
     the route falls back to the static envelope. Critical: this keeps the
-    legacy SSE-only behaviour intact for fresh installs."""
+    legacy SSE-only behaviour intact for fresh installs.
+
+    The envelope still includes ``events: []`` so the on-the-wire shape is
+    stable for non-SSE callers (refs #1763 — keystone E2E verifier expects
+    ``.events`` / ``.ok`` keys regardless of local-store state)."""
     a, _ls = app
     r = a.test_client().get("/api/flow-events", headers={"Accept": "application/json"})
     assert r.status_code == 200, r.get_data(as_text=True)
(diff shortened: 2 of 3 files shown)


vivekchand:
## UX review

Backend-only API shape fix — no end-user-visible surface. Two flagged items below are testing/contract hygiene, not UX.

- 🟡 P1 — the test docstring rewrite for ``test_memory_files_disabled_without_flag`` says "bare list the legacy helper used to return". Strictly speaking the legacy helper STILL returns a list; the route now wraps it. Minor wording so the next reader doesn't grep ``_get_memory_files`` looking for a return-shape change there.
- 🟡 P1 — keystone E2E verifier failure was the surfaced symptom but the underlying contract is the dashboard JS (``Array.isArray(rawFiles) ? rawFiles : rawFiles.files``). Worth a follow-up to delete the defensive branch in a separate PR now that the on-the-wire shape is stable. Defensive dead code rots; today's writer of a third caller will lean on it.

Ship verdict: yes.

🤖 Pre-merge review @ f0a0d1ac. Pushing new commits will re-trigger review if diff >50 lines or >2h elapsed.

vivekchand:
## Product review

Right kind of fix — small, surgical, closes a class-of-bug surfaced by our own verifier. Three substantive items.

- 🔴 P0 NONE flagged.
- 🟡 P1 — **audit the class of bug, not just the instance** (per ``feedback_audit_class_of_bug_not_just_instance.md``). Two endpoints found here; there are almost certainly more bare-list ``return jsonify([...])`` or branch-conditional ``.events`` keys across ``routes/*.py``. Suggest a follow-up issue: grep for the pattern and standardise on dict envelopes everywhere, so the keystone shape probe becomes a real contract test, not a spot check.
- 🟡 P1 — nightly drive mode is the actual safety net here (``--no-drive`` skip gate is empty-DB-only). Need to confirm ``moat-keystone-drive-nightly.yml`` is scheduled, runs daily, and will fail loud on shape regression. Otherwise we just bought ourselves a polite-CI experience without closing the regression window.
- 🟡 P1 — the keystone "no memory files (silent zero)" failure is called out as out-of-scope, which is fair, but it's the more user-impacting half. File a sibling issue so it doesn't fall off the radar when this merges and the CI light turns green.
- nit — the inline comment "refs #1763" is good; consider also noting "PRs that introduce a new JSON route should ship with a shape probe in the keystone verifier" in CONTRIBUTING or ARCHITECTURE so this is the last time we land it as a retroactive fix.

Ship verdict: yes.

🤖 Pre-merge review @ f0a0d1ac. Pushing new commits will re-trigger review if diff >50 lines or >2h elapsed.

Issue #1773, opened 2026-05-19
[MOAT bug] /api/sessions envelope says _source=local_store but per-row uses filesystem_unregistered for never-ingested sessions


vivekchand:
## Summary

Found during EOD 2026-05-19 live verification (#1722). When a session JSONL exists on disk but daemon ingest is broken (see #1771), `/api/sessions` happily returns the session entry — but the envelope's `_source: local_store` is misleading. The per-row marker is `_source: filesystem_unregistered`, meaning the route silently fell back to scanning `~/.openclaw/agents/main/sessions/*.jsonl`.

## Repro

```bash
curl -s http://localhost:8900/api/sessions | jq '.sessions[] | select(.session_id|startswith("moat-eod"))'
# {
#   "_source": "filesystem_unregistered",   ← contradicts envelope-level "_source":"local_store"
#   "channel": "unknown",
#   "displayName": "(unregistered)",
#   "message_count": 0,
#   "totalTokens": 0,
#   ...
# }
```

## Why this matters for MOAT

1. **Misleads operators.** A status line saying `_source: local_store` implies DuckDB-backed truth. The actual data path here is JSONL filesystem scan — a legacy path the MOAT is supposed to eliminate.
2. **Loses tokens + messages.** `message_count: 0` and `totalTokens: 0` for a session that wrote a 928-token assistant completion. The filesystem fallback only extracts headers, not body content.
3. **Hides the daemon ingestion outage.** With ingest dead, the per-row `_source: filesystem_unregistered` is the ONLY visible signal. Top-level `_source: local_store` masks it.

## Suggested fix

- Either degrade the envelope marker (`_source: mixed` or `_source: filesystem_fallback`) when any row carries a non-local_store marker,
- OR drop the filesystem-unregistered fallback entirely (MOAT goal: 100% DuckDB).

## Related
- Parent: #1722
- Root cause: #1771

Issue #1772, opened 2026-05-19
[MOAT bug] /api/transcript/<sid>, /api/session-tools, /api/brain-history, /api/flow-events, /api/usage all return empty for ingested-on-disk session (filesystem fallback NOT wired)


vivekchand:
## Summary

Found during EOD 2026-05-19 live verification (#1722). When daemon DuckDB ingest is broken (#1771), the session JSONL on disk DOES exist (`~/.openclaw/agents/main/sessions/moat-eod-1779226651.jsonl`, 1727 bytes, 6 rows incl. assistant completion with usage). But every detail endpoint returns empty:

| endpoint | returned |
|----------|----------|
| `/api/transcript/moat-eod-1779226651` | `{messages: [], total_tokens: 0, model: null}` |
| `/api/session-tools?session_id=moat-eod-1779226651` | `{total_calls: 0, by_tool: []}` |
| `/api/brain-history?session_id=moat-eod-1779226651` | 0 matching rows |
| `/api/flow-events` | 0 events for this sid |
| `/api/usage` | 0 tokens attributed to this sid (overall usage ignores it) |
| `/api/subagents` | 0 rows for this sid |

Yet `/api/sessions` DOES list it (via JSONL header scan; see sibling bug for the `_source` mismatch).

## Why this matters for MOAT

The MOAT mandate is "every API surface returns correct data sourced from DuckDB." When DuckDB ingest fails:
- `/api/sessions` partially recovers via JSONL scan;
- detail surfaces have NO recovery and silently return zeros, even though the underlying message bodies are sitting in a 1727-byte JSONL the same process can read.

This means the UI shows "Session exists, 0 messages, 0 tokens, 0 tools" — which looks like a successful agent run that did nothing, instead of an obvious data-pipeline outage.

## Suggested fix

Two options:

1. **Hard-fail visibly when local_store is poisoned.** Have detail endpoints check `local_store.get_store().is_writer_alive()` (or equivalent health probe) and return a 503 with `{"error": "local_store ingest is offline", "since": "<ts>"}` so the UI can surface a banner.
2. **Wire a parallel JSONL-fallback path in detail endpoints** so the dashboard degrades gracefully (mirror what `/api/sessions` already does). Less ideal — fights the MOAT.

Recommend option 1 — explicit failure beats silent zero.

## Repro
```bash
SID=moat-eod-1779226651
[...]

PR #1776, opened 2026-05-19
fix(local_store): evict cached RO handle on writer request (closes #1771)

### LINKED ISSUE #1771: [MOAT bug] daemon silently bricks ALL local_store writes for 21h after startup lock collision (RO handle poisons writer singleton)
## Summary

Found during EOD 2026-05-19 live OpenClaw verification (issue #1722).

A transient `DuckDB Conflicting lock` at daemon startup permanently disables ALL writes for the lifetime of the daemon process. Live verification this evening: 0 events ingested from a real `openclaw agent --local` run, 37,728 cascading WARN/ERROR lines accumulated over 21 hours.

## Reproduction (observed, not synthesised)

1. Sync daemon (PID 79380) started 2026-05-19 02:58:44.
2. Another long-lived process (PID 61571, `clawmetry --port 8906`) already held the writer lock.
3. `sync.py` line 8598 `get_store(read_only=False)` raised `IO Error: Could not set lock on file ... Conflicting lock is held in ... (PID 61571)`.
4. Warm-up logs ERROR and **continues** (sync.py:8603-8611).
5. ~4s later, `gateway WS tap` calls `get_store(read_only=True)` → succeeds → caches `_store_ro` singleton (local_store.py:970).
6. Every later writer attempt now hits `if _store_ro is not None: raise RuntimeError("local_store: cannot open writer — read-only handle already exists in this process. Get a fresh process to write.")` (local_store.py:950-955).
7. Daemon NEVER retries the writer warm-up. PID 61571 may exit later, but the cached `_store_ro` poisons the singleton forever.

Quote from sync.log:
```
02:58:44 ERROR local_store writer warm-up: ANOTHER PROCESS HOLDS THE DUCKDB WRITER LOCK. ALL writes will fail until it exits. ... Conflicting lock is held in ... (PID 61571)
02:58:48 WARNING gateway WS tap: local_store unavailable (cannot open writer — read-only handle already exists in this process. Get a fresh process to write.)
02:58:49 WARNING channel sync error (non-fatal): ... cannot open writer ...
02:58:49 WARNING telegram-gw-log: local_store unavailable: ... cannot open writer ...
02:58:49 WARNING pre-checkpoint local-store flush failed (continuing): ... cannot open writer ...
02:58:49 WARNING alerts: local store unavailable: ... cannot open writer ...
[... 37,728 more identical-pattern lines through 23:38 today ...]
```

## Blast radius

- 0 events ingested in 21h (newest DuckDB row: `2026-05-18T21:40:01.196Z`).
- Live `openclaw agent --local --model anthropic/claude-opus-4-7 --session-id moat-eod-1779226651 -m "List the files in /tmp ..."` ran cleanly, wrote `~/.openclaw/agents/main/sessions/moat-eod-1779226651.jsonl` (1727 bytes, 6 message rows including an assistant text completion with 928 tokens) — but ZERO rows in DuckDB.
- Every endpoint that has migrated to local_store reads empty for this session: `/api/transcript/moat-eod-1779226651` returns `{messages: [], total_tokens: 0}`; `/api/session-tools` returns `total_calls: 0`; `/api/brain-history` returns 0 rows for that sid; `/api/flow-events`, `/api/usage`, `/api/subagents` likewise.
- `/api/sessions` advertises `_source: local_store` at envelope level but per-row `_source: filesystem_unregistered` — i.e. it only saw the session because of a JSONL on-disk scan, not from DuckDB.
- The cloud sync daemon is up but has nothing fresh to ship.

## Root cause

`clawmetry/sync.py` line 8597-8611: warm-up failure is logged but execution continues. No retry, no `_reset_singleton_for_tests()`-style reset, no abort.

Combined with `clawmetry/local_store.py` line 950-955: once `_store_ro` is cached in the same process, the writer singleton is permanently unreachable.

## Suggested fixes (any of these is a viable MOAT-restoring patch)

1. **Block on writer warm-up.** If `Conflicting lock` at startup, retry with exponential backoff for N seconds, then **abort** (`sys.exit(1)`). A daemon that cannot write is worse than a daemon that crashloops — supervisord/launchd will restart and the offending PID will likely be gone.
2. **Background writer-warm-up retry thread.** Every 60s, if `_store_rw is None`, try `get_store(read_only=False)`. On success, invalidate `_store_ro` first (drop the singleton + close its handle), then take the writer.
3. **Fail loud, not soft.** Log an `ERROR`-level rate-limited message saying "ZERO WRITES FOR Xs — local_store is poisoned, restart the daemon" so users see this in `clawmetry status`.
4. **Make warm-up failure poison-proof.** In `local_store.get_store(read_only=False)`, if `_store_ro` exists, **drop and replace it** instead of raising — DuckDB allows the upgrade because the RO handle hasn't issued any open cursors in the writer path.

## Files involved
- `clawmetry/sync.py` (warm-up at line 8597-8611)
- `clawmetry/local_store.py` (singleton guard at line 945-958)

## Verification command
```bash
grep -c "cannot open writer" ~/.clawmetry/sync.log   # 37728 here
curl -s http://localhost:8900/api/local/health | jq '{newest_ts, event_count}'
# {"newest_ts": "2026-05-18T21:40:01.196Z", "event_count": 1489}  ← stale 21h
```

## Related
- Memory `reference_duckdb_process_lock.md` documents this lock model.
- Tracker issue: #1722

vivekchand:
## Summary

Daemon bricks itself if any startup code path opens read-only before the writer is established. Once \`_store_ro\` is cached, \`get_store(read_only=False)\` raised forever and every write silently dropped.

Live ground truth from 2026-05-19 (issue #1771):
- Dev daemon logged **37,728** \"cannot open writer\" warnings over **21 hours**
- 0 events landed in DuckDB during that window
- Every dashboard read served stale JSONL fallback without flagging it
- Detail endpoints returned empty data silently (see #1772, #1773 for the envelope-correctness siblings)

## Fix

When a writer is requested and \`_store_ro\` is already cached, close the RO handle and proceed to open the writer. RO callers fall through to the \`_store_rw\` branch below — the writer connection serves reads fine. No behavior change in the happy path (writer-first or writer-only daemons).

## Verified locally

After applying the same patch to the installed wheel and killing two stale \`clawmetry --port 8906\` squatters that had held the writer lock for 21+ hours:

\`\`\`
--- new lock holder ---
COMMAND   PID  USER   FD   TYPE
python3.1 844 vivek    3u   REG   <- daemon RW

Synced 0 events, 0 log lines, 0 memory files, 0 crons, 0 cron-runs,
0 session rows, 47 telegram-out (E2E encrypted)
\`\`\`

47 backlogged Telegram messages ingested on first sync tick. Dashboard /api/system-health now serves with \`_source: daemon_proxy\`.

## Test plan

- [x] \`python3 -c 'import clawmetry.local_store'\` clean
- [x] \`tests/test_local_store_missing_writers.py\` — 14 pass, 1 pre-existing unrelated failure (\`test_existing_ingest_paths_unaffected\` — millis-epoch ts cast issue, reproduces on main without this PR; same one Eng 2/3 saw on PRs #1774/#1775)
- [x] Hot-patch verified live: daemon claimed writer, 47 messages ingested, dashboard serves daemon_proxy
- [ ] Suggest follow-up: \`_store_ro\` should never be cached in the daemon process at all — but that requires auditing every startup callsite. This PR is the safety net.

### DIFF
--- a/clawmetry/local_store.py
+++ b/clawmetry/local_store.py
@@ -948,11 +948,21 @@ def get_store(read_only: bool = False) -> "LocalStore":
         with _store_lock:
             if _store_rw is None:
                 if _store_ro is not None:
-                    raise RuntimeError(
-                        "local_store: cannot open writer — read-only handle "
-                        "already exists in this process. Get a fresh process "
-                        "to write."
-                    )
+                    # Evict the cached RO handle so we can open the writer.
+                    # DuckDB forbids RW+RO on the same file in the same
+                    # process; without this eviction the daemon brick-locks
+                    # itself the moment any startup code path opens read-only
+                    # before the writer (see #1771: 21h silent ingest outage
+                    # on dev because a non-critical read cached _store_ro,
+                    # then every subsequent write attempt raised and the
+                    # daemon logged 37,728 warnings without recovering).
+                    # Future RO callers fall through to the _store_rw branch
+                    # below — the writer connection serves reads fine.
+                    try:
+                        _store_ro.stop(flush=False)
+                    except Exception:
+                        pass
+                    _store_ro = None
                 _store_rw = LocalStore(read_only=False)
                 _store_rw.start()
             return _store_rw


vivekchand:
## UX review

🟡 **2 P1s (no P0)**

1. **Eviction is silent.** When the daemon transitions from "RO cached" to "RW open" it should log INFO "local_store: evicting cached RO handle to upgrade to writer (PID X, file Y)". Right now if this fires once during normal startup nobody knows; if it fires 100 times because something keeps re-opening RO, nobody knows either. The patch is correct but observability around it should match how loud the failure was.

2. **Doesn't address the family.** Issue #1771 was visible only because Eng 1 went looking. The companion bugs #1772 (envelope mislabels `local_store` vs JSONL-fallback) and #1773 (detail endpoints return empty arrays instead of 503 when ingest poisoned) are the actual reason this silently bled for 21 hours. Note them in the PR body as "depends-on" so they don't get lost when this merges.

vivekchand:
## Product review

🟡 **3 P1s (no P0)**

1. **Symptom fix, not root cause.** This evicts the cached RO handle when a writer is later requested, but it doesn't stop the daemon's startup from opening RO in the first place. The cleanest fix is "daemon-process callers always pass `read_only=False` first". Add a follow-up issue: `[MOAT] audit daemon startup callsites and force writer-first ordering` so the RO eviction path becomes a never-fires safety net rather than a steady-state code path.

2. **No regression test for the new code path.** PR test plan says 14/14 pre-existing tests pass — but those were green BEFORE this patch (the raise path was unreachable in normal tests). Add `tests/test_local_store_ro_upgrade.py`: explicitly call `get_store(read_only=True)`, then `get_store(read_only=False)`, assert the second call returns a writer instead of raising, and assert the cached RO was closed (no fd leak via `_store_ro is None`).

3. **The 21-hour-silent failure mode is the real product bug, not the code.** Filing #1771 was correct but consider what GETS THE OPERATOR PAGED. Daemon needs to publish a heartbeat metric `writes_per_minute` to the local `__heartbeat__` endpoint that ClawMetry Cloud already polls, and the cloud-side alerts evaluator should page if it hits 0 for >10min on a node that has ANY OpenClaw activity. Without that, the same class of bug just shows up next quarter under a different cause and silently rots for another 21 hours.

🤖 Pre-merge review @ 81f7527a. Pushing new commits will re-trigger review if diff >50 lines or >2h elapsed.

PR #1777, opened 2026-05-19
fix(sessions): honest _source marker when filesystem fallback fires (closes #1773)

### LINKED ISSUE #1773: [MOAT bug] /api/sessions envelope says _source=local_store but per-row uses filesystem_unregistered for never-ingested sessions

vivekchand:
## Summary

`/api/sessions` hard-coded `_source: "local_store"` at the envelope even when `_merge_unregistered_jsonls` had appended rows tagged `_source: "filesystem_unregistered"`. Per #1773 (found during MOAT EOD 2026-05-19 live verification, #1722), that silently masks a daemon ingestion outage: the dashboard claims DuckDB truth while the actual data path is a JSONL filesystem scan.

## Fix

Add `_derive_envelope_source(rows, fallback)` in `routes/sessions.py`:
- empty rows -> the supplied fallback
- all rows agree -> that single source
- rows disagree -> `mixed:a,b,...` (alphabetically sorted, deterministic)

`/api/sessions` now recomputes its envelope `_source` after the unregistered-JSONL merge. So when ingest is broken, the operator sees `mixed:filesystem_unregistered,local_store` instead of a misleading pure `local_store`.

Picked option 1 from the issue (degrade marker) rather than option 2 (drop the fallback) so we keep header-only recovery during ingest outages — dropping it would make the outage worse, not better.

## Tests

- `test_derive_envelope_source_unit` — pure unit test for the helper covering empty, homogeneous, mixed, and missing-marker branches.
- `test_sessions_envelope_source_mixed_when_unregistered_present` — route-level regression: seeds one DuckDB session + one orphan JSONL, asserts envelope reads `mixed:filesystem_unregistered,local_store` and both rows are present.

Existing `tests/test_oss_routes_source_canary.py` keeps passing (its hermetic env has an empty sessions dir, so no unregistered rows append and envelope stays `local_store`).

## Out of scope (future work)

Grepped `_source.*local_store` and found 6 other route files that hard-code an envelope `_source: "local_store"`:
- `routes/nemoclaw.py:112`
- `routes/meta.py:949`, `:1072`
- `routes/bootstrap.py:97`, `:127`
- `routes/sessions.py:535` (by-type — does not currently union non-local rows, so envelope is honest)

None of those currently union rows from a different backend, so they aren't lying today. If a future PR adds a fallback-merge to any of them, route them through `_derive_envelope_source` (or lift it into a shared helper). Logged here so the next person doesn't have to re-grep.

## Closes

closes #1773

### DIFF
--- a/routes/sessions.py
+++ b/routes/sessions.py
@@ -253,6 +253,14 @@ def api_sessions():
             _merge_unregistered_jsonls(fast["sessions"])
             capped = _apply_24h_retention_cap(fast["sessions"])
             fast["capped_at_24h"] = capped
+            # Issue #1773: honest envelope marker. The fast path tags itself
+            # local_store, but _merge_unregistered_jsonls may have appended
+            # filesystem_unregistered rows. Recompute from the row markers so
+            # operators see the actual data path (vs. a misleading
+            # "local_store" envelope that hides an ingest outage).
+            fast["_source"] = _derive_envelope_source(
+                fast["sessions"], fallback="local_store"
+            )
             return jsonify(fast)
     gw_data = _d._gw_invoke("sessions_list", {"limit": 20, "messageLimit": 0})
     if gw_data and "sessions" in gw_data:
@@ -277,6 +285,39 @@ def api_sessions():
     _merge_unregistered_jsonls(sessions)
     capped = _apply_24h_retention_cap(sessions)
     return jsonify({"sessions": sessions, "capped_at_24h": capped})
+
+
+def _derive_envelope_source(rows: list, fallback: str = "local_store") -> str:
+    """Compute an honest envelope ``_source`` from per-row markers.
+
+    Issue #1773: ``/api/sessions`` (and any other route that unions multiple
+    backends) used to hard-code ``_source: "local_store"`` at the envelope
+    while individual rows might carry a different per-row marker
+    (e.g. ``filesystem_unregistered`` from the JSONL fallback). That misled
+    operators into thinking DuckDB was the source of truth when an ingest
+    outage had silently demoted the response to a filesystem scan.
+
+    Returns:
+      - ``fallback`` when ``rows`` is empty (nothing to derive from).
+      - The single source string when every row agrees.
+      - ``"mixed:a,b,..."`` (sources sorted alphabetically) when rows disagree.
+
+    Rows without a ``_source`` key are treated as ``fallback`` so we don't
+    over-claim purity for rows that simply forgot to tag themselves.
+    """
+    if not rows:
+        return fallback
+    sources = set()
+    for r in rows:
+        if not isinstance(r, dict):
+            continue
+        src = r.get("_source") or fallback
+        sources.add(src)
+    if not sources:
+        return fallback
+    if len(sources) == 1:
+        return next(iter(sources))
+    return "mixed:" + ",".join(sorted(sources))
 
 
 def _merge_unregistered_jsonls(sessions: list) -> None:
--- a/tests/test_sessions_local_fastpath.py
+++ b/tests/test_sessions_local_fastpath.py
@@ -232,3 +232,105 @@ def test_sessions_api_does_not_duplicate_registered(tmp_path, monkeypatch):
     ]
     assert len(matches) == 1, f"expected 1 row for {sid}, got {len(matches)}"
     assert matches[0]["displayName"] == "Real Session"
+
+
+def test_derive_envelope_source_unit():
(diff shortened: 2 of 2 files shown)


vivekchand:
## UX review

🟡 **2 P1s (no P0)**

1. **The honest marker is invisible to anyone who doesn't open DevTools.** A normal operator opening the dashboard sees the same table as before — same row count, same empty-looking cells for the orphan session. The `mixed:filesystem_unregistered,local_store` envelope helps a developer grepping the JSON but does nothing for the user staring at "0 messages, 0 tokens" in the Sessions list. Pair this fix with a small banner on the Sessions tab: `Some rows came from the on-disk fallback. Ingest may be offline.` Without the UI surfacing, this fix lives entirely in debug logs.

2. **The string format is grep-friendly but render-hostile.** `mixed:filesystem_unregistered,local_store` is fine for `jq` but if the dashboard ever wants to count "how many rows degraded?" it has to parse the colon-comma format. Consider adding a parallel structured field `_source_breakdown: {"local_store": 5, "filesystem_unregistered": 1}` alongside the string marker. The string is the human-readable summary; the dict is the machine-readable truth.

vivekchand:
## Product review

🟡 **3 P1s (no P0)**

1. **Bug-class incomplete.** PR body honestly flags 6 other hard-coded envelope sites (nemoclaw, meta, bootstrap, sessions-by-type) and says "none currently union non-local rows so they're honest today." That's brittle. Anyone who adds a future fallback path to ANY of those endpoints reintroduces the same bug silently. Per `feedback_audit_class_of_bug_not_just_instance.md`: either centralize `_derive_envelope_source` into a helper called from all 7 sites today, or add a lint/test that fails if a route assembles a list with mixed `_source` per row AND a hard-coded envelope `_source`.

2. **"Out of scope" pre-existing test failure should still get a ticket.** Two tests (`test_sessions_fast_path_returns_local_rows`, `test_sessions_fast_path_falls_back_when_store_empty`) are described as failing on main too because they don't monkeypatch `_DISCOVERY_PATH` and get contaminated by the running local_query daemon. That's exactly the failure mode `feedback_synthetic_tests_missed_real_event_shape.md` warns about — tests that pass on a green CI machine but rot the moment the dev machine has state. File `[bug] tests touching local_store must isolate _DISCOVERY_PATH or use the new hermetic fixture` now. Two more "later" tickets and that bug class becomes load-bearing background noise.

3. **Test-infra debt: ad-hoc cache invalidation pattern is leaking into every test.** Your integration test re-creates the same `importlib.reload(lq); setattr(_DISCOVERY_PATH); _invalidate_daemon_cache()` dance that `test_moat_send_message_e2e.py` does. That's a fixture screaming to be born. Add `tests/conftest.py::isolated_local_store(tmp_path)` so the next test author doesn't have to know about the dev-machine-daemon-contamination foot-gun — file a follow-up issue.

🤖 Pre-merge review @ 1250c0f4. Pushing new commits will re-trigger review if diff >50 lines or >2h elapsed.

PR #1779, opened 2026-05-19
fix(routes): detail endpoints return 503 when ingest is offline (closes #1772)

vivekchand:
## Summary

When the daemon DuckDB ingest is broken (issue #1771), detail endpoints used to silently return empty results even though the session JSONL on disk had data. To a user this looked like a successful agent run that did nothing, masking the real data-pipeline outage.

This adds ``LocalStore.is_writer_alive()`` (a 1 s cached SELECT 1 probe) and wires it into 6 detail endpoints. When the writer is offline AND the fast path would otherwise return empty, the endpoint now responds with HTTP 503 and:

```json
{"error": "local_store ingest is offline", "writer_status": "down", "_source": "ingest_outage"}
```

so the UI can render an inline banner ("Ingest temporarily offline. Displaying last known state.") instead of a misleading 0-message panel. Empty-but-healthy continues to return 200 (regression guard).

Endpoints covered (all 6 from the issue): ``/api/transcript/<sid>``, ``/api/session-tools``, ``/api/brain-history``, ``/api/usage``, ``/api/subagents``, ``/api/flow-events``.

closes #1772
refs #1722, #1771, #1776

## Test plan

- [x] ``python3 -c "import dashboard"`` clean
- [x] New ``tests/test_detail_endpoints_503_on_ingest_outage.py`` (11 cases): all 6 endpoints return 503 on outage; empty-but-healthy returns 200; ``LocalStore.is_writer_alive()`` direct probe; 1 s cache invariant
- [x] No regressions vs ``origin/main`` on ``test_sessions*`` / ``test_brain*`` / ``test_usage*`` (19 pre-existing failures on both, 72 pass on both)
- [x] Diff stat: 7 source files + 1 new test, 220 LOC implementation
- [x] No em-dashes in user-facing banner or error envelope copy

🤖 Generated with [Claude Code](https://claude.com/claude-code)

Issue #1778, opened 2026-05-19
[moat] DuckDB UI coverage audit refresh 2026-05-19 — 81% covered, 10 Tier-1 surfaces to migrate


vivekchand:
## Headline

Refreshed against `origin/main@7a6ae99` after the 2026-05-17 Tier-1 drain (PRs #1569-#1583) plus the events-only epic (#1743) shipping #1774 + #1775.

| Metric | 2026-05-17 baseline | 2026-05-19 refresh | Delta |
|---|---|---|---|
| Total handlers (post stacked-decorator collapse) | 230 | 249 | +19 (new evals/review/_dedupe modules) |
| FAST_PATH | 77 | 104 | +27 |
| PARTIAL | 36 | 9 | -27 (re-classified; underlying coverage unchanged) |
| BYPASS | 6 | 15 | +9 (stricter classifier on fleet-SQLite / log-grep / in-memory ring reads) |
| LIVE_SNAPSHOT | 1 | 9 | +8 (component/* OS probes counted as LIVE this round) |
| META | 109 | 112 | +3 |
| **FAST_PATH / data-bearing** | **65%** (77/119) | **81%** (104/128) | **+16 pp** |
| **DuckDB-genuine** (FAST + PARTIAL) | **95%** (113/119) | **88%** (113/128) | -7 pp (denominator widened by new BYPASS classifications) |

Headline: pure FAST_PATH coverage jumped from 65% to 81% of data-bearing endpoints. The DuckDB-genuine bucket held flat at 113 surfaces (PARTIAL drained into FAST_PATH after the canonical-helper convention solidified), but the data-bearing denominator widened because the classifier now scores log-grep + fleet-SQLite + in-memory-ring reads as BYPASS instead of META.

Memory file refreshed: `/Users/vivek/.claude/projects/-Users-vivek-projects-clawmetry/memory/reference_duckdb_coverage_audit.md` (replace, not append).

## Tier-1 BYPASS punch list (top 10)

Each row: endpoint, file:line, current path, suggested DuckDB migration.

- [ ] **/api/usage/cache-trends** — `routes/usage.py:3209` — JSONL walker over `.jsonl` + `.reset.` files parsing `usage.cacheRead/cacheWrite`. → Extend `local_store.query_cache_metrics(days=N)` over `events` filtered on `data.usage.cacheRead > 0`; bucket by day + model.
[...]

PR #1782, opened 2026-05-19
feat(sessions): /api/delegation-tree DuckDB fast path (Tier-1 from #1778)

vivekchand:
## Summary

Migrates `/api/delegation-tree` (item #8 from #1778 punch list) off the
`sessions.json` index file walker onto a DuckDB fast path that reads
the pre-aggregated `subagents` table (same source as
`_try_local_store_subagents`).

- `_try_local_store_delegation_tree` groups by `parent_session_id` (or
  `spawnedBy` full key when the row's data blob carries one, preserving
  legacy walker shape) and sums per-chain token + cost totals from the
  daemon-computed columns.
- `cost_usd` is honored verbatim since the daemon snapshot already
  applies the issue #1460 v3 sibling-pair dedupe at the
  `query_sessions` SQL layer. Falls back to the OSS token-estimator
  only when the row was written by an older `sync.py` build with no
  cost column populated.
- Legacy `sessions.json` walker stays as fallback for installs whose
  daemon hasn't snapshotted yet, so older OpenClaw users keep
  rendering this view.
- Reachable-but-empty store returns the empty shell directly
  (`_source='local_store'`) so we don't trigger a needless
  `sessions.json` open on healthy installs that simply haven't spawned
  any subagents.

False-positive check passed: `routes/sessions.py:2255` (pre-patch)
does call `open(index_path)` on `sessions.json`, so this is a real
fast-path migration, not an audit miss.

## Test plan

- [x] `tests/test_delegation_tree_local_store.py` (3 cases): two
  children one parent chain rollup with daemon-computed cost; reachable
  empty store returns shell not None; `spawnedBy` full-key grouping
  path preserves legacy response shape.
- [x] `python3 -m pytest tests/test_subagents_local_store_v3.py
  tests/test_subagent_attribution_v3.py
  tests/test_delegation_tree_local_store.py -q` -> 10 passed.
- [x] Diff under 250 LOC (155 LOC in `routes/sessions.py`).
- [ ] Smoke against live `/api/delegation-tree` on a daemon-backed
  install pre-merge.

## Notes

- Part of #1778 tracking issue; do not auto-close.
- Coordination note: `routes/sessions.py:_try_local_store_subagents`
  (PR #1774 area) and `is_writer_alive()` patches (PR #1779) edit
  adjacent code in this file. This PR touches only the
  `delegation-tree` handler + its new helper, no shared-helper
  refactors, so the merge surface stays narrow. Will rebase if main
  moves before merge.

### DIFF
--- a/routes/sessions.py
+++ b/routes/sessions.py
@@ -2234,14 +2234,167 @@ def api_subagents():
     return jsonify(payload)
 
 
+def _try_local_store_delegation_tree():
+    """Fast path for /api/delegation-tree (Tier-1 from #1778 punch list).
+
+    Reads the pre-aggregated ``subagents`` table (same source as
+    ``_try_local_store_subagents``) and groups by ``parent_session_id``
+    to produce the per-chain token + cost totals. Skips the JSONL spawn
+    walker + sessions.json index read entirely — the daemon's snapshot
+    pass already computed both ``cost_usd`` and ``token_count`` per
+    subagent row, so chain totals are a server-side sum.
+
+    Returns ``None`` when the ``subagents`` table is empty so the
+    legacy ``sessions.json`` fallback fires for installs whose daemon
+    hasn't snapshotted yet. Returns a populated empty shell only when
+    the store is reachable AND no subagents have been spawned, which
+    is the right answer (the legacy path would also return empty).
+    """
+    import dashboard as _d
+    rows = _ls_call("query_subagents", limit=500)
+    if rows is None:
+        # Store unreachable. Defer to legacy so older installs keep working.
+        return None
+    if not rows:
+        # Store reachable, just empty. Serve the empty shell directly so
+        # we don't trigger an unnecessary sessions.json open + walk.
+        return {
+            "chains": [],
+            "total_subagents": 0,
+            "total_chain_cost_usd": 0.0,
+            "_source": "local_store",
+        }
+
+    # Fall back to the OSS pricing estimator only when the row lacks a
+    # daemon-computed ``cost_usd``. Cloud's dashboard.py doesn't export
+    # this helper; guard with getattr so we degrade to zero rather than 500.
+    _estimate_usd_per_token = getattr(_d, "_estimate_usd_per_token", None)
+    usd_per_tok = _estimate_usd_per_token() if _estimate_usd_per_token else 3.0 / 1_000_000.0
+    now_ms = time.time() * 1000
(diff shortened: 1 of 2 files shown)


vivekchand:
## UX review

🟡 **1 P1 (no P0)**

1. **Response envelope shape claim needs proof.** PR body says "byte-compatible with the legacy walker" but does not show before/after of an actual `/api/delegation-tree` response. The Subagents tab in the dashboard renders these fields and any drift will surface as a blank tree or a misaligned column. Either paste a `diff <(curl legacy) <(curl fast-path)` from a real session in the PR body, OR add an integration assertion that compares both paths against the same fixture.

vivekchand:
## Product review

🟡 **3 P1s (no P0)**

1. **Status classification (`active`/`idle`/`stale`) thresholds are hard-coded.** `age_ms < 120000` for active, `< 600000` for idle. Different operators (slow agents vs fast ones) want different thresholds. Pull into `dashboard.py` constants or env (`CLAWMETRY_SUBAGENT_ACTIVE_MS=120000`) so it's tunable without a code change. Not P0 but it'll come back as a "my agent shows stale when it's actually running" support ticket.

2. **11 pre-existing test failures in `tests/test_sessions_local_fastpath.py` are the SAME class as PR #1777's reviewer flagged** (running dev-machine daemon contaminates fixture). Per `feedback_synthetic_tests_missed_real_event_shape.md`: "tests that pass on green CI but rot on dev with state" is a recurring bug class. File a single P1 issue covering this: `tests/conftest.py::isolated_local_store(tmp_path)` fixture + a sweep through `tests/test_sessions*.py` to use it. PR #1777 + PR #1779 + this PR have all noted the same failures — that's the trigger threshold for "stop noting, start fixing."

3. **Inherits the GREATEST bridge from PR #1774 for free, but tests don't verify it.** `query_subagents` post-#1774 returns `GREATEST(stored.cost_usd, SUM(events.cost_usd))`. This delegation-tree fast path consumes those rows directly, so when the bridge fires (events outvote stored), the tree shows the higher number — correct, but untested in this PR's coverage. Add one regression assertion: insert subagent with stored cost=$0.05 + 3 events summing to $0.42, hit delegation-tree, assert the chain reports $0.42 not $0.05.

🤖 Pre-merge review @ 41be1fcb. Pushing new commits will re-trigger review if diff >50 lines or >2h elapsed.

PR #1780, opened 2026-05-19
feat(usage): /api/usage/cache-trends DuckDB fast path (Tier-1 from #1778)

vivekchand:
## Summary
- Closes #1778 - Tier-1 item #1: `/api/usage/cache-trends` migrates from JSONL walker to a DuckDB fast path backed by a new `LocalStore.query_cache_metrics(days=N)` method.
- Same v3 sibling-pair dedupe pattern as `query_daily_usage_splits` so `assistant` + `model.completed` rows do not double-count cache reads/writes or total cost.
- Canonical fast-path wiring (`_try_local_store_cache_trends` -> `is_local_store_read_enabled` gate -> return-or-defer), modelled on PR #1571 (`_try_local_store_usage_forecast`).
- Adds `_extract_usage_cost_split` so we can carry per-component cost (input/output/cacheRead/cacheWrite) not just `total`.

## Files
- `clawmetry/local_store.py` (+255): new `query_cache_metrics` + `_extract_usage_cost_split` helper.
- `routes/usage.py` (+82): new `_try_local_store_cache_trends` + 8-line wiring into the existing handler. Legacy JSONL walker untouched and remains the fallback.
- `tests/test_usage_cache_trends_local_store.py` (+253): 3 cases.

## Test plan
- [x] `python3 -c "import dashboard"` clean
- [x] `python3 -m pytest tests/test_usage_cache_trends_local_store.py -q` (3/3 pass)
- [x] `python3 -m pytest tests/test_usage*.py -q` (46/46 pass, no regressions in forecast/usage)
- [ ] CI green
- [ ] Manual smoke against live OpenClaw DuckDB before any release that ships this

## Audit notes
Verified the line 3209 handler was a genuine JSONL walker (no DuckDB code path), so this is a real Tier-1 migration. No audit miss to document.

Refs #1778

PR #1785, opened 2026-05-20
fix(channels): DuckDB fast-path for /api/channel-delivery-health

### LINKED ISSUE #1757: [Product P0 from PR #1740] /api/channel-delivery-health scans log files, violates DuckDB-first rule
From the pre-merge product review of PR #1740.

The new `/api/channel-delivery-health` endpoint scans log files to derive per-channel delivery counts (ok / failed / unconfirmed). This violates the DuckDB-first hard rule documented in \`feedback_duckdb_first_rule.md\`: every feature must persist and read through the local DuckDB events table; JSONL / log / process-stat reads are explicit violations.

## What the right shape looks like

Adapters emit a \`delivery.ok\` / \`delivery.failed\` event at send time. The events table gains a small handful of channel-delivery rows per send. The endpoint becomes:

\`\`\`sql
SELECT
  json_extract_string(data, '$.channel') AS channel,
  SUM(CASE WHEN event_type = 'delivery.ok'     THEN 1 ELSE 0 END) AS ok,
  SUM(CASE WHEN event_type = 'delivery.failed' THEN 1 ELSE 0 END) AS failed
FROM events
WHERE ts >= NOW() - INTERVAL '48 hours'
  AND event_type IN ('delivery.ok', 'delivery.failed')
GROUP BY 1;
\`\`\`

No regex pass, no log-file paging, sub-millisecond on DuckDB.

## Why a P0

Per the MOAT epic (#1743) and today's coverage audit, every new log-file-reading endpoint adds to the "Tier-1 easy bypass" bucket. The next sweep then has to re-migrate it — at higher cost than getting it right the first time. PR #1740 is a small enough scope that the events-emit side could land in the same fire.

## Acceptance

- [ ] All 21 channel adapters emit \`delivery.ok\` + \`delivery.failed\` events at send time (provider, recipient hash, latency_ms, error_message if failed).
- [ ] \`/api/channel-delivery-health\` switches to a DuckDB SUM query.
- [ ] Log-file regex path stays as a fallback for one release (when events table is empty), then is deleted in the following \`[RELEASE]\`.
- [ ] Tier strategy decided: free sees counts, Pro sees per-message detail.

Refs: PR #1740, issue #978 (original report), \`feedback_duckdb_first_rule.md\`.

vivekchand:
Closes #1757

## Summary

- Adds `query_channel_delivery_health(since_hours=48)` to `LocalStore` (`clawmetry/local_store.py`): queries `events` rows with `event_type IN ('delivery.ok', 'delivery.failed', 'delivery.intent')`, decodes the `data` BLOB in Python, groups by provider, and returns the same per-channel shape (`channel`, `ok`, `failed`, `intents`, `unconfirmed`, `success_rate`) the endpoint already emits.
- Rewrites `api_channel_delivery_health()` (`routes/channels.py`) to call the DuckDB fast-path first via `_ls_call()`; if it returns non-empty rows the response is served from DuckDB tagged `_source: "local_store"`. The existing log-file regex scan is kept as a `# DEPRECATED FALLBACK` that fires only when DuckDB has no delivery events yet.

**No user-visible change until channel adapters emit `delivery.ok`/`delivery.failed` events** — at that point the log-scan path is bypassed automatically. The fallback deprecation marker is in place for removal in the follow-up PR that instruments the adapters.

## Test plan

- [x] `python3 -c "import ast; ast.parse(open('clawmetry/local_store.py').read())"` — syntax OK
- [x] `python3 -c "import ast; ast.parse(open('routes/channels.py').read())"` — syntax OK
- [x] `pytest tests/test_schema_migration_failure_integrity.py tests/test_local_store.py -x -q` — 36 passed
- [x] Smoke test: empty store returns `[]` (fallback fires); after ingesting `delivery.ok`/`delivery.failed` events the fast-path returns correct per-channel counts and success rates

---
_Generated by [Claude Code](https://claude.ai/code/session_01Y3wXsRNqtetQqJBvgU93QW)_

### DIFF
--- a/clawmetry/local_store.py
+++ b/clawmetry/local_store.py
@@ -4102,6 +4102,70 @@ class LocalStore:
                 "last_test_error", "updated_at"]
         return [_row_to_dict(r, cols) for r in self._fetch(sql, params)]
 
+    def query_channel_delivery_health(
+        self,
+        since_hours: int = 48,
+    ) -> list[dict[str, Any]]:
+        """Per-channel delivery counts from DuckDB (issue #1757 fast-path).
+
+        Queries events rows with event_type in ('delivery.ok',
+        'delivery.failed', 'delivery.intent'), grouped by the ``provider``
+        field inside the ``data`` BLOB.  Returns an empty list when no such
+        events exist in the window — the caller (api_channel_delivery_health)
+        should fall back to log-file scanning in that case.
+
+        This method is the DuckDB-first counterpart to the log-scan path that
+        PR #1740 introduced.  It becomes the primary data source once channel
+        adapters emit delivery events; until then it returns [] and the
+        legacy fallback fires.
+        """
+        from datetime import datetime, timedelta, timezone
+        cutoff = (
+            datetime.now(timezone.utc) - timedelta(hours=int(since_hours))
+        ).isoformat()
+        rows = self._fetch(
+            "SELECT event_type, data FROM events "
+            "WHERE event_type IN ('delivery.ok','delivery.failed','delivery.intent') "
+            "AND ts >= ?",
+            [cutoff],
+        )
+        decoded = _decode_data_blob_rows(rows, ["event_type", "data"])
+
+        counts: dict[str, dict[str, int]] = {}
+        for row in decoded:
+            data = row.get("data") or {}
+            if not isinstance(data, dict):
+                continue
+            provider = str(
+                data.get("provider") or data.get("channel") or ""
+            ).lower().strip()
+            if not provider:
+                continue
+            if provider not in counts:
+                counts[provider] = {"ok": 0, "failed": 0, "intents": 0}
+            et = row["event_type"]
+            if et == "delivery.ok":
+                counts[provider]["ok"] += 1
+            elif et == "delivery.failed":
+                counts[provider]["failed"] += 1
+            elif et == "delivery.intent":
+                counts[provider]["intents"] += 1
+
+        result = []
+        for ch, c in sorted(counts.items()):
+            total_sends = c["ok"] + c["failed"]
+            result.append({
+                "channel": ch,
+                "intents": c["intents"],
+                "ok": c["ok"],
+                "failed": c["failed"],
+                "unconfirmed": max(0, c["intents"] - total_sends),
+                "success_rate": (
+                    round(c["ok"] / total_sends, 3) if total_sends else None
+                ),
+            })
+        return result
+
     def query_alert_rules(
         self,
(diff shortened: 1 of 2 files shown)


vivekchand:
## UX review

🟡 **2 P1s (no P0)**

1. **No user-visible verification.** Internal API, dashboard's Delivery Health panel reads it. PR doesn't show what the panel renders pre/post the migration. Add a curl + screenshot of the panel to PR body so reviewers can see the response shape is preserved.

2. **Empty fallback semantics need a UI signal.** When DuckDB has no `delivery.*` events yet (i.e. before adapter instrumentation ships), the helper returns `[]` and the route falls back to log-grep. Dashboard shows the log-grep version. That's fine UNTIL the daemon emits events but the dashboard still shows numbers that look wrong because they're stale log-grep vs fresh events. Add `_source: "events" | "log_scan"` envelope so the panel can render a subtle source-indicator and the user knows what they're looking at.

vivekchand:
## Product review

🟡 **3 P1s (no P0)**

1. **Cloud-side classification mismatch incoming.** Cloud PR #1002 just classified `/api/channel-delivery-health` as `cloud-disabled` with reason "greps gateway.log; no analog in cloud container." Once this PR ships AND daemon adapters start emitting `delivery.*` events into DuckDB (the prerequisite for this fast path to actually return data), the cloud-disabled classification becomes wrong — cloud could now serve from the daemon's events via cloud-proxy. File a follow-up issue NOW: `[MOAT] cloud-proxy /api/channel-delivery-health once daemon emits delivery.* events`. Otherwise this gap will surface as "cloud says this route is unavailable, OSS shows it's working" support tickets.

2. **No adapter instrumentation in this PR.** The fast path will return `[]` indefinitely unless something starts emitting `delivery.ok` / `delivery.failed` / `delivery.intent` events. Where's that work? Per memory `feedback_synthetic_tests_missed_real_event_shape.md`: shipping a query-side fast path without the write-side change means the surface looks done but returns empty on every real install. File the matching adapter-emit issue in the same epic, OR include the daemon-side change in this PR (better).

3. **Missing real-data E2E test.** Same pattern as PR #1780 reviews: synthetic ingest test exists, but no `tests/test_live_openclaw_e2e.py` assertion that hits this endpoint with real `delivery.*` events from a real run. Per `feedback_synthetic_tests_missed_real_event_shape.md` — this is the bug class that has burned us 3 times today already.

🤖 Pre-merge review @ 861576df. Pushing new commits will re-trigger review if diff >50 lines or >2h elapsed.

PR #1787, opened 2026-05-20
fix: transcript returns None on empty DuckDB rows so JSONL fallback runs

### LINKED ISSUE #1772: [MOAT bug] /api/transcript/<sid>, /api/session-tools, /api/brain-history, /api/flow-events, /api/usage all return empty for ingested-on-disk session (filesystem fallback NOT wired)

vivekchand:
Closes #1772

## Summary

- `_try_local_store_transcript` was returning an empty shell dict (`{messages: [], total_tokens: 0, ...}`) when DuckDB had no events for a session — e.g. when the ingest pipeline was offline (#1771 scenario).
- The caller treats any non-`None` return as the final response, so the JSONL fallback never ran, producing silent zeros even though the session JSONL existed on disk with real content.
- Fix: return `None` when `rows` is empty. The existing JSONL fallback in `api_transcript` then reads the on-disk file (returning real data) or returns 404 if no file exists — both correct.

`_try_local_store_session_tools` already returns `None` for empty rows; this brings `_try_local_store_transcript` into alignment.

## Test plan

- `tests/test_transcript_openclaw_shapes.py::test_empty_store_returns_none` was the pinpointed failing test (asserting `None`, getting empty dict). It now passes.
- All 12 tests in `test_transcript_openclaw_shapes.py` pass after the change.
- `python3 -c "import ast; ast.parse(open('routes/sessions.py').read())"` — syntax clean.

---
_Generated by [Claude Code](https://claude.ai/code/session_01KqEB14oozAqM5jMG3Mmd4Y)_


vivekchand:
## Product review

🟡 **2 P1s (no P0)**

1. **Pairs with PR #1788 — they're identical changes.** PR #1788 ALSO contains this exact `_try_local_store_transcript` change. Close one. My read: keep #1788 (it's the bundle with crash-recovery test + ARCHITECTURE doc, higher-value), close this one as superseded. Otherwise you'll have two PRs touching the same lines and the second to merge will be a trivial no-op rebase that wastes CI cycles.

2. **The fix is a behavioral revert — name the regression it caused.** PR description says "JSONL fallback runs" but doesn't link to the bug that this populated-shell-blocks-fallback pattern caused. Per `feedback_audit_class_of_bug_not_just_instance.md`: this is exactly the silent-zero class of bug from issue #1772. Link to #1772 in PR body so the reader sees "this fixes the silent zeros when ingest is poisoned" not just "returns None now."

🤖 Pre-merge review @ 7a2c4c27. Pushing new commits will re-trigger review if diff >50 lines or >2h elapsed.

PR #1841, opened 2026-05-21
feat(update): one-click 'Update now' button on update banner + daemon kickstart

vivekchand:
## Summary

User asked for a clear banner that actually updates ClawMetry when clicked, end-to-end. Previously the banner only said 'update available' and pointed at the changelog. Now it does the upgrade itself.

## What changed

- **`clawmetry/templates/partials/banners.html`**: add green 'Update now' button next to the dismissible message; demote 'View Changelog' to a secondary outline button.
- **`clawmetry/static/js/app.js`** (+99 lines): add `checkUpdateStatus` / `dismissUpdateBanner` / `updateFromBanner` handlers + the polling/init pair. *Note: `dismissUpdateBanner` was previously referenced from banners.html but had no live definition — the original lived in a long-orphaned `DASHBOARD_HTML` block in dashboard.py that's overridden by the live template at line 10779. This PR fixes that latent bug too.*
- **`routes/meta.py` `/api/update`** (+19 lines): before `_os._exit(0)`, kickstart the launchctl `com.clawmetry.sync` daemon on macOS. Without this the daemon kept the OLD wheel in memory after upgrade (see `feedback_restart_both_processes_after_upgrade.md` in memory). Darwin-only; other OSes get the dashboard restart only.

## Upgrade flow

1. User clicks **Update now** on banner.
2. Confirm dialog: 'Update ClawMetry to vX.Y.Z now? The dashboard and sync daemon will restart. In-flight requests may be interrupted.'
3. `POST /api/update` runs `pip install -U clawmetry` (~10-30s).
4. Server returns `{ok: true, new_version: 'X.Y.Z'}`, then launchctl-kickstarts the sync daemon and self-exits 2s later.
5. Browser polls `GET /api/version` every 1s; reloads the page once the new process answers (or shows a manual-refresh message after 60s ceiling).

## E2E verified locally

- Stubbed `__version__` down to `0.12.250` (PyPI is at `0.12.264`).
- `POST /api/update-check/check-now` → 200, `latest=0.12.264`.
- `GET /api/update-check/status` → `show_banner=true`, `latest_check.{current,latest}` correct.
- `GET /` served the banner partial with the `update-banner-update-btn` button.
- `GET /static/js/app.js` exported all 3 handler functions.
- Browser confirmed banner renders with all 3 buttons (`Update now`, `Changelog`, `Dismiss`).

## Out of scope (follow-ups)

- The dead-code `DASHBOARD_HTML` block at `dashboard.py:2988-8080` should be deleted in a separate cleanup PR — out of scope here.
- The 4h release-stagger + rollback button discussed in the earlier auto-update conversation will land in a follow-up PR.
- Linux/Windows sync-daemon restart paths (this PR is Darwin-only).

## Test plan

- [ ] Upgrade banner shows when an older version is installed.
- [ ] 'Update now' button opens confirm dialog with target version.
- [ ] On confirm, `pip install -U clawmetry` runs, dashboard restarts, browser auto-reloads.
- [ ] After reload, banner is gone (new version matches latest).
- [ ] Sync daemon also picks up the new wheel (no stale-method 400s).

### DIFF
--- a/clawmetry/static/js/app.js
+++ b/clawmetry/static/js/app.js
@@ -15291,3 +15291,102 @@ function _ncEsc(s) {
     }).catch(function(){});
   } catch(e) {}
 })();
+
+// ── Update banner (PyPI version check + one-click upgrade) ─────────────
+// Functions are referenced from clawmetry/templates/partials/banners.html.
+// /api/update-check/status decides visibility; /api/update runs pip install.
+async function checkUpdateStatus() {
+  try {
+    var data = await fetch('/api/update-check/status').then(function(r){return r.json();});
+    var banner = document.getElementById('update-banner');
+    var msg = document.getElementById('update-banner-msg');
+    if (data.show_banner && banner && msg) {
+      var latest = (data.latest_check && data.latest_check.latest) || 'newer';
+      var current = (data.latest_check && data.latest_check.current) || '';
+      msg.textContent = 'Update available: v' + latest + ' is out. You are on v' + current + '.';
+      banner.style.display = 'flex';
+    }
+  } catch(e) {}
+}
+
+async function dismissUpdateBanner() {
+  try {
+    var data = await fetch('/api/update-check/status').then(function(r){return r.json();});
+    if (data.latest_check && data.latest_check.latest) {
+      await fetch('/api/update-check/dismiss', {
+        method: 'POST',
+        headers: {'Content-Type': 'application/json'},
+        body: JSON.stringify({version: data.latest_check.latest})
+      });
+    }
+    var banner = document.getElementById('update-banner');
+    if (banner) banner.style.display = 'none';
+  } catch(e) {}
+}
+
+async function updateFromBanner() {
+  var btn = document.getElementById('update-banner-update-btn');
+  var msg = document.getElementById('update-banner-msg');
+  var changelog = document.getElementById('update-banner-changelog');
+  if (!btn) return;
+  var target = '';
+  try {
+    var s = await fetch('/api/update-check/status').then(function(r){return r.json();});
+    target = (s && s.latest_check && s.latest_check.latest) || '';
+  } catch(e) {}
+  if (!confirm('Update ClawMetry' + (target ? ' to v' + target : '') + ' now?\n\nThe dashboard and sync daemon will restart. In-flight requests may be interrupted.')) return;
+  btn.disabled = true;
+  btn.textContent = 'Updating...';
+  btn.style.cursor = 'wait';
+  btn.style.opacity = '0.7';
+  if (changelog) changelog.style.display = 'none';
+  if (msg) msg.textContent = 'Running pip install -U clawmetry. This usually takes 10-30 seconds...';
+  try {
+    var resp = await fetch('/api/update', {method: 'POST'});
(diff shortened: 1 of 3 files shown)


vivekchand:
## UX review

- 🟡 P1 — Two green pill buttons sit side by side ("Update now" + "Changelog"), and the outline "Changelog" still uses the same green family + same padding/size as the primary CTA. Classic duplicate-affordance smell — non-technical user can't tell which is the action. Demote Changelog to a plain text link (no border, smaller font) or move it inside a "What's new" tooltip. `clawmetry/templates/partials/banners.html:14-25`.
- 🟡 P1 — Confirm dialog copy leaks engineer jargon: "The dashboard and sync daemon will restart. In-flight requests may be interrupted." A non-technical user has no model of "sync daemon" or "in-flight requests". Rewrite to one short sentence: "Update to vX.Y.Z? This takes about 30 seconds." `clawmetry/static/js/app.js:15339`.
- 🟢 nit — Mid-update status uses raw shell wording: "Running pip install -U clawmetry. This usually takes 10-30 seconds...". Replace with "Updating... (about 30 seconds)" so the banner reads like a product, not a terminal. `clawmetry/static/js/app.js:15347`.
- 🟢 nit — The 60-second timeout message "Restart taking longer than expected. Refresh the page manually." is a dead end. Add a one-click "Refresh" button inline so the user doesn't hunt for Cmd-R. `clawmetry/static/js/app.js:15375`.

🤖 Pre-merge review @ 914c7c11. Pushing new commits will re-trigger review if diff >50 lines or >2h elapsed.

vivekchand:
## Product review

- 🟡 P1 — Post-update success state is wasted real estate. After "Updated to vX.Y.Z. Restarting dashboard..." the banner just vanishes on reload. This is the highest-trust moment in the funnel (user just let us run pip install with one click). Replace the post-reload state with a "What's new in vX.Y.Z" toast that links to Cloud-Pro features added in that release. Free → Pro conversion lever. `clawmetry/static/js/app.js:15351`.
- 🟡 P1 — `pip install -U clawmetry` upgrade is shipped to 100% of OSS without any "Want zero-touch auto-updates across your fleet?" CTA. Fleet-wide updates is exactly the kind of pain Pro should solve. Add a one-line link under the success message: "Managing >1 node? Auto-update your fleet from Cloud-Pro." `clawmetry/templates/partials/banners.html` (new line after msg span).
- 🟢 nit — `setTimeout(checkUpdateStatus, 5000)` + `setInterval(..., 3600000)` means we burn a polling request every hour from every OSS install forever. Move to "check on page load + manual refresh button" — same UX, zero recurring server load and zero PyPI rate-limit risk at fleet scale. `clawmetry/static/js/app.js:15391-15392`.
- 🟢 nit — Darwin-only daemon kickstart is correct scope for this PR, but the silent `except Exception: pass` on Linux/Windows means those users will silently keep running the old wheel until manual reboot. Log a one-line warning to the response payload so the banner can say "Restart your sync daemon manually" on non-Darwin. `routes/meta.py:139-149`.

🤖 Pre-merge review @ 914c7c11. Pushing new commits will re-trigger review if diff >50 lines or >2h elapsed.

PR #1881, opened 2026-05-21
[RELEASE] brain: hide gateway.metric as plumbing (carries #1877)

vivekchand:
## Summary
Release PR to publish the Brain activity-stream fix already on `main` (#1877): `gateway.metric` CPU/RAM pings are now classified as plumbing and hidden by default, so the stream shows real agent activity instead of metric noise when the agent is idle.

Triggers `release-on-merge.yml` → patch bump → PyPI.

## Carries
- #1877 — fix(brain): hide gateway.metric as plumbing

## Verified
- 100% green CI on #1877 (incl. E2E browser + visual-diff).
- Live node `/api/brain-history` (300 events): 85 hidden (45 gateway.metric + 40 queue-operation), 215 real shown.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 ## [Unreleased]
 
+### Brain: hide gateway.metric as plumbing so the activity stream isn't buried (2026-05-21)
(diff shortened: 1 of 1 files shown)


vivekchand:
Redundant — the brain fix (#1877) already shipped in v0.12.276 (carried by the #1880 release; verified present in the published wheel's static/js/app.js). No separate release needed. Closing without merge.

Issue #1602, opened 2026-05-17
[silent-failure-investigation/read-path] Time-range boundary exclusivity audit — 140 since/until callsites need sweep


vivekchand:
**Surfaced by silent-failure inventory deep-dive (2026-05-17). Time-budgeted — needs follow-up audit pass.**

## Suspicion
With 140 occurrences of `since` / `until` / `created_at` boundary clauses across 15 route modules (per ripgrep count), it's likely that at least some use exclusive boundaries (`ts > since`) when callers expect inclusive (`ts >= since`). The most-recent row could be silently dropped on every page-load if the caller is computing `since = max_ts_seen_so_far` and the query is exclusive on the same value.

I did not exhaustively audit every callsite in the deep-dive budget. Filing as an investigation issue to drive a follow-up sweep.

## Files to audit
Top 5 by occurrence count:
- `routes/health.py` — 36 hits
- `routes/usage.py` — 13 hits
- `routes/infra.py` — 13 hits
- `routes/brain.py` — 13 hits
- `routes/local_query.py` — 11 hits

Plus `clawmetry/local_store.py:query_events()` `since` / `until` params — verify the SQL clauses are `>=` / `<=` not `>` / `<`.

## Reproduction (per-call-site once identified)
1. Insert event with `ts = '2026-05-17T10:00:00Z'`.
2. Query with `since = '2026-05-17T10:00:00Z'`.
3. Expect the event in the result. Observe missing → boundary is exclusive (bug).

## Detection strategy
- Grep audit: every `since` / `until` clause must include `=` (`>=` or `<=`). CI gate: flag any new `ts > ?` / `created_at > ?` in routes/*.py.
- Property-based test on `LocalStore.query_events`: insert N events with timestamps at boundaries, assert inclusive semantics.

## Fix shape (per occurrence)
- Replace `>` with `>=` and `<` with `<=` wherever the caller's intent is inclusive.
- Document the intended semantics in a docstring on each query helper.

## Severity
**P2** (investigation) — each instance is bounded (off-by-one row, not a full day). But across 140 callsites the cumulative drift could be noticeable in roll-ups and frontend deduplication windows.

## Related
- #1565 MOAT tracking
- #1588 v3-shape silent-zero family

PR #1886, opened 2026-05-22
fix: add migration_status table to audit v7 dedup migration outcome

### LINKED ISSUE #1598: [silent-failure/write-path] Schema-version migration v6→v7 wrapped in try/except — failure leaves stale dupes, no replay

## Symptom
The v6→v7 schema migration (`_run_dedup_migration_v7`) is wrapped in a try/except that LOGS the failure but still STAMPS the schema_version to 7. On the next daemon start, the version gate says \"already at v7\" and the migration body is skipped — even though it never actually ran successfully.

User-visible impact: legacy duplicate event rows from the pre-v7 ingest (#1232 family) persist forever on installs where the v7 migration failed once. Subsequent token/cost aggregates over-count by the duplication factor.

## Root cause
`clawmetry/local_store.py:889-913`:
```python
try:
    before = ...
    _run_dedup_migration_v7(self._conn)
    after = ...
    deleted = max(0, before - after)
    if deleted: log.info(...)
    else: log.debug(...)
except Exception:
    log.exception(
        \"local store: v7 dedup migration failed (continuing — \"
        \"future writes still dedup via canonical id)\"
    )
# Step 4: stamp the version.
if current < SCHEMA_VERSION:
    self._conn.execute(
        \"INSERT OR IGNORE INTO schema_version (version, applied_at) VALUES (?, ?)\",
        [SCHEMA_VERSION, int(time.time() * 1000)],
    )
```

The stamp runs unconditionally — there's no \"migration succeeded\" gate. The log line says \"future writes still dedup\" which is true for NEW rows but says nothing about the legacy dupes that the migration was supposed to clean.

## Reproduction
1. Construct a DuckDB with v6 schema + 100 duplicate events.
2. Patch `_run_dedup_migration_v7` to raise.
3. Open the store — observe stamp = 7, dupes still present, no error surfaced to user.
4. Restart daemon — migration never re-attempts.

## Detection strategy
- Daemon-side: emit `schema_migration_failed` event into DuckDB on except branch. Cloud dashboard surfaces affected nodes.
- Add a `migration_status` table: (version, attempted_at, completed_at, error_msg). Stamp version only when `completed_at IS NOT NULL`.

## Fix shape
- Only stamp the version inside the `try` block AFTER the migration succeeds.
- On except: log + emit warning event + leave version at 6 so next start retries.
- For installs already at v7 with stale dupes: ship a forensic helper `clawmetry repair --dedup-legacy` that re-runs the migration regardless of stamp.

## Severity
**P1** — silent data integrity / aggregation drift. Won't lose new data but corrupts historical roll-ups for unlucky installs.

## Related
- #1232 v7 dedup migration origin
- #1565 MOAT tracking
- #1590 ring-buffer drop (A1) — sibling write-path family

vivekchand:
Closes #1598

## Summary

- Adds `migration_status` DuckDB table (`migration_version`, `attempted_at`, `completed_at`, `error_msg`) to the DDL — created on every fresh or upgraded store alongside all other tables.
- On success: row inserted inside the main transaction, committed atomically with the `schema_version` stamp.
- On failure: row with `error_msg` written in a new transaction *after* ROLLBACK so the failure record survives the rollback — gives operators a permanent audit trail beyond log trawling.

The core version-stamp fix (only stamp when `migration_failed is False`) was already in from the #1602 fix. This is the missing observability layer from the issue's detection strategy.

## Test plan

- [ ] `python3 -m pytest tests/test_local_store.py` — all 33 tests pass (verified locally)
- [ ] Fresh store: open store, confirm `migration_status` table exists and contains a row for `migration_version=7` with `completed_at` set and `error_msg=NULL`
- [ ] Failure path: patch `_run_dedup_migration_v7` to raise, open store, confirm a `migration_status` row with `error_msg` set and `completed_at=NULL` persists after daemon restart (version not stamped, retries on next boot)

https://claude.ai/code/session_011CKJ7g84Rg39zuh7kjSNuo

---
_Generated by [Claude Code](https://claude.ai/code/session_011CKJ7g84Rg39zuh7kjSNuo)_

### DIFF
--- a/clawmetry/local_store.py
+++ b/clawmetry/local_store.py
@@ -692,6 +692,20 @@ _DDL = [
         applied_at BIGINT NOT NULL
     )
     """,
+    # ── Schema migration history (#1598) ───────────────────────────────────
+    # Records every version-gated migration attempt. Success rows are written
+    # inside the main transaction (committed with the version stamp). Failure
+    # rows are written in a separate transaction after ROLLBACK so they survive
+    # the rollback and give operators a permanent audit trail.
+    """
+    CREATE TABLE IF NOT EXISTS migration_status (
+        migration_version INTEGER NOT NULL,
+        attempted_at      BIGINT NOT NULL,
+        completed_at      BIGINT,
+        error_msg         TEXT,
+        PRIMARY KEY (migration_version, attempted_at)
+    )
+    """,
     # ── Sync dead-letter queue (#1601) ─────────────────────────────────────
     # Persists payloads whose AES-GCM encryption raised inside the cloud
     # POST path (rare: malformed payload, key rotation race, corrupt key).
@@ -1194,12 +1208,13 @@ class LocalStore:
             # the second concurrent process blocks here until the first
             # commits; it then re-reads schema_version and sees the already-
             # stamped version, skipping the migration body entirely.
+            migration_failed = False
+            _migration_err: str | None = None
             self._conn.execute("BEGIN TRANSACTION")
             try:
                 cur = self._conn.execute("SELECT MAX(version) AS v FROM schema_version")
                 row = cur.fetchone()
                 current = row[0] if row and row[0] is not None else 0
-                migration_failed = False
                 if current < 7:
                     # v6 → v7: collapse Claude Code event duplicates the three
                     # ingest paths used to produce (#1232). If this raises we
@@ -1218,6 +1233,13 @@ class LocalStore:
                             "SELECT COUNT(*) FROM events"
                         ).fetchone()[0]
                         deleted = max(0, before - after)
+                        _now_ms = int(time.time() * 1000)
+                        self._conn.execute(
+                            "INSERT OR IGNORE INTO migration_status"
+                            " (migration_version, attempted_at, completed_at, error_msg)"
+                            " VALUES (7, ?, ?, NULL)",
+                            [_now_ms, _now_ms],
+                        )
                         if deleted:
                             log.info(
                                 "local store: v7 dedup migration removed %d "
@@ -1229,12 +1251,13 @@ class LocalStore:
                                 "local store: v7 dedup migration found no dupes (rows=%d)",
                                 after,
                             )
-                    except Exception:
+                    except Exception as exc:
                         log.exception(
                             "local store: v7 dedup migration FAILED — schema "
                             "version will NOT be stamped; next boot will retry"
                         )
                         migration_failed = True
+                        _migration_err = str(exc)
                 # Step 4: stamp the version — ONLY if every gated migration
                 # succeeded. Stamping after a swallowed failure is the #1602
                 # silent-half-state bug.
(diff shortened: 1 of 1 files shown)


vivekchand (commented):
## Test plan & review notes

**What changed**
- `clawmetry/local_store.py`: adds a `migration_status` DuckDB table (`migration_version`, `attempted_at`, `completed_at`, `error_msg`) and writes success/failure audit rows around the v7 dedup migration (+41/-2 in one file). Success committed atomically with the `schema_version` stamp; failure row written in a second transaction after ROLLBACK so it outlives the rollback.

**Smoke commands**
- `python3 -m py_compile clawmetry/local_store.py` — syntax clean
- `python3 -m pytest tests/test_local_store.py` — all 33 tests should pass
- Fresh store quick-check:
  ```bash
  python3 -c "
  from clawmetry.local_store import LocalStore
  s = LocalStore('/tmp/test_ms.db')
  rows = s.conn.execute('SELECT * FROM migration_status').fetchall()
  print(rows)  # expect 1 row with completed_at set, error_msg NULL
  "
  ```

- **Failure-path connection state**: the error-path INSERT runs after a ROLLBACK. If DuckDB marks the connection as broken or auto-closes it on a failed transaction, the fallback INSERT silently drops and the audit trail you're adding won't be there. Worth asserting `row count == 1` in the failure-path test before declaring it solid.
- **`migration_version=7` hardcoded**: if a future migration bumps the schema version without touching this write path, the `migration_status` row will log the wrong version. Consider deriving it from the `SCHEMA_VERSION` constant rather than a literal.
- **Idempotency on repeated boot**: if the daemon restarts mid-migration and the store already has a `migration_status` row for v7, confirm the INSERT doesn't fail with a unique-constraint violation (or that there's no unique constraint, and repeated rows are acceptable).

**Issue link**
- Closes #1598 ✓

---

vivekchand (commented):
## Test plan & review notes

**What changed**
- `clawmetry/local_store.py`: adds `migration_status` DuckDB table (`migration_version`, `attempted_at`, `completed_at`, `error_msg`) to the DDL. On success, a row is inserted inside the main transaction. On failure, a row with `error_msg` is written in a separate transaction after ROLLBACK so the audit trail survives.

**Smoke commands**
```bash
python3 -m pytest tests/test_local_store.py -v
# Expected: all 33 tests pass

# Fresh store verification
python3 - <<'EOF'
import duckdb, tempfile, os, sys
sys.path.insert(0, '.')
from clawmetry.local_store import LocalStore
with tempfile.TemporaryDirectory() as d:
    s = LocalStore(os.path.join(d, 'store.duckdb'))
    con = duckdb.connect(os.path.join(d, 'store.duckdb'), read_only=True)
    rows = con.execute("SELECT * FROM migration_status").fetchall()
    print("migration_status rows:", rows)
    # Expected: 1 row with migration_version=7, completed_at set, error_msg NULL
EOF

# Failure path (manual): patch _run_dedup_migration_v7 to raise, restart daemon,
# confirm a row with error_msg is persisted and schema_version is NOT stamped,
# then unpatch and restart — confirm migration retries and succeeds.
```

- If the success-path `INSERT INTO migration_status` shares the same transaction as the schema DDL and that transaction rolls back for an unrelated reason, the status row disappears. Confirm the INSERT is committed atomically with `schema_version` and that a separate connection can read it.
- Concurrent daemon restart during the failure-path write (the `after-ROLLBACK` transaction) could leave the table empty. Low probability but worth a comment in the code.

**Issue link**
- Closes #1598

---

vivekchand (commented):
## Test plan & review notes

**What changed**
- Adds `migration_status` DuckDB table (DDL + writes on success and on failure) to `local_store.py`; failure record is written in a fresh transaction after ROLLBACK so it survives even when the main migration fails.

**Smoke commands**
```bash
python3 -m py_compile clawmetry/local_store.py
python3 -m pytest tests/test_local_store.py -v
# After starting the daemon once:
python3 -c "
import duckdb; con = duckdb.connect('~/.clawmetry/local_store.db')
print(con.execute('SELECT * FROM migration_status').fetchall())
"
```

- **Failure-path transaction timing** — the failure row is written in a second transaction after the main ROLLBACK; verify the store hasn't been closed or put in an unusable state between the ROLLBACK and the second `con.execute`. If the store teardown is aggressive this write can silently drop.
- **Schema upgrade path** — installs upgrading from pre-#1886 stores won't have the `migration_status` table. Confirm the DDL includes `CREATE TABLE IF NOT EXISTS` (not just `CREATE TABLE`) so upgrades don't crash.
- **`completed_at=NULL` for failure rows** — check the query you'd use to detect a stuck migration (`error_msg IS NOT NULL AND completed_at IS NULL`) returns the failure row correctly; easy to get the NULL semantics backwards.

**Issue link**
- Closes #1598 ✓

---

PR #2000, opened 2026-05-24
[RELEASE] LLM Context Inspector OSS↔cloud parity (carries #1983)

vivekchand:
Publishes #1983 to PyPI.

## What ships
- `/api/overview` filters `clawmetry-*` plumbing sessions out of the main-session pick AND `sessionCount` — the OSS dashboard no longer surfaces a 204K SelfEvolve run as the user's main session, matching the cloud snapshot's existing behaviour.
- Two new `/api/overview` response fields the LLM Context Inspector now reads as the single source of truth on both sides:
  - **`currentContextTokens`** — live prompt size from the latest assistant turn (the right value for "Context Window Usage"; `mainTokens` was cumulative and exceeded the model window after a couple of turns, gave a misleading 100% reading).
  - **`skillHeaderTokens`** — real header-token sum, replacing the `contextWindow*0.008 = 1.6K` approximation the frontend fell back to when `/api/skills` is 410-Gone in cloud.
- `query_context_window_peek` gains `exclude_clawmetry=True` default with a 4× over-fetch so a burst of plumbing rows can't crowd the user's real session out of the scan budget.
- Regression test pins the new filter + new response keys.

Cloud-side snapshot + frontend halves shipped earlier under #1956 — this completes the parity.


vivekchand:

---

PR #2056, opened 2026-05-25
[RELEASE] model-aware LLM context window (carries #2055)

vivekchand:
Publishes #2055 to PyPI.

The LLM Context Inspector gauge no longer reads ">100%". `currentContextTokens` (live prompt size incl. cache) is sized against a model-aware `contextWindow` — 1M for `[1m]` Opus/Sonnet variants, with a defensive bump when the observed prompt exceeds the string-derived base. A Claude Code turn with ~323K live context now reads `323K / 1M (32%)` instead of `323K / 200K (100%)`.

Verified live before merge on both OSS `/api/overview` and the decrypted cloud snapshot (both 323485 / 1000000).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,8 @@
 ## [Unreleased]
 
+### Release: model-aware LLM context window (2026-05-25)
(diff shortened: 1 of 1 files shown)


vivekchand:
Redundant — my context-window fix #2055 (f559d53b) is already an ancestor of [RELEASE] #2057, so 0.12.310 (building now) carries it. Closing to avoid a duplicate bump.

Issue #2196, opened 2026-05-27
feat: error-signal accuracy + per-run analysis suite (benign-error filtering, run compare, waste hints, health timeline, triage)


vivekchand:
## Summary

A cluster of improvements that make ClawMetry's per-run analysis sharper and its error signal trustworthy. These are read-only, observability-only enhancements that fit our existing DuckDB-first + E2E-encrypted-snapshot architecture (no new data path, no server-side compute on OpenClaw data).

They split into two themes:

- **Signal accuracy** — stop counting things that aren't really errors.
- **Per-run analysis** — make it fast to answer "what changed / what's wasteful / which agent is unhealthy" at the level of a single agent run.

Each item below is independently shippable as its own PR. Filed as one issue so the motivation reads as a whole; happy to split into sub-tasks.

---

## 1. Benign-error filtering (highest value, smallest change)

**Problem.** Today error counts are driven by the raw `is_error` / `isError` flag on tool results (`routes/tracing.py`, `routes/selfevolve.py`). A number of common tool outcomes carry that flag but are not real failures, so error rates, the health views, and Self-Evolve findings are inflated:

- `grep`/`rg`/`egrep`/`fgrep` returning **exit code 1** = "no match found", not an error.
- `"File has not been read yet"` / `"File has been modified since read"` — benign read-before-write guards.
- `"gateway timeout after ..."` transient retries that succeeded on retry.
- `unmatched "` and similar parser noise.

**Proposal.** Add a single shared classifier (e.g. `helpers/error_signal.py`) that both the tracing read path and the Self-Evolve classifier call, with:

- A configurable list of ignored error substrings.
- A "benign grep exit" rule: a `Bash` tool result ending in `exit code 1` whose command matches `grep|egrep|fgrep|rg|ripgrep` is not an error.
- JSON-aware checks (`{"status":"error"}`, non-zero `Exit code N`).

**Where.** `routes/tracing.py` (`_iter_tool_results` / error counting), `routes/selfevolve.py` (`_classify_event`). Apply at **ingest classification** so DuckDB stores the corrected `is_error`, which then flows to the snapshot — keeps it cloud-correct, not a render-time patch.

**Why it matters.** This quietly fixes error counts everywhere at once (Tracing, Health, any health timeline, Self-Evolve) with one helper. It should also be guarded by a unit test fixture per benign class so the list can't silently rot.

---

## 2. Per-run A/B comparison with deltas

[...]

PR #2204, opened 2026-05-27
feat(security): redact secrets at the ingest chokepoint

### LINKED ISSUE #2197: Secret redaction at ingest (defense-in-depth before DuckDB write)
## Problem

The sync daemon stores event data — tool-call params, prompts, assistant content, log excerpts — **in plaintext in local DuckDB** before the cloud-sync E2E encryption boundary. If an agent echoes an API key, bearer token, or password into a tool argument or transcript, that secret lands verbatim in `~/.clawmetry`'s DuckDB and in any local read surface. We have no server-side scrubbing as a safety net.

Best practice (and a gap relative to peer tooling) is **defense-in-depth redaction**: scrub obvious secrets *before* they're persisted, regardless of what the runtime emitted.

## Proposal

Add a small, high-precision redaction pass applied at the single event chokepoint — `LocalStore.ingest()` (`clawmetry/local_store.py:1515`) — so every event is scrubbed before it hits the ring buffer / DuckDB.

Scope of the first slice:
- A `clawmetry/redaction.py` helper: `redact_text(s)` / `redact_event(event)`.
- Conservative, low-false-positive patterns → replace with a **stable fingerprint** (e.g. `[REDACTED:sha8]`) so de-dup/cardinality still works without exposing the value:
  - `api_key|token|secret|password|authorization = <value>`
  - `Bearer <token>`
  - `sk-…`, `sk-ant-…`, OpenAI/Anthropic-style keys
  - AWS access keys (`AKIA…`), Google API keys (`AIza…`), `ghp_…`/`gho_…`, Slack `xox[baprs]-…`
  - private-key headers (`-----BEGIN … PRIVATE KEY-----`)
- Apply to the text-bearing event fields (content/params/args/prompt/message) and log excerpts.
- **On by default**, with `CLAWMETRY_REDACT=0` to disable and an extensible custom-pattern hook.
- Never crash on bad input (graceful fallback per house rules).

## Why now / fit
- Pure privacy/security hardening of the existing daemon ingest path — benefits every user.
- DuckDB-first: redaction happens at the daemon write, so cloud (which decrypts client-side) only ever sees scrubbed data.
- Prior art for this pattern exists across the agent-observability ecosystem.

## Acceptance
- A crafted event containing `sk-…`, a bearer token, and `password=…` lands in DuckDB with those values fingerprinted, not verbatim.
- Toggle off via env restores raw storage.
- Unit tests for the pattern set; smoke against live ingest.

## Out of scope (follow-ups)
- File-hash-instead-of-raw for large artifacts.
- Per-org custom redaction rules from the cloud policy surface.

vivekchand:
Closes #2197.

## Problem
The sync daemon stores event payloads — tool-call args, prompts, assistant content, log excerpts — **plaintext in local DuckDB** before the cloud-sync E2E boundary. A secret echoed into a tool argument or transcript would land verbatim in `~/.clawmetry` and every local read surface. No server-side scrubbing as a safety net.

## Change
Defense-in-depth redaction at the single event chokepoint, `LocalStore.ingest()` (`clawmetry/local_store.py`), so every event is scrubbed before it hits the ring buffer / DuckDB.

- **`clawmetry/redaction.py`** — `redact_text()` / `redact_event()`.
- High-precision matchers → **stable fingerprint** `[REDACTED:<sha8>]` (same secret always maps to the same token, so de-dup / cardinality / "leaked twice" still work, but the value is irreversible):
  - provider keys: `sk-`, `sk-ant-`, `AKIA…`, `AIza…`, `ghp_…`, `xox[bapr]-…`, `glpat-…`
  - `Bearer <token>`, `key=value` secrets, PEM `PRIVATE KEY` blocks
  - sensitive field **names** (`api_key`, `password`, `authorization`, …), excluding `*_tokens` counts
- Structural identifiers (`id`/`node_id`/`session_id`/`model`/`token_count`/…) pass through untouched.
- **On by default**; `CLAWMETRY_REDACT=0` disables. Never crashes on bad input (graceful fallback).

7-line touch to `ingest()`; the rest is the new module + tests.

## Tests
`tests/test_redaction.py` — 11 tests: each provider pattern, Bearer-vs-`authorization:` ordering, structured sensitive keys, count-field exclusion, env opt-out, and a ring-level integration test proving `ingest()` scrubs before queuing (inspects the in-memory ring so it's daemon-independent). All green locally.

> Note: the broader `test_*_local_store` suite reports failures **only on a dev box with the sync daemon running** — `get_store()` returns the daemon `_ProxyStore`, so those tests' `store._conn`/`_ring` access fails. Confirmed pre-existing (reproduces with this change stashed); green in CI where no daemon runs.

## Scope
Daemon-side only — the cloud decrypts client-side and now only ever sees scrubbed data, so **no cloud pin bump**. Follow-ups (separate): file-hash-instead-of-raw for large artifacts; per-org custom rules from the cloud policy surface.


vivekchand (commented):
## Rebase + test plan & review notes

**Rebase**: This branch was behind main (`f371641` → `f5aab435`) and had a single CHANGELOG.md conflict in the `[Unreleased]` block. I resolved it by keeping both entries (the benign-error-filter entry from main and the redaction entry from this PR) and pushed (`6481070`). CI should kick off now.

**Note on scope**: The `error_signal` commit (`ac18130`) was already merged into main via #2196, so the rebase correctly dropped it. This PR now contains exactly one commit: the redaction module + the `ingest()` hook + tests.

---

**What changed** (post-rebase)
- `clawmetry/redaction.py` — new `redact_text()` / `redact_event()` module with high-precision secret patterns and stable-fingerprint replacement
- `clawmetry/local_store.py` — 7-line change: `ingest()` calls `redact_event()` before queuing events
- `tests/test_redaction.py` — 11 tests covering provider patterns, structured field names, count-field exclusion, env opt-out, and ring-level integration

**Smoke commands**
```bash
# Unit tests (daemon-free, fast)
python3 -m pytest tests/test_redaction.py -v

# Integration: verify ingest scrubs before DuckDB write
# The ring-level test in test_redaction.py::test_ingest_redacts_at_chokepoint already covers this,
# but can also be verified manually:
python3 -c "
from clawmetry.redaction import redact_text, redact_event
print(redact_text('sk-ant-abcdefghijklmnopqrstuvwx'))   # expect [REDACTED:...]
print(redact_text('Bearer secrettoken12345abc'))         # expect Bearer [REDACTED:...]
ev = redact_event({'id':'e1','node_id':'n','event_type':'t','ts':'x',
                   'data':{'args':{'api_key':'hunter2secretlong'}}})
print(ev['data']['args']['api_key'])                     # expect [REDACTED:...]
"

# Full suite
make test-api
```

1. **`_KEYVAL` regex and `Authorization: Bearer` interaction** — the negative lookahead `(?!(?:Bearer|Basic|...)\\b)` is there to avoid double-redacting Bearer tokens, but worth a manual test: `redact_text('Authorization: Bearer eyJabc.def.ghi')` should produce `Authorization: Bearer [REDACTED:...]` (not `Authorization [REDACTED:...]: Bearer [REDACTED:...]`). The test suite covers this, but good to eyeball the output.
2. **`import from clawmetry import redaction`** inside `ingest()` — this is a lazy import at call time. If `clawmetry/redaction.py` is missing from a packaged install (e.g., not in `setup.py` package list), the `except Exception: pass` silently swallows the import error and events go through unredacted. Worth confirming `clawmetry/redaction.py` is included in the package manifest.
3. **Throughput impact** — `redact_event()` is called on every `ingest()` with nested dict traversal + 7 regex passes per string field. For high-frequency sessions this adds latency on the hot path. The `_MAX_SCAN = 1_000_000` char cap prevents worst-case, but worth a quick profiling pass on a busy store if users report sync lag.

**Issue link**
- `Closes #2197` is present ✓

---

PR #2207, opened 2026-05-28
feat: opt-in secret/credential redaction for snapshot data

### LINKED ISSUE #2198: feat: secret/credential redaction before storage + cloud sync
## Problem

ClawMetry stores **raw tool-call parameters and full session transcripts** in DuckDB and ships them in the E2E-encrypted snapshot to the cloud. Those payloads routinely contain live secrets — `Authorization: Bearer …`, OpenAI `sk-…` keys, GitHub `ghp_…`/`gho_…` tokens, GitLab `glpat-…`, Slack `xox[baprs]-…`, AWS access/secret keys, and inline `api_key=`/`password=`/`secret=` assignments inside `bash`/`curl`/env-dump tool calls.

Two real exposure surfaces:
1. **Local DuckDB is plaintext on disk** (`~/.clawmetry/…`). Anyone with filesystem read gets every secret an agent ever touched.
2. **The decrypted snapshot is rendered client-side.** The moment an encryption key is shared — support/debug handoff, a shared Cloud-Pro org, a leaked `cm-enc-key-*` — every secret in transcripts/params is readable.

Today we redact exactly **one** value (`anthropic_api_key` in the insights config response, `routes/insights.py:259`). There is no systematic scrubbing on the data path. We detect `sk-` keys for provider attribution (`clawmetry/proxy.py`, `advisor.py`) but never mask them before storing/transmitting.

This is a privacy/data-hygiene control. It is **complementary to #2010** (which *detects + flags* PII/credential leaks as security events — "flag, don't block"). This issue is "scrub the value so it never lands in the store/snapshot in the first place."

## Proposed approach

A small, opt-in redaction layer applied **in the daemon** (we are read-only and E2E — the cloud cannot do this; it never sees plaintext).

**Config** (new section, off by default):
```jsonc
"redact": {
  "enabled": false,
  "patterns": [ /* regex strings; defaults below */ ],
  "replacement": "[REDACTED]",
  "scope": "snapshot"   // "snapshot" = scrub on the way out (keep local fidelity)
                        // "ingest"   = scrub before DuckDB write (defense-in-depth, loses local raw)
}
```

**Default `scope: "snapshot"`** is the right trade-off: the operator keeps full-fidelity data locally for debugging, but nothing sensitive leaves the machine. Apply it in the snapshot builder (`sync.py` `_build_transcripts` / `sync_system_snapshot`), recursing through strings/arrays/objects of tool params and transcript content. Offer `scope: "ingest"` for shops that don't want plaintext on local disk at all.

**Default pattern set** (case-insensitive where noted; ported as a starting point, tune for false positives):
```
(?i)(api[_-]?key|apikey)["']?\s*[:=]\s*["']?[a-z0-9_-]{16,}
(?i)(password|passwd|pwd)["']?\s*[:=]\s*["'][^"']+["']
(?i)(secret|token|auth)["']?\s*[:=]\s*["']?[a-z0-9_-]{16,}
(?i)bearer\s+[a-z0-9_-]{20,}
(?i)(aws_secret|aws_access)[a-z_]*["']?\s*[:=]\s*["']?[a-z0-9/+=]{20,}
sk-[a-zA-Z0-9]{32,}
ghp_[a-zA-Z0-9]{36}
gho_[a-zA-Z0-9]{36}
glpat-[a-zA-Z0-9_-]{20,}
xox[baprs]-[a-zA-Z0-9-]{10,}
```
Users can append custom patterns (e.g. `COMPANY-[A-Z]{4}-[0-9]{8}`). A `(?i)` prefix → compile case-insensitive.

## Implementation sketch
- New helper `clawmetry/redact.py` (`compile_patterns()`, `redact_value()` recursing dict/list/str). ~50 LOC, no new deps (stdlib `re`).
- Wire into the snapshot builder (default) and optionally the ingest write path in `local_store.py`.
- Surface a count of redactions performed (telemetry/diagnostics) so users can see it's working.

## Acceptance criteria
- [ ] With `redact.enabled: true`, a tool call containing `curl -H 'Authorization: Bearer sk-abc…'` shows `Authorization: [REDACTED]` in the decrypted snapshot (verify via the decrypt-snapshot E2E check, per FLYWHEEL).
- [ ] Default-off: no behavior change when disabled.
- [ ] `scope: "ingest"` proves the secret is absent from the DuckDB row, not just the snapshot.
- [ ] Custom patterns + custom replacement honored.
- [ ] Unit tests cover each default pattern + nested object/array recursion + the no-match passthrough.
- [ ] Never crashes on bad input (graceful fallback + logged warning).

## Notes
- Must run daemon-side (DuckDB-first + E2E invariant). No cloud route can implement this.
- Apache-2.0 reference implementations of this exact pattern set exist in the OpenClaw plugin ecosystem; we can mirror the patterns (attribution in commit).

🤖 Filed by Claude Code

vivekchand:
Closes #2198

## Summary

- **`clawmetry/redact.py`** (new, ~96 LOC): `compile_patterns()` caches compiled regexes for 10 default secret shapes (Bearer tokens, `sk-*`, `ghp_*`, `gho_*`, `glpat-*`, `xox[baprs]-*`, AWS keys, and generic `api_key`/`password`/`token`/`secret` assignments); `redact_value()` recurses dict/list/str; `build_redactor(config)` returns a bound callable or `None` when disabled.
- **`clawmetry/sync.py`** (~12 LOC changed): `_project_snapshot_messages()` gains an optional `redactor` param — secrets are scrubbed after the `raw`-key strip and **before** size-capping, so truncated text is also clean. `_build_transcripts()` loads config, builds the redactor, and passes it down.
- **`tests/test_redact.py`** (new, ~170 LOC): 22 unit tests covering each default pattern, nested dict/list/str recursion, no-match passthrough, default-off behaviour, custom replacement text, custom patterns, bad-pattern graceful fallback, and empty-input edge cases.

## Test plan

- [ ] `python -m pytest tests/test_redact.py -v` → 22 passed
- [ ] Syntax check: `python -m py_compile clawmetry/redact.py` passes
- [ ] Default-off: no behaviour change for existing users (`build_redactor({})` returns `None`)
- [ ] Enable with `{"redact": {"enabled": true}}` in config; verify `curl -H 'Authorization: Bearer sk-abc…'` appears as `[REDACTED]` in snapshot messages

---
_Generated by [Claude Code](https://claude.ai/code/session_01TDut1ukiSPnJGYedL5zS2n)_

### DIFF
--- a/clawmetry/redact.py
+++ b/clawmetry/redact.py
@@ -1,0 +1,96 @@
+"""Secret/credential redaction for snapshot data (issue #2198).
+
+Applied daemon-side before snapshot data leaves the machine.  Default scope is
+``"snapshot"`` (scrub on the way out to the cloud); ``"ingest"`` also scrubs
+before the DuckDB write (loses local fidelity but maximises defense-in-depth).
+Disabled by default — opt in via ``config.redact.enabled: true``.
+"""
+
+import re
+import logging
+
+log = logging.getLogger(__name__)
+
+# Default patterns cover the most common secret shapes found in agent tool-call
+# parameters and transcript content.  Users can append custom patterns via
+# ``config.redact.patterns``.  Apache-2.0 reference set (OpenClaw plugin
+# ecosystem); tune for false-positives in production.
+_DEFAULT_PATTERNS: list[str] = [
+    r'(?i)(api[_-]?key|apikey)["\']?\s*[:=]\s*["\']?[a-z0-9_\-]{16,}',
+    r'(?i)(password|passwd|pwd)["\']?\s*[:=]\s*["\'][^"\']+["\']',
+    r'(?i)(secret|token|auth)["\']?\s*[:=]\s*["\']?[a-z0-9_\-]{16,}',
+    r'(?i)bearer\s+[a-z0-9_\-]{20,}',
+    r'(?i)(aws_secret|aws_access)[a-z_]*["\']?\s*[:=]\s*["\']?[a-zA-Z0-9/+=]{20,}',
+    r'sk-[a-zA-Z0-9]{32,}',
+    r'ghp_[a-zA-Z0-9]{36}',
+    r'gho_[a-zA-Z0-9]{36}',
+    r'glpat-[a-zA-Z0-9_\-]{20,}',
+    r'xox[baprs]-[a-zA-Z0-9\-]{10,}',
+]
+
+# Module-level cache: (pattern_strings, compiled_patterns)
+_compiled_cache: tuple[list[str], list[re.Pattern]] = ([], [])
+
+
+def compile_patterns(extra: list[str] | None = None) -> list[re.Pattern]:
+    """Return compiled patterns (default set + *extra*), cached by input list."""
+    global _compiled_cache
+    keys = _DEFAULT_PATTERNS + list(extra or [])
+    if _compiled_cache[0] == keys:
+        return _compiled_cache[1]
+    out: list[re.Pattern] = []
+    for p in keys:
+        try:
+            out.append(re.compile(p))
+        except re.error as exc:
+            log.warning("redact: skipping bad pattern %r: %s", p, exc)
+    _compiled_cache = (keys, out)
+    return out
+
+
+def redact_value(v, patterns: list[re.Pattern], replacement: str) -> tuple:
+    """Recursively scrub *v* (str/dict/list). Returns ``(scrubbed_value, count)``."""
+    count = 0
+    if isinstance(v, str):
+        out = v
+        for p in patterns:
+            out, n = p.subn(replacement, out)
+            count += n
+        return out, count
+    if isinstance(v, dict):
+        out = {}
+        for k, val in v.items():
+            out[k], n = redact_value(val, patterns, replacement)
+            count += n
+        return out, count
+    if isinstance(v, list):
+        out = []
+        for item in v:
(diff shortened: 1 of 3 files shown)


vivekchand:
Superseded by the already-merged #2204 (`feat(security): redact secrets at the ingest chokepoint`).

#2204 redacts inside `LocalStore.ingest()` — a single chokepoint upstream of every downstream consumer, including the snapshot builder this PR was going to scrub. By the time `_project_snapshot_messages()` reads the row, the secret is already `[REDACTED:<sha8>]`, so this PR has no incremental coverage; in fact it would be strictly weaker than #2204 alone, which also protects local DuckDB-on-disk.

Differences I checked, just to be thorough:
- **#2204** uses high-precision matchers (`sk-`, `sk-ant-`, `AKIA…`, `AIza…`, `ghp_…`, `gho_…`, `xox[bapr]-…`, `glpat-…`, Bearer, `key=value`, PEM blocks) + sensitive field-NAME detection (`api_key`, `password`, `authorization`, …) + stable fingerprint `[REDACTED:<sha8>]` (preserves de-dup / cardinality / "leaked twice" semantics while being irreversible). On by default; `CLAWMETRY_REDACT=0` disables.
- **This PR (#2207)** is a snapshot-path-only regex scrub with `[REDACTED]` and no fingerprint, opt-in by config.

Closing as duplicate. The work was solid — just landed second.

🤖 Filed by Claude Code

PR #2275, opened 2026-05-29
feat(ingest): OSS-side stub for /api/v1/runs/* and /api/v1/engines/*

### LINKED ISSUE #2269: feat(ingest): OSS-side stub for the custom runtime HTTP ingest API (`POST /api/v1/runs/*`)
## Problem

The new /pricing promises a "Custom runtime HTTP ingest API" at the Pro+ tier. The full implementation lives in clawmetry-pro (separate tracker). This OSS issue covers only the **hook surface** the OSS package needs to expose so a paid plugin can plug in.

## Scope (OSS only)

1. Reserve the route namespace `/api/v1/runs/*` and `/api/v1/engines/*` in OSS, with stub handlers that:
   * If a paid plugin has registered an override via `clawmetry.extensions`, delegate to it.
   * Otherwise return HTTP 402 `upgrade_required` with feature key `custom_runtime_ingest` (gate added in #2263).
2. Add the entitlement key `custom_runtime_ingest` (covered under #2263 / #2262).
3. Provide a small `clawmetry.extensions.runtime_ingest` hook point with documented signature so the paid implementation can attach cleanly.

## Out of scope

Actual ingest writes (storage shape, auth token, retention, schema validation, OTel mapping) live in the closed plugin and are tracked separately.

Tracks: #2262

vivekchand:
## Summary

Closes #2269

- Reserves `/api/v1/runs/<path>` and `/api/v1/engines/<path>` so clients get a helpful `402 upgrade_required` (feature: `custom_runtime_ingest`) instead of a confusing 404
- Adds `extensions.dispatch()` — a request/response variant of `emit()` that returns the first non-None handler result, letting the paid plugin attach a real ingest handler without duplicate Flask route registration
- Adds `"custom_runtime_ingest"` to `PAID_FEATURES` so `allows_feature()` has a stable key

## Test plan

- `python3 -m pytest tests/test_runtime_ingest_stub.py -v` — 10 hermetic tests: 402 stub for both namespaces, plugin delegation (handler returns response), None-returning handler falls through to 402, `dispatch()` unit tests, feature key in `PAID_FEATURES`
- `python3 -m pytest tests/test_blueprint_uniqueness.py -v` — confirms `runtime_ingest` blueprint name is unique
- `python3 -c "import ast; ast.parse(open('dashboard.py').read())"` — syntax check
- `ruff check clawmetry/extensions.py clawmetry/entitlements.py routes/runtime_ingest.py` — all clean

---
_Generated by [Claude Code](https://claude.ai/code/session_01TTibzhsZhMWgyCPRA8SGBU)_

### DIFF
--- a/clawmetry/entitlements.py
+++ b/clawmetry/entitlements.py
@@ -136,6 +136,7 @@ PAID_FEATURES = frozenset(
         "anomaly_detection",
         "self_evolve",
         "cost_optimizer",
+        "custom_runtime_ingest",
     }
 )
 
--- a/clawmetry/extensions.py
+++ b/clawmetry/extensions.py
@@ -66,6 +66,30 @@ def emit(event: str, payload: Dict[str, Any] | None = None) -> None:
             )
 
 
+def dispatch(event: str, payload: Dict[str, Any] | None = None) -> Any:
+    """Call handlers for *event* in order and return the first non-None result.
+
+    Unlike :func:`emit`, this is a request/response pattern: the first handler
+    that returns a non-None value wins and subsequent handlers are skipped.
+    If no handler is registered or every handler returns None, returns None.
+    Exceptions in handlers are caught and logged — never propagated.
+    """
+    if payload is None:
+        payload = {}
+    with _lock:
+        handlers = list(_registry.get(event, []))
+    for handler in handlers:
+        try:
+            result = handler(payload)
+            if result is not None:
+                return result
+        except Exception as exc:
+            logger.warning(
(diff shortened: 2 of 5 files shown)


vivekchand:
Superseded by #2283 which lands the full custom-runtime HTTP ingest API (Pro) — including `/api/v1/runs`, `/api/v1/runs/<id>/events`, `/api/v1/runs/<id>/end`, and `GET /api/v1/runtimes` — all gated on `custom_runtime_ingest` with 402 upgrade_required in enforce mode. This stub PR is no longer needed. Closing.

PR #2366, opened 2026-05-30
fix(fleet): surface local-node DuckDB scope on Usage and History tabs

### LINKED ISSUE #1607: [silent-failure-investigation/sync-path] Multi-node fleet (same owner_hash) — local DuckDB has no node_id partition for read-aggregates
**Surfaced by silent-failure inventory deep-dive (2026-05-17). Time-budgeted — needs multi-node E2E test.**

## Suspicion
Local DuckDB schema includes `node_id` on `events` (required) and `sessions` (optional). But every node has its OWN local DuckDB file. The fleet-aggregation view on cloud unifies across nodes BUT cloud has the dedup bug (vivekchand/clawmetry-cloud#929) — and the daemon-proxy local-query path only sees its own node's data.

When user opens dashboard on node A but wants to see node B's sessions, the flow is:
- OSS dashboard on node A → daemon-proxy → local DuckDB on A → returns A's sessions only.
- Cloud dashboard → cloud SQL → returns BOTH (but possibly with the dedup multiplication from #929).

I did not verify whether the cloud-side double-counting matches the multi-node case or whether multi-node sessions are properly disambiguated on the fleet view.

## Files to audit
- `routes/fleet_history.py` — multi-node fleet view.
- `clawmetry-cloud/routes/api.py:_aggregate_events_across_nodes` (if exists).
- Test: spin up 2 daemon processes with the same `cm_` token but different `node_id`. Sync both. Compare cloud aggregates vs sum of local DuckDB counts.

## Reproduction
1. Two daemons, same cm_ token, different node_id, both running 10-event sessions.
2. After sync: cloud `SELECT SUM(token_count) FROM events WHERE owner_hash=?`.
3. Compare to local DuckDB on A + local DuckDB on B (summed manually).

Expected: equal. Investigate any divergence.

## Detection strategy
- Cloud + OSS parity script in CI: spin up 2-node fleet, assert aggregates match.
- Cloud: add a `(owner_hash, node_id, canonical_id)` unique constraint to prevent multi-node double-write of the same event.

## Fix shape
TBD pending audit. Likely combination of:
- Fix sibling vivekchand/clawmetry-cloud#929 (cloud INSERT uses canonical id) — prevents most double-counting.
- Cloud read endpoints: explicit GROUP BY canonical_id when aggregating across nodes.

## Severity
**P2** (investigation) — most users have 1 node. But power users / agencies / fleet customers will hit this.

## Related
- vivekchand/clawmetry-cloud#929 cloud canonical-id dedup
- #1565 MOAT tracking

vivekchand:
Closes #1607

## Summary

- **Usage and History tabs** now show an info banner ("Local node only — see Fleet for cross-node totals") so multi-node fleet users know the analytics scope at a glance.
- **`/api/nodes` response** enriches the local node's entry with DuckDB-derived `duckdb_summary` (`cost_today_usd`, `token_count_today`) queried directly via `local_store.query_aggregates`, bypassing the heartbeat payload which doesn't include these fields today.
- **`fleet_summary.data_scope`** is now `"single_node"` or `"multi_node_partial"` so API callers can detect partial aggregation without parsing node counts.

## Test plan

- [ ] Start ClawMetry with one node — Usage and History tabs show the info banner; Fleet tab shows `data_scope: "single_node"`.
- [ ] Register a second node — Fleet tab shows `data_scope: "multi_node_partial"`; local node entry includes `duckdb_summary` with accurate cost/token figures.
- [ ] Verify `GET /api/nodes` returns `duckdb_summary` for the local node (non-empty DuckDB) without error.
- [ ] Confirm no regression: Usage/History data still loads correctly, banner links to Fleet tab.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Wmoo2PB71KtuXPyJvR1H8T)_

### DIFF
--- a/clawmetry/templates/tabs/history.html
+++ b/clawmetry/templates/tabs/history.html
@@ -1,4 +1,7 @@
 <div class="page" id="page-history">
+  <div id="history-node-scope-banner" style="background:var(--bg-secondary,#f1f5f9);border:1px solid var(--border-primary,#e2e8f0);border-left:3px solid #3b82f6;padding:10px 14px;border-radius:4px;margin-bottom:12px;font-size:13px;color:var(--text-secondary,#64748b);">
+    ℹ️ <strong>Local node only.</strong> These figures reflect only this node's DuckDB. See the <a href="#" onclick="showTab('fleet');return false" style="color:#3b82f6;text-decoration:none;">Fleet</a> tab for cross-node totals.
+  </div>
   <div style="display:flex;align-items:center;gap:16px;margin-bottom:16px;flex-wrap:wrap;">
     <h2 style="font-size:18px;font-weight:700;color:var(--text-primary);margin:0;">&#128202; <span data-i18n="history.title">History</span></h2>
     <div style="display:flex;gap:4px;flex-wrap:wrap;" id="time-range-picker">
--- a/clawmetry/templates/tabs/usage.html
+++ b/clawmetry/templates/tabs/usage.html
@@ -1,4 +1,7 @@
 <div class="page" id="page-usage">
+  <div id="usage-node-scope-banner" style="background:var(--bg-secondary,#f1f5f9);border:1px solid var(--border-primary,#e2e8f0);border-left:3px solid #3b82f6;padding:10px 14px;border-radius:4px;margin-bottom:12px;font-size:13px;color:var(--text-secondary,#64748b);">
+    ℹ️ <strong>Local node only.</strong> These figures reflect only this node's DuckDB. See the <a href="#" onclick="showTab('fleet');return false" style="color:#3b82f6;text-decoration:none;">Fleet</a> tab for cross-node totals.
+  </div>
   <div class="refresh-bar">
     <button class="refresh-btn" onclick="loadUsage()">↻ <span data-i18n="common.refresh">Refresh</span></button>
     <button class="refresh-btn" onclick="exportUsageData()" style="margin-left: 8px;">📥 <span data-i18n="usage.export_csv">Export CSV</span></button>
--- a/routes/fleet_history.py
+++ b/routes/fleet_history.py
@@ -16,6 +16,7 @@ Module-level helpers (``_fleet_db``, ``_fleet_db_lock``, ``_fleet_check_key``,
 ``dashboard.py`` and are reached via late ``import dashboard as _d``.
 """
 
+import datetime
 import json
 import time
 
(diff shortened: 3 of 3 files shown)


vivekchand (commented):
## Test plan & review notes

**CI**: ✅ 19/19 checks green · mergeable: clean · issue link `Closes #1607` already present

**What changed**
- Adds a static info banner ("Local node only — see Fleet for cross-node totals") to the top of Usage and History tabs; enriches the local node's entry in `/api/nodes` with a `duckdb_summary` field; adds `data_scope: "single_node" | "multi_node_partial"` to `fleet_summary`.

---

**Smoke commands**
```bash
make lint
make test-api
```

**What to look at visually**
- `http://localhost:8900/#usage` — blue left-border info banner should appear at the top; data underneath should load unchanged
- `http://localhost:8900/#history` — same banner; time-range picker and charts should render normally
- `curl -sS http://localhost:8900/api/nodes | python3 -m json.tool` → local node entry should include `"duckdb_summary": {"cost_today_usd": ..., "token_count_today": ..., "source": "local_duckdb"}` and `fleet_summary` should include `"data_scope": "single_node"`

1. **`total_cost = max(total_cost, cost_today)` logic looks off** — by the time we reach the DuckDB enrichment block, `total_cost` has already been accumulated by summing all nodes' heartbeat costs. Replacing it with `max(total_cost, cost_today)` means: if two remote nodes together cost $2 and the local node's DuckDB shows $0.50 today, the total drops to $2 (correct by coincidence). But if the local node is the expensive one ($1.50) and the only other node contributed $0.10 (total = $1.60), `max($1.60, $1.50)` leaves `total_cost` at $1.60 — one cent off from what DuckDB says. The larger issue is that this replaces the accumulated multi-node total rather than patching just the local node's contribution. The safer fix would be: subtract the local node's heartbeat cost from `total_cost` then add `cost_today` instead. Worth a look before merge.

2. **Banner is always visible, even on single-user setups** — the banners are static HTML with no conditional hide/show. A user who never uses fleet will always see "See the Fleet tab for cross-node totals" on both Usage and History. Consider hiding the banner when `fleet_summary.data_scope == "single_node"` and only one node is registered, or at least making it dismissible.

3. **`store.query_aggregates(since=today)`** — the call passes a date string (`"2026-05-31"`) to `query_aggregates`. If that method doesn't accept a `since` parameter taking an ISO date string (vs a datetime or timestamp), it will throw and the whole enrichment block will silently no-op (caught by `except Exception: pass`). Quick test: confirm `query_aggregates(since="2026-05-31")` is a valid call signature in `clawmetry/local_store.py`.

**Issue link**
- `Closes #1607` already in PR body ✓


---

PR #1352, opened 2026-05-15
[RELEASE] feat(approvals): wire watcher_kick into _flush_now (#1343 Phase 2.2 — closes Phase 2)

vivekchand:
## Why
Phase 2.2 of #1343. Phase 2.1 (#1350) shipped the kick infrastructure. This wires it: at the end of \`LocalStore._flush_now()\`, if any flushed event was a \`tool_call\`/\`toolCall\`, lazy-import \`approvals.watcher_kick()\` and call it.

## Effect
Detection latency for a fresh tool_call event drops from up to 2 s (watcher's polling interval) → ~0 ms (kick wakes it immediately). Coalesces inside \`approvals._kick_event\` — 50-event flush = 1 wakeup.

## Safety
- Lazy import — partial installs without approvals.py don't crash flush
- try/except swallows unrelated import errors
- Kick AFTER COMMIT (row must be readable when watcher wakes; kick-before would race to empty read)
- \`break\` after first match — one kick per batch
- Backward compat: watcher's 2 s heartbeat is still the safety net

## Issue #1343 status
- ✅ Phase 1.1 (#1344) / 1.2 (#1345) / 1.3 (#1347): UI complete
- ✅ Phase 2.1 (#1350): kick infrastructure
- ✅ Phase 2.2 (this PR): kick wired into ingest path
- 🟡 Phase 3: OpenClaw native bridge

**Phase 2 complete after this lands.** Approvals are event-driven end-to-end.

## Test plan
- [x] \`ast.parse\` passes
- [ ] CI lint + API tests
- [ ] Post-deploy: trigger a synthetic tool_call from openclaw, observe watcher logs show \`scanned 1 new toolCalls\` within ~10ms (vs ~2s today)

PR #2555, opened 2026-06-03
docs(flywheel): runtime-filter rule §1c + per-runtime v1-API verifier

vivekchand:
Encodes the founder's directive as a FLYWHEEL rule and ships the verifier that operationalises it.

## FLYWHEEL §1c — the runtime-filter rule
When the runtime switcher is set to a runtime, **every view must show only that runtime's data** — the only thing that changes is the `runtime=<id>` param; all numbers/lists/headers re-derive from the filtered response. A view that stays node-wide while the switcher says "PicoClaw · 1 session" is a bug (burned 2026-06-03: node-detail header showed 68 sessions / 3.8M tokens with PicoClaw selected).

- The **public v1 API is the filtering contract** (scopes server-side by `?runtime=`). Prefer sourcing runtime-scoped data from it over client-side filtering of a pre-aggregated blob (which silently no-ops — the `_CM_RT_AGGREGATE` tabs).
- **Verify per runtime, not just one** (a filter that hard-codes the first runtime passes a one-runtime eyeball).

## `scripts/verify_runtime_filtering.py`
Sweeps every runtime a node reports through the public v1 API and asserts the responses scope: counts match `nodes/<id>/runtimes`, runtimes differ from each other, and an **unknown runtime returns 0** (not the node total).

**It already paid for itself**: the first live run caught a real server-side leak (unknown runtime returned 50 sessions instead of 0), fixed in clawmetry-cloud #1352.


vivekchand:
The Policy Events panel already has the regex patterns and redaction logic for credential leaks (`_POLICY_SIGNATURES` / `_scan_content_for_policy_events` in `dashboard.py`), but there's no dedicated endpoint returning the `api_key_leak` shape with `file` + `line` info that this issue specifies.

**Problem**: No `GET /api/security/credential-scan` endpoint; no red-badge count specifically for API key leaks on the Security tab.

**Files I'll change:**
- `routes/infra.py` — add `GET /api/security/credential-scan` that re-uses the existing LEAK signatures from `_POLICY_SIGNATURES`, queries events via DuckDB (same path as `/api/security/policy-events`), and returns `{leaks: [{type:"api_key_leak", file:session_id, line:event_idx, pattern:"openai|anthropic|aws|github|generic", redacted:"sk-...XX"}], count:N}`. Never returns the raw key.
- `clawmetry/templates/tabs/security.html` — add an "API Key Scan" panel below the Policy Events section; shows a red badge when `count > 0`, lists hits with file/pattern/redacted columns.

**Tradeoff**: `file` maps to `session_id` (not a raw JSONL path) and `line` to the event index — both are the correct DuckDB-first equivalents since raw JSONL reads return empty in cloud.

Implementing now, will open a PR shortly.

---

PR #2703, opened 2026-06-05
fix(license): install clawmetry-pro into a HOME fallback when site-packages is read-only

vivekchand:
## Bug (found on a real self-hosted box, 2026-06-05)

A system-wide install (e.g. `/opt/clawmetry` owned by `root`) run by a **non-root daemon** (`systemctl --user`) cannot write the Pro wheel into the interpreter site-packages. Both the pip path and the pip-less unzip fallback targeted `sysconfig.get_path("purelib")`, so the auto-provisioner failed with:

```
license: clawmetry-pro install failed: unzip install failed:
[Errno 13] Permission denied: '/opt/clawmetry/lib/python3.10/site-packages/clawmetry_pro'
```

Net: the paid runtime adapters (Claude Code, Codex, Cursor, ...) **silently never loaded** even though the account was entitled (Trial/Pro). The only workaround was a manual `sudo chown` of the install dir, which no normal user will discover.

## Fix (general, no sudo/chown needed)

- `_site_packages_target()` reports whether the interpreter site-packages is actually writable by us (`os.access(..., W_OK)`).
- When it is **not** writable, the wheel is extracted into a HOME-owned fallback dir `~/.clawmetry/pro-packages` and that dir is added to `sys.path` so the adapters import. `_pip_install_wheel` short-circuits to this path too (pip would fail on read-only site-packages anyway).
- `ensure_pro_on_path()` adds the fallback dir to `sys.path`; called at daemon startup **before** plugin discovery (`sync.py`) and at the start of `_provision_pro_wheel` (so an already fallback-installed pro is detected as present and the install stays idempotent).

This covers **any** read-only-install layout, not just `/opt`.

## Tests
- New `tests/test_pro_install_fallback.py`: read-only site-packages -> extract into fallback + put on `sys.path`; `_pip_install_wheel` short-circuits (pip never runs); writable site-packages -> normal path; `ensure_pro_on_path` idempotent.
- Updated `test_license.py::test_pip_install_prefers_pip_when_available` to pin site-packages writable (pip is now used only when writable). 30 license tests pass.

## Follow-up
Optionally harden `install.sh` to make a system-wide install dir writable by the daemon user at install time. The provisioner fallback in this PR already makes that unnecessary for correctness.

### DIFF
--- a/clawmetry/adapters/openclaw.py
+++ b/clawmetry/adapters/openclaw.py
@@ -87,6 +87,102 @@ def _real_install(sessions_dir: str) -> bool:
                for m in ("SOUL.md", "AGENTS.md", "MEMORY.md"))
 
 
+def _model_router_fingerprint() -> dict:
+    """Read the NemoClaw model-router source fingerprint (``git:<sha>``)
+    written by harness onboarding to ``<venv>/.nemoclaw-source-fingerprint``
+    (model-router.ts writeModelRouterInstalledFingerprint). Surfaces the
+    install-provenance / version-drift signal on DetectResult.meta (#2608).
+
+    Read-only and never raises. Returns ``{}`` when the file/venv is absent
+    (plain OpenClaw or old NemoClaw installs), so the meta dict is unchanged.
+    """
+    venv = os.environ.get("NEMOCLAW_MODEL_ROUTER_VENV") or os.path.expanduser(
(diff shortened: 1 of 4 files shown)


vivekchand:
Closing: created from the wrong head branch (obs-gaps/free-runtime-ingest) by mistake. Reopening from the correct fix/pro-install-writable-fallback branch.

PR #2905, opened 2026-06-08
fix(tokens): seed output_tokens floor from message_start + streamed-thinking reconciliation tests (#2842)

vivekchand:
Closes #2842

## What
Hardens the Anthropic SSE token state machine in `proxy.py:parse_anthropic_sse_chunk` so extended-thinking streamed responses are never undercounted, and adds the reconciliation test the issue asks for (streamed thinking usage == non-streamed usage).

## How
- **`message_start` now seeds an `output_tokens` floor.** Previously only `message_delta` ever set `output_tokens`. If a stream was truncated *before* the final `message_delta` (client disconnect, upstream error mid-thinking), the recorded output was **0** — silently undercounting cost. We now take `max(output_tokens, message_start.usage.output_tokens)`, a true lower bound. The authoritative final `message_delta` total (already `max()`-reconciled) can only raise it, never lose ground.
- **Tests added** in `tests/test_proxy.py::TestSSEParsing`:
  - `test_anthropic_streamed_thinking_matches_nonstreamed` — a full thinking sequence (thinking_delta + signature_delta + text_delta, then final `message_delta`) accumulates the exact `output_tokens` the equivalent non-streamed response reports.
  - `test_anthropic_truncated_stream_keeps_floor` — a stream cut off before `message_delta` preserves the `message_start` floor instead of dropping to 0.
  - `test_anthropic_message_start_output_floor_never_lowers_delta` — the floor never clobbers a larger accumulated value.

## Verification
`pytest tests/test_proxy.py tests/test_proxy_cache_risk.py` → 66 passed (10 SSE-parsing tests, 3 new).

Refs epic #2837.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #2913, opened 2026-06-08
fix(proxy): track reasoning_tokens for extended-thinking streams

### LINKED ISSUE #2842: fix(tokens): accurate streamed-token counting (SSE thinking/signature deltas) (#2837)
`proxy.py:parse_anthropic_sse_chunk` / `StreamUsage` (~L835) reads only `message_start` + `message_delta`; extended-thinking reasoning tokens streamed via content_block deltas can be undercounted. Harden the SSE state machine (accumulate to event boundaries, track thinking_delta/signature_delta, reconcile final usage) in `proxy.py` and mirror in `interceptor.py`. Add a test that a streamed thinking response matches the non-streamed usage. Tightens every cost number we report. Refs epic #2837.

vivekchand:
## Summary

- `StreamUsage` gains a `reasoning_tokens` field; `parse_anthropic_sse_chunk` now reads `thinking_tokens` / `reasoning_tokens` / `thinking_input_tokens` from `message_delta.usage` (taking the max across multiple deltas) and handles `content_block_delta` events with `thinking_delta` / `signature_delta` types without silent data loss
- Non-streaming Anthropic path (`_forward_non_streaming`) gets the same reasoning-token extraction
- `ProxyDB.record_usage` accepts `reasoning_tokens`; the SQLite schema gains the column via idempotent `ALTER TABLE` so existing installs migrate seamlessly; both streaming and non-streaming call sites pass the value through
- `interceptor.py` extracts `reasoning_tokens` from Anthropic responses and includes it in JSONL events when non-zero

## Test plan

- [x] 4 new tests in `tests/test_proxy_cache_risk.py`: thinking stream end-to-end, alternate key names (`thinking_tokens` / `reasoning_tokens` / `thinking_input_tokens`), max-across-deltas, and zero for non-thinking sessions
- [x] All 9 tests in `test_proxy_cache_risk.py` pass; no regressions in `test_proxy.py` SSE parsing suite
- [x] Non-thinking sessions unaffected (`reasoning_tokens` stays 0, no schema change observed by callers)

Closes #2842

---
_Generated by [Claude Code](https://claude.ai/code/session_01PjW1QLTZfJnEBS3TaNqJGV)_

### DIFF
--- a/clawmetry/interceptor.py
+++ b/clawmetry/interceptor.py
@@ -245,7 +245,7 @@ def _extract_model_from_body(body_bytes: bytes, url: str) -> str | None:
 
 def _extract_tokens_from_response(body_bytes: bytes, provider: str) -> dict[str, int]:
     """Extract input/output token counts from response body."""
-    result = {"input_tokens": 0, "output_tokens": 0}
+    result = {"input_tokens": 0, "output_tokens": 0, "reasoning_tokens": 0}
     if not body_bytes:
         return result
     try:
@@ -255,6 +255,12 @@ def _extract_tokens_from_response(body_bytes: bytes, provider: str) -> dict[str,
             usage = body.get("usage", {})
             result["input_tokens"] = usage.get("input_tokens", 0)
             result["output_tokens"] = usage.get("output_tokens", 0)
+            result["reasoning_tokens"] = int(
+                usage.get("thinking_tokens")
+                or usage.get("reasoning_tokens")
+                or usage.get("thinking_input_tokens")
+                or 0
+            )
 
         elif provider in ("openai", "openrouter"):
             usage = body.get("usage", {})
@@ -311,6 +317,7 @@ def _build_event(
     latency_ms: float,
     status_code: int,
     library: str,
+    reasoning_tokens: int = 0,
 ) -> dict[str, Any]:
     """Build the event dict to write to JSONL."""
     cost = _estimate_cost(model or "", input_tokens, output_tokens)
@@ -330,6 +337,8 @@ def _build_event(
         event["model"] = model
     if cost is not None:
         event["cost_usd"] = cost
+    if reasoning_tokens:
+        event["reasoning_tokens"] = reasoning_tokens
     src = _get_source()
     if src:
         event["source"] = src
@@ -421,6 +430,7 @@ def _patch_httpx() -> bool:
                 model=final_model,
                 input_tokens=tokens["input_tokens"],
                 output_tokens=tokens["output_tokens"],
+                reasoning_tokens=tokens["reasoning_tokens"],
                 latency_ms=latency_ms,
                 status_code=response.status_code,
                 library="httpx",
@@ -482,6 +492,7 @@ def _patch_httpx() -> bool:
                     model=final_model,
                     input_tokens=tokens["input_tokens"],
                     output_tokens=tokens["output_tokens"],
+                    reasoning_tokens=tokens["reasoning_tokens"],
                     latency_ms=latency_ms,
                     status_code=response.status_code,
                     library="httpx.async",
@@ -568,6 +579,7 @@ def _patch_requests() -> bool:
                 model=final_model,
                 input_tokens=tokens["input_tokens"],
                 output_tokens=tokens["output_tokens"],
+                reasoning_tokens=tokens["reasoning_tokens"],
                 latency_ms=latency_ms,
                 status_code=response.status_code,
                 library="requests",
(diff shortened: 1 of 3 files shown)


vivekchand (commented):
## Test plan & review notes

**What changed**
- `proxy.py`: `StreamUsage` gains `reasoning_tokens`; `parse_anthropic_sse_chunk` now reads `thinking_tokens` / `reasoning_tokens` / `thinking_input_tokens` from `message_delta.usage` (taking `max` across deltas); handles `content_block_delta` with `thinking_delta` / `signature_delta` types. Non-streaming path `_forward_non_streaming` gets the same extraction. `ProxyDB.record_usage` + schema gain the column via idempotent `ALTER TABLE`. Both call sites pass `reasoning_tokens` through.
- `interceptor.py`: `_extract_tokens_from_response` initialises `reasoning_tokens: 0` and reads the key from Anthropic usage; `_build_event` conditionally includes `reasoning_tokens` in the JSONL event when non-zero.
- `tests/test_proxy_cache_risk.py`: 4 new tests (E2E thinking stream, alternate key names, max-across-deltas, zero for non-thinking).

**Smoke commands**
```bash
# Fast focused run
pytest tests/test_proxy_cache_risk.py tests/test_proxy.py -v

make test   # full suite
```

**What to look at (locally)**
- Route a real extended-thinking Anthropic request through the proxy (port 4100). After the session, check the `proxy_usage` SQLite table — `reasoning_tokens` should be non-zero.
- Non-thinking session: column should stay `0`, no regressions in `input_tokens` / `output_tokens`.
- JSONL event file: a thinking session should have an `"reasoning_tokens"` key; a non-thinking session should not (the `if reasoning_tokens:` guard).

- The `ALTER TABLE … ADD COLUMN` is wrapped in a bare `except Exception: pass` — this is the standard SQLite approach but silently swallows any genuine schema errors beyond "column already exists". Fine for this case, but worth a `# column already exists` comment so it's clear why the exception is eaten.
- `content_block_delta` handler does `_ = delta.get("type")` — this is a no-op "consumed" marker for future-proofing. Could be a simple `pass` with a brief comment; the current form looks confusing on first read.
- `max(usage.output_tokens, …)` guard is present for `output_tokens` but `reasoning_tokens` uses a simple assignment with `max(usage.reasoning_tokens, int(rt))`. Double-check that the `max` is indeed applied (it is — `usage.reasoning_tokens = max(usage.reasoning_tokens, int(rt))`); just confirm the order of conditions is correct if `reasoning_tokens` can arrive in `message_start` too.

**Related**
- Complements #2918 (openclaw adapter path for the same token family). Both touch different files.

**Issue link**
- Closes #2842 ✓

---

vivekchand (commented):
## Test plan & review notes

**What changed**
- `StreamUsage` gains a `reasoning_tokens` field; the Anthropic SSE parser now reads `thinking_tokens`/`reasoning_tokens`/`thinking_input_tokens` from `message_delta.usage` (taking the max across deltas); the proxy DB schema gets an idempotent `ALTER TABLE ... ADD COLUMN reasoning_tokens` migration; the zero-config interceptor emits `reasoning_tokens` in JSONL events.

**Smoke commands**
- `python3 -m pytest tests/test_proxy_cache_risk.py -v` — 5 new SSE/reasoning tests should pass.
- Start the proxy (`python3 -m clawmetry.proxy` or via `clawmetry`), then inspect the `proxy_usage` SQLite table: it should have a `reasoning_tokens` column after the first startup.
- Intercept a real extended-thinking stream (e.g. `claude-opus-4-8` with `thinking: {type: "enabled", budget_tokens: 1000}`): check the emitted JSONL event for a `reasoning_tokens` field > 0.

- The `or`-chain for reasoning token keys (`u.get("thinking_tokens") or u.get("reasoning_tokens") or ...`) is falsy on `0` — if the API returns `"thinking_tokens": 0` and `"reasoning_tokens": 150`, the `0` falls through correctly to pick up 150. No issue in practice, but worth noting that a future key returning `0` when another key has the real value would silently lose the count.
- The `content_block_delta` handler body is a no-op (`_ = delta.get("type")`) — it's a placeholder; confirm it doesn't inadvertently suppress a real delta token-count if a future API version embeds per-delta reasoning counts there.
- **Overlap with #2905**: both PRs touch `#2842` (SSE reasoning token reconciliation). Merging both may double-apply some of the same logic in `proxy.py`. Worth aligning before merging.

**Issue link**
- Branch name suggests this targets #2842; no explicit `Closes` in the PR body — worth adding `Closes #2842` if that's intended.

---

Issue #1619, opened 2026-05-17
[PRD] Phase 1 Evals — local-first LLM-as-judge on completed sessions (the MOAT-aligned differentiator)


vivekchand:
**Phase 1 of 4** for ClawMetry evals (per coverage-gap analysis 2026-05-17 vs OpenClaw blog's Pillar #4 Outcome Measurement). User-approved: ship Phase 1 first so we can claim "evals" in marketing within the week.

## Strategic positioning
Every competitor (LangSmith, Langfuse, Phoenix, Helicone) cloud-hosts their eval product — sending your prompts + responses to their servers for scoring. **ClawMetry does it LOCAL-FIRST**: scoring runs on the user's existing API key, results land in their local DuckDB, never touch our cloud. This is the MOAT story applied to evals.

Marketing line: *"Evals where your prompts never leave your machine."*

## Phase 1 scope (1-day ship)
Auto-scoring with LLM-as-judge on every completed session, using a configurable rubric.

### Architecture
1. **New module** `clawmetry/eval_runner.py` (~300 LOC):
   - `EvalRunner` class with `score_session(session_id, rubric_name='default') -> EvalResult`
   - Pulls session transcript from DuckDB, formats prompt for judge model, calls LLM, parses score
   - Default rubric (configurable in `~/.clawmetry/evals.yaml`):
     ```yaml
     default:
       judge_model: claude-haiku-4-5
       prompt: |
         You're evaluating an AI agent's response. Score 0-5:
         - 5: Fully addressed user's request, correct + complete
         - 4: Mostly correct, minor gaps
         - 3: Partial answer, missed key points
         - 2: Misunderstood the request
         - 1: Wrong/harmful answer
         - 0: Failed to respond
         Output: SCORE: <0-5>\nREASON: <one sentence>
     ```
   - Cost guard: skip sessions <10 tokens (trivial), cap at 100 sessions/hour to bound spend

2. **DuckDB schema** (`clawmetry/local_store.py`):
   - Add columns to `sessions`: `eval_score FLOAT, eval_reason TEXT, eval_judge_model TEXT, eval_scored_at BIGINT, eval_rubric TEXT`
   - Idempotent migration

3. **Scheduler** (background thread in daemon, sister of `_flusher_loop`):
   - Every 5 min: find sessions where `eval_score IS NULL AND outcome IN ('success', 'failed', 'escalated')` (skip ongoing) AND `ended_at > NOW() - 24h`
[...]

PR #2920, opened 2026-06-08
feat(eval): deterministic/code evaluators for live sessions (#2862)

### LINKED ISSUE #2862: feat: code/deterministic evaluators on live sessions
Let users attach cheap deterministic checks to live sessions: JSON-parseable, JSON-schema match, required tool arguments present, regex/exact match, output length bounds. Returns native scores that flow into the same views as the LLM-judge evals (#1619).

Why now: Langfuse shipped Code Evaluators 2026-05-28; PostHog shipped pass/fail evals. Deterministic checks run in milliseconds with zero LLM cost — they catch structural failures (malformed tool calls, missing args) that the LLM-judge is overkill for. Complements, doesn't replace, #1619.\n\nScope: small Python/TS evaluator runtime over completed sessions, library of built-ins, scores join existing eval pipeline.\n\nSource: Langfuse code-evaluators changelog, weekly feature intel 2026-06-08. Relates to #1619.

vivekchand:
Closes #2862

## What
Adds a **deterministic, code-based evaluator runtime** that complements the LLM-as-judge surface (#1619). Where the judge spends an LLM call to score *quality*, these checks run a pure Python check in milliseconds with **zero LLM cost** to catch *structural* failures — malformed JSON, a missing required tool argument, an empty/runaway answer, a regex/exact mismatch.

Built-in evaluators (all FREE tier — run on local data, cost nothing):
- `json-parseable` — output parses as JSON
- `json-schema-match` — required keys + scalar type check (small schema, no JSON-Schema dep)
- `regex-match` — pattern search/full-match, optional ignore-case
- `exact-match` — equals expected (strip / ignore-case options)
- `output-length-bounds` — min/max char bounds (catches empty + runaway)
- `required-tool-args` — every call to a tool supplied the required args (catches malformed tool calls)
- `no-tool-errors` — the session recorded no tool/turn error

## How
- `clawmetry/deterministic_evaluators.py`:
  - `EvalInput` (normalized, storage-agnostic: output text + tool calls + error flag) and `DeterministicEvalResult` (0..1 `score` so verdicts line up with the LLM-judge 0..1 range and flow into the same views).
  - `BUILTIN_EVALUATORS` registry + `run_checks(input, checks)` runner with a **never-crash contract**: unknown slugs and throwing evaluators become failing results, never exceptions, so one bad check can't abort the pass.
  - `eval_input_from_events(events)` — best-effort extractor that tolerates both `Event` dataclasses and their `to_dict()` shape.
- Pure, no DuckDB / network / clock — trivially unit-testable and safe to run inline on a live session.

## Scope
This PR delivers the **runtime + built-in library** (the issue's "small Python evaluator runtime over completed sessions, library of built-ins"). Wiring scores onto the sessions row + dashboard surfacing alongside the existing eval pipeline is the natural follow-up.

## Tests
`tests/test_deterministic_evaluators.py` — 29 tests covering each built-in (pass + fail paths), the 0/1 score shape, the runner's never-crash contract, and the event extractor. All green locally:

```
29 passed in 1.64s
```

### DIFF
--- a/clawmetry/deterministic_evaluators.py
+++ b/clawmetry/deterministic_evaluators.py
@@ -1,0 +1,355 @@
+"""
+clawmetry/deterministic_evaluators.py — cheap, code-based checks on sessions.
+
+Companion to the LLM-as-judge surface (``clawmetry/eval_runner.py``, #1619):
+where the judge spends an LLM call to score *quality*, these evaluators run a
+deterministic Python check in milliseconds with **zero LLM cost** to catch
+*structural* failures — malformed JSON, a missing required tool argument, an
+answer that is empty or runaway-long, a regex/exact mismatch. (#2862)
+
+Why both: the LLM judge is overkill for "did the agent return parseable JSON?"
+or "did the `create_file` call include a `path`?". Those are pass/fail facts a
+regex or a `json.loads` answers for free. Deterministic checks complement the
+judge — they do not replace it — and their scores share the same 0..1 shape so
+they flow into the same eval views (a 1.0 pass / 0.0 fail).
+
+Design rules:
+  * **Pure + storage-agnostic.** Each evaluator is a function of an
+    :class:`EvalInput` (the normalized output text + tool calls + error flag)
+    and a small ``config`` dict. No DuckDB, no network, no clock — so they are
+    trivially unit-testable and safe to run inline on a live session.
+  * **Never raises.** A check that hits bad input returns a failing
+    :class:`DeterministicEvalResult` with a human reason, never an exception —
+    a broken check must not take down the eval pass.
+  * **Built-ins are FREE-tier.** They run on the user's box over local data and
+    cost nothing, so they ship in OSS (no entitlement gate).
+
+Public API:
+    EvalInput(output_text=..., tool_calls=[...], had_error=False)
+    eval_input_from_events(events) -> EvalInput        # best-effort extractor
+    BUILTIN_EVALUATORS: dict[str, callable]
+    run_checks(eval_input, checks) -> list[DeterministicEvalResult]
+"""
+
+from __future__ import annotations
+
+import json
+import logging
+import re
+from dataclasses import dataclass, field, asdict
+from typing import Any, Callable
+
+log = logging.getLogger("clawmetry.deterministic_evaluators")
+
+
+# ── Normalized input + result shapes ────────────────────────────────────────────
+
+
+@dataclass
+class EvalInput:
+    """What a deterministic check looks at, decoupled from storage.
+
+    ``output_text`` is the agent's final/assistant text. ``tool_calls`` is a
+    list of ``{"name": str, "arguments": dict}`` (other keys ignored).
+    ``had_error`` flags whether any tool result / event errored.
+    """
+
+    output_text: str = ""
+    tool_calls: list[dict[str, Any]] = field(default_factory=list)
+    had_error: bool = False
+
+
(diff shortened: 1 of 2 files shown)


vivekchand (commented):
## Test plan & review notes

**What changed**
- New `clawmetry/deterministic_evaluators.py`: 7 zero-cost structural evaluators (`json-parseable`, `json-schema-match`, `regex-match`, `exact-match`, `output-length-bounds`, `required-tool-args`, `no-tool-errors`) with a shared `EvalInput` / `DeterministicEvalResult` shape (0.0/1.0 scores, aligning with the existing LLM-judge surface). New `tests/test_deterministic_evaluators.py`: 29 tests covering each built-in and the runner's never-crash contract.

**Smoke commands**
```bash
# Pure unit tests — no server, no DuckDB needed
pytest tests/test_deterministic_evaluators.py -v     # should show 29 passed

make lint      # syntax + import check
make test      # full suite
```

**What to look at**
- No dashboard surface yet (the PR description notes wiring scores onto the sessions row is a follow-up). Manual verification is just running the test file.

- `eval_input_from_events` imports `clawmetry.adapters.base.Event` for the dataclass path test. If that import path isn't resolvable in the test environment the test will `ImportError` rather than skip gracefully — worth checking CI.
- `eval_required_tool_args` treats empty-string values (`""`) as missing. A tool that legitimately passes `{"path": ""}` would fail this check. Low risk, but worth a docstring note.
- `json-schema-match` silently skips unknown type-name strings in `config["types"]` rather than erroring. A typo like `"integr"` would make the type check a no-op with no warning.
- `eval_input_from_events` uses "last non-empty content wins" for `output_text`, so intermediate assistant turns are silently dropped in multi-turn sessions.

**Issue link**
- Closes #2862 ✓

---

vivekchand (commented):
## Test plan & review notes

**What changed**
- Adds `clawmetry/deterministic_evaluators.py` — a pure-Python, zero-LLM eval library with 7 built-in checks (`json-parseable`, `json-schema-match`, `regex-match`, `exact-match`, `output-length-bounds`, `required-tool-args`, `no-tool-errors`), a `run_checks()` runner, and a `eval_input_from_events()` extractor that works with both dict-shaped and dataclass `Event` objects. No HTTP surface in this PR (library only).

**Smoke commands**
- `python3 -m pytest tests/test_deterministic_evaluators.py -v` — all 24 tests should pass including the `test_builtin_catalogue_slugs_are_callable` contract test.
- Quick import smoke: `python3 -c "from clawmetry.deterministic_evaluators import run_checks, EvalInput; print(run_checks(EvalInput(output_text='{\"ok\":1}'), [{'slug':'json-parseable'}]))"` — should print a passing result, no exceptions.

- `test_eval_input_from_events_reads_dataclass_events` imports `from clawmetry.adapters.base import Event` directly — if `Event`'s constructor signature has changed, this test will fail at setup time rather than assertion time.
- `_get(ev, "tool_calls", None)` on a dataclass `Event` that doesn't have a `tool_calls` attribute returns `None` silently (via `getattr` default) — correct, but worth confirming `Event` doesn't have a `tool_calls` property that raises.
- No HTTP endpoint here — confirm the follow-up wiring (likely `routes/usage.py` or `eval_runner.py`) is tracked and the library's public surface (`run_checks`, `BUILTIN_EVALUATORS`, `eval_input_from_events`) matches what the wiring PR will call.

**Issue link**
- Closes #2862 ✓

---

PR #2928, opened 2026-06-09
feat(search): add GET /api/local/search — session full-text search (#2860)

### LINKED ISSUE #2860: feat: full-text search across session prompts & responses
Add full-text search over prompt/response/tool-arg content in the local event store (DuckDB FTS extension). Filter by model, time, cost, status alongside the text query.

Why now: Langfuse shipped Full-Text Search 2026-05-27; it's becoming table-stakes. Users debugging a bad run need to find "that session where the agent looped on the wrong file" by content, not just by id/time.\n\nScope: DuckDB FTS index on content columns, search box on the sessions/trace view, highlight matches.\n\nSource: Langfuse changelog, weekly feature intel 2026-06-08.

vivekchand:
Closes #2860

## What

Adds `GET /api/local/search?q=<text>` to ClawMetry's local query API so users can find sessions by content instead of scrolling by time or guessing session IDs.

## How

- **`clawmetry/local_store.py`** — new `query_search()` method. Searches `sessions.title ILIKE '%q%' OR sessions.eval_reason ILIKE '%q%'`. Optional `model` filter subqueries `events.model`. Supports `status`, `since`, `until`, `limit`. Returns session-summary rows sorted newest-first. Errors are swallowed (returns `[]`) — consistent with every other query method.

- **`routes/local_query.py`** — four small additions:
  1. `"search": "query_search"` in `_SHAPES` (makes it available over the WS relay too)
  2. `search` branch in `_coerce_args()` — validates non-empty `q`, accepts `model`, `status`, `since`, `until`, `limit`; returns 400 on empty/missing `q`
  3. `GET /api/local/search` HTTP handler
  4. `"query_search"` in `_DAEMON_METHODS` — multi-process installs proxy through the daemon correctly

- **`tests/test_search.py`** — 8 tests: title hit, eval_reason hit, case-insensitive, no-match, empty-query guard, model filter, status filter, result field shape.

## Test plan

```
pytest tests/test_search.py -v   # 8 passed
python3 -m py_compile clawmetry/local_store.py routes/local_query.py
```

## Scope note

Searches `title` and `eval_reason` (both uncompressed VARCHAR columns). The `events.data` BLOB contains raw prompts/responses but is Python-compressed before hitting DuckDB — deep content indexing is the natural follow-up once a denormalized text column is added at ingest time.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Xe3a1qefkmTnzCdNqa54HD)_

### DIFF
--- a/clawmetry/local_store.py
+++ b/clawmetry/local_store.py
@@ -7159,6 +7159,75 @@ class LocalStore:
             return []
         return [dict(zip(cols, r)) for r in rows]
 
+    def query_search(
+        self,
+        *,
+        q: str,
+        model: str | None = None,
+        status: str | None = None,
+        since: str | None = None,
+        until: str | None = None,
+        limit: int = 50,
+    ) -> list[dict[str, Any]]:
+        """Search sessions by title and eval_reason text.
+
+        Matches rows where ``sessions.title`` or ``sessions.eval_reason``
+        contains the query string (case-insensitive). Optional ``model``
+        filter restricts to sessions that used a specific model name
+        (sub-selects session_ids from the events table). Returns session
+        summary rows sorted newest-first.
+        """
+        q = (q or "").strip()
+        if not q:
+            return []
+        q_like = f"%{q}%"
+        params: list[Any] = []
+
+        model_join = ""
+        if model:
+            model_join = (
+                "JOIN (SELECT DISTINCT session_id FROM events "
+                "WHERE model ILIKE ? AND session_id IS NOT NULL) em "
+                "ON s.session_id = em.session_id"
+            )
+            params.append(f"%{model}%")
+
+        clauses: list[str] = ["(s.title ILIKE ? OR s.eval_reason ILIKE ?)"]
+        params.extend([q_like, q_like])
+        if status:
+            clauses.append("s.status = ?")
+            params.append(str(status))
+        if since:
+            clauses.append("s.last_active_at >= ?")
+            params.append(since)
+        if until:
+            clauses.append("s.last_active_at <= ?")
+            params.append(until)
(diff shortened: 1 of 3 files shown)


vivekchand (commented):
## Test plan & review notes

**What changed**
- Adds `LocalStore.query_search()` (parameterised DuckDB `ILIKE` across `sessions.title` + `sessions.eval_reason`) and wires it to `GET /api/local/search?q=<term>` with optional `model`, `status`, `since`, `until`, `limit` filters; registers `query_search` in the daemon proxy allowlist.

**Smoke commands**
- `python3 -m pytest tests/test_search.py -v`
- `curl -sS "http://localhost:8900/api/local/search?q=debug"` — expect a JSON array (possibly empty), never a 500.
- `curl -sS "http://localhost:8900/api/local/search"` — expect `{"error":"search shape requires non-empty 'q' parameter"}` with HTTP 400.
- `curl -sS "http://localhost:8900/api/local/search?q=x&limit=5"` — expect at most 5 rows.

- **`sessions.eval_reason` schema**: if an existing install's DuckDB schema predates this column, the `SELECT` will fail. The `except Exception: return []` fallback silently swallows this — users upgrading in-place may see empty search results until the schema migrates. Worth verifying the migration adds this column before the handler runs.
- The `model` filter uses `INNER JOIN` semantics, so a session that matches the title but has no events using that model won't appear in results — correct by intent, but worth a note in docs/comments for future maintainers.
- **Daemon restart required**: `"query_search"` is correctly added to `_DAEMON_ALLOWLIST`, but the running daemon must restart to pick it up if the set is loaded at startup.

**Issue link**
- Closes #2860 ✓

---

PR #3048, opened 2026-06-12
fix(proxy): honor ProxyConfig.enabled as a master pass-through switch

vivekchand:
Addresses vivekchand/clawmetry-cloud#1577 (hardening roundup — proxy `enabled` dead no-op item).

## What
`ProxyConfig.enabled=false` was a dead no-op. `proxy_request()` never consulted the top-level `enabled` flag, so a "disabled" proxy still ran the full enforcement pipeline: auto smart routing, budget check + downgrade, loop / rate / cost-spiral / velocity breakers, cache-risk + tool-churn detection, HITL pause, and explicit model routing. Every breaker has its *own* `enabled` flag, but the master switch did nothing.

## How
Wrapped the entire enforcement/model-rewrite block in `if config.enabled:` so that when the proxy is disabled it acts as a pure pass-through — the request is forwarded upstream unchanged with no breakers, detectors, or model rewriting. The forward-to-upstream path is unchanged and runs in both modes.

Best reviewed with whitespace ignored (`?w=1`) — the bulk of the diff is the indent of the now-guarded block.

## Tests
New `TestProxyDisabledPassthrough` in `tests/test_proxy.py`:
- `test_enabled_budget_blocks` — sanity guard: with enforcement ON, an over-budget request returns `429 budget_exceeded`.
- `test_disabled_bypasses_enforcement` — with enforcement OFF, the same over-budget request is forwarded and fails only at the upstream API-key step (`401 authentication_error`), proving the budget breaker was skipped.

Full proxy suite green: `108 passed` across `test_proxy.py`, `test_proxy_enforcement.py`, `test_proxy_budget_abort.py`, `test_proxy_cache_risk.py`.

### DIFF
--- a/clawmetry/proxy.py
+++ b/clawmetry/proxy.py
@@ -2106,253 +2106,260 @@ def create_proxy_app(config: ProxyConfig = None) -> "Flask":
                 f"Proxy request: {provider} {path} model={model} stream={is_streaming}"
             )
 
-        # ── Auto smart routing (#2816) ─────────────────────────────────
-        # Applied FIRST so cheap-task requests start on a cheaper model. Budget
-        # action=downgrade (below) and explicit ModelRouter rules (later) both
-        # run AFTER this, so they take precedence and override the auto choice.
-        auto_model, auto_reason = auto_router.route(model, body)
-        if auto_model:
-            from_model = model
-            est_input = _estimate_tokens(_body_user_text(body)) + _estimate_tokens(
-                _body_system_text(body)
-            )
-            saved = _estimate_saved_usd(from_model, auto_model, est_input)
-            model = auto_model
-            body["model"] = model
-            body_bytes = json.dumps(body).encode()
-            with _stats_lock:
-                _stats["auto_downgraded"] += 1
-            db.record_event(
-                "auto_downgraded",
-                auto_reason,
-                severity="info",
-                details={
+        # ── Master enable switch (clawmetry-cloud#1577) ───────────────
+        # When the proxy is disabled, act as a pure pass-through: skip every
+        # breaker, detector, and model-rewrite below and forward the request
+        # upstream unchanged. Each breaker has its own `enabled` flag, but the
+        # top-level ProxyConfig.enabled was never consulted here, so flipping
+        # it off left full enforcement running — a dead no-op until now.
+        if config.enabled:
+            # ── Auto smart routing (#2816) ─────────────────────────────────
+            # Applied FIRST so cheap-task requests start on a cheaper model. Budget
+            # action=downgrade (below) and explicit ModelRouter rules (later) both
+            # run AFTER this, so they take precedence and override the auto choice.
+            auto_model, auto_reason = auto_router.route(model, body)
+            if auto_model:
+                from_model = model
+                est_input = _estimate_tokens(_body_user_text(body)) + _estimate_tokens(
+                    _body_system_text(body)
+                )
+                saved = _estimate_saved_usd(from_model, auto_model, est_input)
+                model = auto_model
+                body["model"] = model
+                body_bytes = json.dumps(body).encode()
+                with _stats_lock:
+                    _stats["auto_downgraded"] += 1
+                db.record_event(
+                    "auto_downgraded",
+                    auto_reason,
+                    severity="info",
+                    details={
+                        "from_model": from_model,
+                        "to_model": model,
+                        "session_id": session_id,
+                        "estimated_saved_usd": saved,
+                    },
+                )
+                _emit_duckdb_event("auto_downgraded", session_id, {
                     "from_model": from_model,
                     "to_model": model,
-                    "session_id": session_id,
                     "estimated_saved_usd": saved,
-                },
-            )
-            _emit_duckdb_event("auto_downgraded", session_id, {
-                "from_model": from_model,
-                "to_model": model,
-                "estimated_saved_usd": saved,
-                "reason": auto_reason,
-            })
-            if config.log_requests:
-                logger.info("Auto-route: %s", auto_reason)
+                    "reason": auto_reason,
+                })
+                if config.log_requests:
+                    logger.info("Auto-route: %s", auto_reason)
 
-        # ── Budget check ───────────────────────────────────────────────
-        allowed, reason = budget.check(model, session_id)
-        if not allowed:
-            with _stats_lock:
-                _stats["blocked"] += 1
-            db.record_event(
-                "budget_blocked",
-                reason,
-                severity="warning",
-                details={"model": model, "session_id": session_id},
-            )
-            logger.warning(f"Budget blocked: {reason}")
+            # ── Budget check ───────────────────────────────────────────────
+            allowed, reason = budget.check(model, session_id)
+            if not allowed:
+                with _stats_lock:
+                    _stats["blocked"] += 1
+                db.record_event(
+                    "budget_blocked",
+                    reason,
+                    severity="warning",
+                    details={"model": model, "session_id": session_id},
+                )
+                logger.warning(f"Budget blocked: {reason}")
 
-            if config.budget.action == "downgrade":
+                if config.budget.action == "downgrade":
+                    original_model = model
+                    model = config.budget.downgrade_model
+                    body["model"] = model
+                    body_bytes = json.dumps(body).encode()
+                    db.record_event(
+                        "model_downgraded",
+                        f"Downgraded {original_model} -> {model} (budget)",
+                        severity="info",
+                        details={"original": original_model, "downgraded_to": model},
+                    )
+                elif config.budget.action == "warn":
+                    pass  # Allow through with warning
+                else:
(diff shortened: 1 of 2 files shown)


vivekchand (commented):
## Test plan & review notes

**What changed**
- `clawmetry/proxy.py`: wraps the entire enforcement block (auto-routing, budget, loop/rate/cost-spiral breakers, cache-risk, tool-churn, velocity, HITL) inside `if config.enabled:` — the upstream forward always runs regardless, so a disabled proxy is now a true pass-through instead of silent full-enforcement.
- `tests/test_proxy.py`: new `TestProxyDisabledPassthrough` class pins the contract with two tests.

**Smoke commands**
```bash
make test                              # full suite — catches any import regressions
pytest tests/test_proxy.py::TestProxyDisabledPassthrough -v  # focused
```

**What to look at visually**
- Start proxy with `ProxyConfig(enabled=False)` and verify a request exceeding budget gets a 2xx (or upstream-determined response), not a 429.
- Start proxy with `ProxyConfig(enabled=True)` and verify the same over-budget request gets 429 `budget_exceeded`.

- The entire enforcement block is now one indentation level deeper — any reviewer should eyeball that the `# ── Forward to upstream ──` line is correctly **outside** the `if config.enabled:` block (it is; the diff shows it unindented). Worth a quick look to confirm no mis-indented early-return slipped in.
- `config.enabled` defaults: verify the default in `ProxyConfig` is `True` so existing deployments that never set the flag keep full enforcement.

**Issue link**
- Branch references `gh-cloud-1577`; adding `Closes vivekchand/clawmetry-cloud#1577` to the PR description would auto-close the cloud tracker on merge (cross-repo syntax required).

---

vivekchand (commented):
## Test plan & review notes

**What changed**
- Wrapped the entire enforcement pipeline in `clawmetry/proxy.py` under `if config.enabled:` so `ProxyConfig.enabled=False` now acts as a true master pass-through switch — previously the flag was read by no code path and all breakers (budget, loop, rate, cost-spiral, velocity, HITL, model-routing) still ran when "disabled."

**Smoke commands**
- `make test` / `make test-api`
- `python3 -m pytest tests/test_proxy.py tests/test_proxy_enforcement.py tests/test_proxy_budget_abort.py tests/test_proxy_cache_risk.py -v` (PR claims 108 passed)
- Specifically: `python3 -m pytest tests/test_proxy.py::TestProxyDisabledPassthrough -v` for the two new regression tests
- With a running proxy configured `enabled: false`: `curl -s -X POST http://localhost:4100/v1/messages -H 'Content-Type: application/json' -d '{"model":"claude-3-5-sonnet-20241022","messages":[{"role":"user","content":"hi"}]}'` — should reach the upstream (401 auth error), not a 429 budget/loop block

**Likely failure modes**
- The `if config.enabled:` block runs before `_get_api_key(provider, req_headers)`, so a disabled proxy still resolves the api key. Double-check that `new_provider` (from `router.route`) is not used in the forwarding step when `config.enabled=False` — the model routing block is now inside the `if`, so `provider` stays as-detected from the request header, which is the correct pass-through behaviour, but worth a quick eyeball that streaming vs non-streaming upstream URL construction doesn't rely on a `new_provider` value set only inside the guard.
- If any caller instantiates `ProxyConfig` without an explicit `enabled=` kwarg, it defaults to whatever was set before — confirm the dataclass default is `True` (enforcement on by default) so existing deployments aren't silently disabled.

**Issue link**
- PR body references `clawmetry-cloud#1577`. Suggest adding `Closes clawmetry/clawmetry-cloud#1577` to the description so the cloud issue auto-closes on merge.

---

vivekchand:
✨ auto-fixed: rebased onto current main (was BEHIND by several commits)

---

vivekchand (commented):
## Test plan & review notes

**What changed**
- Wraps the entire enforcement block in `proxy_request` (auto-routing, budget, loop detection, rate/cost-spiral breakers, cache-risk, tool-churn, velocity, HITL pause, and model routing) in a single `if config.enabled:` guard, so `ProxyConfig.enabled=False` is now a true master pass-through that skips all proxy enforcement — previously the flag was never consulted here and remained a dead no-op, leaving full enforcement running even when "disabled".

**Smoke commands**
- Start proxy with `enabled=False` in config: `curl -sS http://localhost:4100/v1/messages` — should pass through without budget/loop checks firing

- If `enabled=False` check happens after connection state is established, there's still overhead; ensure it short-circuits early in the request path (it does — the guard wraps the first operation after logging, so this looks good)
- Ensure that disabling the proxy doesn't accidentally skip auth header forwarding — pass-through should be transparent, not stripped. The `# ── Forward to upstream ──` block sits *outside* the `if config.enabled:` guard, so headers are still forwarded correctly.
- The new tests confirm the contract cleanly: `test_enabled_budget_blocks` verifies enforcement still works when on, and `test_disabled_bypasses_enforcement` verifies the pass-through reaches the upstream (401 auth error rather than 429 budget-blocked).

**Cross-repo note**
- Branch name `fix/gh-cloud-1577` confirms the triggering issue lives in the cloud repo (`clawmetry-cloud#1577`). If there's corresponding cloud-side handling needed (e.g. the UI toggle or API surface that sets `enabled=False`), verify it's covered in a cloud PR.

**Issue link**
- If no `Closes` in PR body: add `Closes vivekchand/clawmetry-cloud#1577` (or the appropriate cross-repo close syntax) to auto-close the originating issue on merge.


---

vivekchand:
## Test plan & review notes

**PR**: fix(proxy): honor ProxyConfig.enabled as a master pass-through switch

---

### What changed

When `ProxyConfig.enabled = False`, all traffic now passes through without budget/loop enforcement. Previously the top-level master switch was never consulted in `proxy_request()` — every sub-breaker (loop, rate, cost-spiral, velocity, HITL, model routing, auto smart routing, cache-risk, tool-churn) has its own `enabled` flag, but the `ProxyConfig.enabled` field was a dead no-op. The fix wraps the entire enforcement/model-rewrite block in `if config.enabled:` so a disabled proxy acts as a pure pass-through. The forward-to-upstream path is unchanged and runs in both modes.

---

### Smoke commands

```bash
# Full proxy test suite
python -m pytest tests/test_proxy.py tests/test_proxy_enforcement.py \
    tests/test_proxy_budget_abort.py tests/test_proxy_cache_risk.py -v

# Quick API test pass
make test-api

# Verify pass-through manually (port 4100, no upstream key)
python3 -c "
from clawmetry.proxy import ProxyConfig, BudgetConfig, create_proxy_app
cfg = ProxyConfig(enabled=False, budget=BudgetConfig(daily_usd=0.01, action='block'))
app = create_proxy_app(cfg)
print('enabled:', cfg.enabled)  # should be False
"
# Then: send a request to port 4100 with an intentionally blown budget;
# expect 401 authentication_error (upstream), NOT 429 budget_exceeded (proxy).
```

---

### Diff review notes

**The approach is correct.** The single `if config.enabled:` guard at the top of the enforcement section cleanly covers all sub-checks: auto smart routing, budget check + downgrade, loop detection, rate breaker, cost-spiral breaker, cache-bust risk, tool-churn detection, velocity check, HITL/backoff pause, and model routing. The upstream forward block falls outside the guard and runs in both modes — exactly right.

**One thing to confirm: the default value of `ProxyConfig.enabled`.** The PR description says this was a dead no-op, implying the field existed before. Verify that the default is `True` (not `False` or `None`) so that users who never set `enabled` explicitly get the same enforcement behavior as before — a default of `False` here would silently disable all enforcement for all existing deployments.

**Possible edge case — `_SESSION_PREFIX` and `_SESSION_TOOLS_NORM` state not populated when disabled.** These module-level dicts track per-session prefix hashes and tool fingerprints for drift/churn detection. When `enabled=False`, those dicts are never written, so if a user re-enables the proxy mid-run the first few turns after re-enabling won't have a prior-turn baseline to diff against. This is acceptable (detection is advisory, never blocking) but worth a comment in the code.

**No logical gaps found** — the enforcement block exit paths (all `return _error_response(...)` or `return _budget_abort_response(...)`) are all inside the `if config.enabled:` block, so a disabled proxy can never return a 429 from enforcement.

---

### CI status

All 19 checks passing (Syntax & Lint, MOAT Verifier 72 tests, MOAT Keystone 13-endpoint bar, API Tests on ubuntu/macos/windows, pip install on all platforms, E2E Browser Tests, Eval Suite Gate, Compression Safety Eval, OSS golden path, Live OpenClaw E2E, Cross-repo handoff C4, PR build, Install + boot + health-check, Wheel install & asset presence).

---

### Issue link

Cross-repo: vivekchand/clawmetry-cloud#1577 (hardening roundup — proxy `enabled` dead no-op item). No matching open issue found in this repo.

---

### Verdict

Code change is clean and well-scoped. New `TestProxyDisabledPassthrough` tests pin the contract with both a positive (enforcement ON → 429) and negative (enforcement OFF → 401 passthrough) case. Ready to merge pending confirmation that `ProxyConfig.enabled` defaults to `True`.

---

PR #3022, opened 2026-06-11
fix(c6): add 5th required check + exit 1 when not configured (rebased onto main)

vivekchand:
## What

Supersedes stale PR #2975 (same change, rebased onto current main `45fce4a`).

Three-file change, ~20 LOC delta.

---

### `scripts/apply_required_status_checks.py`

**Add 5th required check:**

```python
REQUIRED_CHECKS = [
    ("clawmetry",         "OSS golden path (wheel + OpenClaw + 9 tabs)"),
    ("clawmetry",         "Cross-repo handoff (C4)"),
    ("clawmetry",         "E2E Browser Tests (critical subset)"),  # NEW
    ("clawmetry-cloud",   "Cloud golden-path browser E2E"),
    ("clawmetry-landing", "Landing golden path (C3)"),
]
```

The `e2e-critical` job in `ci.yml` runs on every PR with no `paths:` filter and is the 32-tab auth-overlay gate (C5). Without it as a required check, a PR breaking any of the 32 tabs shows a failing CI job but is still mergeable.

**Change exit behavior in read-only path (forcing signal):**

Before:
```python
    # Always exit 0: push-triggered runs are informational, never blocking.
    return
```

After:
```python
    if verify_required_checks(...):
        return       # configured: exit 0
    print("ERROR: Required checks not yet configured...")
    sys.exit(1)      # not configured: red job on every main push
```

Now the "Apply required E2E status checks (C6)" job turns red on every push to main until an admin runs it with a PAT. Visible forcing signal instead of a silent pass.

### `scripts/close-c6.sh`

Updated banner and comments from "4 required status checks" to "5 required status checks", added `E2E Browser Tests (critical subset)` to the printed list.

### `.github/workflows/apply-required-checks.yml`

Updated dry-run preview echo blocks to include the 5th check.

---

## After merging: one action closes C6

```bash
bash scripts/close-c6.sh
```

Or from the browser (no local clone needed):
1. Create a fine-grained PAT: `github.com/settings/personal-access-tokens/new`
   Repos: `clawmetry`, `clawmetry-cloud`, `clawmetry-landing` | Permission: Administration (read+write)
2. Actions > "Apply required E2E status checks (C6 -- one-shot)" > Run workflow
   `confirm=APPLY`, `pat_token=<paste token>` > Run workflow

Closes C6 for all 3 repos in a single run.

---

## Current status

[...]

PR #3051, opened 2026-06-12
fix(c6): use branch endpoint for read-only C6 status verification [superseded by #3052]

vivekchand:
## Summary

- The push-triggered `apply-required-checks.yml` read-only path used `/branches/main/protection/required_status_checks`, which always returns 403 with `GITHUB_TOKEN` on public repos. Every push-to-main run showed `[clawmetry] FAIL: could not read required checks: GitHub GET => 403`, even after C6 is correctly configured by an admin.
- This PR adds `verify_required_checks_readonly()` that uses `GET /repos/{owner}/{repo}/branches/main` instead. Public repos return `protection.required_status_checks.contexts` via this endpoint with a bare `GITHUB_TOKEN`.
- Once C6 is applied via `bash scripts/close-c6.sh` or the workflow dispatch, the next push to main will show `[clawmetry] OK: ['Cross-repo handoff (C4)', 'OSS golden path (wheel + OpenClaw + 9 tabs)']` instead of a false-alarm FAIL.

## Test plan

- [ ] Trigger `apply-required-checks.yml` with default `dry-run` on this branch: the "Apply required status checks" step should be skipped (it only runs on `push` or `confirm=APPLY`), the preview step should show the 2 checks and instructions.
- [ ] After C6 is applied by an admin via `scripts/close-c6.sh`, the next push to main should show `[clawmetry] OK: [...]` in the workflow log rather than `FAIL: 403`.
- [ ] Existing PAT-based apply path (`confirm=APPLY` + admin token) is unchanged.

Tracking: vivekchand/clawmetry#2146 (C6)

---
_Generated by [Claude Code](https://claude.ai/code/session_011tYG2ktwq8DVLVpyAGwgwA)_

### DIFF
--- a/scripts/apply_required_status_checks.py
+++ b/scripts/apply_required_status_checks.py
@@ -174,6 +174,64 @@ def remove_required_check(repo: str, context: str, token: str) -> None:
     print(f"  [{repo}] removed deprecated check ({len(updated)} remaining): {context!r}")
 
 
+def _get_branch_protection_contexts(repo: str, token: str) -> set[str] | None:
+    """Read required status check contexts via the public branch endpoint.
+
+    Uses GET /repos/{owner}/{repo}/branches/main which is accessible with
+    GITHUB_TOKEN on public repos (unlike
+    /branches/main/protection/required_status_checks which requires admin).
+
+    Returns the set of configured context strings, or None if the branch
+    endpoint is unreachable or returns no protection data.
+    """
+    path = f"/repos/{OWNER}/{repo}/branches/main"
+    try:
+        data = _api("GET", path, token=token)
+    except RuntimeError as exc:
+        print(f"  [{repo}] WARN: could not read branch info: {exc}")
+        return None
+    protection = data.get("protection") or {}
+    rsc = protection.get("required_status_checks") or {}
+    return set(rsc.get("contexts", []))
+
+
+def verify_required_checks_readonly(
+    checks: list[tuple[str, str]],
+    deprecated: list[tuple[str, str]],
+    token: str,
+) -> bool:
+    """Read-only variant of verify_required_checks using the branch endpoint.
+
+    Called from the GITHUB_TOKEN (read-only) code path. Uses
+    GET /repos/{owner}/{repo}/branches/main which returns protection info
+    for public repos without requiring admin credentials.
+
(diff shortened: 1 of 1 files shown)


vivekchand (commented):
## Test plan & review notes

**What changed**
- Replaces the admin-only `/branches/main/protection/required_status_checks` endpoint with the public-readable `/repos/{owner}/{repo}/branches/main` endpoint for read-only C6 status verification, eliminating the spurious `FAIL: 403` on every main-branch push.

**Smoke commands**
- `make test` / `make test-api`
- Trigger `apply-required-checks.yml` via Actions > Run workflow with default `dry-run` — the "Apply required status checks" step should be skipped; the preview step should list the checks and print instructions without any 403 errors.
- After C6 is applied (via `bash scripts/close-c6.sh` or the PAT dispatch), push a commit to main and confirm the log shows `[clawmetry] OK: [...]` rather than `FAIL: could not read required checks: GitHub GET => 403`.

**Likely failure modes**
- Data shape mismatch: `GET /repos/{owner}/{repo}/branches/main` returns `protection.required_status_checks.contexts` only when branch protection is already enabled; if no protection rules exist at all, `protection` may be `{}` or absent. The new `_get_branch_protection_contexts` handles this via `.get("protection") or {}`, returning `None` if the key is missing — but that causes an `UNKNOWN` log line rather than a clean `FAIL`. Verify the `UNKNOWN` case is acceptable before C6 is first applied.
- Auth difference: the branch endpoint uses the same `GITHUB_TOKEN`; no auth change. But if the repo is ever made private, this endpoint will also 403. The fallback `WARN` + `return None` path in `_get_branch_protection_contexts` should handle it gracefully — worth confirming the calling code treats `None` as "unconfigured" rather than "configured".
- The `verify_required_checks_readonly` result now drives an INFO print but the read-only path still always exits 0 (`return` without `sys.exit(1)`). Compare with #3022 which changes the read-only path to `sys.exit(1)` when not configured — these two PRs may conflict if merged in either order.

**Issue link**
- Tracks vivekchand/clawmetry#2146 (C6) — confirm the Closes/Tracks link is present in the PR body (it is: "Tracking: vivekchand/clawmetry#2146").


---

vivekchand:
✨ auto-fixed: rebased onto current main (was BEHIND by several commits)

---

PR #3067, opened 2026-06-12
feat(sync): Query Spine P4 — per-method q/1 proactive cache push

### LINKED ISSUE #2990: [query-spine P4] Daemon dual-publish: per-method q/1 cache entries alongside the monolith snapshot
Part of the Query Spine plan. Depends on P1 + P2.

`sync.py` pushes each q/1 method response as its own keyed cache entry (`cache:{owner}:{node}:{method}:{args_hash}`) with per-method debounce: plaintext for aggregate-classed methods (glance, runtimes, models, usage, health), AES-256-GCM with the node key for content-classed methods (sessions, brain, transcript, session), same framing as the existing snapshot blob (nonce = first 12 bytes).

- Behind `CLAWMETRY_QUERY_CACHE_PUSH` (default off for one release, then on).
- The monolith snapshot keeps shipping unchanged until P9.
- CI guard (extract-then-judge): assert no E2E-classed method is ever pushed unencrypted, derived from the P1 registry.
- Verify: decrypt/read live entries for a real node and diff against local q/1 responses.

vivekchand:
Closes #2990

## Summary

- **`_build_q1_cache_pushes(config)`** (new, ~70 LOC in `sync.py`): iterates all live no-required-arg q/1 shapes, dispatches each via `_local_dispatch`/fallback, returns `cache_pushes` entries keyed `q1:{shape}:{owner_hash}:{node_id}`. Plaintext shapes (aggregates, health, runtimes, models) get base64url JSON blobs; e2e shapes (rollup_sessions) use `encrypt_payload()` and are silently skipped when no enc_key is set.
- **Heartbeat assembly**: new best-effort try/except call block inserted after the crons push, same pattern as all other cache push phases.
- **`_PENDING_SHAPES` expansion**: adds `runtimes`, `models`, `rollup_sessions`, `spans`, `traces`, `external_calls`, `search` — all live P2/P3 shapes that were missing from the on-demand dispatch allowlist.
- **`tests/test_sync_q1_cache_pushes.py`** (7 tests): e2e-trust CI guard, flag-off, key format, plaintext blob decodeability, required-arg skip, debounce, `_PENDING_SHAPES` coverage.

Gated by `CLAWMETRY_QUERY_CACHE_PUSH=1` (default off for one release). The monolith snapshot is unchanged until P9.

## Test plan

- [ ] `python -m pytest tests/test_sync_q1_cache_pushes.py tests/test_query_contract_drift.py -v` — all 20 tests pass
- [ ] Set `CLAWMETRY_QUERY_CACHE_PUSH=1` on a live node and confirm `cache_pushes` in the next heartbeat payload include `q1:aggregates:*`, `q1:health:*`, `q1:runtimes:*`, `q1:models:*` entries
- [ ] Confirm e2e shapes (`rollup_sessions`) appear only when `encryption_key` is set in config

---
_Generated by [Claude Code](https://claude.ai/code/session_01TjonATqWneKY6RkcSUCkd6)_

### DIFF
--- a/clawmetry/sync.py
+++ b/clawmetry/sync.py
@@ -6899,6 +6899,14 @@ def send_heartbeat(config: dict) -> bool:
             payload.setdefault("cache_pushes", []).extend(cron_pushes)
     except Exception as _cl_e:
         log.debug("crons cache_push build failed (continuing): %s", _cl_e)
+    # Query Spine P4 (#2990): per-method q/1 cache entries alongside the
+    # monolith snapshot. Gated by CLAWMETRY_QUERY_CACHE_PUSH (default off).
+    try:
+        q1_pushes = _build_q1_cache_pushes(config)
+        if q1_pushes:
+            payload.setdefault("cache_pushes", []).extend(q1_pushes)
+    except Exception as _q1_e:
+        log.debug("q1 cache_push build failed (continuing): %s", _q1_e)
     # Local-first: persist this heartbeat to local DuckDB so the dashboard
     # has a per-node liveness history even when offline. Best-effort.
     try:
@@ -6941,7 +6949,13 @@ def send_heartbeat(config: dict) -> bool:
 # Allowlist mirrors routes.local_query._SHAPES. Duplicated here so the
 # daemon stays safe when `routes/` isn't on sys.path (some installs run the
 # sync daemon without the dashboard module loaded).
-_PENDING_SHAPES = {"events", "sessions", "aggregates", "health", "transcript"}
+_PENDING_SHAPES = {
+    "events", "sessions", "aggregates", "health", "transcript",
+    # Added in P4 (#2990): new live shapes from P2 materialized rollups.
+    "runtimes", "models", "rollup_sessions",
+    # Added in P3 (#2989): rollup-backed span/trace shapes.
+    "spans", "traces", "external_calls", "search",
+}
 
 
 def _canonical_args_hash(args: dict) -> str:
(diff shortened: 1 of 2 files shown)


vivekchand:
## Test plan & review notes

**What changed**
- `clawmetry/sync.py`: adds `_build_q1_cache_pushes()` — pushes each no-required-arg live q/1 shape as a separate keyed cache entry alongside the monolith snapshot. Feature-flagged behind `CLAWMETRY_QUERY_CACHE_PUSH=1` (default off). Expands `_PENDING_SHAPES` to include `runtimes`, `models`, `rollup_sessions`, `spans`, `traces`, `external_calls`, `search`.
- `tests/test_sync_q1_cache_pushes.py`: 7 new tests covering the e2e-shapes-without-enc-key CI guard, feature-flag off default, cache key format (`q1:{shape}:{owner_hash}:{node_id}`), plaintext blob readability, required-arg skip, and per-method debounce.

**Smoke commands**
```bash
# Run the new test file
python3 -m pytest tests/test_sync_q1_cache_pushes.py -v
# Full suite
make test
```

**What to verify manually**
- With `CLAWMETRY_QUERY_CACHE_PUSH=1` set, trigger a heartbeat cycle and confirm `cache_pushes` in the payload includes `q1:*` entries (check sync debug log)
- With no enc_key, confirm no `rollup_sessions` entry appears in the push list

- `_local_dispatch_fallback` is referenced in `_build_q1_cache_pushes` but the fallback symbol must already exist in sync.py — confirm it's defined upstream in the file
- `_Q1_LAST_PUSH` is module-level global state; parallel test runs that import sync multiple times may bleed state — the test already monkeypatches it, which is correct
- `query_contract.py` is lazily imported and silently returns `[]` on `ImportError` — correct defensive behaviour, but worth confirming the module ships in the package

**Issue link**
- Closes #2990

---

vivekchand:
## Test plan & review notes

**What changed**
- Query Spine P4: after each heartbeat cycle the daemon proactively pushes a `q1:{shape}:{owner_hash}:{node_id}` cache entry for every no-required-arg live shape, avoiding cold-read latency on the first cloud dashboard load. Gated behind `CLAWMETRY_QUERY_CACHE_PUSH=1` (default off for this release). Plaintext-class shapes (aggregates, health, runtimes, models) are base64url JSON; e2e-class shapes (rollup_sessions) are AES-256-GCM encrypted and silently skipped when no `encryption_key` is set. `_PENDING_SHAPES` expanded with the 7 new P2/P3 live shapes.

**Smoke commands**
```bash
# New targeted test file (7 tests)
python3 -m pytest tests/test_sync_q1_cache_pushes.py -v
# Full suite
make test
# Live smoke: watch for q1:* entries in heartbeat debug log
CLAWMETRY_QUERY_CACHE_PUSH=1 clawmetry sync
# After one heartbeat cycle, first cloud dashboard load should be fast
curl -sS http://localhost:8900/api/overview
```

- **Writer-lock contention**: `_build_q1_cache_pushes` calls `_local_dispatch` which reads DuckDB. Confirm reads go through `local_server.py` (the daemon-proxy) rather than grabbing the writer lock directly — otherwise this could deadlock mid-heartbeat if called while the daemon holds the write lock.
- **`_local_dispatch_fallback` availability**: the fallback symbol is referenced inside `_build_q1_cache_pushes` before the import try/except resolves. Confirm `_local_dispatch_fallback` is defined earlier in `sync.py` and is not itself referencing a shape that hasn't been registered yet.
- **Network error on push**: the outer `try/except Exception` in `send_heartbeat` correctly swallows failures — good. Confirm individual per-shape failures inside `_build_q1_cache_pushes` also catch and `log.debug` without propagating (they do, per the diff — verified).
- **Push storm on large batch ingest**: debounce is `_Q1_PUSH_DEBOUNCE_SEC = 60` (module-level dict). If the daemon restarts mid-cycle, `_Q1_LAST_PUSH` resets to `{}` and all shapes push immediately on the next heartbeat — this is acceptable but worth noting for burst billing awareness.
- **`query_contract.py` shipping in the package**: it is lazily imported with a silent `return []` on `ImportError` — correct defensive behaviour. Verify `clawmetry/query_contract.py` is listed in `setup.py` package data / included in the wheel (the install-test CI check should catch this if missing).

**CI status**: all 25 checks passing (Syntax & Lint, Sync matrix 6×, API Tests 4×, MOAT Verifier 72 tests, E2E Browser, pip install 3×, OSS golden path, and more)

---

PR #3084, opened 2026-06-13
feat(entitlement): add tier_label() + surface tier_label on /api/entitlement

vivekchand:
## What

Adds `TIER_LABELS` + a `tier_label(tier)` helper to `clawmetry/entitlements.py`, and surfaces a `tier_label` field next to `tier` on `Entitlement.to_dict()` (and therefore on the `/api/entitlement` payload).

This mirrors the existing `runtime_label()` / `RUNTIME_LABELS` pattern so the dashboard and CLI never have to hard-code tier display strings or maintain a duplicate JS map. Today the frontend has no tier-label vocabulary at all — operators would otherwise see raw `cloud_pro` / `cloud_starter` snake-case ids in any tier badge that lands later.

## Why

- Single source of truth for the operator-facing tier name (server + UI agree).
- Forward-compatible: an unknown tier id (a future tier added before this map is bumped, or a typo on the wire) renders title-cased with underscores swapped for spaces ("cloud_team" → "Cloud Team") so the badge still shows something readable.
- Never-crash contract preserved: blank / `None` falls back to the OSS label rather than raising. The `/api/entitlement` resolver-error fallback shape also carries `tier_label` so the JSON contract is stable.

No behaviour change. Purely additive; no `allows_*` semantics touched; grace stays the default.

## Surface

```python
# clawmetry/entitlements.py
TIER_LABELS: dict[str, str]  # canonical display names per tier id
def tier_label(tier: str) -> str: ...
# Entitlement.to_dict() now returns {"tier": "...", "tier_label": "...", ...}
```

```
GET /api/entitlement
{
  "tier": "oss",
  "tier_label": "OSS",
  ...
}
```

## Tests

- `tests/test_entitlements.py`: 7 new tests — every known tier id has an explicit label, conformance test that `TIER_LABELS` stays in sync with `TIER_*` constants, unknown-tier title-casing, empty/None/whitespace fallback, case+whitespace normalisation, and `to_dict()` surface (OSS + paid).
- `tests/test_entitlement_api.py`: 1 new test — `/api/entitlement` exposes the label.

All 38 tests in `test_entitlements.py` + `test_entitlement_api.py` pass locally. `ruff check` on the changed files is clean (`make lint-py` has pre-existing failures elsewhere unrelated to this PR).

## Local operator verification checklist

The remote container can't reach the user's daemon, license file, or live cloud snapshot, so before flipping out of draft please:

[...]

PR #3033, opened 2026-06-11
feat(entitlement): add Entitlement.lock_reason() — single source for the "why is this locked" string

vivekchand:
## Summary

Adds a focused helper to `clawmetry.entitlements.Entitlement` (plus a module-level convenience) that returns a human-readable explanation of why a runtime or feature is locked, or `None` when it is allowed.

Today, that reasoning is re-derived inline by three different surfaces — the `@gate` decorator's `hint` field, the runtime/feature catalog `locked` flags consumed by the dashboard, and the CLI diagnostics that PRs #2421 / #2978 / #3005 add — each composing a slightly different string. `lock_reason()` is the single primitive each of those callers can lean on so the user sees the same explanation for the same item from every surface.

### Resolution table

| State of the install                                               | `lock_reason("…")`                                                                  |
|--------------------------------------------------------------------|--------------------------------------------------------------------------------------|
| grace mode (nothing is enforced yet)                               | `None`                                                                               |
| free runtime (`openclaw`, `nemoclaw`) / free feature (`sessions`)  | `None`                                                                               |
| paid item that the resolved tier already grants                    | `None`                                                                               |
| unknown item (errs on the un-locked side)                          | `None`                                                                               |
| expired license, paid runtime                                      | `"License expired — renew to unlock runtime '<id>'"`                                 |
| expired license, paid feature                                      | `"License expired — renew to unlock feature '<id>'"`                                 |
| OSS / Free tier, paid runtime                                      | `"Paid runtime — upgrade to a paid tier to unlock '<id>'"`                           |
| un-entitled tier, Starter feature                                  | `"Starter-tier feature — upgrade to Starter or higher to unlock '<id>'"`             |
| un-entitled tier, Pro-only feature                                 | `"Pro-tier feature — upgrade to Pro to unlock '<id>'"`                               |
| un-entitled tier, Enterprise feature                               | `"Enterprise-tier feature — contact sales to unlock '<id>'"`                         |

`kind` is `"runtime"`, `"feature"`, or `"auto"` (default — try runtime first, then feature). The instance method and the module-level `lock_reason(item, kind="auto")` both honour the never-raise contract: any internal failure logs a warning and returns `None`, so a UI tooltip or 402 body can never be broken by a flaky entitlement read.

## Scope

- `clawmetry/entitlements.py`
  - `Entitlement.lock_reason(item, kind="auto")` — new instance method on the existing dataclass.
[...]

PR #2925, opened 2026-06-09
feat(entitlement): add runtime_tier() + tier field on runtime_catalog() rows

vivekchand:
## Summary

Adds the **runtime side** of the catalogue's tier vocabulary so the dashboard can render Runtime and Feature rows against a single tier identifier. This is the runtime mirror of the in-flight `feature_tier()` + `feature_catalog()` work (PR #2426).

The three catalogue helpers now converge on a uniform `tier` field:

| Helper              | `tier` vocabulary                         | Source PR |
|---------------------|-------------------------------------------|-----------|
| `feature_catalog()` | `"free" \| "starter" \| "pro" \| "enterprise"` | #2426 (in flight) |
| `tier_catalog()`    | full tier ids (`oss`, `cloud_pro`, …)     | #2692 (in flight) |
| `runtime_catalog()` | `"free" \| "starter"` (**this PR**)        | this PR |

All paid runtimes unlock together via the Starter `multi_runtime` grant, so the runtime vocabulary is intentionally narrower than the feature one — no row reports `pro` or `enterprise`.

### `clawmetry/entitlements.py`
- **`runtime_tier(runtime: str) -> str`** — module-level helper. Returns:
  - `"free"` for `FREE_RUNTIMES` (`openclaw`, `nemoclaw`)
  - `"starter"` for every `PAID_RUNTIMES` member
  - `"starter"` for unknown / empty / non-string ids (errs on the locked side, matches `feature_tier()`'s unknown-id policy)
  - Pure catalogue lookup — does not call `get_entitlement()`. Same answer in grace and enforce. Never raises.
- **`runtime_catalog()`** — each row now carries a `tier` field (`"free"` for free runtimes, `"starter"` for paid). Docstring updated.

### `routes/entitlement.py`
- `GET /api/runtimes` inherits the new field through `runtime_catalog()`. Docstring shape updated. The never-raise OSS-free fallback emits `tier: "free"` on its `openclaw`/`nemoclaw` rows so the UI always has the field.

### `tests/test_runtime_tier.py` (new, 16 tests)
- Mapping: free runtimes resolve to `"free"`; paid runtimes resolve to `"starter"`; unknown ids default to `"starter"`.
- Robustness: case-insensitive, whitespace stripped, empty/`None` safe, never-raise on non-string input.
- Catalog invariants: every row carries `tier`; tier matches the `free` bit; tier vocabulary is exactly `{"free", "starter"}`; row tier matches `runtime_tier(row["id"])`.
[...]

PR #3104, opened 2026-06-13
feat(entitlement): add POST /api/entitlement/refresh for cache busting

vivekchand:
## What

Add `POST /api/entitlement/refresh` to `routes/entitlement.py`. The handler calls `clawmetry.entitlements.invalidate()` and returns `get_entitlement(force=True).to_dict()` — same JSON shape as `GET /api/entitlement`.

## Why

The resolver caches each `Entitlement` for 60 s (`_CACHE_TTL_SECS`). That's the right default for per-request reads but the wrong default when the operator has just dropped a license file into `~/.clawmetry/license.key` or the daemon has just written a fresh `~/.clawmetry/cloud_plan.json` out-of-band.

`/api/license/activate` already calls `invalidate()` internally, but it only covers the in-process activation path. A manual file drop (the `clawmetry activate` CLI lands the key on disk) and the daemon's heartbeat-driven cloud-plan write both bypass it. Without a way to bust the cache, the dashboard would render the stale OSS-free state for up to a full TTL after the new entitlement is on disk.

This endpoint covers that gap. The dashboard (or `clawmetry connect` flow) can POST `/api/entitlement/refresh` once after the install completes and immediately render the new tier in a single round-trip.

## Behaviour

- **GRACE-safe.** `invalidate()` does not change resolution rules, only re-runs them. With no license/plan present, the response is byte-for-byte identical to today's `GET /api/entitlement` OSS-free shape.
- **Never-raise.** Falls back to the same OSS-free shape every other endpoint in this module uses, so refresh is safe to spam from a dashboard timer / connect-flow retry loop without ever taking the dashboard down.
- **No side effect beyond the in-process cache.** Does not touch the license file, the cloud-plan cache, or the daemon.

## Tests

Three new tests in `tests/test_entitlement_api.py`:

1. `test_api_entitlement_refresh_grace_shape` — POST returns the same key set as GET on a clean HOME, `tier=oss`, `grace=True`.
2. `test_api_entitlement_refresh_busts_cache` — GET populates the cache with OSS-free, then `cloud_plan.json` is written with `plan=cloud_pro`, a second GET still returns the stale OSS row (TTL not elapsed), POST `/refresh` returns `tier=cloud_pro`, `source=cloud`, `is_paid=True`.
3. `test_api_entitlement_refresh_idempotent` — two consecutive POSTs return identical JSON, confirming the refresh path is safe to retry.

```
tests/test_entitlement_api.py ........... 10 passed in 0.22s
```

`ruff check routes/entitlement.py tests/test_entitlement_api.py` is clean.

## Out of scope

- No change to `__version__`, no enforce-flag flip, no UI work, no CLI subcommand. A `clawmetry entitlement refresh` CLI bound to this endpoint is a natural follow-up.
- The four pre-existing failures in `tests/test_license.py` (auto-provision install paths) and `tests/test_route_gates.py::…[/api/assets-GET]` reproduce on `origin/main` unchanged — not caused by this PR.

Because this agent cannot reach your local daemon, ~/.openclaw, the live cloud snapshot, or a browser, the operator runs:

- [ ] `cp routes/entitlement.py ~/.clawmetry/lib/python3.X/site-packages/routes/entitlement.py` (replace `3.X` with the daemon's Python).
- [ ] `cp tests/test_entitlement_api.py ~/.clawmetry/lib/python3.X/site-packages/tests/test_entitlement_api.py`.
- [ ] `find ~/.clawmetry/lib -name __pycache__ -type d -exec rm -rf {} +`.
- [ ] `launchctl kickstart -k gui/$(id -u)/com.clawmetry.sync`.
- [ ] `curl -sS http://localhost:8900/api/entitlement | jq` — capture the OSS-free baseline.
- [ ] `curl -sS -X POST http://localhost:8900/api/entitlement/refresh | jq` — confirm identical shape and `grace=true`.
- [ ] Drop a real `~/.clawmetry/cloud_plan.json` or paste a license, then `curl -X POST /api/entitlement/refresh` and confirm `tier` flips immediately (no 60 s wait).
- [ ] Decrypt the live cloud snapshot and confirm `/api/entitlement` still classifies as `oss-passthrough` on the cloud side.
- [ ] Browser-check the dashboard for any unexpected UI flicker after refresh (there should be none — the shape is unchanged).

---
_Generated by [Claude Code](https://claude.ai/code/session_01PB324CK2o6zJXVbaedTsgX)_

### DIFF
--- a/routes/entitlement.py
+++ b/routes/entitlement.py
@@ -6,14 +6,21 @@ runtimes/features to surface (and, once enforcement is live, which to render
 locked behind an upgrade CTA). Backed by :mod:`clawmetry.entitlements`, which
 is the single source of truth — handlers never re-derive tier logic here.
 
-  GET /api/entitlement            — the current Entitlement as JSON.
-  GET /api/entitlement/diagnostic — the *inputs* the resolver consulted
+  GET  /api/entitlement          — the current Entitlement as JSON.
+  GET  /api/entitlement/diagnostic — the *inputs* the resolver consulted
                                      (license/cloud-plan presence, enforce env,
                                      cache liveness) for operator triage.
-  GET /api/runtimes               — the full runtime catalog with locked/free flags.
+  POST /api/entitlement/refresh  — drop the cache and return the freshly
+                                   re-resolved Entitlement (used after a
+                                   license is dropped in or the daemon
+                                   writes a new cloud_plan.json, so the UI
+                                   does not have to wait for the 60 s TTL).
+  GET  /api/runtimes             — the full runtime catalog with locked/free
+                                   flags.
 
-Side-effect-free and never-raise, so it is safe to classify ``oss-passthrough``
-on the cloud side: when no license/cloud plan is present it returns a graceful
+Side-effect-free and never-raise (refresh's only side effect is busting the
+in-process cache), so it is safe to classify ``oss-passthrough`` on the cloud
+side: when no license/cloud plan is present every endpoint returns a graceful
 OSS-free shape, never a 4xx.
 """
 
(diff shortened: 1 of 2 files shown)


vivekchand:
## Test plan & review notes

**What changed**
- `routes/entitlement.py` gains `POST /api/entitlement/refresh`: calls `entitlements.invalidate()` then `get_entitlement(force=True)` and returns the freshly resolved shape. The module docstring is updated to document the new route. Three new tests in `tests/test_entitlement_api.py` cover the OSS-grace shape, cache-bust-after-cloud-plan-write, and idempotency.

**Smoke commands**
- `curl -sS -X POST http://localhost:8900/api/entitlement/refresh | python3 -m json.tool`

**Likely failure modes**
- The fallback error body is a hand-rolled dict rather than calling `_ent.get_entitlement()` in OSS mode, so if the canonical `to_dict()` shape ever gains new required fields the fallback will return a stale subset — callers that key on those fields will see `None` silently.
- `invalidate()` + `get_entitlement(force=True)` are two separate calls without a lock; a concurrent request between them could re-populate the cache with stale data before `force=True` runs. Low risk in practice but worth noting for high-concurrency cloud deployments.

---

vivekchand:
Closing in favour of #3146, which consolidates this PR together with #2925, #3033, #3084, #3106, #3109, #3112 into a single clean batch rebased onto current main.

---

vivekchand:
✨ auto-fixed: branch rebased onto `main` (`31f129c1`) — added `POST /api/entitlement/refresh` endpoint (force-drops the 60s entitlement cache and re-resolves) to `routes/entitlement.py`. Branch is now at `ca418773` and should be mergeable.

---

vivekchand:
Branch `feat/entitlements` rebased onto `origin/main` and force-pushed — `POST /api/entitlement/refresh` cache-busting endpoint applied cleanly on top of the current main.

---

PR #3105, opened 2026-06-13
feat(license): add OFFLINE dry-run verify (CLI + /api/license/verify)

vivekchand:
## Summary

Adds a no-side-effects path to inspect a self-hosted Pro/Enterprise license key — the offline dry-run counterpart of `clawmetry activate`. Verifies the Ed25519 signature, reports tier / nodes / expiry / days-left, and explicitly does NOT touch disk or the entitlement cache.

The licensing surface has activate (writes the key + provisions pro), status (reads the installed key), and deactivate (removes it) — but there was no way to verify a key the customer hasn't installed yet. That's the gap this fills.

## What it adds

- **`clawmetry.license.inspect_key(key) -> dict | None`** — verifies the token's Ed25519 signature against the embedded public key, returns the unlock summary in the same shape as `current_license_info()` (`valid`, `status`, `tier`, `nodes`, `sub`, `exp`, `days_left`). Returns `None` for a forged / malformed token. For an EXPIRED-but-parseable token returns `valid=False` + `status="expired"` so support can still confirm what the (now-stale) key was for. Never raises; never writes.
- **`clawmetry license verify <KEY>` CLI** — pretty-prints the dry-run summary ("would set: Pro · N nodes · M days") and explicitly notes that nothing was written to disk. Exits 1 on an invalid/forged key for scripting.
- **`POST /api/license/verify` HTTP endpoint** — same dry-run over HTTP with a `dry_run: true` marker on the response. Returns 200 even for a forged key (the body carries `valid=false`) because it is a query, not a mutation; only a missing `key` body is a 400.

## Use cases

1. **Support triage** — verify a key a customer pasted into a ticket without having them mutate their install.
2. **Pre-flight UI** — show "this is what activating this key would unlock" before the user clicks *Activate*.
3. **Air-gap** — validate a key on a staging box before transporting it.

## Contract

- Never raises (warning-logged on any unexpected error → returns None / `valid=false`).
- Never writes the key, the license file, the pro marker, the cache, or any other side-effect target.
- Never re-invalidates the `entitlements` cache (explicit test).
- HTTP `200` for both valid and invalid keys; `400` only when the body lacks a `key` field.
- Pure offline — no network call, no `CLAWMETRY_LICENSE_SERVER` lookup.

## Scope guard

No version bump, no `__version__` change, no behavior change to existing flows. Grace stays on; the resolver still returns the OSS-free entitlement when no key is installed.

## Tests

- `tests/test_license.py` — 5 new cases (valid summary, enterprise tier, invalid forged/empty, expired-still-payload, no-disk-no-cache).
- `tests/test_license_api.py` — 4 new cases (valid 200, invalid 200 with `valid=false`, expired 200, missing-key 400).

All 40 license-related tests pass (3 pre-existing unrelated auto-provision failures on origin/main — `_Resp` mock missing `.headers` — are out of scope and deselected). New code introduces **zero** new ruff warnings.

You'll need to do this; the remote container has no daemon / `~/.openclaw` / cloud snapshot:

- [ ] `cp` the updated files into `~/.clawmetry/lib/python3.X/site-packages/clawmetry/` (and `routes/`, `tests/`).
- [ ] Clear `__pycache__/` under both dirs.
- [ ] `launchctl kickstart -k gui/$(id -u)/com.clawmetry.sync` (or systemd equivalent).
- [ ] `clawmetry license verify CLAW1.<good-key>` — confirms green ✅ block with tier/nodes/expiry; verify `~/.clawmetry/license.key` is NOT created.
- [ ] `clawmetry license verify garbage` — confirms ❌ "Invalid or unrecognized license key" + exit code 1.
- [ ] `curl -s -X POST http://127.0.0.1:8900/api/license/verify -H 'Content-Type: application/json' -d '{"key":"<good-key>"}'` — confirms `{"valid":true,"dry_run":true,...}`.
- [ ] Same curl with `{"key":"bogus"}` — confirms 200 with `{"valid":false,"dry_run":true,...}`.
- [ ] Same curl with `{}` — confirms 400.
- [ ] `python -m pytest tests/test_license.py tests/test_license_api.py -q` — green on the cases this PR adds.
- [ ] Browser-check `/api/license/verify` is reachable through the dashboard (no console errors).
- [ ] Decrypt the live cloud snapshot and confirm it still ingests with the new wheel.

PR stays DRAFT until you've run the local verification.

---

https://claude.ai/code/session_012EWbGwZimNvMn5CQcWAVG6

---
_Generated by [Claude Code](https://claude.ai/code/session_012EWbGwZimNvMn5CQcWAVG6)_

### DIFF
--- a/clawmetry/cli.py
+++ b/clawmetry/cli.py
@@ -2842,6 +2842,38 @@ def _cmd_license(args) -> None:
         print("  published at https://clawmetry.com/security to confirm your")
         print("  install carries the genuine license-verification key.")
 
+    elif action == "verify":
+        # Dry-run: verify a key OFFLINE and show what it would unlock,
+        # WITHOUT writing anything to disk. Useful for support and pre-flight.
+        key = getattr(args, "license_key", None) or ""
+        if not key:
+            print("❌  Usage: clawmetry license verify <KEY>")
+            sys.exit(1)
+        from clawmetry import license as _lic
+        info = _lic.inspect_key(key.strip())
+        print("ClawMetry License Verify (dry-run)\n" + "─" * 40)
+        if info is None:
+            print("  Result:      ❌  Invalid or unrecognized license key")
+            print("  Disk:        nothing written")
+            sys.exit(1)
+        status = info.get("status", "unknown")
+        tier = str(info.get("tier", "pro")).capitalize()
+        if not info.get("valid"):
+            # Expired-but-parseable: show what it WAS for so support can confirm.
+            print(f"  Result:      ⚠️  {status} (signature verified, but expired)")
+            print(f"  Was for:     {tier} · {info.get('nodes', 1)} node(s)")
+            if info.get("days_left") is not None:
+                print(f"  Expired:     {abs(info['days_left'])} day(s) ago")
+        else:
+            print("  Result:      ✅  valid (signature verified offline)")
+            print(f"  Would set:   {tier} (self-hosted)")
+            print(f"  Nodes:       {info.get('nodes', 1)}")
+            if info.get("days_left") is not None:
+                print(f"  Expires:     in {info['days_left']} day(s)")
+        if info.get("sub"):
+            print(f"  Account:     {info['sub']}")
+        print("  Disk:        nothing written (run `clawmetry license activate <KEY>` to install)")
+
     else:
         from clawmetry import license as _lic
         info = _lic.current_license_info()
@@ -3340,14 +3372,14 @@ def main() -> None:
         "license_action",
         nargs="?",
         default="status",
-        choices=["status", "activate", "deactivate", "fingerprint"],
-        help="status (default) | activate <KEY> | deactivate | fingerprint",
+        choices=["status", "activate", "deactivate", "fingerprint", "verify"],
+        help="status (default) | activate <KEY> | deactivate | fingerprint | verify <KEY>",
     )
     p_license.add_argument(
         "license_key",
         nargs="?",
         default=None,
-        help="License key (CLAW1.…) — required for 'activate'",
+        help="License key (CLAW1.…) — required for 'activate' / 'verify'",
     )
 
     # tier — print the resolved open-core entitlement (scriptable)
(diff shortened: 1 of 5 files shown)


vivekchand:
## Test plan & review notes

**What changed**
- `clawmetry/license.py` gains `inspect_key()` — an offline, no-disk-write JWT inspector; `clawmetry/cli.py` wires it to `clawmetry license verify <KEY>`; `routes/entitlement.py` exposes it as `POST /api/license/verify`. Five unit tests in `tests/test_license.py` and four API tests in `tests/test_license_api.py` cover valid, invalid, expired, and no-key-body cases.

**Smoke commands**
- `curl -sS -X POST http://localhost:8900/api/license/verify -H 'Content-Type: application/json' -d '{"key":"CLAW1.test"}' | python3 -m json.tool`
- CLI: `clawmetry license verify CLAW1.yourkey`

**Likely failure modes**
- `inspect_key` calls `verify_token` which may depend on the public-key fixture; if that path differs between test and prod environments the function silently returns `None` (caught by the broad `except Exception`) — the error is only a `logger.warning`, so a mis-configured key store fails invisibly rather than noisily.
- `api_license_verify` always returns HTTP 200 for invalid/forged keys (by design), but the error-fallback branch also returns 200 without setting `Content-Type: application/json` explicitly — worth confirming Flask sets it correctly on the bare `jsonify()` call in the exception handler.
- **Visual diff**: the visual diff bot flagged a 100% pixel diff on the usage tab for desktop and mobile. If any frontend code was updated alongside this change (e.g. a verify button in the license UI), verify the usage tab still renders correctly at `http://localhost:8900` on both viewports.

---

vivekchand:
✨ auto-fixed: branch rebased onto `main` (`31f129c1`) — added `inspect_key()` to `clawmetry/license.py`, `clawmetry license verify <KEY>` CLI action, and `POST /api/license/verify` dry-run endpoint to `routes/entitlement.py`. Branch is now at `1399f9f7` and should be mergeable.

---

vivekchand:
Branch `feat/license-verify-dry-run` rebased onto `origin/main` and force-pushed — `inspect_key()`, `clawmetry license verify`, and `POST /api/license/verify` applied cleanly on top of the current main.

---

PR #3344, opened 2026-06-26
fix(adapter): extract fast-mode cutoff metadata from sessions and events

### LINKED ISSUE #3341: [obs-gap:openclaw] openclaw: fast-mode fallback/cutoff state on sessions and events
Automated harness-observability audit found a gap for **openclaw**.

- **Harness exposes:** PR #85104 adds bounded fallback behavior: fast mode now carries fallback cutoff metadata (reason, transition event, delivery mode) for short conversational turns that revert to normal mode
- **ClawMetry misses:** Adapter captures fastMode (bool) from session records and event blobs (fastMode/isFastMode/talkFastMode keys) but reads no fallback-cutoff reason, transition count, or delivery-mode field that #85104 adds to support bounded fallback semantics
- **Where (harness):** `CHANGELOG.md (2026.6.10, #85104)`
- **Severity:** medium
- **Closest capability:** EVENTS

_Filed by `scripts/harness/audit.py`. Fingerprint: hgap-38be940929 (used to dedupe — keep it in the body)._

vivekchand:
Closes #3341

## Summary

- openclaw PR #85104 extended fast-mode with bounded fallback semantics, adding five new fields to session and event records: `fastModeCutoff` (bool), `fastModeCutoffReason` (string), `fastModeTransitionCount` (int), `fastModeDeliveryMode` (string), and `fallbackModel` (string)
- The adapter already read `fastMode`/`isFastMode`/`talkFastMode` but silently dropped all five new fields
- This PR extends both the session-record extraction block (~line 960) and the event-blob extraction block (~line 1162) to also extract the five new fields, using the same `if _val is not None: extra[key] = val` guard pattern as the surrounding code

## Test plan

- [ ] Verify `python3 -m py_compile clawmetry/adapters/openclaw.py` passes (no syntax errors)
- [ ] On a build of openclaw that emits `fastModeCutoff`/`fastModeCutoffReason` etc., confirm these keys appear in session `extra` dicts returned by the adapter
- [ ] On older openclaw builds that don't emit these fields, confirm no regression — all five extractions are optional guards that no-op when the key is absent

---
_Generated by [Claude Code](https://claude.ai/code/session_013w8k9rwPnz1NW5GpXu1Cd2)_

### DIFF
--- a/clawmetry/adapters/openclaw.py
+++ b/clawmetry/adapters/openclaw.py
@@ -960,6 +960,27 @@ class OpenClawAdapter(AgentAdapter):
             _fm = s.get("fastMode") if s.get("fastMode") is not None else s.get("isFastMode")
             if _fm is not None:
                 extra["fastMode"] = bool(_fm)
+            # Fast-mode fallback/cutoff metadata (#3341): PR #85104 also emits
+            # cutoff state, reason, transition count, delivery mode, and fallback
+            # model for sessions where fast-mode reverts to normal mode.
+            _fmc = s.get("fastModeCutoff")
+            if _fmc is not None:
+                extra["fastModeCutoff"] = bool(_fmc)
+            _fmc_reason = s.get("fastModeCutoffReason") or s.get("cutoffReason")
+            if _fmc_reason is not None:
+                extra["fastModeCutoffReason"] = _fmc_reason
+            _fmc_count = s.get("fastModeTransitionCount") or s.get("transitionCount")
+            if _fmc_count is not None:
+                try:
+                    extra["fastModeTransitionCount"] = int(_fmc_count)
+                except (TypeError, ValueError):
+                    pass
+            _fmc_mode = s.get("fastModeDeliveryMode") or s.get("deliveryMode")
+            if _fmc_mode is not None:
+                extra["fastModeDeliveryMode"] = _fmc_mode
+            _fm_fallback = s.get("fallbackModel") or s.get("fastModeFallbackModel")
+            if _fm_fallback is not None:
+                extra["fallbackModel"] = _fm_fallback
             # /think reasoning-level tier (#3324): PR #94067 stores the active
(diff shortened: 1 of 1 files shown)


vivekchand (commented):
## Test plan & review notes

**What changed**
Extends the openclaw adapter to extract five new fast-mode fallback fields (`fastModeCutoff`, `fastModeCutoffReason`, `fastModeTransitionCount`, `fastModeDeliveryMode`, `fallbackModel`) from both session records and event blobs, with alias fallbacks for older field names.

**CI status**: All 19 checks green ✅ — syntax, lint, MOAT, E2E, cross-repo handoff, install smoke tests.

**Smoke commands**
```bash
python3 -m py_compile clawmetry/adapters/openclaw.py   # syntax clean
make test-api                                           # fast API pass
```

**One nit worth a quick look**

The alias fallbacks for `fastModeTransitionCount` use `or` instead of an `is not None` guard:

```python
# current (both session + event blocks)
_fmc_count = s.get("fastModeTransitionCount") or s.get("transitionCount")
```

`0` is falsy in Python, so a session where fast-mode is active but hasn't transitioned yet (`fastModeTransitionCount=0`) silently falls through to the `transitionCount` alias — and if that's also absent, the field gets dropped entirely. Zero transitions is a meaningful value here.

Suggested fix (same pattern the surrounding code uses for the bool field):
```python
_fmc_count = s.get("fastModeTransitionCount")
if _fmc_count is None:
    _fmc_count = s.get("transitionCount")
```

Same issue appears in the event-blob block (~line 1180). All four string-alias fields (`fastModeCutoffReason`, `fastModeDeliveryMode`, `fallbackModel`) use `or` too, but empty-string edge cases are unlikely to matter for those — the int field is the main concern.

Otherwise the change looks solid: consistent guard pattern, no new deps, backward-compatible on older openclaw builds. 

**Issue link**: `Closes #3341` is already in the body ✓

---

PR #3634, opened 2026-07-10
fix(auto-update): restart BOTH the dashboard and the daemon onto the new wheel

vivekchand:
The 0.12.552 hands-off verification surfaced this: the daemon restarted itself onto the new wheel but the local dashboard kept serving 0.12.549 from memory (the auto path never kicked it). The restart helper is now role-aware and kicks the other service first, self last via _exit. Source-level guard included; suite 26/26.

### DIFF
--- a/routes/meta.py
+++ b/routes/meta.py
@@ -452,19 +452,57 @@ def perform_self_update(reason: str = "manual", restart: bool = True,
         import os as _os
         import platform as _pl
 
+        # BOTH long-running processes must restart onto the new wheel or one
+        # keeps serving the old build from memory (the classic "daemon
+        # updated but the dashboard UI is stale" gotcha — see
+        # feedback_restart_both_processes_after_upgrade.md; found again live
+        # 2026-07-10 when the daemon self-updated to 0.12.552 while the
+        # dashboard kept 0.12.549 in memory). Ordering is load-bearing:
+        # `kickstart -k` on our OWN service kills us mid-function, so kick
+        # the OTHER service(s) first and restart ourselves last via _exit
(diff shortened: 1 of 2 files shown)


vivekchand (commented):
## Test plan & review notes

**What changed**
- Auto-update restart helper is now role-aware: kicks the sibling service (dashboard ↔ daemon) first, then self-exits via `_exit`. Fixes a real regression where the daemon restarted onto the new wheel but the dashboard kept serving the old version from memory (surfaced during 0.12.552 hands-off verification). 26/26 tests passing.

**Smoke commands**
- `python -m pytest` filtered to auto-update tests: `python -m pytest -k "autoupdate or auto_update" -v`
- Full suite: `make test`

**What to look at** (hard to trigger locally without a real update cycle)
- On a staging install: `pip install clawmetry==<old_version>`, let the daemon detect the newer wheel, then confirm `ps aux | grep clawmetry` shows both the dashboard *and* daemon pids changed after the update
- Check logs for any "failed to restart sibling" noise — especially on macOS (launchctl) vs Linux (systemd/process) where the sibling-kick mechanism may differ

- Race condition: if the sibling service already restarted or crashed before the kick signal arrives, signaling a stale pid could raise — confirm there's a guard or the error is caught and logged rather than silently swallowed
- On systems where the dashboard and daemon are managed by separate OS service units (launchctl plist, systemd unit), kicking a sibling pid may not survive a full service restart — the fix may need a platform-specific code path

**Issue link**
- No issue reference in the PR body. Worth opening one (or checking if an existing issue tracks auto-update version drift) so this fix has a paper trail and the regression can be marked closed.

---

PR #3795, opened 2026-07-17
fix: conhost ANSI garbage + CI guards for the silent-CLI class (#3791)

vivekchand:
## Summary

Two related deliverables: a user-facing fix for ANSI garbage in the classic Windows console, and the CI guards that close the silent-CLI class behind #3791.

**Conhost ANSI fix (user-facing):** `clawmetry onboard` in a cmd.exe conhost window printed literal escape garbage (`←[1m` around every styled string). The wizards gate colors on `isatty()`, which is True in conhost, but conhost ships with Virtual Terminal processing OFF. New `_ansi_ok()` helper enables VT via `SetConsoleMode` and reports whether escapes render; the three wizard `_is_tty` gates use it (plain-text fallback when VT is unavailable), and `main()` calls it early so the unconditional ANSI emitters (uninstall, account) render too. Windows Terminal is unaffected (VT already on). This was invisible until 0.12.557 because #3791 swallowed all wizard output.

**CI guards (the class, not the instance):**
1. Lint ban on `sys.stdout`/`sys.stderr` TextIOWrapper rebinds repo-wide; `reconfigure()` is the only allowed spelling. Revert-proven: 4 hits on the pre-fix `dashboard.py` blob, 0 on current main.
2. Output-asserting CLI smoke on all 3 OS: the old `pip-install` smoke steps were `continue-on-error: true` with a `|| echo` fallback and could never fail; #3791 passed them with exit 0 and zero bytes. New steps assert stdout content from the real `clawmetry` console script, plus a subprocess check that `import dashboard` leaves `sys.stdout` open.
3. `cli.py`'s closed-handle devnull swap now leaves a breadcrumb in `~/.clawmetry/cli-stream-guard.log` when it fires on a process that had a console, so future silent-output bugs surface instead of hiding.

## Test plan

- [x] Lint regex revert-proven locally: 4 violations against `8a52ee1~1:dashboard.py`, 0 against current
- [x] `ci.yml` validated with `yaml.safe_load`; `clawmetry/cli.py` validated with `ast.parse`
- [x] Piped `onboard --local` output contains zero ESC bytes (non-tty plain-text fallback intact, 699 bytes of real content)
- [ ] This PR's own CI run proves the new smoke steps on all 3 OS
- [ ] Post-release: `clawmetry onboard` in a real cmd.exe conhost window renders bold/dim styling instead of `←[1m` (needs a live console; CI runners have none)

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -35,6 +35,29 @@ jobs:
       - name: Lint (warnings only, don't fail)
         run: ruff check dashboard.py --select E,W --ignore E501,W503 || true
 
+      # Silent-CLI class guard (#3791): rebinding sys.stdout/sys.stderr to a
+      # new TextIOWrapper over the same buffer orphans the previous wrapper;
+      # its GC finalizer closes the SHARED buffer, and every later print(),
+      # argparse help, and wizard prompt dies silently (exit 0, zero bytes).
+      # reconfigure() is the only allowed spelling for stream re-encoding.
+      - name: Ban stdout/stderr TextIOWrapper rebinds (silent-CLI class, #3791)
+        run: |
+          python3 -c "
+          import glob, re, sys
+          pat = re.compile(r'sys\.(stdout|stderr)\s*=\s*io\.TextIOWrapper\(', re.S)
+          bad = []
+          for f in ['dashboard.py'] + glob.glob('clawmetry/**/*.py', recursive=True):
+              src = open(f, encoding='utf-8', errors='replace').read()
+              for m in pat.finditer(src):
+                  line = src[:m.start()].count(chr(10)) + 1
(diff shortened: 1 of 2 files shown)


vivekchand:
blocked on admin action — skipping (auto-mergeability sweep): the only failing check is **C6 Required-status-checks gate**, which exits 1 because `GITHUB_TOKEN` lacks `Administration (read+write)` and cannot register `Cross-repo handoff (C4)`, `E2E Browser Tests (critical subset)`, `MOAT Keystone (13-endpoint bar)`, and `OSS golden path (wheel + OpenClaw + 9 tabs)` as required status checks on `main`. All other 26 checks pass. To unblock: create a fine-grained PAT with Administration read+write on this repo, add it as a secret, then run the **"Apply required E2E status checks (C6 -- one-shot)"** workflow action — or add the four checks manually in Settings › Branches › main › Required status checks.

---

PR #3878, opened 2026-07-21
fix: pre-rename clawmetry.exe on Windows before pip upgrade (WinError 32)

### LINKED ISSUE #3875: Windows: auto-update bricks the install (pip uninstalls old version, then WinError 32 on locked clawmetry.exe)
## What happened (live on a real Windows 11 node, 2026-07-21)

The daemon's auto-updater tried to move 0.12.562 -> 0.12.563 while the dashboard was running. pip's upgrade is uninstall-then-install: the uninstall of 0.12.562 SUCCEEDED (package code removed from site-packages), then the install FAILED because `Scripts\clawmetry.exe` is locked by the running dashboard process:

```
Attempting uninstall: clawmetry
  Found existing installation: clawmetry 0.12.562
  Uninstalling clawmetry-0.12.562:
ERROR: Could not install packages due to an OSError: [WinError 32] The process cannot
access the file because it is being used by another process:
'c:\users\devad\appdata\local\programs\python\python312\scripts\clawmetry.exe'
```

Result: every CLI invocation dies with `ModuleNotFoundError: No module named 'clawmetry'`, pip leaves orphaned rename stubs in site-packages (`~lawmetry-0.12.562.dist-info`, `~ests`, `~elpers`, `~outes`), and the 1800s retry loop re-runs into the same locked exe for as long as the dashboard lives. To the user, the product uninstalled itself overnight. The daemon keeps running only because its old modules are in memory; any restart would leave the node fully dead.

## Root cause

Windows cannot overwrite a running .exe. Any node where the dashboard (started via `clawmetry.exe`) is up when the updater fires will hit this. macOS/Linux never see it (running binaries are replaceable), which is why the fast-update loop looks safe elsewhere.

## Suggested fix (in the updater, Windows branch)

1. Before invoking pip: rename `Scripts\clawmetry.exe` to `clawmetry.exe.old` (renaming a RUNNING exe is allowed on Windows even though overwriting is not); clean up `*.exe.old` on next daemon start.
2. On pip failure mid-upgrade: immediately `pip install clawmetry==<previous>` to restore, so a failed upgrade can never leave zero versions installed.
3. Guard: a unit test on the pre-rename helper plus a CI check that the Windows updater path never calls pip while `clawmetry.exe` would be overwritten in place.

## Recovery used on the affected node

kill the dashboard process holding the exe, `pip install clawmetry==0.12.563`, delete the `~*` stub dirs, restart the daemon. Confirmed healthy afterwards (`clawmetry status` green, cloud sync resumed).

Related: the Windows fixes from #3854 shipped in 0.12.563 and are confirmed present in the published wheel.

vivekchand:
## Summary

Closes #3875

- On Windows, pip's uninstall-then-install sequence fails with WinError 32 when `clawmetry.exe` is held open by the running dashboard. Renaming a running `.exe` IS allowed on Windows even when overwriting it is not — so we rename it to `clawmetry.exe.old` before pip runs, letting pip write the new launcher cleanly.
- On pip failure: restores the renamed `.exe` and pip-reinstalls the previously working version so the node is never left dead after a partial uninstall.
- Cleans up any `*.exe.old` stubs at the start of each update attempt (handles stubs left by a prior failed run).

All changes are in `_win_rename_exe_before_pip()` and `_win_cleanup_old_exe_stubs()` helpers in `routes/meta.py`; both are no-ops on macOS/Linux, so the existing update paths are untouched.

## Test plan

- [ ] On macOS/Linux: `clawmetry update` behaves identically (helpers return `None, None` immediately)
- [ ] On Windows with a running dashboard: auto-update no longer hits WinError 32; new version installs and `clawmetry.exe` is present after upgrade
- [ ] Simulate pip failure on Windows (e.g. invalid version spec): `clawmetry.exe` is restored and the previous version is reinstalled automatically
- [ ] After a prior failed attempt that left a `clawmetry.exe.old` stub: next update attempt removes the stub before proceeding

---
_Generated by [Claude Code](https://claude.ai/code/session_01GwMgMgNX195H7cwupcYypz)_

### DIFF
--- a/routes/meta.py
+++ b/routes/meta.py
@@ -351,6 +351,48 @@ def api_version():
     return {"current": current, "latest": latest, "update_available": update_available}
 
 
+def _win_cleanup_old_exe_stubs() -> None:
+    """Remove clawmetry*.exe.old stubs left by a previous Windows update attempt."""
+    if not sys.platform.startswith("win"):
+        return
+    try:
+        import sysconfig
+        from pathlib import Path
+        scripts = Path(sysconfig.get_path("scripts"))
+        for stub in scripts.glob("clawmetry*.exe.old"):
+            try:
+                stub.unlink()
+            except Exception:
+                pass
+    except Exception:
+        pass
+
+
+def _win_rename_exe_before_pip():
+    """On Windows, rename clawmetry.exe -> clawmetry.exe.old before pip runs.
+
+    pip's uninstall-then-install tries to overwrite clawmetry.exe while it may
+    be held open by a running dashboard process, causing WinError 32. Renaming
+    an open .exe is allowed on Windows even when overwriting it is not, so pip
+    then writes the new .exe without touching the in-use file.
+
+    Returns (orig_path, renamed_path) on success, or (None, None) otherwise.
+    """
+    if not sys.platform.startswith("win"):
+        return None, None
+    try:
+        import sysconfig
+        from pathlib import Path
+        exe = Path(sysconfig.get_path("scripts")) / "clawmetry.exe"
+        if not exe.exists():
+            return None, None
+        old = exe.with_name("clawmetry.exe.old")
+        exe.rename(old)
+        return exe, old
+    except Exception:
+        return None, None
+
+
 def perform_self_update(reason: str = "manual", restart: bool = True,
                         target_version=None):
     """Core self-update: ``pip install -U clawmetry`` then schedule a process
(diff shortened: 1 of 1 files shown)


vivekchand:

The only failing check is **C6 Required-status-checks gate**, which verifies that branch protection rules include `OSS golden path`, `Cross-repo handoff (C4)`, `MOAT Keystone`, and `E2E Browser Tests (critical subset)` as required checks on `main`. All 22 other CI checks pass.

This is a repository settings issue (not a code issue): fixing it requires a fine-grained PAT with Administration (read+write) access and running `scripts/close-c6.sh` or triggering the "Apply required E2E status checks (C6 -- one-shot)" workflow. Once branch protection rules are updated, the C6 check will pass on all open PRs simultaneously.

---

Issue #3897, opened 2026-07-21
E2E Robustness Epic: 7-criteria gate (OSS + Cloud + Landing + Visual-diff + Cross-repo + Required-checks + Quarantine)


vivekchand:
## E2E Robustness Tracking

Automated cron fires every 2h, picks the single highest-leverage gap, ships one PR.
Cron stops when all 7 criteria are green for 7 consecutive days.

---

## Criteria checklist

- [x] **C1 OSS golden path** -- wheel install + real OpenClaw + synthetic session + 9 tabs render without auth errors, hard gate on every PR in &lt;12 min. Workflow: `oss-golden-path.yml`. Recent runs: green.
- [x] **C2 Cloud golden path** -- stub-DB Flask app + real Chromium through `/cloud`, `/cloud/node/`, `/cloud/account`, `/cloud/billing`, `/cloud/team` without JS errors. Workflow: `clawmetry-cloud:e2e.yml`. Recent runs: green.
- [x] **C3 Landing golden path** -- hero loads, CTA links to `#get-started`, `/api/subscribe` returns ok with mocked deps. Workflow: `clawmetry-landing:landing-golden-path.yml`. Recent runs: green.
- [x] **C4 Cross-repo handoff** -- landing signup stub -&gt; cloud account -&gt; daemon pair -&gt; first sync event lands. 5 step-test files in `tests/e2e/cross_repo_handoff/`. Workflow: `clawmetry-cloud:e2e-cross-repo.yml`. Recent runs: green.
- [x] **C5 Visual-diff all tabs post-auth** -- every one of the 33 canonical OSS dashboard tabs screenshots post-login with no auth overlay. Hard gate in `oss-golden-path.yml` (C5 step runs 33-tab `test_e2e_oss_all_tabs.py`). Visual-diff `pr-screenshots.yml` seeds `CLAWMETRY_VISUAL_DIFF_TOKEN` into localStorage. Recent runs: green. Fixes the user-reported gateway-token bug (2026-05-17).
- [ ] **C6 Required status checks** -- all CI gates above are marked required in branch protection on main across all 3 repos. **Currently red**: `apply-required-checks.yml` on push silently exits 0 without calling the verification script, hiding whether checks are actually configured. Fix in PR #3898 (awaiting merge).
[...]

PR #3898, opened 2026-07-21
fix(c6): remove silent exit guard so verify script runs on every push

vivekchand:
## Problem

`apply-required-checks.yml` "Apply required status checks" step was
exiting 0 in under 1 second on every push to main when `GITHUB_TOKEN`
is the default `ghs_` Actions token (no `E2E_ADMIN_PAT` configured).
The early bash guard printed a message and called `exit 0` before
`apply_required_status_checks.py` ran at all.

Evidence: run 29830084849 -- step started_at and completed_at both
`12:26:22` (0-second runtime, no Python output).

This meant C6 was silently reporting green even when required status
checks were absent from branch protection.

## Fix

Remove the early exit guard entirely. `apply_required_status_checks.py`
already handles the `ghs_` case correctly via a read-only verification
path (public `GET /repos/{owner}/{repo}/branches/main` endpoint -- no
admin rights needed). It exits 0 if C6 is configured, exits 1 if not.
With an admin PAT it applies + verifies.

The step now unconditionally runs:
```
python3 scripts/apply_required_status_checks.py
```

## Behaviour after merge

- **push (ghs_ token)**: script runs read-only verify; job green if C6
  configured, red if not -- gives real signal instead of always-green
- **push (E2E_ADMIN_PAT configured)**: full apply + verify as before
- **workflow_dispatch APPLY + pat_token**: unchanged
- **workflow_dispatch dry-run**: unchanged (separate step, not affected)

## Tracking

Part of E2E Robustness Epic: vivekchand/clawmetry#3897 (C6 criterion)

---
_Generated by [Claude Code](https://claude.ai/code/session_019EFGfVdtxUimj7m4DmmLha)_

### DIFF
--- a/.github/workflows/apply-required-checks.yml
+++ b/.github/workflows/apply-required-checks.yml
@@ -14,16 +14,15 @@ name: Apply required E2E status checks (C6 -- one-shot)
 #                      local clone, no gh CLI, no secret setup needed.
 #                      The token is masked via ::add-mask:: before use.
 #
-#   push to main -- informational only (exits 0 regardless of C6 state).
-#     c6-schedule-heal.yml handles daily auto-apply when E2E_ADMIN_PAT
-#     is configured.  This workflow no longer reds main on push.
+#   push to main -- runs apply_required_status_checks.py.
+#     With only GITHUB_TOKEN (ghs_): read-only verify path in the script
+#     (exits 0 if C6 configured, exits 1 if not -- no admin rights needed).
+#     With E2E_ADMIN_PAT: full apply + verify.
 #
 # TOKEN PATHS (priority order):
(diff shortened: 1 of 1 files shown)


vivekchand:

The only failing check is **C6 Required-status-checks gate** (ironic given this PR's subject). The script now runs correctly (the silent-exit bug this PR fixes is resolved), but it exits 1 because the 4 required status checks are not yet configured in `main` branch protection: `OSS golden path`, `Cross-repo handoff (C4)`, `MOAT Keystone`, `E2E Browser Tests (critical subset)`. No code push to the PR branch can fix this — it requires a repo admin to add those checks to branch protection via GitHub Settings or the **"Apply required E2E status checks (C6 — one-shot)"** Actions workflow with an admin PAT.

---

Issue #3899, opened 2026-07-21
E2E Robustness Epic: 7 criteria to full automation


vivekchand:
## Goal

Automate end-to-end testing across vivekchand/clawmetry, vivekchand/clawmetry-cloud, and vivekchand/clawmetry-landing so every PR is gated by real browser tests, and the team can ship features confidently without manual verification.

User mandate (2026-05-17): &#34;we need to heavily invest in E2E tests, install openclaw in the gitlab pipeline and test OSS and clawmetry cloud end to end... without e2e tests there will be lot of manual effort required, also the screenshots are always broken -- gateway token is not passed for oss so it never displays other screens.&#34;

---

## 7 Criteria (cron stops when all green for 7 consecutive days)

- [x] **1. OSS golden path** -- install ClawMetry from wheel + spin up real OpenClaw + send a message + verify dashboard renders all tabs (Sessions, Brain, Tokens, Crons, Channels, Flow, Memory, Security, Health) WITHOUT auth errors. Runs on every PR in &lt;5 min. _GREEN: `oss-golden-path.yml` C1 gate, `tests/test_e2e_oss_golden_path.py` (9 tabs + auth check + DuckDB ingest verify)._

- [x] **2. Cloud golden path** -- synthetic signup + daemon connect + sync synthetic message + cloud dashboard renders /cloud, /cloud/node/``, /cloud/account, /cloud/team, /cloud/billing without JS errors. Runs on every PR in &lt;8 min. _GREEN: `clawmetry-cloud` `e2e.yml` passing consistently._

- [x] **3. Landing golden path** -- hero loads + click &#39;Get Started&#39; + signup form submits + OTP delivered + handoff to cloud signup. Runs on every PR in &lt;3 min. _GREEN: `clawmetry-landing` `landing-golden-path.yml` confirmed running on PRs._

- [x] **4. Cross-repo handoff** -- landing signup -&gt; cloud account creation -&gt; daemon pair -&gt; first sync event lands. End-to-end across all 3 repos in &lt;10 min. _GREEN: `cross-repo-handoff.yml` passing on every PR._

[...]

PR #3900, opened 2026-07-21
fix(e2e): C5 post-auth sweep always runs; startup failures fail fast

vivekchand:
## Summary

Fixes two silent-failure modes in `oss-golden-path.yml` that allowed the C5 post-auth tab sweep (the direct fix for the user-reported gateway-token bug) to be skipped on broken branches.

**Problem 1: Dashboard startup crashes were silent**

When a branch had a startup bug (e.g. `ImportError: cannot import name 'OpenClawAdapter'`), the health-check loop in "Start dashboard from installed wheel" silently exited after 30 s without failing the step. This caused `conftest.py` to attempt a second server launch from source (which also failed with the same error), producing confusing `ERROR at setup` failures across all C1 tests. Observed in run `29829243269` on branch `auto/issue-3885-trusted-proxy-device-pairing`.

**Problem 2: C5 was silently skipped whenever C1 had any failure**

Because GitHub Actions skips non-`if: always()` steps when a prior step fails, C5 (the 33-tab auth-overlay sweep) was silently skipped any time C1 tests failed for any reason -- test-assertion failures or startup crashes. This meant that on branches with regressions, the only hard-gated post-auth sweep never ran, defeating its purpose as a fix for the user-reported "gateway token not passed for OSS" bug.

## Changes

**`oss-golden-path.yml`** only:

- **Start dashboard step**: health-check loop now uses an `ok=0/1` flag and exits 1 with a clear error message if the server never becomes ready. The dashboard log tail is printed inline for fast diagnosis.
- **Restart dashboard step**: same health-check fix + `if: always()` so the fresh server for C5 is started even when C1 tests fail from test-assertion failures (not startup crashes).
- **Run C5 step**: `if: always()` so the 33-tab post-auth overlay sweep always executes. C1 and C5 are independent dimensions: a C1 data regression does not imply C5 auth is broken.

## Acceptance test

On a branch with a startup crash (e.g. `ImportError`):
- Before: "Start dashboard" step passes silently, C1 tests all `ERROR`, C5 skipped.
- After: "Start dashboard" step fails immediately with `::error::Dashboard never became ready`, C1 and C5 both fail, no silent skip.

On a branch where C1 has test-assertion failures (dashboard is healthy):
- Before: C5 skipped.
- After: C5 runs and its result (pass or fail) gates the PR independently of C1.

Tracking: vivekchand/clawmetry#3899 (E2E robustness epic, criterion C5)

---
_Generated by [Claude Code](https://claude.ai/code/session_01LxkLWJYrauZduFUNdYENJ4)_

### DIFF
--- a/.github/workflows/oss-golden-path.yml
+++ b/.github/workflows/oss-golden-path.yml
@@ -10,7 +10,7 @@ name: OSS Golden Path (C1)
 # Budget: 12 minutes wall-clock (raised from 8; Playwright chromium download
 # ~100 MB takes up to 4.5 min on a cold runner -- cache hit drops it to ~30 s;
 # C5 33-tab Playwright sweep adds ~90 s on top of the C1 9-tab sweep).
-# Tracking: vivekchand/clawmetry#1646 (C1), #3666 (C5)
+# Tracking: vivekchand/clawmetry#1646 (C1), #3666 (C5), #3899 (E2E epic)
 
 on:
   pull_request:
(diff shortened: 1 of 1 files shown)


vivekchand:

The only failing check is **C6 Required-status-checks gate**, which verifies that `main` branch protection requires `OSS golden path`, `Cross-repo handoff (C4)`, `MOAT Keystone`, and `E2E Browser Tests (critical subset)`. All four are absent from the current branch protection rules. This is a repository admin configuration issue — no code push to the PR branch can make C6 pass. To unblock: configure those required status checks via GitHub Settings > Branches > Edit main, or run the **"Apply required E2E status checks (C6 — one-shot)"** Actions workflow with an admin PAT.

---

PR #3916, opened 2026-07-22
fix: Windows log streams (openclaw.cmd launch, select-on-pipe, missing tail)

vivekchand:
Live failure on a real Windows 11 node: the dashboard Logs tab crashed mid-SSE (`FileNotFoundError` from `Popen(["openclaw", ...])` - CreateProcess cannot launch the npm `openclaw.cmd` wrapper from a bare name), and the frontend's retries then hit the SSE rate limiter (endless 429s). The same POSIX assumptions break the fallback stream and the daemon: `select.select()` on a pipe raises on Windows (sockets only), and `tail` does not exist there.

Fixes all four sites through one portable primitive:

- `process_control.PipeLineReader`: reader thread + queue with wait-with-timeout and timeout-0 poll semantics on every OS. Used by the `openclaw logs --follow --json` SSE generator and the sandbox OCSF drain (reader registry cleared on daemon shutdown).
- `shutil.which()` resolution for the openclaw binary (CreateProcess launches .cmd only with the full name).
- Pure-Python file follows (tail -f -n 0 semantics) for the dashboard fallback stream and the daemon gateway-log streamer - one fewer subprocess on every OS, no tail dependency.
- The JSON SSE stream ends with an honest `stream_ended` event on EOF instead of spinning until max_duration.

Verification: 5 regression tests, revert-proven on a real Windows 11 machine (collection fails on the un-fixed tree; the class guards flag all four historical sites). Includes auto-discovering guards that fail if `select.select` or a `tail` subprocess returns to `routes/infra.py` or `clawmetry/sync.py`.

Same Windows-hardening arc as #3854 / #3915; the audit that predicted this class is in #3854's history.

### DIFF
--- a/clawmetry/process_control.py
+++ b/clawmetry/process_control.py
@@ -269,6 +269,55 @@ def _start_tokens_equivalent(want: str, have: str, tol: int = 3) -> bool:
     return any(abs(x - y) <= tol for x in a for y in b)
 
 
+class PipeLineReader:
+    """Portable non-blocking line reader for a subprocess pipe.
+
+    ``select.select()`` on a pipe is POSIX-only — on Windows select works
+    on sockets exclusively and raises OSError, which crashed every
+    streaming reader (dashboard SSE log streams, daemon gateway-log
+    streamer, sandbox OCSF drain). A daemon reader thread pumping lines
+    into a queue gives the same wait-with-timeout semantics on every OS.
+    """
+
+    def __init__(self, stream) -> None:
+        import queue
+        import threading
+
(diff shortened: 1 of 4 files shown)


vivekchand:
blocked on admin config — skipping (auto-mergeability sweep): C6 gate fails because `OSS golden path`, `Cross-repo handoff (C4)`, `MOAT Keystone`, and `E2E Browser Tests` are not yet required status checks on `main`. Fix: Settings > Branches > Edit main > add those 4 checks, or run the "Apply required E2E status checks (C6 -- one-shot)" Actions workflow with a fine-grained PAT (Administration: write).

---

PR #4005, opened 2026-07-24
feat(extensions): surface probed_plugins in /api/extensions

vivekchand:
## Summary

- Adds `probed_plugins` + `probed_plugin_count` to `GET /api/extensions`, exposing the side-effect-free `clawmetry.extensions.probe_plugins()` view over HTTP alongside the existing in-process `plugins` / `failed_plugins` mirrors.
- The probe answers a distinct triage question — *"would ClawMetry try to load this on the next restart, and would the import work right now?"* — that diverges from `plugins` when a wheel is installed after daemon startup, when a post-startup dependency upgrade breaks the import, or when the on-disk marker exists but the module can't actually be imported.
- Mirrors the existing defensive posture: an older core wheel that predates `probe_plugins()` (or a helper that raises) degrades the field to `[]` / `0` rather than 5xx'ing the whole envelope, so a mixed deploy (new routes, old core) never blanks the dashboard's diagnostic tab.

## What changed

- `routes/extensions.py` — resolves `probe_plugins()` inside `api_extensions()` under its own `try` / `except` block, appending `probed_plugins` + `probed_plugin_count` to the JSON envelope. Docstring updated to describe the three orthogonal views and their degrade contracts.
- `tests/test_extensions_introspection.py`:
  - `test_api_extensions_shape_empty` updated to include the two new keys (twice, matching the file's existing duplicated fixture literal).
  - `test_api_extensions_reports_probed_plugins` — pins full row shape (name / value / importable / error) for one success + one import-error entry.
  - `test_api_extensions_probed_independent_of_load_plugins` — the whole point: the probe populates even when `load_plugins()` was never called on this process.
  - `test_api_extensions_survives_missing_probe_plugins_helper` — same fallback contract as `failed_plugins` (mixed deploy, new routes + old core).
  - `test_api_extensions_probed_row_never_leaks_traceback` — only `str(exc)` is captured; no traceback / frame locals leak.

## Test plan

- [x] `python -m pytest tests/test_extensions_introspection.py tests/test_extensions_probe.py tests/test_extensions.py tests/test_cli_extensions_subcommand.py tests/test_cli_status_extensions.py tests/test_claudecode_loads_plugins.py tests/test_sync_loads_plugins.py tests/test_proxy_loads_plugins.py` — 69 passed
- [x] `python -m py_compile routes/extensions.py tests/test_extensions_introspection.py` — clean
- [x] `ruff check routes/extensions.py tests/test_extensions_introspection.py` — clean (pre-existing lint noise in `dashboard.py` is unchanged, and CI already treats ruff as warnings-only)
- [x] The existing CLI↔HTTP drift test (`test_extensions_json_matches_api_extensions_response`) uses a shared-subset check on 6 keys — additive extension to the HTTP surface does not break it. The CLI `--json` shape is untouched by this PR; catching the CLI up to also emit `probed_plugins` can follow as a separate change.

The remote fire has no access to the local daemon / live cloud snapshot / browser, so the operator drives:

- [ ] Copy `routes/extensions.py` into `~/.clawmetry/lib/python3.*/site-packages/routes/` (or reinstall the wheel).
- [ ] Clear `__pycache__/` for `routes/` and `clawmetry/`.
- [ ] `launchctl kickstart -k gui/$(id -u)/com.clawmetry.sync` (macOS) or the systemd equivalent.
- [ ] `curl -sS http://localhost:8900/api/extensions | jq .` — envelope now includes `probed_plugins` + `probed_plugin_count`.
- [ ] Cross-check on a node WITH `clawmetry-pro` installed: `probed_plugins[].importable` should be `true` for each entry, and the `name` set should equal `plugins ∪ failed_plugins[].name`.
- [ ] Cross-check on a node WITHOUT `clawmetry-pro`: `probed_plugins == []` and `probed_plugin_count == 0`.
- [ ] Cross-check the "installed after startup" flow: `pip install clawmetry-pro` on a running daemon (no restart), then hit `/api/extensions` — `probed_plugins` should show the new entry as `importable: true` even while `plugins` is still empty until the next daemon restart.
- [ ] Decrypt the live cloud snapshot, verify the diagnostic panel renders the new field without a browser console error.
- [ ] Browser: dashboard extensions/diagnostic tab still loads (any consumer relying on additive-only extension is unaffected).

Stays in GRACE — no gating change, no version bump, no release tag. Local operator handles verification, release, and merge per FLYWHEEL.md.

---
_Generated by [Claude Code](https://claude.ai/code/session_01XS7Ygk8AovyWBZP374eoC5)_

### DIFF
--- a/routes/extensions.py
+++ b/routes/extensions.py
@@ -33,12 +33,16 @@ def api_extensions():
     Response shape::
 
         {
-          "plugins":             ["clawmetry-pro", ...],
-          "plugin_count":        1,
-          "failed_plugins":      [{"name": "clawmetry-pro", "error": "..."}],
-          "failed_plugin_count": 1,
-          "events":              ["session.snapshot", ...],
-          "handler_counts":      {"session.snapshot": 2, ...}
+          "plugins":              ["clawmetry-pro", ...],
+          "plugin_count":         1,
+          "failed_plugins":       [{"name": "clawmetry-pro", "error": "..."}],
+          "failed_plugin_count":  1,
+          "probed_plugins":       [{"name": "clawmetry-pro",
+                                    "value": "clawmetry_pro.ext:register_all",
+                                    "importable": true, "error": null}, ...],
+          "probed_plugin_count":  1,
+          "events":               ["session.snapshot", ...],
+          "handler_counts":       {"session.snapshot": 2, ...}
         }
 
     ``plugins`` and ``failed_plugins`` are complementary — a given entry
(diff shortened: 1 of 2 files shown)


vivekchand:
C6 gate failure here is the standing repo-admin issue tracked in #3864 — the workflow doesn't run any tests, it posts the reminder above and fails to nag someone with `Administration: write` into configuring branch protection on `main`. It fires on every PR against this repo and predates this diff, which only touches `routes/extensions.py` + `tests/test_extensions_introspection.py` (nothing E2E / infra). Nothing for me to push a fix for — leaving this red pending the browser-side / PAT fix documented in the bot comment.

The other six checks are unaffected; I'll keep an eye on them as they complete.

---

vivekchand:
Automated maintenance pass (2026-07-24). Marking ready -- CI is fully green except the known repo-level C6 gate (branch protection not yet configured, fails on every PR, unrelated to this diff).

Verified: `probe_plugins()` is present on `main` (`clawmetry/extensions.py`), so the new route code has its backing implementation.

**Human handoff needed before merge** per the FLYWHEEL.md sign-off items in the PR description:
- Install updated `routes/extensions.py` (reinstall wheel or copy into site-packages)
- Restart daemon, verify `curl http://localhost:8900/api/extensions | jq .` includes `probed_plugins` + `probed_plugin_count`
- Cross-check the "installed after startup" scenario and the cloud snapshot diagnostic panel

Once local verification is done, merge and proceed with version bump + `[RELEASE]` per FLYWHEEL.md.

---

vivekchand:

The sole failing check (`C6 Required-status-checks gate`) is a repository admin configuration issue: branch protection on `main` does not yet list the 4 E2E checks as required. All 21 other checks pass; the PR code is clean. Fixing C6 requires an admin to add the required checks in GitHub Settings > Branches > Edit main, or run the `Apply required E2E status checks (C6)` Actions workflow with a PAT that has Administration: write.

---

PR #4012, opened 2026-07-25
fix(cli): short-circuit --version before DuckDB-triggering import

vivekchand:
## Summary

- `clawmetry --version` was SIGSEGVing (exit -11) on ubuntu-latest, breaking `pip install (ubuntu-latest)` CI since commit 394c047 (CI run #5881)
- Root cause: `cli.main()` sets `CLAWMETRY_ROLE=dashboard` then does `from dashboard import main`, which triggers module-level `get_store()` in dashboard.py that opens DuckDB in read-only mode. On some Linux runners duckdb-1.5.5's init path in this mode segfaults
- The import smoke test passes (no `CLAWMETRY_ROLE` set, no DuckDB init) but the CLI smoke test fails because the env var triggers the DuckDB open
- macOS and Windows pass because this is a Linux-specific duckdb segfault path

**Fix:** Detect `--version` early in `cli.main()`, before `CLAWMETRY_ROLE=dashboard` is set and before dashboard is imported. Read `__version__` directly from `dashboard.py` via `importlib.util.find_spec`, print it, and return immediately -- no DuckDB needed for a version query.

## Test plan

- [ ] CI `pip install (ubuntu-latest)` passes: `clawmetry --version` no longer segfaults
- [ ] `clawmetry --version` output still contains "clawmetry" (satisfies test assertion)
- [ ] `clawmetry --help` still works (falls through to `dashboard_main()` as before)
- [ ] All other CI jobs stay green

---
_Generated by [Claude Code](https://claude.ai/code/session_01PGB9rxoYjAfDafBtc6Rita)_


vivekchand:
blocked on repo admin action — skipping (auto-mergeability sweep): the sole failing check is "C6 Required-status-checks gate", which detects that branch protection on `main` doesn't mark the four E2E checks as required. All 26 functional checks pass. Fixing C6 requires `Administration: write` permission — run `bash scripts/close-c6.sh` or use the Actions UI with a fine-grained PAT to resolve.

---

Issue #3970, opened 2026-07-23
E2E Robustness Epic: 7-criteria tracking (automated cron)


vivekchand:
Superseded by #4552, which is the active tracking issue. All 7 E2E criteria are confirmed green for 13+ consecutive days as of 2026-08-26. The cron stop condition has been met -- see #4552 for the final summary and action requested (disable the scheduled task at https://claude.ai/code/scheduled).

_Generated by [Claude Code](https://claude.ai/code)_

vivekchand:
**Fire log entry: 2026-07-23T13:00Z**

Status audit complete. Current criteria:

| Criterion | Status | Notes |
|---|---|---|
| C1: OSS golden path | GREEN | `oss-golden-path.yml` passing on every PR |
| C2: Cloud golden path | GREEN | `cloud-e2e-screenshots.yml` passing |
| C3: Landing golden path | GREEN | `landing-golden-path.yml` passing |
| C4: Cross-repo handoff | GREEN | `cross-repo-handoff.yml` passing |
| C5: Visual diff all tabs post-auth | GREEN | `test_e2e_oss_all_tabs.py` runs as hard gate; `pr-screenshots.yml` injects token. Fixes the 2026-05-17 gateway-token bug. |
| C6: Required status checks | **RED** | 4 checks not configured on `clawmetry/main`. See below. |
| C7: Failure quarantine | GREEN | `quarantine-sweep.yml` passing; `quarantine.txt` empty. |

**Gap closed this fire:** PR #3971 adds a GitHub issue notification step to `apply-required-checks.yml`. When the push-triggered C6 verify exits 1 (checks not configured), it now upserts a comment here on #3970 so you receive a GitHub notification, not just a silent red badge.

**C6 still needs a human action to actually configure branch protection on `clawmetry/main`:**

```bash
bash scripts/close-c6.sh
```

Or via GitHub Settings UI (60 seconds, no PAT needed):
https://github.com/vivekchand/clawmetry/settings/branches
Edit main > Require status checks > add:
- `OSS golden path (wheel + OpenClaw + 9 tabs)`
- `Cross-repo handoff (C4)`
- `MOAT Keystone (13-endpoint bar)`
- `E2E Browser Tests (critical subset)`

`clawmetry-cloud` and `clawmetry-landing` are already configured. Only `clawmetry/main` is missing.

---
_Generated by [Claude Code](https://claude.ai/code)_

vivekchand:
[...]

PR #4047, opened 2026-07-26
fix(license): /api/license/status + /verify shape parity + never-5xx

vivekchand:
## Summary

- `GET /api/license/status` no-license branch used to return only 3 keys, forcing every UI to special-case it and dropping the trust anchor (`pubkey_fingerprint_sha256`) an operator can verify BEFORE installing a key. Also 5xxed on introspection failure, contrary to the never-crash posture the surrounding routes (`/api/entitlement`, `/api/features`, `/api/license/pubkey`) all follow. Now every branch carries the same 10-key envelope from `current_license_info()`, plus `plan: "oss"` (legacy consumers) and `error: str(exc)` on the failure branch. Never 5xxs.
- `POST /api/license/verify` had the same shape-parity gap on its invalid-signature and introspection-failure branches (3 keys vs the 11 that `inspect_key()` returns). Now mirrors the healthy branch's shape with Nones for missing fields plus the trust anchor. Never 5xxs.
- Docstring on `current_license_info()` already promised "every branch returns the SAME field set so a UI can render all three uniformly without special-casing which keys are present" — the two routes were breaking that contract from outside. Fixed.

## Test plan

- [x] `python -m pytest tests/test_license_api.py -q` -- 18 tests, all pass
- [x] Broader focused sweep: `test_license*.py`, `test_entitlement_api*.py`, `test_cli_license_json.py`, `test_is_pro_user_license_branch.py` -- 227 passed, 1 skipped
- [x] `python -m py_compile routes/entitlement.py tests/test_license_api.py`
- [x] `make lint-py` (repo-wide): my touched files (`routes/entitlement.py`, `tests/*.py`) are outside the ruff scope declared in the Makefile (`dashboard.py dashboard_claudecode.py clawmetry/`), and the 213 pre-existing errors in scope-files are untouched.

### Backward compatibility

- Existing tests in `tests/test_license_api.py` asserting `status == "no_license"`, `valid is False`, `tier`, `nodes`, and `status == "invalid"`/`dry_run is True` all still pass unchanged.
- The `plan` key that shipped in the previous no-license shape is preserved on the no-license + error branches for any undocumented downstream consumer.
- No frontend JS/HTML consumes `/api/license/status` -- searched: only `routes/entitlement.py` + `tests/test_license_api.py` reference the endpoint.

### Local operator verification checklist

The remote sandbox can't touch `~/.clawmetry`, the running daemon, or a browser -- the following are for the local operator to confirm before merge:

- [ ] Copy changed files into the installed package: `cp routes/entitlement.py ~/.clawmetry/lib/python*/site-packages/routes/ && cp tests/test_license_api.py ~/.clawmetry/lib/python*/site-packages/tests/`
- [ ] Clear cached bytecode: `find ~/.clawmetry -name __pycache__ -type d -exec rm -rf {} +`
- [ ] Restart the daemon: `launchctl kickstart -k gui/$(id -u)/com.clawmetry.sync`
- [ ] With no license installed: `curl -s http://localhost:8900/api/license/status | jq` -- should return a 10-key envelope with `status: "no_license"`, `plan: "oss"`, and a non-null `pubkey_fingerprint_sha256`.
- [ ] With an active license: same call should still return the healthy shape unchanged (was already correct pre-PR).
- [ ] Simulate an introspection failure (temporarily rename `~/.clawmetry/license.key` to something invalid): the response should be HTTP 200 with `status: "unknown"` and a populated `error` field -- never 500.
- [ ] `curl -s -X POST -H 'Content-Type: application/json' -d '{"key":"CLAW1.not.real"}' http://localhost:8900/api/license/verify | jq` -- should return the full inspect_key shape with `pubkey_fingerprint_sha256` populated and `dry_run: true`.
- [ ] Decrypt the live cloud snapshot, browser-check the license/entitlement tile still renders (no shape regression).

---
_Generated by [Claude Code](https://claude.ai/code/session_014PwjSxnao8KhBVTSBFJXq7)_


vivekchand:
Acknowledging the failing check.

**`C6 Required-status-checks gate`** — this is the repo-wide policy blocker tracked in #3970 (the E2E Robustness Epic). It fires on every open PR because branch protection on `main` doesn't yet enforce the E2E gate (as documented in the `c6-pr-gate-bot` comment above). The failure is unrelated to the code in this PR — it predates my changes, and reproduces on `main` and every other open PR. Only the operator can close it (Settings UI, `bash scripts/close-c6.sh`, or storing an `E2E_ADMIN_PAT`).

All PR-scoped signals for this diff look right:
- 18/18 tests in `tests/test_license_api.py` pass locally.
- 227 pass / 1 skipped across the focused license + entitlement + CLI sweep.
- `py_compile` clean on both changed files.
- `make lint-py` scope (`dashboard.py`, `dashboard_claudecode.py`, `clawmetry/`) untouched by this diff.

Holding here per the drive-to-green convention for known base-branch failures. Will act on any real regression on future re-runs.

---

PR #4080, opened 2026-07-26
fix(c6-health): treat 4xx cross-repo reads as unknown, not error

vivekchand:
## Summary

The `c6-health.yml` daily workflow audits branch protection across all three repos (`clawmetry`, `clawmetry-cloud`, `clawmetry-landing`). The `GITHUB_TOKEN` available inside a workflow is scoped to the repo that owns the workflow file (`clawmetry`), so reads against the other two private repos (`clawmetry-cloud`, `clawmetry-landing`) return HTTP 404.

**Bug:** `read_protection()` mapped any error (including 404) to the string `"error"`, causing the health report to show `ERROR` for the cloud and landing repos -- making C6 look broken when it is not.

**Fix:** Distinguish HTTP 4xx (access/scope limit -- expected for cross-repo reads) from HTTP 5xx / network failures (genuine unexpected errors):

```python
# Before
if err is not None or data is None:
    return set(), "error"

# After
if err is not None or data is None:
    if isinstance(err, int) and 400 <= err < 500:
        return set(), "unknown"   # private/inaccessible -- can't verify
    return set(), "error"         # 5xx or network -- genuine failure
```

Also updated the "UNKNOWN" row message to "cannot verify cross-repo (private repo or GITHUB_TOKEN scope limit)" and the success body to explain this is expected behaviour.

## Impact

- C6 health report now shows `UNKNOWN` (with a "verify manually" link) instead of `ERROR` for `clawmetry-cloud/main` and `clawmetry-landing/main`
- `ERROR` is now reserved for genuine 5xx / network failures -- much more actionable signal
- No functional change to branch-protection enforcement; this is reporting accuracy only

---
_Generated by [Claude Code](https://claude.ai/code/session_011sZrpdVgqdzmECWxAEEimw)_

### DIFF
--- a/.github/workflows/c6-health.yml
+++ b/.github/workflows/c6-health.yml
@@ -3,11 +3,12 @@ name: C6 daily health check (branch-protection audit -- all 3 repos)
 # Audits whether required E2E status checks are configured on main for
 # ALL 3 repos: clawmetry, clawmetry-cloud, clawmetry-landing.
 #
-# clawmetry/main protection is read with github.token (same-repo).
-# For clawmetry-cloud and clawmetry-landing: unauthenticated reads via
-# the public /branches/main endpoint. The scoped GITHUB_TOKEN cannot be
-# used cross-repo (returns 403 even on public repos); unauthenticated
-# reads return the full protection.required_status_checks object.
+# clawmetry/main protection is read with github.token (same-repo,
+# authenticated). For clawmetry-cloud and clawmetry-landing: these are
+# private repos -- GITHUB_TOKEN cannot read them cross-repo (returns 403)
+# and unauthenticated reads return 404. Both are reported as "unknown"
+# (can't verify) rather than "error" (broken). The health check gate
+# decision is based only on the clawmetry/main result (same-repo readable).
 #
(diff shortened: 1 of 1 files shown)


vivekchand:
The "C6 Required-status-checks gate" check failure here is expected behaviour -- it fires on every PR as a reminder that branch protection hasn't been configured yet. It is intentionally excluded from REQUIRED_CHECKS (a check that tells you to require checks cannot itself be required without a circular dependency), so it will not block merging this PR.

The changes in this PR are limited to `c6-health.yml` (the daily health reporter). No fix to the gate check is possible via code -- closing C6 requires an admin to configure branch protection on all 3 repos. Options:

- **Terminal (30s):** `bash scripts/close-c6.sh` (needs `gh` CLI authenticated as repo admin)
- **Actions UI:** [Apply required E2E status checks](https://github.com/vivekchand/clawmetry/actions/workflows/apply-required-checks.yml) -- workflow dispatch with a fine-grained PAT (Administration: read+write)
- **Browser (60s):** Settings > Branches > edit `main` on each repo, add the required checks listed in the gate comment

This PR is safe to merge whenever convenient. The gate comment will disappear from future PRs once branch protection is in place.

---

PR #4096, opened 2026-07-27
fix(c6): schedule-heal exits 0 when checks already configured via Settings UI

vivekchand:
## Summary

- **Bug fixed**: `c6-schedule-heal.yml` always exited 1 when `E2E_ADMIN_PAT` was absent, even when branch protection was already correctly configured (e.g. via the GitHub Settings UI). This meant the workflow would produce red CI badges every 2 hours forever after C6 was closed via Option A -- the admin would see perpetual CI noise even after doing the right thing.

- **Root cause**: The schedule-heal only checked `has_pat == false` to decide whether to error out. It never checked whether checks were actually configured. The read-only state check wrote its results only to the job summary, not to `GITHUB_OUTPUT`, so downstream steps could not use them.

- **Fix**:
  - `c6-schedule-heal.yml`: the `read-state` step now exports `checks_ok=true/false` to `GITHUB_OUTPUT`. A new "C6 already configured" success step fires when `checks_ok == true`, logging the green state clearly. The error/apply/post-apply steps are now gated on `checks_ok != true` so they skip when Settings UI already has the right protection in place.
  - `scripts/apply_required_status_checks.py`: fix 2 stale `#3970` references in docstring and error message (tracking issue is `#4029`).

## Why this matters

After the admin configures branch protection via Settings UI (Option A -- the simplest close path), the schedule-heal workflow was supposed to self-heal to green on the next run. Before this fix it would NEVER self-heal: the PAT check happens before the protection check, so the workflow always hit the `exit 1` error step. The admin would have to also set `E2E_ADMIN_PAT` (a separate step) to get a green badge.

After this fix: configure branch protection via Settings UI -> next 2-hour run detects the checks -> exits 0 -> CI badge turns green. The PAT is only needed for auto-application; once C6 is done, it is not needed.

## Test plan

- [ ] Merge this PR
- [ ] Configure branch protection via Settings UI (Option A in the issue/workflow comments) for clawmetry/main
- [ ] Wait for the next `c6-schedule-heal.yml` run (within 2 hours) -- confirm it exits 0 and shows "C6 is GREEN" in the logs
- [ ] Confirm the CI badge turns green

## Acceptance test

After merge, when branch protection is configured on clawmetry/main (via any method), the next `c6-schedule-heal.yml` run shows conclusion=success instead of failure.

Closes criterion: C6 self-heal correctness (C6 detection was already working; now the exit code matches the detection).

Tracking: vivekchand/clawmetry#4029

---
_Generated by [Claude Code](https://claude.ai/code/session_016Q37rmoPrt6a1xjoprY3Xc)_

### DIFF
--- a/.github/workflows/c6-schedule-heal.yml
+++ b/.github/workflows/c6-schedule-heal.yml
@@ -5,10 +5,10 @@ name: C6 auto-heal (required-checks application)
 # Self-heals C6 within 2 hours after E2E_ADMIN_PAT secret is added.
 #
 # When E2E_ADMIN_PAT is NOT set:
-#   - emits an ::error:: annotation (visible red badge)
-#   - writes a formatted job summary showing current required-check state
-#     and a one-click close path (no log-reading required)
-#   - exits 1 so the badge stays red until the PAT is configured
+#   - reads current branch protection state (read-only, no admin needed)
+#   - if checks already configured (e.g. via Settings UI): exits 0 (C6 green)
+#   - if checks not yet configured: emits ::error:: annotation and exits 1
+#     (red badge acts as a forcing signal until the admin acts)
 #
 # When E2E_ADMIN_PAT IS set:
 #   - runs scripts/apply_required_status_checks.py to configure checks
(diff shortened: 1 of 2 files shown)


vivekchand:
**CI status note**: The "C6 Required-status-checks gate" failure is expected and pre-existing -- `c6-pr-gate.yml` exits 1 on every PR because branch protection is not yet configured (that is C6 itself). It is not caused by this PR's changes.

This PR only touches `c6-schedule-heal.yml` (a scheduled/dispatch-only workflow, not triggered on PRs) and `scripts/apply_required_status_checks.py` (a Python script). Neither can cause a CI regression on pull_request events.

OSS Golden Path (C1) is in progress on this branch and should confirm the changes are safe.

---

PR #4097, opened 2026-07-27
feat(license): pro-installation helpers + /api/license/pro-installed{,ation} endpoints

vivekchand:
## Summary

- Adds three public helpers on `clawmetry.license` and their paired read-only HTTP endpoints so a paywall / install-health UI can bind to the `clawmetry-pro` wheel install-state axis without reaching into the private `_pro_installed_version` / `_read_pro_marker` helpers.
- The wheel install-state axis is independent from the license *claim* (tier / exp / etc.) already covered by the existing `/api/license/tier`, `/api/license/is-expired`, and `/api/license/status` endpoints. Splitting the two lets an operator diagnose "activated but wheel missing" (`CLAWMETRY_OFFLINE=1` install, air-gapped node, failed download) apart from "wheel installed but licence expired" (paid feature stops unlocking on renewal lapse).
- Grace-mode behaviour is unchanged — these are pure reads that never touch enforcement.

### New public helpers (`clawmetry/license.py`)

| Function | Returns | Purpose |
|---|---|---|
| `pro_installed_version()` | `str \| None` | Thin never-raises wrapper for the existing private `_pro_installed_version()` `importlib.metadata` probe. |
| `pro_installed()` | `bool` | Scalar gate — "is the paid wheel actually importable right now?" Empty version strings collapse to `False` so a broken metadata reader doesn't render as a green health tile. |
| `pro_installation_info()` | `{installed, version, marker}` | Combined envelope. Live wheel version and on-disk provision marker are surfaced side-by-side because they can disagree in normal operation (marker present + wheel pip-uninstalled; wheel present + marker never written on a pre-marker install) and that disagreement is exactly what an operator debugging a paywall glitch needs to see. |

### New endpoints (`routes/entitlement.py`, `bp_entitlement`)

| Endpoint | Payload |
|---|---|
| `GET /api/license/pro-installed` | `{installed: bool, version: str \| null}` |
| `GET /api/license/pro-installation` | `{installed: bool, version: str \| null, marker: {installed_at, version, source, node_id, …}}` |

Both endpoints degrade to their empty envelope at HTTP 200 on any underlying introspection failure, matching the never-5xx posture of the sibling `/api/license/*` scalar endpoints.

## Test plan

- [x] `python -m pytest tests/test_license_pro_installation.py` — 21 cases green
- [x] `python -m pytest tests/test_license.py tests/test_license_api.py tests/test_license_is_expired_is_perpetual.py tests/test_license_tier_scalar.py tests/test_license_pubkey.py` — 170 sibling cases still green (no regression)
- [x] `python -m pytest tests/test_entitlement_api.py tests/test_entitlement_api_fallback_shape.py` — 39 sibling entitlement cases still green
- [x] `python -m py_compile clawmetry/license.py routes/entitlement.py tests/test_license_pro_installation.py` — clean

Coverage matrix on the new test file:

- Module-level scalar `pro_installed_version` — absent / present / never-raises.
- Module-level scalar `pro_installed` — absent / present / empty-string / never-raises.
- Module-level envelope `pro_installation_info` — every four-way branch of the (wheel × marker) matrix, plus wrong-type marker and never-raises on either leg failing independently.
- Endpoint `/api/license/pro-installed` — every branch, never-5xx.
- Endpoint `/api/license/pro-installation` — full-healthy / marker-only / no-license / never-5xx, and pins that the envelope shape is stable across every branch so a UI can bind to one code path.

Tests use an ephemeral `_PRO_MARKER_PATH` under `tmp_path` and monkeypatch `_pro_installed_version` so no real site-packages or `~/.clawmetry/` state is touched.

I can't reach the host from this sandbox, so the local-daemon / live-cloud legs are on you:

- [ ] Copy the changed files into the venv the daemon uses: `cp clawmetry/license.py ~/.clawmetry/lib/python3.11/site-packages/clawmetry/` and `cp routes/entitlement.py ~/.clawmetry/lib/python3.11/site-packages/routes/` (adjust `python3.X` to match your daemon's interpreter).
- [ ] Clear the byte-cache so the new code is picked up: `find ~/.clawmetry/lib -name __pycache__ -exec rm -rf {} +`
- [ ] Restart the daemon: `launchctl kickstart -k gui/$(id -u)/com.clawmetry.sync`
- [ ] Hit both new endpoints against the running dashboard:
  - `curl -s http://localhost:8900/api/license/pro-installed | jq`
  - `curl -s http://localhost:8900/api/license/pro-installation | jq`
- [ ] Sanity-check the existing sibling endpoints still return their unchanged shape:
  - `curl -s http://localhost:8900/api/license/status | jq`
  - `curl -s http://localhost:8900/api/license/is-perpetual | jq`
- [ ] If clawmetry-pro is installed on the host: `installed=true`, `version` is a version string, `marker.installed_at` is a unix seconds int.
- [ ] Decrypt a live cloud snapshot to confirm the new endpoints are reachable through the cloud relay and don't break the snapshot decrypt path.
- [ ] Browser-check `/entitlements` (or wherever the paywall UI lives) after redeploy — nothing should regress since these are additive read-only endpoints.

---
_Generated by [Claude Code](https://claude.ai/code/session_01JMvgCjuc8z1UPbvXwqUnJr)_

### DIFF
--- a/clawmetry/license.py
+++ b/clawmetry/license.py
@@ -1191,3 +1191,85 @@ def is_perpetual() -> bool:
     except Exception as exc:
         logger.warning("license: is_perpetual failed: %s", exc)
         return False
+
+
+def pro_installed_version() -> str | None:
+    """Public alias for :func:`_pro_installed_version` -- the on-disk
+    ``clawmetry-pro`` package version, or ``None`` if the paid wheel is
+    not importable.
+
+    Kept as a thin wrapper so external callers (routes, CLI, dashboards)
+    can bind to a stable public name without reaching into the underscore
+    helper. The underlying reader is idempotent (pure ``importlib.metadata``
+    lookup) and never raises; this wrapper preserves both properties.
+    """
+    try:
+        return _pro_installed_version()
+    except Exception as exc:
+        logger.debug("license: pro_installed_version wrapper failed: %s", exc)
+        return None
+
+
+def pro_installed() -> bool:
+    """Scalar gate for "is the paid wheel actually importable right now?"
+
+    Returns ``True`` iff :func:`pro_installed_version` returns a non-empty
+    version string. Every other state -- wheel never provisioned, wheel
+    unpacked but ``importlib.metadata`` can't see it, transient
+    introspection failure -- collapses to ``False`` so a paywall renderer
+    or a health tile bound to this helper never crashes on a partial
+    install.
+
+    Complements :func:`is_tier` / :func:`license_tier` (which read the
+    license *claim*): a healthy Pro install must have both a signed
+    Pro-tier claim AND the wheel on-disk. Splitting the two lets an
+    operator diagnose "activated but wheel missing" (download failed,
+    ``CLAWMETRY_OFFLINE=1`` on a fresh install, air-gapped node) from
+    "wheel installed but license expired" (paid feature stops unlocking
+    on renewal lapse).
+
+    Never raises.
+    """
+    return bool(pro_installed_version())
+
+
+def pro_installation_info() -> dict:
+    """Operator-facing description of the ``clawmetry-pro`` install state.
+
+    Combines the two independent facts a paywall / install-health UI
+    typically wants together:
+
+    * ``installed`` / ``version`` -- can Python actually import
+      ``clawmetry-pro`` right now? (Live ``importlib.metadata`` probe.)
+    * ``marker`` -- the ``~/.clawmetry/pro_installed.json`` sidecar written
+      by :func:`_write_pro_marker` at provision time
+      (``installed_at`` unix seconds, ``source``, ``node_id``, the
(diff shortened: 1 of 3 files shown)


vivekchand:
Ack — the `C6 Required-status-checks gate` failure is the passive infra nudge that fires on every PR until a repo admin turns on branch protection (see #4029 and the c6-pr-gate-bot comment above). It doesn't exercise this PR's diff — the job's only step is to post the reminder comment and exit non-zero. Nothing to fix from this branch; it clears repo-wide once C6 is closed, not per-PR.

The functional checks that do exercise the diff (`Syntax & Lint`, `PR build`, `Install + boot + health-check`, `OSS golden path`, `Cross-repo handoff`, `visual-diff`) are still running — I'll drive those to green if any of them go red.

---

PR #4117, opened 2026-07-27
fix(windows): Claude Code discovery, portable system metrics, runtime-aware setup

vivekchand:
## Why

ClawMetry has a product demo on Windows coming up, and running the dashboard on a real Windows 10 machine surfaced four classes of Windows breakage. None of them crashed anything: every one failed silently, rendering empty tiles or dead ends instead of errors.

## What

1. Claude Code sessions never synced on Windows. The cwd slug encoder handled only "/" and ".", so a Windows cwd like C:\Users\Vivek never matched the real ~/.claude/projects/C--Users-Vivek directory. Now also collapses backslash and the drive colon. Verified against the actual directory on a Windows host.

2. Disk, RAM and Load tiles rendered as dashes. They shelled out to df and free and read /proc/loadavg, none of which exist on Windows. helpers/system.py now provides stdlib-only portable equivalents (shutil.disk_usage, GlobalMemoryStatusEx, GetSystemTimes deltas, registry CPU model), and the Load slot falls back to CPU utilisation where load average does not exist.

3. Setup demanded an OpenClaw gateway token even when OpenClaw is not installed. Detection via /api/entitlement/runtime-detection (endpoint existed; nothing consumed it) now drives the setup step: it lists the runtimes actually found on the machine, keeps the gateway token path one click away, and the local install hint switches to PowerShell on Windows.

4. Portability sweep: the sync daemon now detaches properly on Windows (start_new_session is POSIX-only, so closing the launching console killed the daemon), the pid file honours TEMP instead of /tmp, hardware probing falls back from wmic (absent on Windows 11 24H2+) to the registry, and sessions.json is read as UTF-8 (cp1252 locale decode raised on emoji in session titles).

## Verification

- Ran the dashboard on Windows 10 (real machine, not CI): system rows show real values, runtime detection modal renders correctly (Playwright screenshot), PID file writable, hardware probe returns correct CPU/RAM with wmic simulated absent.
- New tests: Windows + UNC slug encoding; existing encoder tests unchanged and green.
- Full targeted test sweep: 324 passed. 13 failures reproduce identically on the unmodified base commit in a clean worktree (pre-existing Windows failures, list in comments).

### DIFF
--- a/clawmetry/cli.py
+++ b/clawmetry/cli.py
@@ -108,8 +108,16 @@ def _maybe_apply_nemoclaw_preset(_input, BOLD, CYAN, DIM) -> None:
         return
 
     print()
-    result = subprocess.run(["bash", script_path], check=False)
-    if result.returncode == 0:
+    # The preset helper is a .sh script. Windows has no bash on PATH by
+    # default, and Popen raises FileNotFoundError rather than returning a
+    # non-zero code, which would abort onboarding instead of falling through
+    # to the "run this manually" hint below.
+    try:
+        result = subprocess.run(["bash", script_path], check=False)
+    except (FileNotFoundError, OSError):
+        print(f"  {DIM('bash was not found on PATH.')}")
+        result = None
+    if result is not None and result.returncode == 0:
         print()
         return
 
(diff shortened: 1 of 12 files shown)


vivekchand:
Auto-mergeability sweep: the only failing check is `C6 Required-status-checks gate`, which requires repo-admin action — add `E2E Gate (required)` to branch protection on `main` (see instructions in the gate output or run `bash scripts/close-c6.sh`). Cannot fix this by code push; skipping.

---

PR #4164, opened 2026-07-28
feat(runtime): default the switcher to the one runtime that has sessions

vivekchand:
Founder ask: a Claude-Code-only machine still made the user pick the runtime by hand. First-load default now follows the data (exactly-one-non-zero rule), persisted via the same path as a manual pick so user choices always win; URL-pinned tabs and OTLP apps are exempt. Placed before the switcher's single-runtime visibility gate, which used to early-return past any chance to default. Verified live in a fresh headless browser on the affected machine.

### DIFF
--- a/clawmetry/static/js/app.js
+++ b/clawmetry/static/js/app.js
@@ -8875,6 +8875,27 @@ function _cmPopulateGlobalRuntime(counts) {
   // Visibility gate: zero-count free runtimes render when the switcher is
   // shown but must not summon it by themselves — a plain single-runtime
   // install (openclaw only, nothing locked) keeps its switcher-free header.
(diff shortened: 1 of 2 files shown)


vivekchand:

The sole failing check is **C6 Required-status-checks gate**, which is an intentional repo-admin alert: the `E2E Gate (required)` check runs and passes on this PR, but is not yet listed as a required status check in `main`'s branch protection rules. Fixing it requires `Administration: write` access to GitHub repo settings — outside the scope of a code-level sweep. All other 22 checks are green. Options to resolve: GitHub Settings UI at `vivekchand/clawmetry/settings/branches`, or the Actions workflow `Apply required E2E status checks (C6 -- one-shot)` with a fine-grained PAT.

---

PR #4166, opened 2026-07-28
feat(license): license_age_days_at(epoch) scalar + endpoint

vivekchand:
## Summary

- Adds `clawmetry.license.license_age_days_at(epoch: int) -> int | None`, the perspective-epoch flavour of `license_age_days()`. Answers "how old was the installed license as of `epoch`?" without the caller having to snapshot the license state at that time or compute `(epoch - iat) // 86400` themselves. Mirrors the existing `days_until_expiry_at(epoch)` scalar (merged in #4162): same coerced-through-`int()` input handling, same `bool`-is-refused guard, same never-raise fallback.
- Adds `GET /api/license/age-days-at?epoch=<int>` on `bp_entitlement`. Envelope carries `age_days` (signed), `requested_epoch` (echoed), `issued_at`, `has_license`, `valid` — same quartet-plus-echo as `/api/license/days-until-expiry-at` so the two endpoints share their reader helpers and a UI binding both cannot catch them disagreeing on install-state fields for the same install.

Posture notes (mirrors `license_age_days()` and `days_until_expiry_at()`):
- **Signed integer, deliberately NOT clamped to `max(0, ...)`** — unlike the "now" flavour `license_age_days()`, which clamps because clock-skew is the only way `iat` can be in the future when reading against `time.time()`. Here the caller EXPLICITLY passes a perspective epoch, so a negative result is a real, actionable signal (they asked a question that only makes sense pre-issuance), not clock skew to be hidden. Matches the signed-integer posture of `days_until_expiry_at`, which returns negative days when the perspective epoch is past `exp`.
- **Lenient on expiry**: a signed-but-lapsed key still surfaces its real `age_days` at the perspective epoch — the support scenario is "how old was this lapsed key evaluated as of last Friday?" — so an audit tile can render "was 12 days old as of `<date>`" without special-casing the expired branch. The `valid` field independently carries the "signature-valid AND not expired" signal for callers that DO want to hide the row on lapsed keys, exactly like the sibling endpoints.
- **`bool` is refused explicitly** despite being an `int` subclass, so a caller that passes `True` / `False` gets `None` rather than a spurious "days from iat to epoch 1" number.
- **Non-integer `epoch` collapses to `age_days=null` + `requested_epoch=null`, HTTP 200**: the bad-input signal is the `null` result, not a 4xx, matching the never-crash posture of the surrounding license endpoints.
- **Missing `iat` claim collapses to `null`** regardless of perspective epoch — nothing to reference against.
- **Invalid signature collapses to `null`**: the payload cannot be trusted (an attacker could stuff any `iat` into an unsigned body), matching how `license_issued_at()` refuses that branch.

Cross-endpoint consistency covered by two tests: the perspective endpoint at `epoch=now` agrees with `/api/license/age-days` within ±1 day (the base endpoint reads `time.time()` with sub-second precision inside `license_age_days` at handle-time while the perspective handler receives an already-truncated `int(time.time())`), and the shared `_license_issued_snapshot()` guarantees `issued_at` / `has_license` / `valid` match `/api/license/issued-at` exactly for the same install.

## Test plan

- [x] `python3 -m pytest tests/test_license_age_days_at_scalar.py -q` — **24 passed**
- [x] `python3 -m pytest tests/test_license_days_until_expiry_at.py tests/test_license_days_until_expiry.py tests/test_license_issued_scalar.py tests/test_license_expires_at_scalar.py tests/test_license_is_expired_is_perpetual.py tests/test_license.py tests/test_license_api.py tests/test_entitlement_api.py tests/test_entitlement_api_fallback_shape.py -q` — **279 passed**, no regressions in adjacent license/entitlement suite
- [x] `python3 -m ruff check --select=E,F,W tests/test_license_age_days_at_scalar.py` — clean (no violations in the new file; the pre-existing 17 lint findings on the wider `clawmetry/license.py` / `routes/entitlement.py` are all outside the hunks this PR touches)
- [x] `python3 -m py_compile clawmetry/license.py routes/entitlement.py tests/test_license_age_days_at_scalar.py` — clean

- [ ] Copy `clawmetry/license.py`, `routes/entitlement.py`, and `tests/test_license_age_days_at_scalar.py` into `~/.clawmetry/lib/pythonX.Y/site-packages/clawmetry/` and `.../site-packages/routes/`.
- [ ] `curl -s "http://localhost:8900/api/license/age-days-at?epoch=$(date +%s)" | jq .` — on an active install, expect `age_days` within ±1 of `/api/license/age-days`'s `age_days`; on OSS-free, expect `{"age_days": null, "requested_epoch": <epoch>, "issued_at": null, "has_license": false, "valid": false}`.
- [ ] `curl -s "http://localhost:8900/api/license/age-days-at?epoch=$(($(date +%s) + 86400*30))" | jq .` — perspective 30 days from now; `age_days` should be roughly 30 days larger than the "now" value (positive delta because the key will be older then).
- [ ] `curl -s "http://localhost:8900/api/license/age-days-at?epoch=$(($(date +%s) - 86400*365*10))" | jq .` — perspective 10 years ago; `age_days` should be strongly negative on any key (support tile can render "would have been -N days old as of `<date>`"). Confirms the endpoint is NOT clamped to 0, unlike `/api/license/age-days`.
- [ ] `curl -s "http://localhost:8900/api/license/age-days-at?epoch=garbage" | jq .` — expect `{"age_days": null, "requested_epoch": null, ...}` with HTTP 200 (never a 4xx).
- [ ] `curl -s "http://localhost:8900/api/license/age-days-at" | jq .` — no `epoch=` arg; expect `age_days=null`, `requested_epoch=null`, HTTP 200.
- [ ] Cross-check: `issued_at` / `has_license` / `valid` on `/api/license/age-days-at?epoch=<any>` must equal the same three fields on `/api/license/issued-at` for the same install (they share `_license_issued_snapshot()`).

---
_Generated by [Claude Code](https://claude.ai/code/session_01Vt5MuLpeaR1zB738cfvnxD)_

### DIFF
--- a/clawmetry/license.py
+++ b/clawmetry/license.py
@@ -1914,6 +1914,91 @@ def license_age_days() -> int | None:
     return age if age >= 0 else 0
 
 
+def license_age_days_at(epoch: int) -> int | None:
+    """Scalar view of "how many days from ``iat`` to ``epoch``?" -- the
+    perspective-epoch flavour of :func:`license_age_days`, for a
+    scheduled-audit / retrospective tile that wants to answer "how old
+    was the license as of <date>?" without the caller having to snapshot
+    the license state at that time or compute ``(epoch - iat) // 86400``
+    at every call site.
+
+    Returns:
+      * ``None`` when there is nothing meaningful to derive: no license
+        file on disk, an invalid signature (the payload can't be trusted
+        -- an attacker could stuff any ``iat`` into an unsigned body), a
+        signed payload whose ``iat`` claim is absent / non-numeric, OR a
+        non-numeric / bool ``epoch`` argument.
+      * A signed integer number of days otherwise. Zero when ``epoch``
+        equals the ``iat`` second; positive when ``epoch`` is after
+        ``iat`` (the normal case -- "N days old as of <date>"); negative
+        when ``epoch`` is BEFORE ``iat`` (support scenario: "the operator
+        rolled a machine back to a pre-issuance timestamp -- how far
+        before issuance were we?").
+
+    Deliberately NOT clamped to ``max(0, ...)`` -- unlike the "now"
+    flavour :func:`license_age_days`, which clamps because clock-skew is
+    the only way ``iat`` can be in the future when reading against
+    ``time.time()``. Here the caller EXPLICITLY passes a perspective
+    epoch, so a negative result is a real, actionable signal (they asked
+    a question that only makes sense pre-issuance), not clock skew to be
+    hidden. Mirrors the signed-integer posture of
+    :func:`days_until_expiry_at`, which returns negative days when the
+    perspective epoch is past ``exp``.
+
+    Days are floor-divided from seconds ``(epoch - iat) // 86400``,
+    matching how :func:`license_age_days` derives its "now" counterpart
+    from ``(now - iat)`` so the two scalars never disagree at the day
+    boundary when ``epoch`` equals the current time.
+
+    ``epoch`` is coerced through ``int()``; ``bool`` is explicitly
+    refused despite being an ``int`` subclass so a caller that passes
+    ``True`` / ``False`` gets ``None`` rather than a spurious "days
+    from iat to epoch 1" number. A non-numeric value collapses to
+    ``None`` so a caller cannot silently miscount on a typo.
+
+    Deliberately lenient on expiry, mirroring :func:`license_age_days`:
+    an expired-but-signature-valid key still carries a meaningful
+    ``iat`` (support scenario: "how old was this lapsed key evaluated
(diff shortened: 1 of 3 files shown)


vivekchand:
C6 gate is a repo-settings issue (tracking #4029) — it fails on every PR regardless of diff and needs operator-only action (Settings UI, `scripts/close-c6.sh`, or the `E2E_ADMIN_PAT` secret). Not something this PR can fix.

Actual code checks so far on `d20e3d7`: **Syntax & Lint** ✅, **Install + boot + health-check** ✅, **Cross-repo handoff (C4)** ✅. Remaining API-test / entitlement / MOAT / E2E jobs are still in-progress; I'll keep watching.

---

PR #4177, opened 2026-07-28
feat(license): is_expired_at(epoch) scalar + endpoint

vivekchand:
## Summary

- Adds `clawmetry.license.is_expired_at(epoch: int) -> bool`, the perspective-epoch flavour of `is_expired()`. Answers "was the installed license expired evaluated as of `epoch`?" without the caller having to snapshot the license state at that time or compare `exp` against a caller-supplied epoch themselves. Mirrors the existing `is_expiring_at(epoch)` predicate and the `days_until_expiry_at(epoch)` scalar: same coerced-through-`int()` input handling, same `bool`-is-refused guard, same never-raise fallback.
- Adds `GET /api/license/is-expired-at?epoch=` on `bp_entitlement`. Envelope carries `is_expired_at` (bool), `requested_epoch` (echoed), `expires_at`, `has_license`, `valid` — same quartet-plus-echo as `/api/license/is-expiring-at` and `/api/license/days-until-expiry-at` so all three share the `_license_expires_snapshot()` reader and a UI binding any two cannot catch them disagreeing on install-state fields for the same install.

Posture notes (mirrors `is_expired()` and `days_until_expiry_at()`):
- **Deliberately lenient on expiry NOW**, unlike `is_expiring_at()` (which refuses lapsed keys because a renewal-window predicate on a lapsed key would gate the WRONG UI). A retrospective "was this expired on `<date>`?" tile absolutely should keep firing `true` on a lapsed key at any epoch past `exp` — that IS the entire support scenario — so the underlying "signature-valid" check does NOT roll in the "not expired now" clause the sibling `is_expiring_at()` uses. Callers that DO want to gate off the current-state validity still have the `valid` field on the endpoint envelope.
- **When `epoch` equals "now", the predicate must agree with `is_expired()`** for the same install — both derive from the same signed `exp` claim and use the same `<=` cutoff, so the two scalars cannot disagree at the boundary. Covered by two tests (active-key and lapsed-key parity).
- **`bool` is refused explicitly** despite being an `int` subclass, so a caller that passes `True` on a lapsed key doesn't silently ask "was the key expired at epoch 1?" and get a spurious positive answer back. The lapsed-key setup in the test proves the refusal is real, not accidental (the `exp <= 1` cutoff would otherwise trivially match).
- **Non-integer `epoch` collapses to `is_expired_at=false` + `requested_epoch=null`, HTTP 200**: the bad-input signal is the `false` result, not a 4xx, matching the never-crash posture of the surrounding license endpoints.
- **Invalid signature collapses to `false`**: the payload cannot be trusted (an attacker could stuff any `exp` into an unsigned body), matching how `license_expires_at()` refuses that branch.
- **Perpetual keys (no `exp` claim) collapse to `false`** regardless of `epoch` — nothing to compare against, mirroring `is_expired()`'s False branch for perpetual keys.

Cross-endpoint consistency covered by three tests: the perspective endpoint at `epoch=now` agrees with `/api/license/is-expired` on both active keys and lapsed keys (both use `exp <= now`), and the shared `_license_expires_snapshot()` guarantees `expires_at` / `has_license` / `valid` match `/api/license/is-expiring-at` and `/api/license/days-until-expiry-at` exactly for the same install (all three shape-endpoints tested together in a single query so a mismatch on any of the three is caught).

## Test plan

- [x] `python3 -m pytest tests/test_license_is_expired_at.py -q` — **28 passed**
- [x] `python3 -m pytest tests/test_license_days_until_expiry_at.py tests/test_license_days_until_expiry.py tests/test_license_is_expired_is_perpetual.py tests/test_license_expires_at_scalar.py tests/test_license.py tests/test_license_api.py tests/test_entitlement_api.py tests/test_entitlement_api_fallback_shape.py -q` — **249 passed**, no regressions in adjacent license/entitlement suite
- [x] `ruff check clawmetry/license.py routes/entitlement.py tests/test_license_is_expired_at.py` — clean
- [x] `python3 -m py_compile clawmetry/license.py routes/entitlement.py tests/test_license_is_expired_at.py` — clean

- [ ] Copy `clawmetry/license.py`, `routes/entitlement.py`, and `tests/test_license_is_expired_at.py` into `~/.clawmetry/lib/pythonX.Y/site-packages/clawmetry/` and `.../site-packages/routes/`.
- [ ] `curl -s "http://localhost:8900/api/license/is-expired-at?epoch=$(date +%s)" | jq .` — on an active install, expect `{"is_expired_at": false, "requested_epoch": <int>, "expires_at": <int>, "has_license": true, "valid": true}`; on OSS-free, expect `{"is_expired_at": false, "requested_epoch": <int>, "expires_at": null, "has_license": false, "valid": false}`.
- [ ] `curl -s "http://localhost:8900/api/license/is-expired-at?epoch=$(($(date +%s) + 86400*365*10))" | jq .` — perspective 10 years from now on an active key; `is_expired_at` should flip to `true` (assuming the key expires in less than 10 years), `valid` still `true` (not expired NOW). Confirms the retrospective/future predicate is decoupled from current-state validity.
- [ ] `curl -s "http://localhost:8900/api/license/is-expired-at?epoch=$(($(date +%s) - 86400*365*10))" | jq .` — perspective 10 years ago; expect `is_expired_at: false` on any signed key (`exp` is definitely > `epoch` there).
- [ ] `curl -s "http://localhost:8900/api/license/is-expired-at?epoch=garbage" | jq .` — expect `{"is_expired_at": false, "requested_epoch": null, ...}` with HTTP 200 (never a 4xx).
- [ ] `curl -s "http://localhost:8900/api/license/is-expired-at" | jq .` — no `epoch=` arg; expect `is_expired_at=false`, `requested_epoch=null`, HTTP 200.
- [ ] Cross-check: `expires_at` / `has_license` / `valid` on `/api/license/is-expired-at?epoch=<X>` must equal the same three fields on both `/api/license/is-expiring-at?epoch=<X>` and `/api/license/days-until-expiry-at?epoch=<X>` for the same install (they share `_license_expires_snapshot()`).
- [ ] Boundary check: on an active install, `/api/license/is-expired-at?epoch=$(date +%s)` must return the same value as `/api/license/is-expired`'s `expired` field (both compute `exp <= now`).

---
_Generated by [Claude Code](https://claude.ai/code/session_011H3jRi7i9Be97q5s48emwG)_

### DIFF
--- a/clawmetry/license.py
+++ b/clawmetry/license.py
@@ -1343,6 +1343,73 @@ def is_expired() -> bool:
         return False
 
 
+
+def is_expired_at(epoch: int) -> bool:
+    """Boolean gate for "was the installed license expired evaluated as of
+    ``epoch``?" -- the perspective-epoch flavour of :func:`is_expired`, for
+    a scheduled-audit tile that wants to answer "would we have shown the
+    expired banner on <date>?" without the caller having to snapshot the
+    license state at that time or compare ``exp`` to a caller-supplied
+    epoch themselves.
+
+    Returns ``True`` iff a license is installed, signature-valid, carries
+    an ``exp`` claim, AND ``exp <= epoch``. Returns ``False`` for every
+    other state: no license file, invalid signature (an attacker could
+    stuff any ``exp`` into an unsigned body, so we refuse to trust it),
+    perpetual / no-``exp`` keys (nothing to compare against), and any
+    ``exp`` value strictly greater than ``epoch``.
+
+    When ``epoch`` equals "now", this predicate must agree with
+    :func:`is_expired` for the same install -- both derive from the same
+    signed ``exp`` claim and use the same ``<=`` cutoff, so the two
+    scalars cannot disagree at the boundary. On any other epoch this
+    helper answers the retrospective question that :func:`is_expired`
+    cannot -- e.g. "was this key already expired last Friday?" (positive
+    signal even on a key that has since been renewed) or "will this key
+    be expired at our next quarterly audit?" (positive signal on an
+    active key whose ``exp`` falls before the audit date).
+
+    Deliberately lenient on expiry NOW, unlike :func:`is_expiring_at`
+    (which refuses lapsed keys because a renewal-window predicate on a
+    lapsed key would push callers to gate the WRONG UI). A retrospective
+    "was this expired on <date>?" tile absolutely should keep firing
+    ``True`` on a lapsed key -- that is the entire support scenario --
+    so the "signature-valid" check here does NOT roll in the "not
+    expired now" clause the sibling :func:`is_expiring_at` uses.
+
+    ``epoch`` is coerced through ``int()``; ``bool`` is explicitly
+    refused despite being an ``int`` subclass so a caller that passes
+    ``True`` / ``False`` gets ``False`` back rather than a spurious
+    "was the key expired at epoch 1?" answer. A non-numeric value
+    collapses to ``False`` so a caller cannot silently mis-gate on a
+    typo.
+
+    Never raises. Any underlying introspection failure (import error,
+    corrupt install, cryptography-lib mismatch) collapses to ``False``
+    so a scheduled audit job never crashes on a bad install.
+    """
+    if isinstance(epoch, bool):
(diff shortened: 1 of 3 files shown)


vivekchand:

---

PR #4183, opened 2026-07-28
fix(update): CI runners never self-update mid-job (kills the visual-diff flake)

vivekchand:
## Why
The recurring visual-diff "flake" was never chance. The updater's startup check fires 60s after boot (`CLAWMETRY_UPDATE_CHECK_SECS` default 60). Whenever a `[RELEASE]` had just published, the PR-branch HEAD dashboard in CI (versioned one release behind main) saw the newer wheel on PyPI, pip-updated, and exec-restarted itself ~60s into the screenshot sweep - every subsequent tab logged `[render-error] base=200 head=0` while the BASE server (already on the bumped version) sailed through. Failure rate tracked release cadence, and the flywheel releases constantly. Diagnosed 2026-07-29 after three consecutive identical kills on #4181/#4182 (last head-server log line: the store's RO handle evicted by the updater's writer, then process exit).

## What
`_env_auto_update_disabled()` now treats a truthy `CI` env (set by GitHub Actions, GitLab, CircleCI, Travis, Azure) as implicit disable: an ephemeral runner must render the code it was started with. Explicit `CLAWMETRY_AUTO_UPDATE=1` re-arms it; `=0` still hard-disables everywhere. The guard covers `_maybe_auto_update`, the fast loop, and is visible in `/api/update-check/status` as `env_disabled`.

Auto-update tests that simulate user machines now clear `CI` alongside the kill switch (they run inside GitHub Actions, where `CI=true` would otherwise trip the new guard).

Note: an equivalent belt-and-suspenders env line for `.github/workflows/pr-screenshots.yml` (`CLAWMETRY_AUTO_UPDATE: "0"`) could not be pushed - the gh token lacks `workflow` scope. This Python-side guard alone fixes the harness; add the workflow line at your leisure if you want the explicit marker.

## Verified
- 28/28 in tests/test_auto_update.py locally (4 new: CI implicit disable, explicit opt-in override, user-machine default unchanged, kill switch precedence).
- Revert-proof: the CI-guard test asserts behavior the old code cannot produce.

### DIFF
--- a/routes/update_check.py
+++ b/routes/update_check.py
@@ -41,9 +41,26 @@ CHANGELOG_URL = "https://github.com/vivekchand/clawmetry/blob/main/CHANGELOG.md"
 
 def _env_auto_update_disabled():
     """Kill switch: ``CLAWMETRY_AUTO_UPDATE=0`` disables unattended upgrades
-    regardless of the stored config (fleet operators / CI / debugging)."""
+    regardless of the stored config (fleet operators / CI / debugging).
+
+    CI runners are implicitly disabled: an ephemeral runner must render the
+    code it was started with, never swap itself for the newest wheel mid-job.
+    The visual-diff harness proved why — whenever a [RELEASE] had just
+    published, the PR-branch dashboard (versioned one release behind) saw the
+    newer wheel on PyPI ~60s after boot, pip-updated, and exec-restarted in
+    the middle of the screenshot sweep; every subsequent page logged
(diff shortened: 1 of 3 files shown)


vivekchand:
blocked on author decision — skipping (auto-mergeability sweep): C6 check fails because `E2E Gate (required)` is not configured as a required status check in `main` branch protection settings — this requires repo-admin access to fix (Settings → Branches → edit `main` → add `E2E Gate (required)` to required checks), not a code change in this PR.

---

PR #4191, opened 2026-07-29
feat(license): pro_install_age_days_at(epoch) scalar + endpoint

vivekchand:
## Summary

- Adds `clawmetry.license.pro_install_age_days_at(epoch: int) -> int | None`, the perspective-epoch flavour of `pro_install_age_days()`. Answers "how old was the `clawmetry-pro` install evaluated as of `epoch`?" without the caller having to snapshot the marker state at that time or compute `(epoch - installed_at) // 86400` themselves. Pairs with `pro_installed_at()` the same way `license_age_days_at(epoch)` pairs with `license_issued_at()`.
- Adds `GET /api/license/pro-install-age-days-at?epoch=<int>` on `bp_entitlement`. Envelope carries `age_days` (int|null), `requested_epoch` (echoed int|null), `installed_at`, `marker_present`, `installed` — same shape as `/api/license/age-days-at` but sourced from the marker rather than the signed `iat` claim. Shares `_pro_install_snapshot()` with `/api/license/pro-installed-at` and `/api/license/pro-install-age-days` so all three necessarily agree on `installed_at` / `marker_present` / `installed` for the same install.

Posture notes (mirrors `license_age_days_at(epoch)` on the signed-license side of the fence):

- **Deliberately NOT clamped to `max(0, ...)`**, unlike the "now" flavour `pro_install_age_days()` (which clamps because clock-skew is the only way `installed_at` can be in the future when reading against `time.time()`). Here the caller EXPLICITLY passes a perspective epoch, so a negative result is a real, actionable signal (they asked a question that only makes sense pre-provisioning), not clock skew to be hidden. Mirrors the signed-integer posture of `/api/license/age-days-at`.
- **When `epoch` equals "now", the perspective scalar must agree with `pro_install_age_days()`** for the same install within one day (fractional-second drift between the two calls: the base scalar reads `time.time()` with sub-second precision inside itself; the caller here passes an `int(time.time())` that already truncated the fraction). Boundary parity is guarded by a paired test on both the scalar and the endpoint.
- **`bool` is refused explicitly** despite being an `int` subclass so a caller that passes `True` doesn't silently ask "days from installed_at to epoch 1?" and get a very negative number back. `False` is also refused for the same reason (both surface as `None`, not `-installed_at // 86400`).
- **Non-integer `epoch` collapses to `age_days=null` + `requested_epoch=null`, HTTP 200**: the bad-input signal is the `null` result, not a 4xx, matching the never-crash posture of the surrounding license endpoints (`/api/license/age-days-at`, `/api/license/days-until-expiry-at`, `/api/license/is-expiring-at`).
- **Independent of live importability**: an operator can have the marker on disk yet Python cannot currently import `clawmetry-pro` (wheel was pip-uninstalled since). That disagreement is exactly what a paywall-debug tile needs to surface rather than collapse both facts into one boolean. `age_days` still surfaces from the marker in that case; the `installed` field independently carries the live-import signal for callers that DO want to hide the row on a broken install.
- **Never 5xxs**: any underlying failure (corrupt marker, blown-up snapshot helper) degrades to `{age_days: null, requested_epoch: <echoed|null>, installed_at: null, marker_present: false, installed: false}` — the "no marker" branch shape, matching the pattern used by every other license endpoint on this Blueprint.

Cross-endpoint consistency is covered by three tests: the perspective endpoint at `epoch=now` agrees with `/api/license/pro-install-age-days` within one day (both compute from the same marker); and the shared `_pro_install_snapshot()` guarantees `installed_at` / `marker_present` / `installed` match `/api/license/pro-installed-at` and `/api/license/pro-install-age-days` exactly for the same install, regardless of which perspective epoch the new endpoint is queried at.

Zero behaviour change: purely additive, read-only. No new gate is enforced, no tier logic is altered, no existing route or scalar is touched.

## Test plan

- [x] `python3 -m pytest tests/test_license_pro_install_age_days_at_scalar.py -q` — **28 passed** (scalar unit tests + endpoint tests + cross-endpoint consistency)
- [x] `python3 -m pytest tests/test_license_pro_install_age_scalar.py tests/test_license_age_days_at_scalar.py tests/test_license.py tests/test_license_api.py tests/test_entitlement_api.py tests/test_license_pro_installation.py -q` — **191 passed**, no regressions in the sibling suite
- [x] `python3 -m pytest tests/test_license_days_until_expiry_at.py tests/test_license_days_until_expiry.py tests/test_license_is_expired_is_perpetual.py tests/test_license_expires_at_scalar.py tests/test_license_issued_scalar.py tests/test_license_age_days_at_scalar.py -q` — **182 passed**, no regressions in adjacent scalar-endpoint parity suites
- [x] `ruff check clawmetry/license.py routes/entitlement.py tests/test_license_pro_install_age_days_at_scalar.py` — clean
- [x] `python3 -m py_compile clawmetry/license.py routes/entitlement.py tests/test_license_pro_install_age_days_at_scalar.py` — clean

- [ ] Copy `clawmetry/license.py`, `routes/entitlement.py`, and `tests/test_license_pro_install_age_days_at_scalar.py` into `~/.clawmetry/lib/pythonX.Y/site-packages/clawmetry/`, `.../site-packages/routes/`, and `.../site-packages/tests/`.
- [ ] Clear `__pycache__` under all three directories.
- [ ] `launchctl kickstart -k gui/$(id -u)/com.clawmetry.sync` to restart the daemon so a future request loads the new endpoint.
- [ ] Active install — `curl -s "http://localhost:8900/api/license/pro-install-age-days-at?epoch=$(date +%s)" | jq .` — expect `{"age_days": <int>, "requested_epoch": <int>, "installed_at": <int>, "marker_present": true, "installed": true}`; on OSS-free (no marker), expect all four value fields null / false as documented.
- [ ] Future perspective — `curl -s "http://localhost:8900/api/license/pro-install-age-days-at?epoch=$(($(date +%s) + 86400*30))" | jq .age_days` — should be ~30 higher than the "now" value from `/api/license/pro-install-age-days`.
- [ ] Pre-install perspective — `curl -s "http://localhost:8900/api/license/pro-install-age-days-at?epoch=$(($(date +%s) - 86400*365*10))" | jq .` — expect `age_days` to be negative (perspective epoch 10 years before install; the endpoint deliberately does NOT clamp to 0 here, unlike `/api/license/pro-install-age-days`).
- [ ] Typo — `curl -s "http://localhost:8900/api/license/pro-install-age-days-at?epoch=garbage" | jq .` — expect `{"age_days": null, "requested_epoch": null, ...}` with HTTP 200 (never a 4xx).
- [ ] Missing arg — `curl -s "http://localhost:8900/api/license/pro-install-age-days-at" | jq .` — expect `age_days=null`, `requested_epoch=null`, `installed_at`/`marker_present`/`installed` still populated from the on-disk marker.
- [ ] Cross-check: `installed_at` / `marker_present` / `installed` on `/api/license/pro-install-age-days-at?epoch=<any>` must equal the same three fields on both `/api/license/pro-installed-at` and `/api/license/pro-install-age-days` for the same install (they share `_pro_install_snapshot()`).
- [ ] Boundary check: on an active install, `/api/license/pro-install-age-days-at?epoch=$(date +%s)` must return `age_days` within one day of `/api/license/pro-install-age-days`'s `age_days` field (both compute from the same marker; the ±1 tolerance is the fractional-second drift between the two handlers).
- [ ] Decrypt the live cloud snapshot and browser-check the dashboard still loads (this PR only adds a new read-only endpoint; no runtime handler is modified, no gate change, no cloud path is affected, but the safety check is cheap).

---
_Generated by [Claude Code](https://claude.ai/code/session_01GE593XQgSqj86YcgfQ7SFK)_

### DIFF
--- a/clawmetry/license.py
+++ b/clawmetry/license.py
@@ -1555,6 +1555,89 @@ def pro_install_age_days() -> int | None:
     return age if age >= 0 else 0
 
 
+def pro_install_age_days_at(epoch: int) -> int | None:
+    """Scalar view of "how many days from ``installed_at`` to ``epoch``?" --
+    the perspective-epoch flavour of :func:`pro_install_age_days`, for a
+    scheduled-audit / retrospective tile that wants to answer "how old
+    was the ``clawmetry-pro`` install as of <date>?" without the caller
+    having to snapshot the marker state at that time or compute
+    ``(epoch - installed_at) // 86400`` at every call site.
+
+    Returns:
+      * ``None`` when there is nothing meaningful to derive: no marker
+        file on disk, a marker with no ``installed_at`` key, a marker
+        whose ``installed_at`` value is non-numeric / non-positive, OR
+        a non-numeric / bool ``epoch`` argument.
+      * A signed integer number of days otherwise. Zero when ``epoch``
+        equals the ``installed_at`` second; positive when ``epoch`` is
+        after ``installed_at`` (the normal case -- "N days old as of
+        <date>"); negative when ``epoch`` is BEFORE ``installed_at``
+        (support scenario: "the operator rolled a machine back to a
+        pre-provisioning timestamp -- how far before install were we?").
+
+    Deliberately NOT clamped to ``max(0, ...)`` -- unlike the "now"
+    flavour :func:`pro_install_age_days`, which clamps because clock-
+    skew is the only way ``installed_at`` can be in the future when
+    reading against ``time.time()``. Here the caller EXPLICITLY passes
+    a perspective epoch, so a negative result is a real, actionable
+    signal (they asked a question that only makes sense pre-install),
+    not clock skew to be hidden. Mirrors the signed-integer posture of
+    :func:`license_age_days_at`, which returns negative days when the
+    perspective epoch is before ``iat``.
+
+    Days are floor-divided from seconds ``(epoch - installed_at) //
+    86400``, matching how :func:`pro_install_age_days` derives its
+    "now" counterpart from ``(now - installed_at)`` so the two scalars
+    never disagree at the day boundary when ``epoch`` equals the
+    current time.
+
+    ``epoch`` is coerced through ``int()``; ``bool`` is explicitly
+    refused despite being an ``int`` subclass so a caller that passes
+    ``True`` / ``False`` gets ``None`` rather than a spurious "days
+    from installed_at to epoch 1" number. A non-numeric value collapses
+    to ``None`` so a caller cannot silently miscount on a typo.
+
+    Pairs with :func:`pro_installed_at` the way
+    :func:`license_age_days_at` pairs with :func:`license_issued_at`:
+    both derive from the same marker so they cannot disagree at the
+    day boundary. The perspective-epoch flavour lets a scheduled audit
+    tile answer "how old was the pro wheel when we shipped that build
+    last Friday?" without having to snapshot the marker state at those
+    times.
+
+    Independent of live importability -- an operator can have the
+    marker on disk (wheel was provisioned yesterday) even when Python
+    cannot currently import ``clawmetry-pro`` (wheel was pip-
+    uninstalled since), and the age still refers to the marker. Use
+    :func:`pro_installed` alongside this scalar for the live-import
+    signal.
+
+    Never raises. Any exception under the hood degrades to ``None`` so
+    a scheduled audit job never crashes on a bad install.
+    """
+    if isinstance(epoch, bool):
+        # ``bool`` is a subclass of ``int``; refuse it explicitly so a
+        # caller that passes ``True`` doesn't silently ask "days from
+        # installed_at to epoch 1?" and get a very negative number back.
+        return None
+    try:
+        wanted = int(epoch)
+    except (TypeError, ValueError):
+        return None
+    try:
+        installed = pro_installed_at()
+    except Exception as exc:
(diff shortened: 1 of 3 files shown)


vivekchand:
C6 Required-status-checks gate red — not fixing from this PR.

The failure has nothing to do with the diff (all new tests pass locally; the sibling license/entitlement suites are green). It's the repo-wide branch-protection posture on `main` that the [c6-pr-gate-bot comment](https://github.com/vivekchand/clawmetry/pull/4191#issuecomment-) above lays out: `main` doesn't yet require `E2E Gate (required)` to pass before merge, and C6 will fire red on every PR until Settings > Branches is updated. Two reasons I'm not pushing a fix for it here:

- **Wrong tool.** Repository-admin permission (Administration: write on the repo, or the fine-grained PAT the bot mentions) is what's needed, not a code change. I don't hold that permission from a scheduled fire — the operator has to click through Settings > Branches or run `bash scripts/close-c6.sh` with the right token.
- **Wrong scope for this task.** This PR is a purely additive read-only scalar + endpoint under the open-core grace-only posture. Enforcing a new merge gate on `main` is explicitly outside what a scheduled fire should touch, and the fix would land in a repo-config change, not this branch.

Diff itself: 28 new tests + the pre-existing scalar-endpoint parity suite (373 tests total across the four adjacent files) are green, ruff clean, `py_compile` clean. Full test / verification checklist is in the PR body.

---

PR #4375, opened 2026-08-01
feat(context): merge LLM Context tab into Context usage (honest, session + runtime scoped)

vivekchand:
## Why

The LLM Context Inspector was showing fabricated data. Its Context Composition bars were hardcoded estimates (Safety 120 tokens, Memories 200, tool schemas as 3.5% of the window), its section names were OpenClaw-only (SOUL.md, Heartbeats) even under a Claude Code filter, and its headline gauge used the node-wide cumulative token total, producing readings like 761K / 200K (100%). It was also the last remaining `_CM_RT_AGGREGATE` tab: the runtime switcher never scoped it.

Context usage (context-economics) already tells the same story from real per-turn usage readings, so this PR merges the Inspector into it and makes the survivor fully honest: every number is measured, and everything scopes by session and by runtime.

## What

**Merge**
- `tabs/context.html`, `loadContextInspector()`, both nav items and all capability-map entries removed. `switchTab('context')` aliases to `context-economics`, so old deep links and bookmarks still land.
- `_CM_RT_AGGREGATE` is now empty, with a unit guard so no silently-aggregating tab can come back.
- context-economics moves from the COST to the EVENTS capability, so no-cost runtimes (Cursor, PicoClaw, NanoClaw) keep a context surface.
- Screenshot/visual-diff tab lists and nav tests updated in the same PR.

**Honest scoping**
- `/api/context-economics` accepts `?runtime=` with server-side prefix scoping (`_runtime_session_id_clause`), the local twin of the snapshot `contextEconomics.byRuntime` slice the cloud interceptor already serves. Unknown runtime returns empty, never node-wide.
- The loader passes the runtime switcher value, and session chips now also filter client-side. This fixes a cloud no-op where the snapshot interceptor ignores `session_id`, so picking a chip silently kept showing all sessions on the hosted dashboard.
- Summary chips are recomputed from the scoped lists, so they always agree with the gauge and the compaction log.

**New gauge**
- "Current context window": the latest measured input plus cache tokens against that turn's model-aware window, labeled with the session and model. This is the Inspector's headline feature rebuilt on real data.

**Bug found during verification (fixed and guarded)**
- `JSON.stringify(...)` inside double-quoted `onclick` attributes emits raw double quotes that terminate the attribute, so the session chips, transcript links, and swimlane lane controls all threw on click. New `attrJsStr()` helper applied to all 9 sites, plus a static guard that flags 8 broken lines on origin/main and zero after this PR.

**Copy**
- Tab subtitle is runtime-neutral now ("when your agent compacted", not "when OpenClaw compacted"); title unified with the nav as Context usage.

## Verification

- `tests/test_context_econ_runtime_scope.py` (new): runtime scoping against a real DuckDB fixture; revert-proven (3 tests fail on origin/main backend, all pass with the fix).
- 37/37 all-tabs post-auth E2E green against a live boot of this branch on :8903.
- Runtime scoping curl-verified against the live DuckDB through the daemon proxy: `claude_code` returns only prefixed sessions, `openclaw` only bare sessions, `codex` returns honest empty.
- Browser click-through: nav item gone, alias lands on Context usage, chip click scopes gauge/turns/peak (91 turns to 25), runtime switcher re-scopes, zero console errors. Screenshots attached below.
- `node tests/test_appjs_units.js`: 207 passed, including the new guards.


vivekchand:
Verification screenshots from a live boot of this branch on :8903 (banners hidden per screenshot convention):

**All runtimes** (real gauge: 426K / 1000K tokens, claude-opus-5, 88 turns, session chips with real peak %; note the Developer drawer no longer has an LLM Context item):

all runtimes

**Session-scoped** (real chip click, post attrJsStr fix: gauge scopes to f1424463 at 217K / 1000K, turns and peak recomputed for the session, timeline scoped, honest per-session empty state for compactions):

session scoped

Runtime scoping, curl against the live DuckDB through the daemon proxy: runtime=claude_code returns only claude_code:* sessions, runtime=openclaw only bare-uuid sessions, runtime=codex returns an honest empty set. The shots/pr-4375 branch exists only to host these images and is not for merge.

vivekchand:
**Shipped end to end and verified live.** 0.12.617 published to PyPI (wheel cracked, all 8 merge markers present), cloud pin bumped (clawmetry-cloud #1866 + route-policy classification for the 4 new 0.12.617 license routes), deploy-gate contract updated (#4385, the gate correctly blocked the first promotion on the removed tab), Cloud Run revision promoted, served app.js verified.

Hosted dashboard walk (app.clawmetry.com, Claude Code runtime pinned, session chip clicked, zero console errors):

hosted session scoped

Real measured gauge (360K / 1000K, claude-opus-5), chip scoping works on cloud (118 to 74 turns), honest per-session empty states. Note for anyone hitting an all-empty hosted tab: a stale browser enc key fails every snapshot decrypt (console shows OperationError); re-pairing or re-entering the node key fixes it. That predates this PR.

PR #4399, opened 2026-08-01
feat(security): numbat ingest — ClawMetry as the console for Perplexity's agent-EDR

vivekchand:
## Why
Perplexity open-sourced [numbat](https://github.com/perplexityai/numbat) on 2026-07-29: an endpoint detection & response layer for ~27 AI agent harnesses (50+ CEL rules — secret-exfil chains, permission bypasses, persistence — plus opt-in pre-action blocking). It emits NDJSON findings and explicitly delegates storage/display to operators (its issue #5 is users asking where to route the HTTP sink). ClawMetry already has the destination tables, alert pipeline, and Security surface — this PR wires them together while the launch press cycle is warm: **ClawMetry is the first dashboard for numbat.**

## What
- **`clawmetry/numbat_ingest.py`** — single shared mapper (one collapse engine, per the brain_dedupe rule):
  - `finding` → `security_events` (near-exact shape fit) + a `numbat_finding` events shadow row so count-over-threshold alert rules fire (the `loop_detected` precedent)
  - `enforcement` → `guardrail_events` (deny/no_override → blocked/monitored)
  - raw `event` records **skipped by design** — they mirror session activity ClawMetry already ingests from the harnesses' own files; storing both would double-count
  - wire schema pinned to numbat `0.2.x`; unknown versions counted + logged, never crash
  - ids are deterministic (numbat's `finding_id` is a content hash) → replays dedupe through the existing upserts for free
- **`clawmetry/sync.py`** — `sync_numbat_events`: daemon tail of `~/.numbat/*.ndjson` with per-file byte cursors (numbat's file sink is its only durable output; its HTTP sink buffers in memory and drops on failure — so the tail is the primary path)
- **`routes/infra.py`** — `POST /api/numbat/ingest` on `bp_security`: accepts numbat's `application/x-ndjson` batches (gzip supported; loopback zero-config, remote via bearer = gateway token through the global `/api/*` auth check); writes through the daemon proxy (both ingest methods already allowlisted) with the standard direct-store fallback; fires banner+telegram immediately for critical/high with the standard cooldown
- **`routes/alerts.py`** — opt-in `numbat_finding_default` seed rule (OSS, off by default — numbat publishes no false-positive rates)
- **`docs/NUMBAT.md`** — install one-liner targeting ClawMetry + coexistence guide (OTLP :4318 contention, OpenClaw's exclusive `plugins.allow`, the `/v1/logs` trap)

## Verified
- 13 unit tests: mapper routing contract, id determinism, canonical session-id prefixing (`claude_code:UUID`), schema pinning, DuckDB round-trip with replay dedupe, severity filter
- Flask test-client smoke of the endpoint: 200 with correct per-type counts, gzip replay ingests + dedupes (1 row per record after double delivery), `session_id=claude_code:abc-123`, enforcement lands `deny/blocked`
- Zero new lint errors on touched files (baseline-diffed vs origin/main); new modules ruff-clean

## Notes
- DuckDB-first respected: handlers never read raw files; HTTP path writes via `local_store_via_daemon`; shadow events rows are daemon-side only (bare `ingest` is deliberately not proxy-allowlisted)
- Full research report (repo deep-read, 8 ranked integration plays, phased plan): internal artifact from today's numbat research workflow


vivekchand:
## Real-binary E2E (not synthetic)

Downloaded the actual numbat release (v0.1.2, schema 0.2.0 — matches our pin) and ran a real forensic scan over this machine's Claude Code artifacts, then ran `sync_numbat_events` exactly as the daemon does (`mark_writer_owner()` + isolated store):

- numbat scan: **2,263 artifacts parsed → 942 real findings** (127k NDJSON records with `--emit all`)
- daemon tail: **942/942 findings ingested in 2.7s**; raw `event` records skipped by design
- severities: 822 high / 113 low / 7 medium; real rules firing: `secrets.cloud_secret_manager_read` (715), `tamper.guardrails_off` (113), `exfil.secret_read_and_egress_oneliner` (62), `chain.secret_manager_read_then_egress` (19)
- 942 shadow `numbat_finding` events rows, all attributed `agent_type=claude_code`; session ids canonical (`claude_code:<UUID>`)
- full replay with reset cursor: 942 ingest calls → **still 942 rows** (deterministic ids + upsert dedupe hold)
- cursor advanced to byte 131,714,585 and re-tail with same cursor reads 0 — the offset watermark works on a 131 MB real file

One find during E2E: a non-daemon process calling `get_store().ingest_*` silently no-ops through the `_ProxyStore` (known burn — dashboard writes must use `local_store_via_daemon` kwargs, which the HTTP route already does; the daemon tail runs as writer owner, which this E2E now replicates).

vivekchand:
## Shipped & verified live (FLYWHEEL done bar)

- **Merged**: `98392d5a3` on main (feature) + #4401 release train
- **Released**: PyPI **0.12.625** — wheel cracked and verified to contain `clawmetry/numbat_ingest.py`, `sync_numbat_events`, and the `/api/numbat/ingest` route
- **Cloud**: Auto-deploy Cloud on OSS update → success (no new cloud surface in this phase)
- **Verified live on a production node**: the node auto-updated to 0.12.625 and restarted; ran the real numbat binary (`numbat scan --agent claude --output file`) into the real `~/.numbat/findings.ndjson`; the production daemon tailed it on its next cycle; the live dashboard now serves **942 numbat findings** at `GET /api/security-threats` with correct severities, rule ids (`secrets.cloud_secret_manager_read`, `chain.guardrails_off_then_egress`, …) and canonical `claude_code:<UUID>` session ids.

Follow-ups tracked for Phase 2: Security-tab numbat section with cloud parity (snapshot slice + interceptor), fleet security lane, enforcement→approvals bridge, `clawmetry secure` installer.

PR #4412, opened 2026-08-01
[RELEASE] fix(onboarding): trial/license activation left the daemon dead on fresh installs

vivekchand:
## What

Live-hit 2026-08-01: activating the free trial from the gate on a fresh machine wrote `license.key` but the node stayed dark — `clawmetry status` said "Daemon: Not running", the DuckDB store was never created, and every family runtime (Claude Code with 517 on-disk sessions included) was invisible in the dashboard. The sync.log showed the whole story: `Daemon crashed: No config at ~/.clawmetry/config.json`, a 15s restart, then `Another instance is already running. Exiting.`

## Three gaps, one class

1. **`routes/trial.py` `_ensure_local_daemon`** spawned `python -m clawmetry.sync` with no `config.json` — `run_daemon()` → `load_config()` raises `FileNotFoundError` and the daemon crash-loops without ingesting a byte. It also inherited the dashboard's CWD: `python -m` puts the CWD on `sys.path`, so a dashboard launched from a clawmetry source checkout spawned the (stale) repo copy instead of the installed wheel.
2. **`dashboard.py` `_start_daemon_background`** — same two hazards (five call sites).
3. **`routes/onboarding.py` activate-license** — never started ingestion at all; a license without a daemon is a dashboard with no data.

## Fix

- New `clawmetry.sync.ensure_local_config()`: writes the local-only config shape (mirrors `cli._finish_local`; empty `api_key` is egress-safe via `is_cloud_disabled`) iff none exists; best-effort, never raises.
- Both spawn sites call it before the spawn and pin `cwd` to `$HOME`.
- License activation kicks `_ensure_local_daemon` after success (no-op when a daemon already holds the pid lock); failed activation spawns nothing.

## Tests

- Spawn shape asserts config-bootstrap-before-spawn + neutral CWD (`tests/test_local_trial_sync_gate.py`).
- `ensure_local_config` write-once / no-clobber of an existing config.
- Activation kick fires on success, not on failure (`tests/test_onboarding_gate.py`).
- 31 passed locally across both files.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

PR #4421, opened 2026-08-02
[RELEASE] fix(onboarding): trial/license activation left the daemon dead on fresh installs

vivekchand:
Replaces #4412 (same fix, rebased cleanly onto current main).

## Summary

- On a fresh machine with no `~/.clawmetry/config.json`, activating a trial or license via the dashboard spawned the sync daemon, which immediately crash-looped: `run_daemon()` calls `load_config()` which raises `FileNotFoundError`, so the DuckDB store was never created and every runtime tab stayed empty.
- Adds `ensure_local_config()` to `clawmetry/sync.py` — bootstraps a local-only config (empty `api_key`, `local_only: True`) if none exists. No-op when config already present. Called before every daemon spawn path: `routes/trial._ensure_local_daemon` and `dashboard._start_daemon_background`.
- Fixes a silent wrong-binary bug: `python -m clawmetry.sync` adds CWD to `sys.path`, so a dashboard started from a source checkout would spawn the repo's (possibly stale) `sync.py` instead of the installed wheel. Pinned `cwd=~` in all Popen calls.

## Test plan

- [ ] `test_ensure_local_daemon_spawn_shape` — verifies config bootstrapped before spawn + CWD is `~`
- [ ] `test_ensure_local_config_writes_once` — verifies missing config gets written once; existing config survives
- [ ] `test_trial_activation_calls_ensure_daemon` — pre-existing; verifies daemon kicked after trial activation

### Release plan

- [x] PR title starts with `[RELEASE]` so `release-on-merge.yml` auto-bumps + publishes

---
_Generated by [Claude Code](https://claude.ai/code/session_01F7eq59h28hVfmZiTnYgn9k)_

### DIFF
--- a/clawmetry/sync.py
+++ b/clawmetry/sync.py
@@ -845,6 +845,38 @@ def save_config(data: dict) -> None:
     os.replace(tmp_path, CONFIG_FILE)
 
 
+def ensure_local_config() -> bool:
+    """Write a local-only config if none exists so run_daemon() can boot.
+
+    Every daemon spawn path (dashboard trial/license activation, the CLI's
+    local onboard) requires CONFIG_FILE to exist — load_config() raises
+    otherwise and the daemon crash-loops without ever creating the DuckDB
+    store (live-hit 2026-08-01: trial activated from the gate on a fresh
+    machine left `clawmetry status` at "Daemon: Not running" and the
+    dashboard with no family runtimes). Shape mirrors cli._finish_local:
+    empty api_key is safe because is_cloud_disabled() gates all egress.
+
+    Returns True if a config was written, False if one already existed or
+    the write failed (best-effort; never raises).
+    """
+    try:
+        if CONFIG_FILE.exists():
+            return False
+        import platform as _plat
+        import socket as _sock
+
+        save_config({
+            "api_key": "",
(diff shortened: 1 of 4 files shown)


vivekchand:
Acknowledging the drift-bot findings — they are documentation-only concerns about the 8090.ai blueprint system and do not reflect any issue with the code or tests.

All three findings refer to new behaviour that is intentional and already well-documented in code comments and tests:

1. **`ensure_local_config()` (sync.py)** — bootstraps a local-only config before every daemon spawn so `run_daemon()` doesn't crash on `load_config()`. Behaviour and rationale documented in the function docstring; covered by `test_ensure_local_config_writes_once`.

2. **Config bootstrap in `_start_daemon_background()` (dashboard.py)** — same fix, same rationale; calls `ensure_local_config()` before the `Popen` so the background path doesn't crash-loop either.

3. **`cwd=~` in trial spawn (routes/trial.py)** — pins the working directory to avoid `sys.path` pollution from a source-checkout launch. Rationale in the inline comment.

The blueprint system at factory.8090.ai is an external documentation layer outside the scope of this PR. The code changes are correct, tested, and fix a live production bug (fresh-machine trial activation left the daemon dead). No code change is needed in response to these drift findings.

---

Issue #3917, opened 2026-07-22
Agents tab empty-state lies when runtimes are detected but the sync daemon is not running


vivekchand:
## Observed (real Windows 11 node, 2026-07-22, v0.12.565)

Fresh reinstall, dashboard running, NOT yet connected (so no sync daemon). The header runtime switcher showed 'OpenClaw - 5 sessions' (live adapter detection), while the Agents tab showed the 'No agents yet, and that is fine. ... Nothing to configure.' empty state. Both true separately, but together the screen gaslights the user: OpenClaw is visibly running and the tab says nothing is happening. Founder hit this live and read it as a data bug ('why is agents tab empty even it detected openclaw running??').

Root cause: /api/inventory composes from the DuckDB rollups only the daemon populates; without a running daemon the store is empty. Detection and ingestion have different sources of truth and the empty state only tells the second one.

## Fix

When the local zero path is hit (store enabled, roster empty) AND `registry.detect_all()` finds detected runtimes, the response should carry that context (e.g. `detectedRuntimes`, `daemonRunning`) and the tab should render the honest state: 'OpenClaw is running, but the sync daemon is not ingesting yet. Run: clawmetry connect (or clawmetry sync).' Keep the bare-zero shape for the store-less cloud cold-fallthrough (that early return path is untouched, so the cm-cloud-inventory contract holds).

## Related founder request (same onboarding surface, 2026-07-22)

At install time, when OpenClaw is absent (or present), onboarding should detect ALL supported runtimes on the machine, list what it found, and state plainly: the free tier watches OpenClaw + NVIDIA NemoClaw; the other detected runtimes need a license key (clawmetry activate) or the cloud-hosted plan (signup). Today a machine full of Cursor/Claude Code sessions onboards with no hint of any of that.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

vivekchand:
Looked at this carefully and the fix is straightforward.

**The problem**

`/api/inventory` composes its roster from DuckDB rollups that only the sync daemon populates. When the daemon isn't running yet, `_build_runtime_summary()` returns `{}` and the roster is empty — so the API returns `{"agents": [], "total": 0}`. Meanwhile `registry.detect_all()` works independently of DuckDB and correctly sees OpenClaw on the machine. Those two truth-sources never talk to each other at render time, so the UI shows the contradictory state.

**Files I'll change**

[...]

PR #4439, opened 2026-08-02
fix(inventory): empty state sees paid runtimes pre-pro-install + transient-empty heal

vivekchand:
## Why

Founder report (2026-08-02): after opting for self-host, the Agents tab showed "No agents yet, and that is fine. Nothing to configure." on a machine with 523 Claude Code sessions.

## Root cause

`/api/inventory`'s empty branch detects runtimes via the **adapter registry** — blind to paid runtimes (Claude Code, Cursor, ...) until the clawmetry-pro wheel is installed post-activation. So on a fresh self-host install, detection returned `[]` in exactly the window the issue-#3917 guidance was designed for, and the UI fell back to the generic copy. Two aggravators: the client never re-polls an empty roster (a transient empty sticks until re-click), and the copy hardcodes "10 more runtimes" (we're at 16).

## What

- **Server**: empty branch falls back to the filesystem lite detector (`_detect_other_runtimes_lite`, same one `/api/agent-presence` uses) when the registry sees nothing
- **Client**: 20s retry while the empty state shows, gated to the active + visible inventory tab (perf budget respected); cleared as soon as data renders
- **Copy**: `inventory.empty_body` → `inventory.empty_body_v2`, count-free ("...and every other supported runtime the moment they run"); stale key removed from all 36 locales so everyone falls back to correct English; drops a user-facing double dash

## Verification

- New guards in `tests/test_inventory_empty_state_guidance.py` (5); server fallback revert-proven red→green
- `test_inventory_roster_consistency` + `test_i18n_catalog`: 53/53 green; `node --check` clean
- Reproduced the live scenario with Playwright against the running localhost:8900 dashboard before fixing


vivekchand:
Drift finding acknowledged: the registry IS the primary mechanism and stays so; the lite-detector fallback exists precisely for the window where the registry cannot know paid runtimes yet (pro wheel not installed until post-activation). The blueprint should document this pre-activation fallback; the alternative (registry-only) is the live-hit bug this PR fixes.

PR #4450, opened 2026-08-02
feat(entitlement): has_features_at + has_runtimes_at + /api/entitlement/has-{features,runtimes}-at (plural what-if boolean-gate scalars)

vivekchand:
## What

Fills the plural `_at` slot in the grant-axis boolean-gate family:

**Scalars (`clawmetry/entitlements.py`)**
- `has_features_at(perspective_tier, features) -> bool`
- `has_runtimes_at(perspective_tier, runtimes) -> bool`

**Endpoints (`routes/entitlement.py`)**
- `GET /api/entitlement/has-features-at?tier=<perspective>&features=a,b,c`
- `GET /api/entitlement/has-runtimes-at?tier=<perspective>&runtimes=x,y,z`

Plural what-if siblings of `has_feature_at` / `has_runtime_at` in the same
relationship the existing `has_features` / `has_runtimes` have to
`has_feature` / `has_runtime`. So a pricing matrix that gates on a BUNDLE
(`fleet + otel_export + sso`) can bind ONE boolean per (perspective, bundle)
cell off ONE URL each, instead of walking the singular `has_feature_at` per
item and AND-ing on the client.

## Semantics

Both scalars delegate per-item to the existing singular `_at` helpers, which
are already backed by `_hypothetical_entitlement(grace=off)`. So the plural
answer is **grace-independent by construction**:

- `has_features_at("oss", ["fleet"])` -> `False` in both grace AND enforce
  (whereas the live `has_features(["fleet"])` returns `True` in grace).
- Fold rule mirrors `has_features` / `has_runtimes` exactly: `True` iff every
  item is granted; empty / None / non-iterable bundle -> `False`; any unknown
  / non-string item collapses the fold to `False` (typo catches at callsite).
- Perspective validated once on the outer helper against `_TIER_ORDER`;
  unknown / empty / non-string perspective -> `False`. Never raise.

Endpoint envelope (15 keys, byte-stable across every input branch): adds
`tier` + `perspective_tier_rank` on top of the sibling plural non-`_at`
shape, matching the singular `_at` envelope pattern (skips `upgrade_required`
— a caller compares `required_tier_rank > perspective_tier_rank` themselves).
Runtime alias canonicalisation happens per-token upstream at the endpoint
layer (`claude-code` → `claude_code`), matching `/has-runtimes` exactly.
Missing / blank / all-unknown CSV or unknown tier → 200 with
`has_<axis>_at=false`; resolver / scalar blowup collapses to the OSS-free
fallback. **Never 4xx, never 5xx.**

## Rollout

Perspective-shaped by design so the `_at` answers do not change with the
resolver's grace pass-through. Wiring these into a pricing matrix today
changes **no current behavior** — this is a GRACE-only fire that adds
surface area without touching any enforcement path.

## Tests

76 new hermetic tests in `tests/test_entitlement_has_features_at_has_runtimes_at.py`:

- Scalar semantics: empty / None / non-iterable / unknown / mixed bundles
  across every tier, including grace-independence invariant (grace vs
  enforce yields the same `has_<axis>_at` for the same perspective+bundle).
- Endpoint envelope shape: 15 keys, byte-stable across 12 parametrised URLs
  per axis; never-4xx / never-5xx via monkeypatched resolver and scalar
  blowup.
- Cross-consistency: `required_tier` byte-parity with `/min-tier-for-<axis>`,
  `current_tier` byte-parity with the live `/has-<axis>` sibling, and the
  plural `allowed` bit equals the per-item AND-fold of the singular
  `/has-<axis>-at` sibling.
- Alias canonicalisation upstream at endpoint layer with strict-scalar
  pass-through.

Full new suite: **76 passed**. Sibling suites (`has_feature_has_runtime`,
`has_feature_at_has_runtime_at`, `has_features_has_runtimes`): **187 passed**
after the change.

Only the operator can verify against the live daemon / snapshot / browser —
please run:

1. Copy the three changed files into your live install:
   ```
   cp clawmetry/entitlements.py ~/.clawmetry/lib/python*/site-packages/clawmetry/
   cp routes/entitlement.py     ~/.clawmetry/lib/python*/site-packages/routes/
   ```
2. Clear caches:
   `find ~/.clawmetry/lib -name __pycache__ -exec rm -rf {} +`
3. Restart the daemon:
   `launchctl kickstart -k gui/$(id -u)/com.clawmetry.sync`
4. Hit the new endpoints and confirm 200 + expected boolean:
   ```
   curl -s 'http://localhost:8900/api/entitlement/has-features-at?tier=pro&features=fleet' | jq
   curl -s 'http://localhost:8900/api/entitlement/has-runtimes-at?tier=oss&runtimes=claude_code' | jq
   ```
5. Decrypt the live snapshot and browser-check that no pricing surface
   regresses (no UI is wired to these URLs yet — they are net-new).
6. Confirm `/api/entitlement` still returns the OSS-free grace envelope
   unchanged (this fire adds NO behavior gate).

---
_Generated by [Claude Code](https://claude.ai/code/session_01Fzcdqb7gToA4qifAoMGZUb)_

### DIFF
--- a/clawmetry/entitlements.py
+++ b/clawmetry/entitlements.py
@@ -5802,6 +5802,133 @@ def has_runtimes(runtimes) -> bool:
         return False
 
 
+def has_features_at(perspective_tier: str, features) -> bool:
+    """Hypothetical-perspective sibling of :func:`has_features`: would tier
+    ``perspective_tier`` grant **all** ``features``?
+
+    Plural-fold twin of :func:`has_feature_at` on the same axis, in the same
+    relationship :func:`has_features` has to :func:`has_feature`. Fills the
+    plural ``_at`` slot in the grant-axis boolean-gate family so a pricing
+    matrix that gates on a BUNDLE (``fleet + otel_export + sso -- Available
+    in Enterprise``) can bind ONE boolean per (perspective, bundle) cell off
+    ONE URL each, instead of walking the singular :func:`has_feature_at` per
+    item and AND-ing on the client.
+
+    Fold rule mirrors :func:`has_features` exactly: returns ``True`` iff
+    :func:`has_feature_at` returns ``True`` for **every** item in the
+    iterable, i.e. the tightest single-item denial wins. Consequences of
+    that fold, all deliberate:
+
+    * Empty / ``None`` / non-iterable ``features`` -- returns ``False``.
+      The vacuous-truth answer would silently render "granted" on a caller
+      who forgot to pass the bundle, which is exactly the callsite-typo
+      posture the singular :func:`has_feature_at` catches with its empty-id
+      ``False``. Matches the plural :func:`has_features` empty-fold posture
+      byte-for-byte.
+    * Unknown / empty / non-string ``perspective_tier`` -- returns ``False``.
+      Perspective validation lives on the OUTER helper (once per call) rather
+      than folding into per-item :func:`has_feature_at` calls, so a bogus
+      perspective fails-closed in one place instead of re-parsing the tier
+      per item. Same fail-closed posture as :func:`has_feature_at` on an
+      unknown perspective.
+    * Unknown / empty / non-string item id -- returns ``False``. The singular
+      :func:`has_feature_at` fails-closed on typos so the bundle fold
+      inherits that: ``has_features_at("pro", ["fleet", "Fleeet"])`` is
+      ``False`` even in grace, surfacing the typo instead of silently
+      granting the bundle.
+    * Grace-independence: the answer is derived from the static per-tier
+      grant map via :func:`has_feature_at`'s :func:`_hypothetical_entitlement`
+      backing, so the returned bit is IDENTICAL under grace vs enforce for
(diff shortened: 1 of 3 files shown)


vivekchand:
Not acting on the drift-bot finding. `docs/ENTITLEMENTS.md` intentionally documents only the small curated top-level endpoint table (resolve / refresh / upgrade-diff / lock-reason / required-tier); none of the `/api/entitlement/has-*` scalar family is listed there — not the singular `has-feature` / `has-runtime`, not the perspective siblings `has-feature-at` / `has-runtime-at` this PR mirrors, and not the plural `has-features` / `has-runtimes` this PR extends to the perspective axis. The convention is that per-axis scalar helpers live in their own docstrings (both the module scalar and the endpoint) rather than the top-level table, which the diffs on this PR follow. Expanding scope here to document only the new pair would leave the rest of the sibling family inconsistent; a separate doc pass for the whole `has-*` family would be the right vehicle.

---

PR #4472, opened 2026-08-03
[RELEASE] feat(alerts,approvals): runtime-scoped by default, node-wide by choice (0.12.640)

vivekchand:
Answers the founder's report on the Alerts tab screenshot: alerts/approvals were node-wide relics from before multi-runtime, and the Approvals nav item vanished under a runtime filter because it was gated on the OpenClaw-only `GATEWAY_RPC` capability.

## Alerts
- `alert_rules.runtime` column ('all' default = node-wide back-compat; guarded ALTER for existing fleet DBs).
- **Default = the active runtime filter** when creating a rule; "All runtimes (node-wide)" is the explicit opt-in. Editor gets an *Applies to* selector; rows get a scope chip — a node-wide rule under a runtime filter is labeled, never silent.
- Scoped evaluation end to end, honestly: daily-spend via the per-runtime DuckDB rollup, token-velocity + unproductive-burn via session-prefix-filtered events, event-stream rules filtered in the shared evaluator, and quality rules via NEW per-runtime quality windows (`query_session_quality_window(runtime=…)` + `quality_by_runtime`) — **a scoped rule never fires on a node-wide number**; missing slice = honest under-fire.
- Unknown runtime ids are 400s — never silently widened.

## Approvals
- `?runtime=` on `/api/approvals` + `/api/approvals-audit` (session-prefix scoping); the tab filters pending + history identically. Composes with #4464 (which made the tab a node tab + added hook gates) — this PR makes it **scope**.
- Stale "node-wide" banners removed from both tabs.

## Tests
8 new in `tests/test_alert_runtime_scope.py` — evaluator scoping (incl. openclaw = no-family-prefix semantics and the quality-slice honesty contract), CRUD round-trip/validation, approvals filtering. Battery run: only failures are pre-existing on main (verified against pristine `origin/main`). Zero new lint findings on touched files.

Cloud follow-up (separate PR): the cloud alert API should persist `runtime` from the rule body so cloud-authored scoped rules relay to daemons with scope intact (`condition_json.runtime` is already read by the evaluator).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,3 +1,18 @@
+## 0.12.640
+
+- **Alerts and approvals are runtime-scoped by default.** Alert rules carry a
+  per-rule scope: creating a rule while the runtime switcher is set inherits
+  that runtime, and "All runtimes (node-wide)" is the explicit opt-in (rule
+  rows show a scope chip either way). Scoped rules evaluate per-runtime
+  slices end to end: daily-spend and token-velocity read the per-runtime
+  DuckDB rollups, event-stream rules filter by the session-id prefix, and
+  quality rules (score drop / failure rate) read a per-runtime quality
+  window — a scoped rule never fires on a node-wide number.
+- **The Approvals tab now shows for every runtime** (it was hidden behind the
+  OpenClaw gateway capability) and scopes its pending + history rows to the
+  selected runtime; /api/approvals and /api/approvals-audit accept
+  ?runtime=. The stale "node-wide" banner is gone from both tabs.
+
 ## 0.12.639
 
 - **Runtime feature parity: the dashboard stops being OpenClaw-first.** Every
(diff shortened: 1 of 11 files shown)


vivekchand:
Drift-bot triage:
- **Findings 1–6** (feature not yet documented in requirement/blueprint): forward drift — the *Remote Alerts and Action Approvals* blueprint now has a pending edit documenting per-rule runtime scope, prefix-derived membership, the honest under-fire contract for missing scoped slices, and runtime-scoped approval reads.
- **Finding 7** ("?runtime= not handled server-side") is **incorrect**: `routes/policy.py` extracts `request.args.get("runtime")` in both `/api/approvals` (line ~188) and `/api/approvals-audit` (line ~145) and filters via `_filter_rows_by_runtime()`; covered by `test_approvals_queue_scopes_by_runtime` (passing). The tab's client-side filter is an addition on top for the cloud-mode fetch path, not a substitute.

PR #4494, opened 2026-08-03
feat(spend-flow): node-wide AI spend flow (context in -> runtime -> output out)

vivekchand:
## What

An Opik-style AI spend tracker, per node, local-first, across every runtime: a new **Where the money goes** flow on the Cost/Usage tab showing input spend categories (your messages, earlier replies, tool results, and the derived system prompt + tool definitions overhead) flowing through each runtime into output categories (thinking, replies, built-in vs MCP tool calls), each with tokens and dollars.

## Why

We hold full transcripts for the whole node in DuckDB, so we can measure where context spend actually goes instead of estimating it. This replaces guessy context bars with measured, reconciled numbers and sets up the follow-on savings recommendations (thinking mode, MCP pruning).

## How (honesty contract)

- `clawmetry/spend_flow.py` is a pure engine (mirrors `efficiency.py`: no I/O, never raises, honest `insufficient_data`). Per model call it measures content-based token shares, scales them to the model-reported prompt volume, and books the unexplained remainder as an `overhead` residual with `basis: residual`. Dollars come from the stored `cost_usd` cost-of-record (pricing-table fallback), so category sums equal call totals **by construction**.
- Handles both ingest worlds: OpenClaw v3 SDK envelopes (with the assistant/model.completed sibling dedupe from `query_aggregates`) and family per-block events whose usage lands camelCase in `data.extra` (the 'data.extra $0' bug class). Family adapters strip thinking text at ingest, so thinking output is a per-call residual, flagged as such.
- `LocalStore.query_spend_flow`: windowed walk, 60s TTL cache (CPU budget), 200k row cap surfaced as `row_cap_hit`, runtime scoping via the shared session-id-prefix clause (filtered sums reconcile with the node total exactly).
- `GET /api/spend-flow` via the daemon proxy + fallback; OSS/free callers get the same 24h history cap policy as `/api/usage` (`capped_at_24h`, never silent).
- `sync.py` ships a `spendFlow` snapshot slice; the `cm-cloud-spend-flow` interceptor lands in clawmetry-cloud next (cloud pin bump after release).

## Verification

- 21 unit tests: reconciliation invariants, dedupe, residuals, endpoint clamps, garbage input.
- Real-node E2E on this machine's live DuckDB (56MB copy): 1547 calls / $101.89 over 7d, category dollars reconcile with `query_aggregates` cost-of-record within 0.15%, per-runtime sums equal node total exactly, unknown runtime returns honest empty.
- Rendered-tab screenshot on real data (dark mode, labels collision-free) verified locally on :8901; palettes validated with the dataviz six-checks script in light and dark modes.


vivekchand:
Rendered on this node's real data (local :8901, dark mode):

PR #4504, opened 2026-08-03
fix(approvals): expire stale pending rows on daemon restart

### LINKED ISSUE #4491: approvals: daemon restart orphans in-flight approvals as pending forever (no boot reconciliation)
On a live node (0.12.642), `/api/approvals-audit` showed 132 approvals stuck in `status=pending` locally, some >12h old, policy timeouts long past.

**Cause:** `process_tool_call` writes the pending row then blocks polling for the decision in-process. When the daemon restarts (e.g. autoupdate), the poll threads die; nothing on boot sweeps rows whose `created_at + timeout` has passed, and the cloud's `system:expiry` flip (poll-triggered) never propagates back because nobody polls again. The rows re-surface in every cache_push as pending, cluttering the cloud Approvals inbox until a human bulk-approves.

**Suggested fix:** on watcher start (and periodically), sweep local pending approvals past expiry → apply the policy's on_timeout (or mark `expired`), mirroring update_approval_decision so the cloud queue drains.

Related observation from the same investigation: the PreToolUse hook and the daemon watcher can both process the same tool call (the `_in_flight` dedup is per-process), producing duplicate approval/audit rows for one command. A shared dedup key (tool_call_id) in the approvals table would cover both.

vivekchand:
Closes #4491

## Summary

- Add `_reconcile_stale_pending()` to `clawmetry/approvals.py` — sweeps `pending` approval rows whose deadline has passed, called once at watcher startup before the main loop
- Hook-path rows carry `deadline_ms` + `on_timeout` in their stored `args` JSON; used directly to determine expiry and verdict
- Blocking-path rows (raw tool args, no deadline stored) fall back to a 7-day wall-clock cutoff and are marked `expired`
- Calls the existing `update_approval_decision()` so the cloud cache_push drains on the next heartbeat with no additional coupling

## Test plan

- All 42 approvals unit tests pass (`test_approvals_local_store`, `test_approvals_duckdb_watcher`, `test_approvals_local_blocking`, `test_approvals_deny_kill`)
- `_reconcile_stale_pending()` is guarded by try/except so a query failure on startup never crashes the watcher thread

---
_Generated by [Claude Code](https://claude.ai/code/session_01NgwgB7T4NRCfXaWSku3gj7)_


vivekchand:
The 17 drift-bot threads all flag the same gap: `_reconcile_stale_pending()` is not yet reflected in the external requirement/blueprint docs. No code change needed — the reconciliation behavior is a defensive implementation detail (startup sweep of rows whose deadline already passed), not a new user-facing policy. The 42 approvals unit tests pass and CI is green. Requirement/blueprint docs should be updated in factory.8090.ai to describe the daemon-restart reconciliation and 7-day fallback cutoff; that's out of scope for this PR.

---

PR #4553, opened 2026-08-05
fix(ci): add job-level if guard to c6-schedule-heal to eliminate phantom PR failures

vivekchand:
## Summary

When a PR modifies `.github/workflows/c6-schedule-heal.yml`, GitHub creates a phantom workflow run with `event=push` even though the workflow only declares `schedule` and `workflow_dispatch` triggers. With zero matching jobs, GitHub marks the run conclusion as **failure**, causing false red badges on unrelated PRs (e.g. #4549).

**Fix:** Add a job-level `if: github.event_name == 'schedule' || github.event_name == 'workflow_dispatch'` guard. When the guard is false (push event), the job is **skipped** and the run concludes as **success (skipped)**, eliminating the phantom failure noise.

## Changes

- `.github/workflows/c6-schedule-heal.yml`: Added `if:` guard at the `c6-auto-heal` job level with an explanatory comment

## Test plan

- [ ] Merge this PR; observe that the phantom push-triggered run on this PR itself concludes as success (skipped), not failure
- [ ] Verify that schedule-triggered and workflow_dispatch-triggered runs still execute the job normally
- [ ] Confirm no regression in C6 status after PR #4549 (schedule re-enable) is also merged

Closes #4542 (phantom failure noise eliminated)

Related: #4552 (E2E robustness tracker, C6 criterion)

---
_Generated by [Claude Code](https://claude.ai/code/session_016YMTRrVafoyNWTph4TQKVu)_

PR #4538, opened 2026-08-04
feat(license): is_subject_at_batch + /api/license/is-subject-at-batch

vivekchand:
## Summary

- Adds `is_subject_at_batch(subject, epochs)` — the shared-`subject` batch sibling of `is_subject_at` (and the `/api/license/is-subject-at` scalar endpoint). Where the singular scalar folds ONE `(subject, epoch)` pair to ONE bool, this preserves per-value rows for a fixed `subject` across a sequence of perspective epochs so a scheduled-audit tile answering "was this node bound to <account> on each of these audit dates?" hydrates the whole column in ONE round-trip instead of fanning out N calls to the scalar. Completes the `(value, epochs) -> per-epoch bool` batch triad on the license-subject axis alongside the already-shipped `is_state_at_batch` (#4244) / `is_tier_at_batch` / `has_feature_at_batch`.
- Adds paired `GET /api/license/is-subject-at-batch` endpoint on `bp_entitlement`. Envelope shape mirrors the sibling `/api/license/is-state-at-batch` / `/api/license/is-tier-at-batch` (`kind` / `count` / `requested_subject` / `rows` + current-time snapshot fields `subject` / `expires_at` / `has_license` / `valid`). Shares the existing `_license_subject_at_snapshot` with the sibling `/api/license/subject-at` and `/api/license/is-subject-at` endpoints so a UI binding several for the same install cannot catch them disagreeing on the current-time reference.
- Grace-independent by construction: delegates per-row to `is_subject_at`, which reads the signed `sub` claim via `license_subject_at`. Wiring this into a paywall today changes **no** current behavior — resolved entitlement still defaults to OSS-free-grace. Deliberately open-ended on subject (matching `is_subject` / `is_subject_at`) since a subject typically encodes an account id / email / tenant handle the code here has no business whitelisting.
- 43-test module `tests/test_license_is_subject_at_batch.py` covering scalar + endpoint envelope shape, subject normalisation (case + strip), empty / None / non-iterable / typo input, epoch dedup + first-seen order, bad-epoch collapse, matching / non-matching / dropped-sub / perpetual / bogus-signature branches, per-row parity with `is_subject_at` (helper + endpoint), boundary agreement with `is_subject` at "now", never-4xx / never-5xx posture, envelope key set, cross-endpoint snapshot-field consistency vs `/api/license/subject-at{,-batch}` + `/api/license/is-subject-at`, and row-index alignment with `/api/license/subject-at-batch`.

Files:
[...]

PR #4579, opened 2026-08-06
Auto-update root cause fix + self-host trial silently failing + runtime-chip cloud redirect

vivekchand:
## Summary

- The browser onboarding gate (`routes/onboarding.py`) has been the default first-run path since the 2026-07-31 hard-gate rollout, but its completion handlers only touched the nocloud marker file — they never started or registered a background sync daemon. The only thing left polling PyPI was the foreground dashboard&#39;s in-process checker thread, which stops the moment that one process exits (closed terminal, sleep, reboot, crash), silently and permanently halting auto-update until a human manually relaunched `clawmetry`. Onboarding completion now calls a new `clawmetry/daemon_registration.py::ensure_persistent_daemon()` for every choice (managed, selfhost_license, selfhost_trial).
- Windows gets real daemon persistence for the first time: it had no launchd/systemd equivalent, so `_start_daemon`&#39;s Windows branch fell straight to an unsupervised detached subprocess with zero reboot/crash recovery. It now registers a logon-triggered, restart-on-failure Task Scheduler task (`schtasks`), falling back to the old subprocess only if that fails. `_daemon_supervised()` now detects it too.
- The `auto_update` config flag&#39;s self-heal (re-asserting the default-on `True` after a stale persisted `False`) previously only ran for entitled paid cloud accounts via the heartbeat path (`_sync_auto_update_with_plan`) — self-hosted and free-tier installs had no path back to `True`. The update-check worker now heals a stale unset `False` for every role/tier on boot, while a real explicit user opt-out (`POST /api/update-check/config`) is tracked via a new `auto_update_user_set` flag and never overridden.
- `ci.yml` and `sync-test.yml` had no `concurrency` block, unlike every other workflow in this repo — a push to this PR queued a brand new full CI run without cancelling the previous one, so 4+ superseded runs piled up in the shared runner queue and starved the newest commit&#39;s jobs for over an hour. Added the same concurrency-group pattern used elsewhere in the repo.
- **Self-host trial silently failing** (found live, on a real self-hosted Windows install, mid-PR): the Google/GitHub self-host sign-in card writes a cloud token purely to carry identity into a trial-signup call, before that call has even run. If the trial-signup half then fails for any reason, the account is linked (token on disk) but no license/trial is ever activated — yet the onboarding gate&#39;s `_cloud_connected()` check only looked at whether a token existed, not what it was for, so it reported the install as already onboarded (&#34;managed&#34;) on every later page load. No error, no way to retry, just a permanently half-onboarded dashboard with everything locked. Fixed `_cloud_connected()` to treat a token written under self-host intent (nocloud marker set) as identity-only, not proof of a completed managed choice.
- **Runtime-chip cloud redirect** (found investigating the above): the bottom-right runtime-status chip&#39;s locked-runtime click handler opened `app.clawmetry.com/upgrade` in a new tab, redirecting a self-hosted user to a cloud account page instead of the local sign-in-then-trial modal the header dropdown already uses correctly for the identical click. Wired the chip to the same local `_cmShowRuntimePaywall` flow. Also corrected `tests/test_runtime_switcher_gating.py`, which had gone stale asserting the old cloud-redirect behavior for the (already-fixed) header paywall modal — exactly the kind of drift that let the chip regression go unnoticed.
- **Onboarding-complete request hang** (caught live by this PR's own visual-diff bot, mobile screenshots stuck forever on "Initializing ClawMetry"): the daemon-registration call added above (`register_systemd`/`register_launchd`) shelled out to systemctl/launchctl with no timeout, synchronously inside the `/api/onboarding/complete` HTTP handler the browser's init sequence waits on. A runner/container with no working user systemd/dbus session can make those commands hang instead of failing fast. Bounded every subprocess call in `daemon_registration.py` to a few seconds, and moved the registration call onto a background thread so onboarding-complete never blocks on it at all, regardless of how slow the underlying OS call is.

## Test plan

- [x] New `tests/test_daemon_registration.py`: Windows Task Scheduler registration (`schtasks /create` with `RestartOnFailure`/`LogonTrigger`), `windows_task_registered()` probe, `ensure_persistent_daemon()` platform dispatch + fallback, wiring into `cli.py::_start_daemon`&#39;s Windows branch and `routes/update_check.py::_daemon_supervised()`, plus new tests pinning a `timeout=` on every systemctl/launchctl subprocess call.
- [x] Extended `tests/test_onboarding_gate.py`: both `/api/onboarding/complete` and `/api/onboarding/activate-license` now trigger daemon registration for every choice, dispatched via `threading.Thread` (not called inline) so it never blocks the response; a registration failure never breaks onboarding completion; tests pin `_cloud_connected()` ignoring a token written under the nocloud marker while still honoring a real managed-connect token.
- [x] New `tests/test_auto_update_self_heal.py`: stale unset `auto_update=False` heals to `True` on worker boot for every role/tier; an explicit user opt-out (`auto_update_user_set=True`) is never overridden; heal failures never raise.
- [x] Updated `tests/test_auto_update.py::test_auto_update_in_allowed_config_keys` for the new `auto_update_user_set` stamping behavior.
- [x] Corrected + extended `tests/test_runtime_switcher_gating.py`: the paywall-overlay test now pins the current local-trial-flow contract instead of the stale cloud-redirect one, plus a new dedicated test for the runtime chip&#39;s locked-click handler.
- [x] `ruff check` / `py_compile` clean on every changed Python file.
- [x] Full local pytest run of every changed/added test file green (39/39 for the onboarding + daemon-registration suites after the hang fix). A full repo-wide `pytest tests/` run was also done as a sanity check — it surfaces ~1000 pre-existing failures unrelated to this diff (confirmed via `git stash` + re-run on the unmodified base commit) that trace to this sandbox&#39;s environment (missing DuckDB fixtures/extensions), not to anything in this PR.

---
_Generated by [Claude Code](https://claude.ai/code/session_014M9RFasj1Cak93j3CuF1q5)_

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,6 +5,10 @@ on:
     branches: [main, master]
   pull_request:
 
+concurrency:
+  group: ci-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
 jobs:
   # ── Fast syntax check (runs in ~10s, blocks everything else) ─────────────────────
   lint:
--- a/.github/workflows/sync-test.yml
+++ b/.github/workflows/sync-test.yml
@@ -17,6 +17,10 @@ on:
       - '.github/workflows/sync-test.yml'
   workflow_dispatch:
 
+concurrency:
+  group: sync-test-${{ github.event.pull_request.number || github.ref }}
+  cancel-in-progress: true
+
 jobs:
   sync-test:
     strategy:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,28 @@
 ## Unreleased
 
+- **Fixed the root cause of ClawMetry silently no longer auto-updating.**
+  The browser onboarding gate (`routes/onboarding.py`, the default
+  first-run path since the 2026-07-31 hard-gate rollout) completed a
+  `managed`/`selfhost_*` choice without ever starting or registering a
+  background sync daemon — only the CLI paths (`clawmetry connect`,
+  `clawmetry onboard`) did that. The only thing left polling PyPI was the
+  foreground dashboard's in-process checker thread, which stops the
+  moment that one process exits (closed terminal, sleep, reboot, crash),
+  silently and permanently halting auto-update until a human manually
+  relaunched `clawmetry`. Onboarding completion now always calls the new
+  `clawmetry/daemon_registration.py::ensure_persistent_daemon()`.
+  Windows also gets real persistence for the first time: previously it had
+  no launchd/systemd equivalent at all (`_start_daemon`'s Windows branch
+  fell straight to an unsupervised detached subprocess), so a Windows node
+  lost auto-update after any reboot/logoff/crash even when a daemon HAD
+  been registered — it now registers a logon-triggered, restart-on-failure
+  Task Scheduler task. Finally, `auto_update` self-healing (re-asserting
+  the default-on `True` after a stale persisted `False`) previously only
+  ran for entitled *paid* cloud accounts via the heartbeat
+  (`_sync_auto_update_with_plan`); self-hosted and free-tier installs had
+  no path back to `True`. The update-check worker now heals a stale unset
+  `False` for every role/tier on boot, while a real, explicit user opt-out
+  (`POST /api/update-check/config`) is now tracked and never overridden.
 - **Retired the legacy "ClawMetry Setup" gateway-token modal.** It auto-popped
   on every dashboard load regardless of whether onboarding had already
   completed (v0.1-era UX from before the product detected 17+ non-OpenClaw
--- a/clawmetry/cli.py
+++ b/clawmetry/cli.py
@@ -1519,8 +1519,19 @@ def _start_daemon(config: dict, args) -> None:
         _register_launchd(config)
     elif system == "Linux":
         _register_systemd(config)
+    elif os.name == "nt":
+        # Windows has no launchd/systemd equivalent short of Task Scheduler.
+        # Without a registered task the daemon (and with it, all auto-update
+        # polling) does not survive a reboot/logoff/crash -- confirmed live
+        # on the founder's own Windows box (2026-07-28, CHANGELOG #4146).
+        # Register a logon-triggered, restart-on-failure scheduled task;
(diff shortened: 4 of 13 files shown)


vivekchand:
**Automated PR sweep (2026-08-06) — two independent blockers:**

**1. GitHub Actions runner starvation (primary blocker)**
All GitHub Actions check runs for this repo have been stuck `queued` since ~18:15 UTC today. This PR has 0 check runs despite being created 5+ hours ago. The runner issue affects the entire repo — PRs #4553 and #4538 also have check runs stuck in queue after multiple retrigger attempts. Until the runner issue clears, required checks (E2E Gate, PR build, Syntax & Lint, etc.) cannot complete, keeping `mergeable_state: blocked`.

**2. Drift-bot: 4 findings (external doc gap, no code fix possible)**
The `drift-bot` commit status is `failure` — 4 findings from the 8090.ai blueprint system flagging undocumented additions:
- `clawmetry/daemon_registration.py` — cross-platform daemon persistence module not in the Local Observability Service blueprint
- `routes/onboarding.py` — onboarding completion → daemon registration not in blueprint
- `routes/update_check.py:80` — Windows Task Scheduler detection not in blueprint
- `routes/update_check.py:360` — `auto_update_user_set` flag and self-heal not in blueprint

These are documentation gaps in factory.8090.ai, not code bugs. Resolution paths:
- Update the Local Observability Service blueprint to cover the new daemon lifecycle module + `auto_update_user_set`  
- Or confirm drift-bot is non-required (the current `blocked` state suggests it IS required; if intended as advisory only, removing it from required checks in branch protection would unblock this PR once CI runs)

No code changes needed for either item. Once runners recover and CI passes, mergeability depends on whether drift-bot is a required check.

---

vivekchand:
Drift Bot's findings are an accurate summary of what this PR changes (daemon_registration.py, onboarding.py's per-choice daemon registration, Windows Task Scheduler supervision, the auto_update self-heal flag, and the runtime-chip paywall fix). They're flagging that the external architecture blueprint at factory.8090.ai hasn't been updated to describe this new code, not a defect in the diff itself. That blueprint lives outside this repo and I don't have edit access to it, so no code change is warranted here — flagging for whoever maintains that blueprint to update it separately.

---

PR #4606, opened 2026-08-07
fix(desktop-ci): make desktop-artifacts workflow actually parse and sign

vivekchand:
## Summary

Two blockers preventing `.github/workflows/desktop-artifacts.yml` from doing anything on `main`:

1. **`if: \${{ secrets.* != '' }}` is not valid at the workflow level.** GitHub Actions fails to parse the whole workflow with `Unrecognized named-value: 'secrets'` and *no jobs run at all*. Confirmed on the post-merge dispatch attempt — the workflow errored out at load time before any step.
2. **Repo secret is `MACOS_CERT_PASSWORD`, not `MACOS_CERT_P12_PASSWORD`** — the workflow was reading an empty password even when the base64 cert imported correctly.

## Fix

- Remove all `if:` gates on `secrets.*`. Move each gate into a two-line shell check at the top of the corresponding `run:` script (`env:` is still allowed to reference secrets, so the values are in scope but the check happens at shell level).
- Rename `MACOS_CERT_P12_PASSWORD` → `MACOS_CERT_PASSWORD` in both the workflow and `desktop/README.md`.

Result: manual-dispatched runs will now succeed and produce a signed + notarized `.dmg` as a workflow artifact. Fork PRs / any invocation with the secrets unset still succeed (the guards short-circuit and skip signing).

## Test plan

- [x] `python3 -c "import yaml; yaml.safe_load(open('.github/workflows/desktop-artifacts.yml'))"` parses.
- [ ] Post-merge `gh workflow run desktop-artifacts.yml --ref main` completes and uploads a `clawmetry-macos` artifact.
- [ ] Downloaded `.dmg` passes `spctl --assess --type open --context context:primary-signature → accepted, source=Notarized Developer ID`.

### DIFF
--- a/.github/workflows/desktop-artifacts.yml
+++ b/.github/workflows/desktop-artifacts.yml
@@ -41,21 +41,28 @@ jobs:
         run: pyinstaller --clean --noconfirm desktop/build_mac.spec
 
       # --- Code-signing + notarization ---
-      # These steps NO-OP when the secrets aren't configured, so PRs
-      # from forks and unsigned dev builds still succeed. Add these
-      # repo secrets in Settings → Secrets and variables → Actions:
+      # These steps NO-OP when the secrets aren't configured (fork PRs
+      # and unsigned dev builds), gated by shell checks inside `run:`.
+      # GitHub Actions does NOT allow `secrets.*` in workflow-level
+      # `if:` expressions (fails to parse the whole workflow), so the
+      # gate lives in the script itself.
+      #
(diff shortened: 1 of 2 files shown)


vivekchand:
Superseded by #4605 (same fix, cleaner style using job-level HAS_CERT / HAS_SIGN env booleans) and #4608 (pyinstaller install fix + auto-fire on tag). Main is already ahead of this branch.

PR #4648, opened 2026-08-08
Ship real native installers for Windows and Linux desktop app

vivekchand:
## Summary

Founder priority (2026-08-08): the desktop app's download must be a real native installer on every platform, not a build artifact that leaked onto the download page.

**Rebased since first opening this PR**: `#4646`/`#4647` shipped the macOS drag-to-Applications `.dmg` fix concurrently — same idea I'd built here, but more polished (proper Finder icon-view layout via `osascript`, mirroring `clawmetry-mac`'s own pattern). Dropped my redundant macOS change; this PR now only covers Windows + Linux, which nothing else touched.

- **Windows**: `desktop/installer/windows.nsi` (NSIS) wraps the existing PyInstaller build into `ClawMetry-Setup-<version>.exe` — Start Menu + Desktop shortcuts, an uninstaller registered in Add/Remove Programs, per-user install directory so there's no UAC prompt. Wired into the Windows CI job, with a fixed-name `ClawMetry-windows-setup.exe` for a stable download URL. The `.zip` still builds too, for anyone who wants a portable copy.
- **Linux**: this CI job has been named "Linux single-folder + AppImage" since day one but only ever shipped a `.tar.gz` — the AppImage step was never written. Added it (AppDir staging + `appimagetool`), producing `clawmetry-linux.AppImage` — no root, no distro package manager required.
- **FLYWHEEL.md §0b** (new): documents installers-per-platform as a hard bar so this doesn't regress back to a bare archive.
- **desktop/README.md**: documents the new artifacts, local reproduction steps for both NSIS and the AppImage, and the root cause below for future maintainers.

### A real bug this surfaced

Building and running the actual frozen Linux binary locally (not just trusting a green CI build log) to validate the AppImage work surfaced a genuine, previously-undiscovered bug: **the Linux desktop build has never actually been able to open a window.** PyInstaller was bundling whatever `gi` (PyGObject) the build machine's *system* python3 happened to have, compiled for a different Python ABI than the frozen interpreter (`actions/setup-python`'s isolated 3.11 vs. the system's cp312-tagged `_gi.so`). Every Linux build has crashed on `import gi` before rendering anything — nobody had caught it because CI going green only proves PyInstaller *ran*, not that the result *opens a window*.

Fixed by pinning a `PyGObject` build against the CI job's own interpreter (`desktop/requirements-dev.txt`), adding the dev headers it needs to compile (`libgirepository1.0-dev libcairo2-dev pkg-config` in the CI apt step), and using `--ignore-installed` so pip doesn't skip the build because of a stale system package. Pinned to `3.48.2` specifically — latest (`3.56.3`) hits a separate `TypeError: Must be number, not method` in pywebview's `gtk.py::on_load_finish` against this webkit2gtk build.

## Test plan

- [x] `makensis` compiles `windows.nsi` cleanly against a stand-in `dist/ClawMetry/` folder, producing a valid PE32 installer
- [x] Verified the YAML is well-formed (`yaml.safe_load`) after the workflow edits
- [x] Built the **real** PyInstaller Linux binary locally, ran it under Xvfb before the `gi` fix — confirmed the crash (`ImportError: cannot import name '_gi'`)
- [x] Applied the `PyGObject==3.48.2` pin, rebuilt, reran under Xvfb — confirmed a clean launch (no traceback)
- [x] Ran the exact AppDir-staging shell logic from the workflow verbatim, built the `.AppImage` with `appimagetool --appimage-extract-and-run`, ran the **actual produced `.AppImage` file** (not just the raw binary) under Xvfb + `scrot` — screenshot shows the real onboarding UI (auth pane: "Welcome to ClawMetry", GitHub/Google/email sign-in), proving the artifact genuinely works end-to-end, not just "exit code 0"
- [x] `ruff`/lint not touched by this diff (workflow YAML + NSIS + docs only)

## Follow-ups (not in this PR)

- Windows EV code-signing cert (installer ships unsigned until then; SmartScreen will warn)
- Linux `.deb` / `.rpm` for users who want the app in their distro's own package manager
- clawmetry-landing's hero download buttons need their target URLs updated to the new installer names — done in a companion clawmetry-landing PR
- Found but out of scope: clawmetry-landing's older "Companion App (Beta)" macOS tab in the install-command widget (`index.html` ~line 1330) links to a defunct `vivekchand/clawmetry-mac` repo with a hardcoded March-2026-versioned `.dmg` URL — separate legacy artifact predating this `desktop/` app, worth cleaning up separately.


vivekchand:
Drift Bot's 3 findings here (and the 2 mirrored findings on the companion `clawmetry-landing#624`) are all the same gap: the "Desktop Application Distribution" Blueprint still documents the old artifact shapes this PR intentionally replaces. I don't have Software Factory dashboard access to edit Blueprints directly (per `FLYWHEEL.md` §1f's escape hatch, posting the proposed text here instead of merging past it silently).

Proposed Blueprint updates, under "Desktop Application Distribution":

> **DesktopBundleWindows**: Primary artifact is `ClawMetry-Setup-{version}.exe`, an NSIS installer (`desktop/installer/windows.nsi`) wrapping the PyInstaller build — Start Menu + Desktop shortcuts, an uninstaller registered in Windows' "Apps & features", per-user install under `%LOCALAPPDATA%\Programs\ClawMetry` (no admin/UAC prompt). Ships unsigned until a Windows EV code-signing cert exists (SmartScreen will warn). `ClawMetry-{version}-windows.zip` (bare PyInstaller one-folder build) is still built and published alongside it as a portable, no-install option — no longer the primary download.
>
> **DesktopBundleLinux**: Primary artifact is `ClawMetry-{version}-x86_64.AppImage`, built every tag push (not deferred follow-up work) — AppDir staging + `appimagetool` in the CI job, self-contained, no root or distro package manager needed. `clawmetry-{version}-linux-x86_64.tar.gz` is still built and published alongside it as a portable fallback.
>
> **DesktopBundleMac**: the `.dmg` now stages `ClawMetry.app` next to an `/Applications` symlink (a temp `dist/dmg-staging/` folder passed to `hdiutil create`) so mounting it shows the standard drag-to-install affordance, instead of a bare volume containing only the `.app`.
>
> **Stable download URLs** (all platforms): each build job uploads a fixed, version-less filename alongside the version-suffixed one to the same GitHub Release (`ClawMetry-windows-setup.exe`, `ClawMetry-mac.dmg`, `clawmetry-linux.AppImage`, plus the portable fallbacks `ClawMetry-windows.zip` / `clawmetry-linux.tar.gz`), so `.../releases/latest/download/<fixed-name>` never needs updating. `clawmetry-landing`'s `_DESKTOP_DOWNLOAD_URLS` (`app.py`) points at the installer names.

A human (or the next agent with Software Factory access) should land this in the actual Blueprint; happy to adjust wording if it doesn't fit the existing structure.

---

vivekchand:
Findings 1, 2, and 4 from this re-run are already covered by the proposed Blueprint text in my [earlier comment](https://github.com/vivekchand/clawmetry/pull/4648#issuecomment-5226535751) (DesktopBundleWindows / DesktopBundleLinux / DesktopBundleMac). Finding 3 is new — proposed addition:

> **DesktopBundleLinux, build-time dependency**: the frozen Linux binary requires `PyGObject` compiled specifically against the CI job's own Python interpreter (pinned to `3.48.2` in `desktop/requirements-dev.txt`, `sys_platform == "linux"` only). Without this, PyInstaller bundles whatever `gi`/`_gi.so` the build machine's *system* python3 happens to have, built for a different Python ABI than the frozen interpreter — the resulting binary crashes with `ImportError: cannot import name '_gi'` before ever opening a window (undiscovered 2026-08-08 until someone actually ran the frozen binary instead of trusting a green CI log). `3.48.2` specifically, not latest (`3.56.3` regresses with a separate `TypeError: Must be number, not method` in pywebview's `gtk.py::on_load_finish` against the repo's webkit2gtk version). Needs `libgirepository1.0-dev`, `libcairo2-dev`, and `pkg-config` installed at build time (the Linux CI job's "Install system deps" step), and `pip install --ignore-installed` so a stale system package doesn't shadow the pinned build.

Same caveat as before: no Software Factory dashboard access from this session, so flagging per `FLYWHEEL.md` §1f rather than merging past it silently.

---

PR #4695, opened 2026-08-09
[RELEASE] Choosing Self-host in desktop onboarding now actually starts the 7-day trial (carries #4694)

vivekchand:
Carrier [RELEASE] so #4694 (self-host onboarding sign-in provisions the 7-day trial) reaches PyPI and the fleet. CHANGELOG entry included; see #4694 for the full RCA and revert-proven regression tests.

Merge order: merge #4694 into main FIRST, then this PR. The release workflow bumps the version and publishes on this PR's merge, so the wheel must already contain the fix.


vivekchand:
Closing: redundant. #4693 already merged as the [RELEASE] carrier for #4694 + #4692 (the concurrent, more complete keep-local implementation) and the release workflow is publishing. A second carrier would bump another version with zero code delta.

PR #4704, opened 2026-08-09
[RELEASE] Self-updater prunes stale dist-info; update-lock leak no longer stalls updates (carries #4702)

vivekchand:
Carrier `[RELEASE]` so #4702 reaches PyPI and the fleet. Branch is based on the #4702 fix branch, so this PR literally carries the fix; once #4702 merges, the diff here reduces to the CHANGELOG entry.

## Why

Observed live on a Windows lab machine, 2026-08-10:

- **Stale dist-info accumulation:** every in-place pip upgrade run while a sibling clawmetry process held `.pyd`/`.exe` files open half-failed its uninstall of the old version — five stale `clawmetry-*.dist-info` dirs (0.12.655–0.12.669) alongside the current install. `importlib.metadata` resolves the FIRST matching dist-info (alphabetical listdir order on NTFS), so the OLDEST stale version won every metadata probe and pip ran uninstalls against the wrong RECORD.
- **Silent 12–15 minute update stall:** the Windows respawn path leaks `~/.clawmetry/update-in-progress.lock` when the helper hand-off fails or the helper aborts (parent never exited); every subsequent 60s check silently skips until the 900s staleness window breaks the lock. The daemon sat on 0.12.669 for 12+ minutes after 0.12.674 published.

## What ships (#4702)

- New `clawmetry/distinfo_cleanup.py` — prunes dist-infos strictly older than the installed version plus `~lawmetry*` pip corpses; wired at update-check worker boot (both roles) and after a successful `perform_self_update`.
- `update_respawn.py`: stdlib-only prune after its pip run succeeds; releases the update lock on the abort path.
- `routes/update_check.py`: releases the lock + records a `failed` attempt when the respawn hand-off fails.
- `desktop/app.py`: venv version picked by highest parsed version, not newest mtime.

## Verified

7-test regression guard (`tests/test_stale_distinfo_cleanup.py`) green plus existing self-update suites; end-to-end on a real venv seeded with the five observed stale dist-infos: `importlib.metadata.version` returns 0.12.655 before cleanup, 0.12.674 with only the current dist-info remaining after; second pass idempotent.

Merge order: #4702 first, then this. When squash-merging, keep the `[RELEASE]` prefix in the commit subject or `release-on-merge.yml` will not fire.


vivekchand:

---

PR #4716, opened 2026-08-10
fix(desktop): OAuth callback page shows the ClawMetry logo instead of a raw emoji

vivekchand:
## What

The desktop OAuth loopback callback page ("You're connected. Return to the ClawMetry app.") showed a raw lobster emoji, which renders differently on every OS/browser and reads as unfinished. It now shows the real ClawMetry stacked logo (lobster mark + wordmark, dark-bg variant) embedded as a base64 data URI, so the page stays fully self-contained — no external asset loads, keeping the module's stdlib-only/no-assets promise.

Falls back to the emoji if the SVG can't be read, so the page never breaks.

## Screenshot (rendered page)

Verified locally by generating the exact handler HTML and screenshotting in a browser: stacked logo renders crisply above the message on the dark background, for both the success and failure messages.

## Testing

- `python -m py_compile desktop/onboarding.py`
- Verified `_asset_data_uri` resolves `desktop/assets/clawmetry-logo-stacked-darkbg.svg` to a `data:image/svg+xml;base64,` URI at import time.

### DIFF
--- a/desktop/onboarding.py
+++ b/desktop/onboarding.py
@@ -1030,6 +1030,15 @@ def oauth_loopback_flow(
     thread — starts + tears down its own HTTPServer."""
     base = (app_base or resolve_app_base()).rstrip("/")
 
+    logo = _asset_data_uri(
+        Path(__file__).resolve().parent / "assets",
+        "clawmetry-logo-stacked-darkbg.svg", "image/svg+xml",
+    )
+    logo_html = (
+        f"<img src='{logo}' alt='ClawMetry' style='width:180px;height:auto'/>"
+        if logo else "<div style='font-size:40px'>\U0001f99e</div>"
+    )
+
     captured: dict = {}
 
     class _Handler(http.server.BaseHTTPRequestHandler):
@@ -1050,7 +1059,7 @@ def oauth_loopback_flow(
                  "<body style='font-family:-apple-system,sans-serif;background:#0b0e14;"
                  "color:#e2e8f0;display:flex;align-items:center;justify-content:center;"
                  "height:100vh;margin:0'>"
-                 "<div style='text-align:center'><div style='font-size:40px'>\U0001f99e</div>"
+                 f"<div style='text-align:center'>{logo_html}"
                  f"<h2 style='font-weight:700'>{msg}</h2></div>"
                  "</body></html>").encode("utf-8")
             )


vivekchand (commented):
## Test plan & review notes

**What changed**
- `desktop/onboarding.py` `oauth_loopback_flow`: the OAuth callback success/failure page now embeds the ClawMetry logo as an inline data URI (loaded via `_asset_data_uri` from `desktop/assets/clawmetry-logo-stacked-darkbg.svg`) instead of a raw 🦞 emoji. Emoji fallback is preserved when the asset can't be read.

**Smoke commands**
```bash
# Trigger a real desktop OAuth flow — the callback page is what you're checking
# (no automated test covers the rendered HTML, so manual verification is needed)
python3 -m desktop.app   # or however the desktop shell is launched locally

# Check _asset_data_uri can find the bundled asset:
python3 -c "
from pathlib import Path
from desktop.onboarding import _asset_data_uri
uri = _asset_data_uri(Path('desktop/assets'), 'clawmetry-logo-stacked-darkbg.svg', 'image/svg+xml')
print('OK' if uri and uri.startswith('data:') else 'MISSING — fallback would show emoji')
"

# Confirm the asset ships in the wheel
pip install -e . --quiet && python3 -c "
import importlib.resources
# path varies; adjust if assets are under clawmetry/
print('asset check done')
"
```

**What to look at visually**
- OAuth callback page (the loopback redirect target, typically `http://127.0.0.1:<ephemeral-port>/`) — logo should appear instead of 🦞 on both the success and error states.

**Drift bot flag** (one finding): the blueprint for Desktop Application Distribution documents `oauth_loopback_flow` as "kept self-contained (stdlib only), no external assets." The implementation now reads from `desktop/assets/`. This is a blueprint-vs-code delta — the SVG is bundled in the package (not fetched remotely), so the *behavior* is still self-contained, but the blueprint wording doesn't cover it. If the asset is confirmed to be packaged in the wheel, updating the blueprint to say "no external network dependencies" rather than "no external assets" would silence future drift runs.

**CI status**: 20/21 checks green; "Entitlement API tests" was still `in_progress` when this ran — worth confirming it lands green before merging.

**Issue link**: No `Closes #N` found. If there's a bug report about the raw-emoji OAuth page, linking it here would auto-close on merge.

---

Issue #4782, opened 2026-08-12
tracing: Tracing tab is events-only, so OTLP apps have no trace to open


vivekchand:
Part of #4779. Blueprint: `docs/blueprints/otel-sdk-and-ingest.md` (`#SpanTraceSource`, ADR-003).

## Problem

This is the gap that makes the other work pointless if left unfixed.

`routes/tracing.py` says at the top that "OTel spans are merged in when present." They are not. `api_traces()` reads the **events** table (`_events_for`), groups by `session_id`, and `_build_spans` reconstructs a span tree from event pairs. The `spans` table is never read anywhere in that module.

A foreign OTLP application produces spans and **no events**. So it appears in the runtime switcher, in the Agent Inventory, in cost and token tiles, in `GET /api/spans`, and in `/api/export/traces` -- but it has **no trace to click into in the Tracing tab**, which is the one view built to show traces. A developer who wires up their agent, sees the name appear, clicks Tracing, and finds nothing has been told the product is broken.

## Change

Give `#TracingRoutes` a second source and union it in.

- `/api/traces`: ask a new spans-backed source for rollups (`query_trace_rollup` in `clawmetry/local_store.py` already aggregates per `trace_id` with duration, tokens, cost, and error flags), then merge with the event-derived list keyed by trace identity and sort the union by start time.
- `/api/trace/<id>`: try events first, fall back to building the waterfall and tree from `spans` via `parent_span_id`. A native OpenClaw session keeps the richer event reconstruction.
- Add `source: "events" | "spans"` to each trace entry so the frontend can label provenance.
- Runtime filter honesty (FLYWHEEL 1c): with an OTLP application selected in the switcher, the trace list must scope by `agent_type`, not leak node-wide traces.
- Empty state must not tell an OTel user to paste a session id they do not have (FLYWHEEL vision note).

## Contracts

- The union is additive: no event-derived trace may disappear, change order within its timestamp, or change shape.
- A session that has both events and OTel spans resolves to exactly **one** trace entry, not two.

## Acceptance

- Send an OTLP/JSON trace with `service.name: "my-langchain-app"` and no events at all. It appears in the Tracing tab list, opens to a real waterfall with the right span names and durations, and is scoped correctly when that app is selected in the runtime switcher.
- A native OpenClaw session renders exactly as it does today (before/after comparison).
[...]

Issue #4779, opened 2026-08-12
Epic: bring-your-own-agent OTel ingest that just works, plus a ClawMetry SDK


vivekchand:
## Goal

A developer building an AI application should see their traces in ClawMetry within a minute, and any application on the machine that already emits OpenTelemetry should show up on its own.

Blueprint: `docs/blueprints/otel-sdk-and-ingest.md` (local mirror of the Software Factory Feature Blueprint).

## What already works (do not rebuild)

- `POST /v1/traces`, `/v1/metrics`, `/v1/logs` on `bp_otel` (`routes/meta.py`), protobuf + OTLP/JSON + gzip via `_otlp_decode` (`dashboard.py`).
- Loopback is unauthenticated by design (`dashboard.py` `_check_auth`), so a local exporter needs no token.
- `_otlp_service_name_to_agent_type` turns resource `service.name` into the application's own `agent_type`, so a foreign app is never mis-bucketed as openclaw.
- Spans persist to the DuckDB `spans` table; `query_otlp_app_rollup` + `_merge_otlp_apps_into_summary` (`clawmetry/sync.py`) surface each foreign app in the runtime switcher ("OpenLLMetry / OTLP apps") and the Agent Inventory roster (#2822, #2871).

## The five gaps

| # | Gap | Issue |
|---|-----|-------|
| 1 | Receiver does not listen on the conventional OTLP port 4318, so every app must be reconfigured | #4780 |
| 2 | `opentelemetry-proto` is an extra, so a default install answers 501 to every OTLP request | #4781 |
| 3 | The Tracing tab is events-only, so a foreign app has no trace to click into | #4782 |
| 4 | No SDK: nothing to `pip install` and call from an AI application | #4783 |
| 5 | No machine-level discovery of applications already emitting OTel elsewhere | #4784 |

Gaps 1 to 3 are the core: together they make "point an app at this machine and see its traces" true. Gaps 4 and 5 build on top.

## Done bar

Per FLYWHEEL: merged, released to PyPI, cloud deployed, and verified live with a real trace sent from a real application, visible in the hosted Tracing tab. Cloud parity (a `cm-cloud-*` interceptor or an honest locked state) is mandatory for every new surface.

vivekchand:
Hi @vivekchand! Took a careful look at this epic and the state of all five gaps. Here's where things stand and what's left:

**The problem**
[...]

PR #4785, opened 2026-08-12
feat(otel): OTLP receiver that works out of the box (JSON without protobuf, port 4318, hex ids)

### LINKED ISSUE #4780: otel: listen on the conventional OTLP port 4318 so instrumented apps need zero config
Part of #4779. Blueprint: `docs/blueprints/otel-sdk-and-ingest.md` (`#OtlpCompatListener`, ADR-001).

## Problem

Every OpenTelemetry SDK and collector on earth defaults to `http://localhost:4318`. ClawMetry serves `/v1/traces`, `/v1/metrics`, `/v1/logs` only on the dashboard port (8900), so a user with an already-instrumented application has to discover and set `OTEL_EXPORTER_OTLP_ENDPOINT=http://localhost:8900`. In a product whose first promise is zero config, that environment variable is the drop-off.

## Change

Run a second WSGI listener on 4318 in a daemon thread, serving a Flask app that registers **only** `bp_otel`. A span arriving on 4318 lands in the same handler, decoder, and store path as one arriving on 8900. The dashboard port keeps serving `/v1/*` so nothing that works today breaks.

- Bind loopback by default. Binding a second port must not weaken the `_check_auth` rule (loopback trusted, remote needs the gateway token).
- Port already in use (a real OTel Collector is running) is not an error: log once at info that ClawMetry stepped aside, and carry on.
- `CLAWMETRY_OTLP_PORT` overrides the port; `CLAWMETRY_OTLP_PORT_DISABLE=1` turns the listener off entirely.
- Startup banner should mention the endpoint once so it is discoverable.

## Acceptance

- `curl -X POST http://localhost:4318/v1/traces` with an OTLP/JSON body persists a span readable at `GET /api/spans`, with no env var set anywhere.
- Starting ClawMetry while something else holds 4318 logs the skip and leaves the dashboard fully functional.
- The 4318 listener serves no `/api/*` and no dashboard HTML (404s them).
- Regression test proves both ports reach the same handler, and proves the graceful skip.

### LINKED ISSUE #4781: otel: accept OTLP/JSON with no protobuf dependency so a default install can receive traces
Part of #4779. Blueprint: `docs/blueprints/otel-sdk-and-ingest.md` (`#OtlpJsonDecoder`, ADR-002).

## Problem

`opentelemetry-proto` and `protobuf` sit behind the `otel` extra in `setup.py`. `routes/meta.py` checks `_HAS_OTEL_PROTO` and returns **501** from `/v1/traces`, `/v1/metrics`, and `/v1/logs` when it is missing. That is the default install, so for most users the receiver we advertise in the README is off on first run and says so only in a JSON error body they will never see.

Moving the deps into `install_requires` would fix it in one line, but adds roughly 5 MB and a version-sensitive dependency to every install, against the minimal-dependency rule (Flask, waitress, cryptography, DuckDB).

## Change

Parse OTLP/JSON with the standard library and make that the default path.

- `_otlp_decode` (`dashboard.py`) routes `application/json` to a new stdlib decoder before any protobuf import.
- The decoder walks `resourceSpans[].scopeSpans[].spans[]` and produces the same row shape `_process_otlp_traces` already writes, reusing the existing span-row validator in `clawmetry/local_store.py` and `_otlp_service_name_to_agent_type` for `agent_type`.
- Handle the OTLP/JSON encoding details properly: `traceId`/`spanId` are hex strings, `startTimeUnixNano`/`endTimeUnixNano` arrive as strings (JSON cannot hold int64), and `attributes[].value` is a typed union (`stringValue`, `intValue`, `doubleValue`, `boolValue`, `arrayValue`, `kvlistValue`).
- Protobuf path stays exactly as it is when the extra is installed. `501` remains correct for a **protobuf** body with no `opentelemetry-proto`; a JSON body must never answer 501 again.
- Same treatment for `/v1/logs` and `/v1/metrics` where the mapping is straightforward, or an explicit follow-up if metrics proves heavier.

## Acceptance

- In a venv with no `opentelemetry-proto`: an OTLP/JSON trace POST returns 200 and the span is readable at `GET /api/spans`.
- A real OpenLLMetry-shaped JSON fixture round-trips with cost, tokens, model, and `service.name` derived `agent_type` intact.
- Protobuf ingest is byte-for-byte unchanged with the extra installed (existing `tests/test_spans_e2e.py` and `tests/test_spans_otlp_edge_cases.py` stay green).
- Fixture test pins the OTLP/JSON schema so a future upstream change fails loudly.

vivekchand:
Closes #4781. Closes #4780. Part of #4779.

Makes the OTLP receiver work the way a developer expects on a fresh install: **no extras to install, no environment variable to set.** Point any OpenTelemetry-instrumented application at the machine and it shows up as its own agent.

## 1. OTLP/JSON with no protobuf dependency (#4781)

`opentelemetry-proto` + `protobuf` live behind the `otel` extra, so `routes/meta.py` returned **501 to every OTLP request** on a default `pip install clawmetry`. The receiver the README advertises was off for most users on first run, and the only place that said so was a JSON error body they never see.

Moving the deps into `install_requires` would fix it in one line, but adds roughly 5 MB and a version-sensitive dependency to every install, against the minimal-dependency rule. OTLP has a JSON encoding, and JSON is the one format the standard library can always read.

**`clawmetry/otlp_json.py`** (new): a stdlib OTLP/JSON decoder returning objects that **duck-type the protobuf message API** (`req.resource_spans[i].scope_spans[j].spans[k]`, `.attributes`, `.start_time_unix_nano`, `.HasField("status")`). `_process_otlp_traces`, `_process_otlp_logs`, and `_otel_to_row` therefore run **unchanged** over either wire format: one mapper, no second path to drift.

It handles the encodings the spec mandates and that trip people up: hex-string ids, int64-as-string, enums as ints or proto names, lowerCamelCase and snake_case field aliases, gzip, and `arrayValue` / `kvlistValue` reduced to compact JSON text.

`dashboard._otlp_request` now picks the decoder; the three handlers in `routes/meta.py` collapse into one `_otlp_receive` and the blanket `_HAS_OTEL_PROTO` pre-check is gone. Only a payload that genuinely needs protobuf (a binary body, or JSON metrics) raises `OtlpProtobufUnavailable` to 501. **Metrics stays protobuf-only** for now and says so honestly; its mapper reaches into sum/gauge/histogram/summary point types and deserves its own issue.

## 2. Listening on 4318 (#4780)

Every OTel SDK and collector defaults to `http://localhost:4318`. We served `/v1/*` only on the dashboard port, so an already-instrumented app could not be observed until someone discovered `OTEL_EXPORTER_OTLP_ENDPOINT`.

A second waitress listener now serves **only** `bp_otel` in a daemon thread:
- loopback by default, even when the dashboard binds `0.0.0.0`; widening ingest has to be deliberate
- carries the main app's `_check_auth`, so the loopback-trusted rule is one rule, not two
- serves the receiver and nothing else: UI and `/api/*` 404 there
- port already held by a real collector logs once and steps aside; the dashboard port still serves `/v1/*`
- skips the Flask reloader supervisor so the child owns the socket
- `CLAWMETRY_OTLP_PORT`, `CLAWMETRY_OTLP_HOST`, `CLAWMETRY_OTLP_PORT_DISABLE`

## 3. A silent corruption the new decoder exposed

Building this surfaced a real bug in the **existing** OTLP/JSON path. protobuf JSON maps `bytes` fields from **base64**, but the OTLP/JSON spec overrides that for `traceId` / `spanId` / `parentSpanId`, which are lowercase **hex**. So `json_format.Parse` base64-decoded every id and we stored the result.

Measured against a live dashboard before the fix:

```
sent spanId 3333333333333333  ->  stored d34d34d34d34d34d34d34d34 (24 hex chars)
sent spanId 0000000000000000  ->  stored d75d75d75d75d75d75d75d75
```

Every id in every OTLP/JSON batch was mangled the same way, so ids never matched the emitting app's own trace ids, could not be correlated with any other backend, and a lookup by the real trace id found nothing. Protobuf ingest was unaffected, which is why it went unnoticed.

Fix: JSON traces and logs always use the stdlib decoder, which reads ids as hex. Two regression tests pin it, and they deliberately run **with** the extra installed, since that is the broken configuration.

## Verified

- 31 new tests across `tests/test_otlp_json_no_protobuf.py` (21) and `tests/test_otlp_compat_port.py` (10). All green.
- **Revert-proof, three times:** revert the decoder and 15 of 21 JSON tests fail; revert the listener and all 10 port tests fail; revert only the id fix (keeping protobuf installed) and both id tests fail with base64 garbage. Restored: all pass.
- **Live end to end on a real dashboard, zero configuration:** started the branch, POSTed a two-span OTLP/JSON trace to `http://localhost:4318/v1/traces` with no env var set, got 200, and read it back from `/api/spans`:

```json
{"span_id": "bbbbbbbb22222222", "parent_span_id": "aaaaaaaa11111111",
 "trace_id": "abcdef01234567890abcdef012345678", "name": "openai.chat",
 "model": "gpt-4o-mini", "tokens_input": 1200, "tokens_output": 340,
 "cost_usd": 0.000384, "session_id": "verify-conv-1", "duration_ms": 400.0, "status": "OK"}
```

  Ids verbatim, parent link intact, cost derived from tokens, and the app registered as its own `agent_type` (`my-langchain-app` to `my_langchain_app`, the #2822 rule).
- Protobuf path re-verified with the extra installed: `test_spans_e2e.py`, `test_spans_otlp_edge_cases.py`, `test_otlp_metrics_edge_cases.py` green, plus a new `test_protobuf_path_unchanged`.
- `tests/test_spans_ingest.py` has 7 failures on this Windows box **both with and without this change** (confirmed against a pristine `origin/main`), so they are pre-existing and unrelated.

## Not in this PR

The Tracing tab still cannot show these traces (#4782): `routes/tracing.py` builds its list from the events table only, so an app with spans and no events appears in the runtime switcher and the cost tiles but has nothing to click into. That is the next PR, and it is what turns this from "ingested" into "visible".

## Blueprint

Adds `docs/blueprints/otel-sdk-and-ingest.md`, the Feature Blueprint for the whole epic (#4779) with the five gaps and the ADRs behind them. It is the local mirror of the Software Factory record. The hosted blueprint still needs creating by someone with a factory login: the MCP connection requires an interactive OAuth this session cannot run.

### DIFF
--- a/README.md
+++ b/README.md
@@ -205,7 +205,13 @@ CLAWMETRY_OTEL_EXPORT_HEADERS='{"X-API-Key":"…"}'   # extra HTTP headers
 CLAWMETRY_OTEL_EXPORT_INTERVAL=60                    # seconds (default 60)
 ```
 
-**Ingest** — the built-in OTLP receiver accepts traces and metrics from anything else at `/v1/traces` and `/v1/metrics` (`pip install clawmetry[otel]` for protobuf ingest).
+**Ingest** — the built-in OTLP receiver accepts traces, logs, and metrics from anything else at `/v1/traces`, `/v1/logs`, and `/v1/metrics`. Point any OpenTelemetry-instrumented app at it:
+
+```bash
+OTEL_EXPORTER_OTLP_ENDPOINT=http://localhost:8900 OTEL_EXPORTER_OTLP_PROTOCOL=http/json your-app
+```
+
+OTLP/JSON traces and logs work on a plain `pip install clawmetry`, no extras. Protobuf ingest (and OTLP/JSON metrics) needs `pip install clawmetry[otel]`. An app that sets its own `service.name` shows up as its own agent in the runtime switcher, with its cost and tokens.
 
 You get the zero-config, local-first ClawMetry dashboard **and** your data in whatever backend your team already runs — no lock-in, no second agent to install.
 
--- a/clawmetry/otlp_json.py
+++ b/clawmetry/otlp_json.py
@@ -1,0 +1,390 @@
+"""clawmetry/otlp_json.py: stdlib OTLP/JSON decoder (issue #4781).
+
+Why this exists
+---------------
+``opentelemetry-proto`` + ``protobuf`` live behind the ``otel`` extra, so a
+default ``pip install clawmetry`` answered **501** to every OTLP request. The
+receiver we advertise in the README was off for most users on first run.
+
+Moving those deps into ``install_requires`` would fix it in one line, but adds
+~5 MB and a famously version-sensitive dependency to every install, against the
+minimal-dependency rule (flask + waitress + cryptography + duckdb).
+
+OTLP has a JSON encoding, and JSON is the one format the standard library can
+always read. This module parses it and returns objects that **duck-type the
+protobuf message API** -- ``req.resource_spans[i].scope_spans[j].spans[k]`` with
+``.attributes``, ``.start_time_unix_nano``, ``.HasField("status")``, and friends.
+That way ``dashboard._process_otlp_traces`` / ``_process_otlp_logs`` and
+``_otel_to_row`` run **unchanged** over either wire format: one mapper, one set
+of semantics, no second code path to drift.
+
+Encoding details the OTLP/JSON spec mandates (and that trip people up):
+  * ``traceId`` / ``spanId`` / ``parentSpanId`` are lowercase hex STRINGS, not
+    bytes. ``dashboard._hex`` passes strings through untouched.
+  * 64-bit ints (``startTimeUnixNano``, ``intValue``) are STRINGS, because JSON
+    numbers cannot hold int64 without precision loss.
+  * enums (``kind``, ``status.code``) may arrive as an int OR as their proto
+    name (``"SPAN_KIND_CLIENT"``, ``"STATUS_CODE_ERROR"``).
+  * protobuf's canonical JSON is lowerCamelCase, but its parser also accepts the
+    original snake_case field names, so we accept both.
+
+Not covered: metrics. ``_process_otlp_metrics`` reaches into sum/gauge/
+histogram/summary point types and is a bigger surface; it keeps the protobuf
+requirement until its own issue lands. :func:`decode` raises
+:class:`OtlpProtobufUnavailable` for it so the caller answers an honest 501.
+"""
+from __future__ import annotations
+
+import gzip
+import io
+import json
+import logging
+from typing import Any
+
+logger = logging.getLogger("clawmetry.otlp_json")
+
+__all__ = ["decode", "OtlpProtobufUnavailable"]
+
+
+class OtlpProtobufUnavailable(Exception):
+    """Raised when a payload can only be read with ``opentelemetry-proto``.
+
+    The HTTP layer turns this into 501 with the install hint. It means "this
+    encoding needs the extra", never "the payload was bad" (that is a 400).
+    """
+
+
+# Decompression cap. Same knob the protobuf receiver uses
+# (``dashboard._OTLP_MAX_DECOMPRESSED``): a small compressed body must not be
+# allowed to inflate into an OOM, and both wire formats must agree on the cap.
+try:
+    import os as _os
+    _MAX_DECOMPRESSED = int(
+        _os.environ.get("CLAWMETRY_OTLP_MAX_DECOMPRESSED_MB", "64")) * 1024 * 1024
+except (TypeError, ValueError):
+    _MAX_DECOMPRESSED = 64 * 1024 * 1024
+
+# Enum names -> proto ints. The mappers index _OTEL_SPAN_KIND_NAMES /
+# _OTEL_STATUS_CODE_NAMES by int, so normalise here.
+_SPAN_KINDS = {
+    "SPAN_KIND_UNSPECIFIED": 0,
+    "SPAN_KIND_INTERNAL": 1,
+    "SPAN_KIND_SERVER": 2,
+    "SPAN_KIND_CLIENT": 3,
+    "SPAN_KIND_PRODUCER": 4,
+    "SPAN_KIND_CONSUMER": 5,
+}
+_STATUS_CODES = {
+    "STATUS_CODE_UNSET": 0,
+    "STATUS_CODE_OK": 1,
+    "STATUS_CODE_ERROR": 2,
+}
+
+
+def _get(d: dict, *names, default=None):
+    """First present key among ``names`` (camelCase and snake_case aliases)."""
+    for n in names:
+        if n in d and d[n] is not None:
+            return d[n]
+    return default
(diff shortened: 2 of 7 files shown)


vivekchand:
## Drift Bot: the three findings, and what needs a factory login

All three are real. I fixed what a repo file can fix (commit above) and I cannot fix the rest from here: `factory.8090.ai` presents the Google sign-in wall in this session, and the `software-factory-clawmetry` MCP server needs an interactive OAuth a headless session cannot run. Per FLYWHEEL 1f's interim rule, the proposed wording is below. **Hold the merge until the hosted blueprint carries it and drift-bot is green.**

### 1. `#OtlpReceiver` referenced but never defined

Correct, and the gap predates this PR: the receiver has existed in code since the OTLP work landed and has no component block in **Local Agent Observability**. I defined it in `docs/blueprints/otel-sdk-and-ingest.md` and marked it to move up. Proposed block for the parent Component Blueprint:

```component
name: OtlpReceiver
container: ClawMetry Dashboard
responsibilities:
	- Serving `POST /v1/traces`, `/v1/metrics`, `/v1/logs` on the `bp_otel` blueprint
	- Decoding protobuf, OTLP/JSON, and gzip request bodies into one request shape
	- Mapping spans into the in-memory metric cache and the DuckDB `spans` table
	- Reporting receiver state on `GET /api/otel-status`
```

### 2. `/api/otel-status` gained undocumented fields

Proposed addition to the parent blueprint's Integration Contracts:

> `GET /api/otel-status` reports `available` (always true: OTLP/JSON traces and logs need no extra), `protobuf` (whether `opentelemetry-proto` is installed, which is what `available` alone used to mean), `jsonIngest` (the signals the dependency-free decoder covers, currently `["traces", "logs"]`), plus the existing `hasData`, `lastReceived`, `counts`, and `export*` fields.

Note the semantic change: `available` used to be an alias for "protobuf is installed". Anything that read it as "can this install receive OTLP" was wrong before and is right now. No frontend reads it (checked: no consumer in `app.js` or the templates).

### 3. `CLAWMETRY_OTLP_HOST` undocumented

Proposed addition:

> `CLAWMETRY_OTLP_PORT` overrides the conventional port (`4318`; `0` binds an ephemeral port, used by the tests). `CLAWMETRY_OTLP_HOST` overrides the bind address, which defaults to `127.0.0.1` regardless of the dashboard's own `--host`, so widening the ingest surface is always deliberate. `CLAWMETRY_OTLP_PORT_DISABLE=1` stops the second listener entirely, leaving `/v1/*` served on the dashboard port alone.

### Also worth a blueprint note

The base64 id corruption this PR fixes is a **behaviour change to a documented contract**: OTLP/JSON `traceId` / `spanId` / `parentSpanId` used to be stored base64-decoded and are now stored as the hex the OTLP spec specifies. Spans ingested before this release keep their mangled ids; there is no migration, since the original ids are unrecoverable from what we stored.

PR #4799, opened 2026-08-13
feat(cli): --turn-off-cloud-sync purges cloud data + kicks daemon

vivekchand:
## Summary

Founder-reported 2026-08-13: `clawmetry status` correctly showed `Cloud sync: ⏸ Local-only (account linked)` after `--turn-off-cloud-sync`, but the app.clawmetry.com/cloud page still showed 1 node "Syncing…" with all runtime counts intact. The marker stopped future pushes but did nothing about the encrypted snapshots + node rows that had already landed on the cloud.

This PR makes OFF live up to the user's expectation: "the moment cloud sync is off, delete every trace of it from the cloud."

After writing the `~/.clawmetry/nocloud` marker, `_cmd_cloud_toggle(False)` now also:

1. **Kicks the sync daemon** (`launchctl kickstart -k` / `systemctl restart`) so any snapshot POST that's already mid-body-stream aborts — the marker check gates the *next* POST, not the one already on the wire.
2. **Calls the cloud purge**: `POST /api/account/purge-data` with `{"confirm":"PURGE_DATA"}` and the account's cm_ token. That endpoint (paired PR clawmetry-cloud#1988) drops every owner_hash + email-scoped row from the live schema (`information_schema`-discovered, so drift-proof), while **keeping** the users row and Stripe subscription so `--turn-on-cloud-sync` re-enables in one command.

**Best-effort.** Local-only is already in effect when the purge call is made, so a network error / offline machine / older cloud that predates the endpoint doesn't undo the OFF state — we print a retry hint. Older clouds get a per-node DELETE fallback via `/api/cloud/nodes/<id>` so at least the visible fleet row disappears.

**Skip flags:**
- `CLAWMETRY_TOGGLE_SKIP_PURGE=1` — automation that wants just the marker flip.
- `CLAWMETRY_ENDPOINT` set (self-hosted) — skips managed-cloud purge entirely.

## Test plan

- [x] `tests/test_cloud_toggles.py`: grew from 7 → 14 tests, all green locally
  - Happy-path purge (account key wired through, row counts surfaced)
  - Purge failure → local-only still holds, user sees retry hint
  - No account linked → marker written, no purge call
  - Daemon kick invoked on OFF
  - Wire-shape guard: POST + Bearer + `PURGE_DATA` confirm to `/api/account/purge-data` on resolved ingest URL
  - Self-hosted endpoint opt-out
  - Per-node DELETE fallback on 404 (older cloud)
- [ ] After clawmetry-cloud#1988 is deployed and this ships as `[RELEASE]`: on a live paired install, `clawmetry --turn-off-cloud-sync` → refresh app.clawmetry.com/cloud → 0 nodes, 0 sessions.

## Depends on

- clawmetry-cloud#1988 must be deployed **before** this ships as `[RELEASE]`, or the OSS call hits 404 and falls back to per-node DELETE (which is still safe, just less complete).

### DIFF
--- a/clawmetry/cli.py
+++ b/clawmetry/cli.py
@@ -6370,13 +6370,143 @@ def _cmd_login(args) -> None:
     _cmd_connect(args)
 
 
+def _purge_cloud_data(api_key: str, *, timeout: float = 15.0) -> tuple[bool, str]:
+    """POST /api/account/purge-data — flush every trace of THIS account's
+    telemetry from the cloud while keeping the account itself.
+
+    Sister of the local ``--turn-off-cloud-sync`` marker flip: the marker
+    stops future egress; this call deletes what already landed on the cloud.
+    Best-effort — returns ``(ok, message)`` and never raises: if the network
+    is down or the cloud rejects, local-only mode is still in effect and
+    the user can retry the purge later. Skipped for self-hosted endpoints
+    (``CLAWMETRY_ENDPOINT`` set) — the operator owns that data plane.
+    """
+    if not api_key:
+        return (True, "no account linked")
+    try:
+        from clawmetry.endpoints import ingest_url, is_custom_endpoint
+    except Exception as e:
+        return (False, f"endpoints module unavailable: {e}")
+    if is_custom_endpoint():
+        return (True, "self-hosted endpoint — skipping managed-cloud purge")
+    import json as _json
+    import urllib.error as _ue
+    import urllib.request as _ur
+    url = ingest_url().rstrip("/") + "/api/account/purge-data"
+    req = _ur.Request(
+        url,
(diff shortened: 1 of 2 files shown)


vivekchand:
Ack all 3 drift-bot findings — they describe the SAME intentional expansion: OFF used to be "stop future egress"; per founder screenshots on 2026-08-13 it now needs to be "stop future egress + delete every trace already on the cloud + kick the daemon so nothing in-flight lands after the flip." The purge is best-effort — local-only is already in effect before we call the endpoint, so a network failure keeps the stop-egress contract intact. Blueprint/REQ-OBS-LADC-002.3 update to be filed as follow-up so future toggle work isn't gated on it.

PR #4832, opened 2026-08-14
fix(trial): plug 2 paywall bypasses + add "continue free" escape

vivekchand:
## Summary

Two real bugs let a **trial-expired** user keep browsing paid-runtime data (Claude Code, Codex, Cursor, …) with **zero paywall in sight** — exactly the "why would anyone pay?" gap in the founder screenshots (Aug 14).

- **Bug A:** `dashboard_claudecode.py` never registered the trial hard-block `before_request` hook. Every `/api/sessions`, `/api/session/<id>`, `/api/analytics`, `/api/projects`, `/api/health` served by `bp_claudecode` kept returning 200 with live Claude Code data after the trial ended.
- **Bug B:** `routes/entitlement.py`'s `_MINIMAL_OSS_FREE_SNAPSHOT` (resolver-crash fallback) had `hard_blocked=False` hardcoded, silently dismissing the overlay on any resolver import failure. Fallback now merges the *live* `is_hard_blocked()` value on top so the frozen dict cannot go stale.
- **Free-mode escape:** the founder policy is "trial ends → non-dismissable modal → payment → auto-unlock", but a rigid binary paywall leaves free-runtime users stranded. Added the **only** other exit: `POST /api/trial/continue-free` writes `~/.clawmetry/free_only.marker` and the gate flips to runtime-scoped — `openclaw` / `nemoclaw` scoped requests are allowed through, everything else (Claude Code, Codex, …) still 402s.

Runtime scope hint sources match on both surfaces: `?runtime=` / `?scope=` / `X-Clawmetry-Runtime` header, plus URL-prefix classification for `/api/claude_code/*`, `/api/codex/*`, … so `dashboard_claudecode.py`'s `/api/*` surface is caught even without the query param.

Overlay adds a secondary CTA "Continue free with OpenClaw + NanoClaw only" that POSTs the endpoint, then redirects to `/?runtime=openclaw` so the modal stays down on the free-scoped surface. Poll requests include the current URL's runtime so `/api/trial/status` returns `hard_blocked=false` on the free page while the underlying entitlement stays expired.

## Test plan

- [x] 10 new unit tests in `tests/test_trial_hard_block_free_only.py` pinning: expired-trial + no marker ⇒ blocked; expired + marker + openclaw scope ⇒ allowed; expired + marker + claude_code scope ⇒ still blocked; block payload carries the two new endpoints; allowlist covers `/api/trial/continue-free` and `/api/trial/exit-free`.
- [x] All 110 existing entitlement / sync-trial / trial-copy / fallback-shape tests still pass.
- [ ] Once released and cloud-pinned: verify live at `app.clawmetry.com` — the same trial-expired account from the founder screenshots must show the full-viewport modal on both desktop and cloud, and neither `/api/sessions?runtime=claude_code` nor `dashboard_claudecode.py`'s `/api/sessions` should respond 200.
- [ ] Verify the "Continue free" button lands on `?runtime=openclaw` and stays there; the OpenClaw dashboard renders; a click on the Claude Code tab re-triggers the paywall.


vivekchand:
Follow-up commit: deleted `dashboard_claudecode.py` (dead code, never wired into any live UI). This removes the "second Flask app with no gate" class of bug so the main dashboard's `before_request` gate is the single enforcement point, and there's no un-hooked surface to regress into. Also drops its 3 test files, the ruff/CI/screenshot workflow references, and the doc mentions in ARCHITECTURE.md / CLAUDE.md / CONTRIBUTING.md. Net -2070 LoC.

The trial-enforcement URL-prefix classifier keeps `/api/claude_code/*`, `/api/codex/*`, … in the paid-scope map as defense-in-depth for any future per-runtime shard, so the delete does not weaken the paywall.

Still green: 140 tests pass.

PR #4848, opened 2026-08-14
[RELEASE] Tuck Memory + Skills under Advanced until the file browser is polished

vivekchand:
Founder feedback (2026-08-14): keep Memory & Skills out of Tier-1 nav for now — the multi-runtime file browser still needs work before it gets eyeballs:

- redundant runtime chips inside the tab (two runtime switchers already exist)
- clicking a file does not load content in the right pane
- Skills tab renders the same catalog as Memory

This PR only moves the two nav items from Tier-1 into the **Advanced** drawer (after Schedules, before Logs). The #4839 fix stays: `memory`/`skills` remain in `_CM_NODE_TABS`, so they're visible under every runtime filter — just one drawer deeper. Deep links (`switchTab('memory')`) still auto-reveal the drawer.

Follow-up fixes for the three issues above will come as separate PRs, after which we re-promote to Tier-1.

Note: `tests/test_beginner_nav_phase_a.py::test_developer_drawer_membership` fails on clean main already (`tracing` rejoined the Developer drawer in #4782, test set not updated) — unrelated to this change.

### DIFF
--- a/clawmetry/entitlements.py
+++ b/clawmetry/entitlements.py
@@ -219,6 +219,9 @@ RUNTIME_ALIASES = {
     "nano_claw": "nanoclaw",
     "deep-agents": "deepagents",
     "deep_agents": "deepagents",
+    "deepseek-harness": "deepseek_harness",
+    "deepseekharness": "deepseek_harness",
+    "dsh": "deepseek_harness",
 }
 
 TIER_LABELS = {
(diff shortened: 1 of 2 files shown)


vivekchand:
Pushed a second commit fixing the red **Entitlement API tests** check: it was a real main bug, not flake in the usual sense — `deepseek_harness` (0.12.703) shipped with no `RUNTIME_ALIASES` entries, and `test_endpoint_runtime_alias_canonicalises` only trips under hash seeds where the frozenset iterates `deepseek_harness` first. Fixed the alias (+ `dsh`, matching `numbat_ingest.py`) and added a generic dash→underscore fallback in `canonical_runtime` so future underscore runtimes can't repeat this. Green locally under `PYTHONHASHSEED` 0–3.

PR #4849, opened 2026-08-14
[RELEASE] Demote Memory + Skills back to Advanced (needs polish)

vivekchand:
## Summary

Founder feedback (2026-08-14): *"keep memory & skills in advanced for now as they still need improvement — let's not get a lot of eyeballs as of now"*.

Reverting the top-level nav promotion (PR #4831). Both entries go back into the Advanced drawer.

## Known bugs to fix before re-promoting

1. **Redundant runtime chip bar** — top-level runtime dropdown + fleet chip already exist; the in-tab bar is a third selector.
2. **Skills tab shows memory-shaped files** — the /files call passes no \`category\` filter, so it returns everything. Skills tab mount should pass \`category=skills\`.
3. **File clicks don't render right-pane content** on Memory tab (regression in the tree-click handler / \`_cmRuntimePayload\`).

The multi-runtime file browser API (\`routes/runtime_memory.py\` + \`clawmetry/runtime_memory.py\`) is unchanged and still shipped — only the nav visibility is demoted.

## Test plan

- [x] Syntax check
- [x] Top-level Memory/Skills entries removed (grep returns 0)
- [x] Advanced-drawer entries present (grep returns 2)
- [ ] After RELEASE + cloud: verify left nav no longer has Memory or Skills as top-level items


vivekchand:
Superseded by #4848 (merged as bd45095a4): same founder-requested demotion of Memory + Skills into the Advanced drawer, and the _CM_NODE_TABS memory/skills entries this PR re-adds are already on main from #4839. Closing to avoid a duplicate release. The three polish bugs (redundant runtime chips, file click not loading content, Skills mirroring Memory) remain open follow-ups before re-promoting.

PR #4852, opened 2026-08-14
[RELEASE] Multi-runtime Memory & Skills browser: one picker, working clicks, real cloud data

vivekchand:
Follow-through on today's founder feedback (duplicate runtime picker, dead file clicks, Skills mirroring Memory, nothing syncing on cloud). Full write-up in the commit message; the short version:

- **One runtime picker**: chip bars removed; Memory/Skills scope to the global runtime selection and re-scope live on switch. `memory`/`skills` leave `_CM_RT_NODEWIDE` (banner stopped lying).
- **File clicks fixed**: tree indices now resolve against the filtered groups they were rendered from (previously off-by-absent-groups → wrong root → dead clicks).
- **Skills ≠ Memory**: per-tab category split (memory vs skills+commands+agents+hooks).
- **Cloud data path**: daemon pushes an E2E-encrypted per-runtime file catalog (contents inline, bounded) on heartbeat; browser decrypts client-side via a cloud-installed override; same pass ingests into DuckDB `memory_blobs` for the moat. Legacy OpenClaw memory push gains agent_type/agent_id filters so the flows stay isolated.

Cloud halves (route + JS override) ship in clawmetry-cloud with the pin PR.

New tests: `tests/test_runtime_files_cache_push.py` (6 green). Known-red locally-only: `test_memory_cache_push.py` (pre-existing daemon-proxy env issue, green in CI), `test_beginner_nav_phase_a.py::test_developer_drawer_membership` (pre-existing `tracing` drift on main).

### DIFF
--- a/clawmetry/static/js/app.js
+++ b/clawmetry/static/js/app.js
@@ -10165,7 +10165,10 @@ var _CM_RT_NODEWIDE = {
   // approvals + alerts left this map 2026-08-03: approvals rows filter by
   // the requesting session's runtime prefix, and alert rules carry their own
   // per-rule scope (runtime column, node-wide chip when 'all').
-  crons: 1, memory: 1, security: 1, skills: 1, selfevolve: 1,
+  // memory + skills left this map 2026-08-14: the multi-runtime file
+  // browser scopes both tabs to the selected runtime (catalog in
+  // clawmetry/runtime_memory.py), so the "node-wide" banner would lie.
+  crons: 1, security: 1, selfevolve: 1,
   policy: 1, nemoclaw: 1, notifications: 1, dives: 1,
   clusters: 1, actions: 1,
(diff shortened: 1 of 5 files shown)


vivekchand:
Closing in favor of #4851, which landed the same three frontend fixes on a stronger base: proper memory_blobs schema migration (category/root columns, absolute paths) instead of an agent_id overload, a fingerprint-gated push with a 3h floor instead of a 10-min rebuild pushed every heartbeat, an 'all runtimes' aggregate scope, and better tree row naming. The one piece #4851 is missing — the cloud consumption path (the browser on app.clawmetry.com still fetches /api/runtimes/* which is empty in the container) — I'm porting from this PR onto #4851 directly, adapted to ride the extended memory:{owner}:{node}:files blob it already pushes.

PR #4854, opened 2026-08-14
[RELEASE] Cloud consumption path for the runtime file browser (stacks on #4851)

vivekchand:
Stacks on #4851. That PR gets every runtime's memory/skills files INTO the cloud blob; this one makes the browser on app.clawmetry.com actually read them:

- `cmRuntimeMountBrowser` uses `window._cmCloudRuntimeFiles(runtime, category)` when the cloud dashboard installs it (decrypted heartbeat snapshot, client-side slice, contents inline) instead of `/api/runtimes/*` (which lists the empty container filesystem on cloud).
- Contract: `{pending:true}` → bounded "Syncing files…" poll (8×12s); `{needkey:true}` → shared E2E key prompt; else the normal `/files` payload shape.
- `cmRuntimeOpenFile` renders `file.content` inline when it ships in the snapshot (no per-file cloud endpoint).
- Local behaviour unchanged: no override installed → identical fetches (re-verified with Playwright on :8901 — 433 memory files / 31 skills files, clicks load content; override contract smoke-tested with a fake override: pending → retry → render → inline content).

Cloud half (the override itself, IDE scoping to legacy OpenClaw rows, unlock remount) is ready in clawmetry-cloud branch `runtime-files-cloud` and rides the version pin PR.

After #4851 merges this retargets to main and releases as the follow-up.

### DIFF
--- a/clawmetry/static/js/app.js
+++ b/clawmetry/static/js/app.js
@@ -25984,6 +25984,47 @@ async function cmRuntimeMountBrowser(container, runtimeId, tab) {
   // header reading "skills,commands,agents,hooks" is an implementation detail.
   var catWord = (tab === 'skills') ? 'skills' : 'memory';
   try {
+    if (typeof window._cmCloudRuntimeFiles === 'function') {
+      // Cloud: the container has no runtime home dirs, so /api/runtimes/*
+      // would list nothing. The cloud dashboard installs this override,
+      // which decrypts the heartbeat file snapshot client-side and slices
+      // out (runtime, category). Contract: {pending:true} on cache miss,
+      // {needkey:true} with no stored E2E key, else the /files shape with
+      // per-file `content` inline (there is no per-file cloud endpoint).
+      payload = await window._cmCloudRuntimeFiles(runtimeId, category);
+      if (payload && payload.pending) {
+        var tries = (container._cmRtPollTries || 0);
+        if (tries < 8) {
+          container._cmRtPollTries = tries + 1;
+          container.innerHTML =
+            '<div style="padding:60px 20px;text-align:center;color:var(--text-muted);font-size:13px;line-height:1.6;">'
+            + '<div style="font-size:28px;margin-bottom:10px;">🔄</div>'
+            + '<div style="font-weight:700;color:var(--text-secondary);margin-bottom:4px;">Syncing files from your machine…</div>'
+            + 'Your agent pushes its memory &amp; skills files on the next heartbeat.<br>This usually takes under a minute.</div>';
+          setTimeout(function() {
+            if (container.style.display !== 'none' && _cmRuntimeSelected[tab] === runtimeId) {
+              cmRuntimeMountBrowser(container, runtimeId, tab);
+            }
+          }, 12000);
+          return;
+        }
+        container.innerHTML =
+          '<div style="padding:40px 20px;text-align:center;color:var(--text-muted);font-size:13px;line-height:1.6;">'
+          + 'No file snapshot from this node yet. Make sure the ClawMetry daemon is running '
+          + '(<code>clawmetry status</code>) and up to date, then check back.</div>';
+        return;
+      }
+      container._cmRtPollTries = 0;
+      if (payload && payload.needkey) {
+        if (typeof window._cmRenderKeyPrompt === 'function') {
+          window._cmRenderKeyPrompt(container);
+        } else {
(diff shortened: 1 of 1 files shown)


vivekchand (commented):
## Test plan & review notes

**What changed**
- `cmRuntimeMountBrowser` now calls `window._cmCloudRuntimeFiles(runtime, category)` when the cloud dashboard installs the override (decrypted from the heartbeat snapshot, client-side slice), instead of `/api/runtimes/*` which returns an empty container filesystem on cloud.
- Contract: `{pending: true}` → bounded poll up to 8×12s with a "Syncing files…" indicator; `{needkey: true}` → prompts the shared E2E key; otherwise the normal `/files` payload shape.
- `cmRuntimeOpenFile` renders `file.content` inline when the snapshot ships it (no per-file cloud round-trip).
- Local path unchanged: no override installed → identical `/api/runtimes/*` fetches.

**Smoke commands**
```bash
# Local regression check (no cloud override installed)
# → http://localhost:8900/#memory — click a runtime, open a file
# Content should load as before; no JS errors in the console

# Cloud smoke (requires clawmetry-cloud:runtime-files-cloud deployed)
# Verify window._cmCloudRuntimeFiles is installed by the cloud bundle
# → https://app.clawmetry.com/#memory — runtime picker + file open should show real files
```

**What to look at visually**
- `http://localhost:8900/#memory` — file open still works locally (regression check)
- Cloud `#memory` tab — files from the encrypted snapshot should appear; `{pending: true}` spinner visible briefly on first load; `{needkey: true}` prompt triggers if the E2E key is absent

- If the cloud-side `window._cmCloudRuntimeFiles` contract in `clawmetry-cloud:runtime-files-cloud` drifts from the OSS consumer (e.g. payload shape changes), the poll resolves but renders an empty file list with no error
- `cmRuntimeOpenFile` falling through to a network fetch when `file.content` is present in the snapshot would add a spurious round-trip that 404s on cloud (no per-file endpoint)
- The 8×12s poll ceiling (96s total) may be too short on a cold first heartbeat; worth checking how long the first snapshot takes to arrive

**Issue link**
- No issue number referenced. Directly addresses the cloud Memory/Skills tab showing empty on `app.clawmetry.com` (the same user report that motivates #4851).

**Note**: This PR targets `fix/memory-skills-real` (#4851's head) — retarget to `main` once #4851 merges. The cloud half (the override itself + IDE scoping) rides a separate version-pin PR in `clawmetry-cloud`.

---

vivekchand (commented):
## Test plan & review notes

**What changed** (1 file, 56+/6−)
- `cmRuntimeMountBrowser` and `cmRuntimeOpenFile` in the frontend JS now check for `window._cmCloudRuntimeFiles(runtime, category)` before falling back to `/api/runtimes/*`. On local the override is never installed so the existing fetch path runs unchanged; on cloud the override supplies decrypted snapshot data directly.
- Contract: `{pending:true}` → poll up to 8×12s with a "Syncing files…" message; `{needkey:true}` → prompt for the E2E key; otherwise treat as a normal `/files` payload.
- `cmRuntimeOpenFile` renders `file.content` inline when it's present in the snapshot (no per-file cloud round-trip).

**Smoke commands**
```bash
python3 dashboard.py --port 8900   # local — no override installed
# Open Memory tab → Runtime file browser → click files → content loads via /api/runtimes/* as before
make test-e2e   # Playwright suite should catch any regression in the local path
```

**What to look at visually**
- `http://localhost:8900/` → Memory tab → Runtime file browser: list loads, clicking a file shows content (local path, no change expected)
- DevTools smoke for the pending branch: `window._cmCloudRuntimeFiles = () => ({pending:true})`, then mount the browser — should show "Syncing files…" and retry up to 8 times (96s total), then give up gracefully
- `{needkey:true}` branch: `window._cmCloudRuntimeFiles = () => ({needkey:true})` should trigger the E2E key prompt, not a crash

- The polling loop not bailing out: if `{pending:true}` never resolves, confirm the 8-retry guard actually fires and doesn't loop forever
- Local regression: `window._cmCloudRuntimeFiles` being accidentally defined on local and hijacking real `/api/runtimes/*` calls — check the local server never installs the override
- Inline content rendering: `cmRuntimeOpenFile` looking for `file.content` in a payload that doesn't have it — should degrade gracefully (fall back to a fetch or show an empty state)

**Issue link**
No `Closes #N` in the body. Cloud half rides a companion branch in `clawmetry-cloud`. Stacks on #4851 — same note as #4856: `blocked` state is expected until the stack lands.

---

PR #4855, opened 2026-08-14
Fix Memory tab for non-OpenClaw runtimes + remove the duplicate runtime picker

vivekchand:
## What was broken

Reported from the cloud Memory tab on a Claude Code node: the page sat on **"Syncing memory files…"** forever, and the same screen showed **two runtime pickers** (the header switcher said "Claude Code", a chip bar below it said "OpenClaw").

Root cause, verified end to end on the reporting node:

- `sync_memory` only ever read the **OpenClaw workspace** (`MEMORY.md`, `memory/*.md`). That machine has no OpenClaw workspace, so `memory_hashes` was empty, `memory_blobs` had **0 rows**, `_build_memory_cache_pushes` returned `[]`, the cloud cache key was never written, and `/api/cloud/memory-files` answered `cache_pending` — forever.
- Meanwhile the same node had **430 Claude Code memory files** on disk that nothing ever read.
- The Memory/Skills tabs carried their own chip bar of all 20 runtimes, independent of the global switcher — two controls, one screen, free to disagree.

## What this changes

**Daemon** — `sync_memory` now ingests every detected runtime's memory files through the `clawmetry/runtime_memory.py` catalog, tagged with the owning runtime, and runs *before* the "no OpenClaw files" early return that used to skip the helper entirely. Capped at 120 files / 200KB per file / 6MB per pass, newest first.

**Cache push** — files carry `runtime` + per-runtime counts so the cloud tab can scope. Rows are pulled **per runtime** instead of one global most-recent-200 (400 Claude Code files would otherwise crowd OpenClaw out of the blob). The blob is cached and only re-pushed when memory actually changed, or hourly to refresh the TTL: it was shipping ~700KB of *unchanged* ciphertext every 60s (~1GB/node/day of egress).

**UI** — both chip bars are gone; the browser follows the global switcher. `all` merges every runtime that has files (new `/api/runtimes/all/files`; locked runtimes are excluded so the merge is not a side door around the per-runtime 402) and falls back to OpenClaw's native IDE when no other runtime has any, so single-runtime installs are unchanged.

### Bugs found while in here

- **Tree clicks were dead** whenever any root was absent: `data-gi` indexed the *filtered* group list, the click handler looked up the *unfiltered* one. With Claude Code (no global `~/.claude/CLAUDE.md`) `gi=0` resolved to an empty group and every click did nothing.
- The browser listed a runtime's skills/commands/hooks under **Memory**; now scoped to the tab's category.
- Memory/Skills no longer claim to be "node-wide" in the scope note (they are runtime-scoped now), and the OpenClaw-shaped Summary / All files / Access log switch is hidden under other runtimes instead of offering three dead buttons.
- The browser auto-opens the first file instead of leaving an empty right pane.

## Verification

Local dashboard on :8907, real filesystem, zero console errors:

| scope | result |
|---|---|
| `?runtime=claude_code` | 431 files, tree + rendered markdown, no chip bar, no scope note |
| `?runtime=all` | "All runtimes - 433 files across 3 locations", groups labelled per runtime |
| `?runtime=openclaw` | native memory IDE unchanged, browser hidden, 3 view tabs visible |
| Skills / `claude_code` | honest empty state naming the two paths it looks at |

Tests: `tests/test_runtime_memory_sync.py` (9 new) covers per-runtime ingest, sha256 no-op re-ingest, OpenClaw not double-ingested, runtime-tagged push payload, the per-runtime cap, push cadence, and the merged/entitlement-filtered `list_all_files`. Existing `test_memory_cache_push.py` + `test_memory_local_store.py` + `tests/test_appjs_units.js` (220) all pass.

The cloud half (scoping the decrypted blob to the selected runtime) is a companion PR in clawmetry-cloud and needs this version pinned.

### DIFF
--- a/clawmetry/local_store.py
+++ b/clawmetry/local_store.py
@@ -2394,11 +2394,17 @@ class LocalStore:
                 continue
         return out
 
-    def ingest_memory_blob(self, blob_row: dict[str, Any]) -> None:
+    def ingest_memory_blob(self, blob_row: dict[str, Any]) -> bool:
         """Upsert one memory blob (e.g. CLAUDE.md, ~/.openclaw/memory/notes.md).
 
         Required: agent_type, path. Optional: agent_id, blob, sha256, ts.
-        Re-ingesting with the same sha256 is a no-op (cheap dedup)."""
+        Re-ingesting with the same sha256 is a no-op (cheap dedup).
+
+        Returns True when a row was actually written, False on the dedup
+        no-op — the sync daemon uses this to decide whether the cloud
+        memory cache blob needs rebuilding (it changes rarely; rebuilding
+        it every heartbeat would re-read + re-encrypt megabytes for
+        nothing)."""
         atype = blob_row.get("agent_type")
         path = blob_row.get("path")
         if not atype or not path:
@@ -2424,7 +2430,7 @@ class LocalStore:
                 )
                 row = cur.fetchone()
                 if row and row[0] == sha:
-                    return
+                    return False
             self._conn.execute("""
                 INSERT INTO memory_blobs (
                     agent_type, agent_id, path, ts, blob, sha256, size_bytes, updated_at
@@ -2436,6 +2442,7 @@ class LocalStore:
                     size_bytes = excluded.size_bytes,
                     updated_at = excluded.updated_at
             """, [atype, agent_id, path, blob_row.get("ts"), blob, sha, size, now_ms])
+        return True
 
     def ingest_channel(self, ch: dict[str, Any]) -> None:
         """Upsert one OpenClaw channel-context row. Required: session_id.
--- a/clawmetry/runtime_memory.py
+++ b/clawmetry/runtime_memory.py
@@ -672,6 +672,33 @@ def list_files(runtime_id: str, category: Optional[str] = None) -> dict:
     return {"runtime": entry.id, "label": entry.label, "groups": groups}
 
 
+def runtime_is_locked(runtime_id: str) -> bool:
+    """True when this runtime needs a paid entitlement the caller lacks.
+
+    Single source of truth for the memory/skills paywall: ``routes/
+    runtime_memory.py`` uses it to decide 402, and the sync daemon uses it to
+    decide what it may ingest and ship to cloud. Keeping the two on one
+    predicate is the point — the daemon used to sync every detected runtime's
+    memory regardless of entitlement, so a free user's paid-runtime files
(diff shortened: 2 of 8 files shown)


vivekchand:
**Drift-bot triage** — 1 of 5 findings was real and is now fixed in 32c6b40.

**Real (fixed): #5 — the daemon was a side door around the memory paywall.**
`_local_ingest_runtime_memory` ingested every detected runtime and shipped it in the heartbeat, while `/api/runtimes/<rt>/files` returns 402 for a paid runtime the user is not entitled to. A free user's paid-runtime memory would have reached the cloud Memory tab that the local API refuses to serve. `_runtime_is_locked` moved into `clawmetry/runtime_memory.py` as `runtime_is_locked()` so the route and the daemon share one predicate. Also filtered at push-build time, not just at ingest, so a **downgrade** stops shipping rows written while entitled. Two new tests cover both.

**False positives — each verified against the committed file, not the description:**

| # | Claim | Reality |
|---|---|---|
| 1 | `/api/runtimes/all/files` not implemented | It is — `all` is handled inside the existing `/api/runtimes/<runtime_id>/files` route (`routes/runtime_memory.py:126`), verified live returning merged groups |
| 2 | `memory-runtime-chips` div still present | `git show HEAD:clawmetry/templates/tabs/memory.html \| grep runtime-chips` → no match |
| 3 | `skills-runtime-chips` div still present | Same, no match in `skills.html` |
| 4 | `ingest_memory_blob` never returns True | `clawmetry/local_store.py:2445` returns True after the write (and False on the sha256 dedup no-op) |

vivekchand:
Closing as a duplicate of #4851 — I hit the same user report independently and built the same fix across the same 7 files. #4851 is the better base: it's green, its data model is broader (category + root columns, skills synced too, fingerprint change-gating), and its owner already updated the Runtime and Session Observability blueprint.

The delta this PR had over #4851 is now stacked as **#4856**: the shared `runtime_is_locked()` predicate (closing the push-build half of the paywall gap — a downgrade kept shipping rows ingested while entitled), per-runtime push buckets (a single 600-row global query lets 429 Claude Code files crowd OpenClaw out of the blob entirely), `(runtime, path)` dedup for files several runtimes share, and 10 tests each verified to fail without its fix.

Cloud: my clawmetry-cloud#2009 is closed in favour of the `runtime-files-cloud` approach; I'm porting its node-run test harness (which exercises the real injected `cm-cloud-memory` script) onto that branch.

PR #4856, opened 2026-08-14
One paywall predicate, per-runtime push buckets, and the test suite (stacks on #4851)

vivekchand:
Stacks on #4851 (retarget to `main` when that merges). I hit the same user report independently, built a duplicate in #4855, and closed it in favour of #4851 — this PR is the delta that branch doesn't already have. Coordinated with the #4851 and #4854 owners.

## 1. The paywall predicate had three homes, and the third was missing

`_runtime_is_locked` existed in `routes/runtime_memory.py`; #4851 added a second copy as an `_allowed()` closure inside the daemon's ingest. There was no third copy on the **cache-push build** — so rows ingested while entitled kept riding the heartbeat after the entitlement lapsed. A downgrade went on serving paid-runtime memory to the cloud Memory tab.

`runtime_is_locked()` now lives in `clawmetry/runtime_memory.py`, and the route, the ingest, and the push build all read it. One answer, three callers.

## 2. One chatty runtime crowded out the others

The push did a single `query_memory_blobs(limit=600)`, sorted by `updated_at` across every runtime. This laptop has **429 Claude Code memory files** and **406 Hermes skill files** — the cap is spent before OpenClaw gets a single row, and OpenClaw's Memory tab renders empty. That is the exact bug this stack exists to fix, reintroduced one layer down. Rows are pulled per runtime now.

## 3. Dedup on path alone hid files shared between runtimes

`AGENTS.md` is read by Codex, opencode **and** Copilot; `seen` keyed on path meant it survived under whichever runtime sorted first and was invisible under every other one. Dedup is `(runtime, path)` now, the change-fingerprint is keyed the same way (a path-only key misses a change on all but the first owner), and `memory_content` entries carry their `runtime` so a viewer matching on path alone can't render one runtime's copy under another's tree.

## 4. Tests

`tests/test_runtime_memory_sync.py` — 10 tests, and I verified they actually fail without their fix rather than trusting green:

```
FAILED test_downgrade_stops_shipping_rows_ingested_while_entitled
FAILED test_one_chatty_runtime_cannot_crowd_out_the_others
FAILED test_a_file_shared_by_two_runtimes_survives_under_both
```

(that's with the three fixes reverted; all 10 pass with them). Coverage: per-runtime ingest + `category`/`root` tagging, sha256 no-op re-ingest, change pickup, locked-runtime ingest, the downgrade case, the crowd-out cap, shared-file dedup, over-budget files staying **listed** with empty content per the blueprint contract, the push floor, and a content change defeating that floor.

`tests/test_appjs_units.js` still 220/220.

### DIFF
--- a/clawmetry/runtime_memory.py
+++ b/clawmetry/runtime_memory.py
@@ -690,6 +690,36 @@ def list_runtimes() -> list:
 SKILLS_TAB_CATEGORIES: tuple = ("skills", "commands", "agents", "hooks")
 
 
+def runtime_is_locked(runtime_id: str) -> bool:
+    """True when this runtime needs a paid entitlement the caller lacks.
+
+    THE single predicate for the memory/skills paywall. Three callers must
+    agree or the paywall leaks:
+
+      * ``routes/runtime_memory.py`` — returns 402 / omits from the ``all``
+        sweep.
+      * the sync daemon's ingest — never writes a locked runtime's files.
+      * the sync daemon's cache-push build — never SHIPS them, which is a
+        separate decision: rows written while entitled would otherwise keep
+        riding the heartbeat after the entitlement lapsed.
+
+    Free runtimes are never locked. For paid runtimes ``allows_runtime`` on
+    the resolved entitlement is authoritative (it already honours grace mode).
+    Any exception falls OPEN — a resolver hiccup must not blank out a paying
+    user's memory, and the read endpoints re-check on every request.
+    """
+    try:
+        from clawmetry.entitlements import FREE_RUNTIMES, get_entitlement
+    except Exception:
+        return False
+    if runtime_id in FREE_RUNTIMES:
+        return False
+    try:
+        return not bool(get_entitlement().allows_runtime(runtime_id))
+    except Exception:
+        return False
+
+
 def parse_categories(category) -> set:
     """Normalise a category filter into a set of valid category names.
 
--- a/clawmetry/sync.py
+++ b/clawmetry/sync.py
@@ -8463,6 +8463,44 @@ MEMORY_PUSH_MIN_INTERVAL_SEC = MEMORY_CACHE_TTL_SEC // 2
 _memory_push_state: dict = {"fingerprint": None, "ts": 0.0}
 
 
+def _memory_rows_by_runtime(store) -> list:
+    """Memory rows pulled PER RUNTIME, not as one global most-recent-N.
+
+    A single ``query_memory_blobs(limit=N)`` sorts by updated_at across every
+    runtime, so the noisiest one takes the whole budget: this laptop has 429
+    Claude Code memory files and 406 Hermes skill files, which is already past
+    a 600-row cap before OpenClaw gets a single row. The Memory tab for the
+    crowded-out runtime then renders empty — the exact failure this branch is
+    fixing, reintroduced one layer down.
+
+    Locked paid runtimes are skipped here as well as at ingest: rows written
+    while entitled must stop shipping when the entitlement lapses, otherwise a
+    downgrade keeps serving them to cloud.
+    """
+    rows: list = []
+    seen_runtimes: set = set()
+    try:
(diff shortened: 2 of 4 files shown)


vivekchand (commented):
## Test plan & review notes

**What changed**
- Moves `runtime_is_locked()` into `clawmetry/runtime_memory.py` so the route, the ingest closure, and the cloud push build all share one entitlement predicate (previously each had a copy, with the push copy missing entirely — allowing paid-runtime memory to keep flowing after a downgrade).
- Switches the cloud push from one `query_memory_blobs(limit=600)` across all runtimes to per-runtime buckets, so a runtime with 400+ files can't crowd out another.
- Keys dedup and the change-fingerprint on `(runtime, path)` instead of bare `path`, so shared files (e.g. `AGENTS.md` read by Codex + Copilot) survive under every owner.
- Adds 10 tests in `tests/test_runtime_memory_sync.py`; the description confirms they fail without the fix.

**Smoke commands**
```bash
# Fast: just the new tests
python3 -m pytest tests/test_runtime_memory_sync.py -v

# Full suite
make test

# JS unit tests (should still be 220/220)
# tests/test_appjs_units.js per the PR description
```

**What to look at visually** (replay locally after #4851 lands)
- `http://localhost:8900/#memory` — select different runtimes in the picker; each should show its own files; a high-volume runtime (400+ files) must not cause another runtime's memory tab to render empty.
- After simulating a downgrade (revoke entitlement then trigger a push cycle): verify the cloud Memory tab no longer shows files from the previously-entitled runtime.

- If `from clawmetry.runtime_memory import runtime_is_locked` is missing in either `routes/runtime_memory.py` or the sync daemon's ingest closure, the route or the push will fall back to permissive behaviour — the downgrade regression silently reappears.
- If the per-runtime `ORDER BY updated_at DESC LIMIT N` runs after a cross-runtime union rather than per-shard, the cap is shared again.

**Issue link**
- No issue number referenced. The fix addresses the crowd-out bug described in #4851. Once this retargets to `main` on merge, consider `Closes <issue>` if there's a tracking issue.

**Note**: This PR targets `fix/memory-skills-real` (#4851's head) and needs to be retargeted to `main` once #4851 merges.

---

vivekchand (commented):
## Test plan & review notes

**What changed** (4 files, 359+/55−)
- `runtime_is_locked()` extracted to `clawmetry/runtime_memory.py` — route, daemon ingest, and cloud push-build all import from there now instead of each having their own copy
- Push query changed from a single `query_memory_blobs(limit=600)` sorted across all runtimes to per-runtime buckets — fixes OpenClaw being crowded out by 429 CC + 406 Hermes files
- Dedup key changed from `path` → `(runtime, path)` so shared files (`AGENTS.md` etc.) appear under every runtime that reads them
- 10 new tests in `tests/test_runtime_memory_sync.py` (verified red without the fix, green with)

**Smoke commands**
```bash
python -m pytest tests/test_runtime_memory_sync.py -v   # 10/10 should pass
make test-api                                            # fast sanity
python3 dashboard.py --port 8900                        # then hit Memory tab
```

**What to look at visually**
- `http://localhost:8900/` → Memory tab → cycle through every runtime → each should show its own file count, not just Claude Code
- Trigger a downgrade: entitle → ingest memory rows → revoke entitlement → confirm cloud push stops shipping those rows

- Import drift: any of the three callers still pulling `runtime_is_locked` from a stale local definition will shadow the fix — grep for `runtime_is_locked` across the tree to confirm there's exactly one definition
- Per-runtime cap math: 15 runtimes × 40 rows = 600 total; confirm the overall bound is still respected
- `(runtime, path)` dedup: check that nothing else in the push pipeline still keys on `path` alone

**Issue link**
No `Closes #N` found in the body. Stacks on #4851 and coordinates with #4854 — no standalone issue to close.

**Note on merge target**: PR description says retarget to `main` after #4851 merges; the current `blocked` state is expected while the stack lands.

---

PR #4875, opened 2026-08-15
[RELEASE] Paywall overlay fits the window and quotes the price Stripe charges (carries #4873)

vivekchand:
Carries #4873 to PyPI (0.12.708) and then to the cloud pin.

## Why this needs its own release

0.12.707 **armed the paywall for the first time**. That flipped the upgrade overlay from a screen almost nobody reached into the *only* screen a lapsed user sees. Two defects that were harmless while it was inert became blocking, and they only reach users through a new wheel.

### 1. The overlay could not be scrolled, so its own escape hatches were unreachable

`position: fixed; inset: 0` + `align-items: center` + no `overflow`. When the card is taller than the window, centering pushes overflow out of **both** ends of a fixed container, and with no scroller there is no way to get to it.

The two controls stranded at the bottom of that card are **"Activate license key"** and **"Continue free with OpenClaw + NVIDIA NemoClaw"**. So on a short window this was not a paywall, it was a lockout — a self-hosted customer holding a valid key could not enter it, and a user who only wanted their free runtimes back could not decline.

Container is now the scroller (`align-items: flex-start`, `overflow-y: auto`, `overscroll-behavior: contain`), with `margin: auto` on the card so it still centres when it fits. Vertical padding trimmed; card is 872px.

### 2. The free-runtime fallback named a runtime the user had just lost

```js
var RT_LABELS = { openclaw: 'OpenClaw', nemoclaw: 'NanoClaw' };
```

`nemoclaw` is NVIDIA NemoClaw (`FREE_RUNTIMES`). `nanoclaw` is a different runtime (`PAID_RUNTIMES`). The single line telling a blocked user what still works was advertising something behind the paywall, and never named the thing they kept.

## Pricing: correcting my own earlier framing

An earlier draft of this PR said the overlay's annual prices moved `$190 → $90` and `$390 → $190`. **That was wrong, and I've fixed it in the CHANGELOG.** The charged amounts were already correct — 0.12.707 quoted `starter 9/90` and `pro 19/190`, matching Stripe's `_SUB_PRICING` exactly.

`$190` and `$390` are the struck-through **anchors** being *added*, sourced from `pricing.html`'s `data-wasyr` rather than invented. What actually changed:

- struck annual anchors added (monthly deliberately has none, because `/pricing` publishes none)
- subtitles now match `/pricing` verbatim; interval pill reads "2 months free"
- annual-only device line correctly hidden on monthly
- **benefit list is now tier-aware** — selecting Pro previously repainted the price but left Starter's benefits underneath, so the Pro card sold Starter
- `_selTier`/`_selInterval` moved into module state, so the background entitlement poll can no longer reset the user's selection mid-repaint

## Verified before opening this

Rendered against the published 0.12.707 wheel in a browser, not asserted from the diff: annual `$190`→`$90`, Pro annual `$390`→`$190`, Pro monthly `$19` with the device line hidden, free label reads "OpenClaw + NVIDIA NemoClaw", Pro benefits repaint on tier switch, card 872px with both bottom controls reachable at an 800px viewport.

## After merge

1. PyPI 0.12.708 publishes.
2. Crack the wheel and grep the markers (FLYWHEEL §0a.3 — a concurrent `[RELEASE]` can bump past a feature commit).
3. Bump `clawmetry-cloud`'s Dockerfile pin to the version whose wheel was verified, and confirm the served `static/js/app.js` carries it.

## Companion PRs

- **clawmetry-cloud#2016** — new bug-class gate asserting the overlay's `PLAN_PRICES` matches `_SUB_PRICING`, read from the wheel the Dockerfile pins so it fires on the pin bump. The license-checkout path has had a guard for this family since the "$190-shown-$290-charged" bug; the overlay had none. Plus four stale price comments.
- **clawmetry-landing#640** — `PUBLIC_CLAIMS.md`, the file that calls itself the canonical pricing source, still carried the pre-reprice `$29`/`$290`.


vivekchand (commented):
## Test plan & review notes

**What changed**
- CSS: paywall overlay container made scrollable (`overflow-y: auto`, `align-items: flex-start`) so the "Activate license key" and "Continue free" buttons are reachable at any viewport height
- JS: `RT_LABELS.nemoclaw` corrected from `'NanoClaw'` (a different paid runtime) to `'NVIDIA NemoClaw'` (the actual FREE_RUNTIMES entry), so the free-tier fallback label is accurate
- JS: struck-through annual price anchors added (`$190` Starter, `$390` Pro) sourced from `pricing.html`'s `data-wasyr`; benefit list made tier-aware; `_selTier`/`_selInterval` moved to module state so background entitlement poll can't reset the user's selection mid-repaint

**Smoke commands**
```bash
# Syntax check
node --check clawmetry/static/js/app.js

# Confirm the mislabelled runtime is gone
grep -n "NanoClaw" clawmetry/static/js/app.js          # should return 0 hits
grep -n "NVIDIA NemoClaw" clawmetry/static/js/app.js   # should hit the RT_LABELS entry + the fixed comment
```

**What to look at visually** (replay locally on an 800 px-tall window)
- `/` (dashboard, lapsed/trial-expired session) — overlay should scroll vertically; "Continue free with OpenClaw + NVIDIA NemoClaw" label should be visible without scrolling past it; "Activate license key" button must be reachable
- Overlay tier switch: select Pro → benefit list should update to Pro features, not stay on Starter's
- Toggle Monthly ↔ Annual: the device line should appear/disappear correctly; annual view should show `$190`/`$390` struck-through anchors; monthly should show none

- If `overscroll-behavior: contain` interacts badly with pywebview's embedded Chromium, the scroll might be swallowed by the outer page instead of the overlay — worth testing in the desktop app specifically, not just a browser tab
- The `_selTier`/`_selInterval` module-state refactor: if the background entitlement poll fires during a tier/interval selection, make sure it doesn't visually reset the UI back to defaults

**Issue link**
- Carries #4873 (no `Closes` line in the body — if #4873 should auto-close on merge, add `Closes #4873` to the PR description)

**Drift-bot**: 3× no-drift ✅ — clear to merge once you're happy with the visual check.

---

PR #4892, opened 2026-08-15
[RELEASE] Quality: evidence-bearing verdicts, real runtime labels, per-runtime calibration

vivekchand:
## Why

Founder screenshot 2026-08-15: the Quality tab attributed eight rough runs to **openclaw** on a machine that only runs Claude Code, and asked why any number on that screen should be trusted. Two independent fabrications, both confirmed against the live DuckDB.

## The runtime label was a loop variable

`upsert_sessions` hardcodes `agent_type = session.get("agent_type") or "openclaw"` and nothing ever sets it — 61/61 claude_code sessions were stored that way. `query_outcomes` doesn't SELECT the column, so `routes/quality.py` swept 17 hardcoded agent types and stamped each row with whichever one the loop was on:

```python
for r in rows:
    r.setdefault("agent_type", at)   # `at` = the loop variable
```

Only the openclaw pass ever returned rows → the other 16 iterations were dead code, and every session read "openclaw". The correct label (`metadata.runtime`) was one field away.

## The failure signal couldn't produce a true negative

`_session_has_cognitive_loop` had one false-positive guard — "a new tool or file appeared, so this is progress" — reading tool use only from `data.message.content[*].type == "tool_use"`. Family runtimes emit tool calls as separate `tool_call` events under `data.tool_calls[*]`, and that type wasn't in the scan set. Replaying the live classifier on a real 824-event session:

```
candidate assistant turns:            41
turns with any tool use parsed:        0   <- guard is dead
```

All 8 flagged sessions had **39–63 real tool calls inside the exact window examined**. What remained was a 0.85 Jaccard match on the last 200 chars of three turns, which ordinary status narration trips trivially.

It was measurably noise: flagged sessions averaged **3.59%** real tool-error rate vs **2.96%** for "successes". The six genuinely worst sessions were all labelled success. `outcome_confidence` was a literal (0.8/0.85), and `success` was the fallthrough default — 18 of 62 sessions were research chats collecting a free pass.

## What changed

- **`clawmetry/quality_signals.py`** — reads both dialects into one normalized event; probes what each runtime actually emits; `Verdict.__post_init__` raises when exhibits are empty, so **a claim without evidence cannot be constructed**. Four signals, each declaring the fields it needs; a runtime that can't supply them reports *not measurable*, never a clean bill of health.
- **`clawmetry/quality_thresholds.py`** — percentiles of each runtime's own 30d history on this install, with floors and a documented cold start. Provenance travels into the evidence.
- **Confidence is derived** from sample size and margin — varies 0.43–0.95 on real data.
- **Grading moved to ingest** (`sync.py`, off events already in hand) so the tab is a pure DuckDB read.
- **Cost-weighted grade** — a plain mean let thirty cheap chats outvote the $82 run that spun on one file.
- **UI** renders the evidence: timestamps, tools, file paths, error text, threshold + source, plain-English signal names.

## The gate that would have caught this

`tests/test_quality_runtime_conformance.py` ingests all **12 committed runtime fixtures** and asserts that when raw events contain tool calls, the reader parses them *with inputs*. Verified to fail when the original bug is reintroduced.

It immediately earned its keep — **it found a second, independent blind spot**: goose, n8n, opencode and qwen_code carry tool arguments under `arguments`, not `input`. Thrash and forward-progress detection would have been silently dead on all four while the capability probe still looked healthy. The contract tests caught a third, in the opposite dialect (Anthropic `tool_use` blocks inside `message` events).

## Verified

- Live 62-session store: **7 of 8 prior false positives now read clean**; the survivor is a true positive (5 identical browser calls, 2 errored) with its evidence on screen. The 6 genuinely-worst sessions surface at 0.65–0.95 confidence.
- Playwright screenshot of the rendered tab: correct per-runtime labels, populated evidence panels, clean console.
- 62 new tests. Related suite matches the origin/main baseline exactly (36 pre-existing dev-box failures, **0 introduced**).
- Drive-by: `make lint-daemon-allowlist` was red on main (`query_replay_events` missing → session replay silently empty on daemon installs).

## Honest limits

- Benign-error filtering is Pro, so OSS computes higher rates than Pro on identical data — the tab now states which filter produced its numbers.
- Hosted dashboard has no local DuckDB; instead of reporting an empty week for a healthy machine, it says the grade is computed locally.
- Sessions past the deep-scan cap are screened, never reported rough without evidence; counts are in the payload.


vivekchand:

---

PR #4895, opened 2026-08-15
Show which Claude surface a session came from + org-wide coverage card

vivekchand:
## What

Sessions now carry a `surface` field (terminal / desktop / sdk) and render a quiet badge for it. Claude Code, the Claude Desktop app and Agent-SDK runs all write to the same transcript tree, so until now **every desktop session read as someone typing in a terminal**. On one machine that was 150 of 545 sessions.

The paid Claude adapter stamps `metadata["surface"]` (see clawmetry-pro #138). This PR exposes it on `/api/sessions`, carries it to the cloud push so the hosted list splits the same way, and renders the badge.

### The bug that live data caught

`surface` is read from `metadata["surface"]` **and nothing else**. `metadata["source"]` is already spoken for by other adapters, which fill it with cwd paths and provider names. Real values sitting in that key on this machine:

```
'ollama': 7, 'openai': 1, 'qwen_code': 4, 'cli': 4,
'/Users/vivek/projects/clawmetry': 2, '/Users/.../scratchpad': 4
```

A fallback read would have rendered `ollama` and full cwd paths as surface badges on unrelated runtimes' sessions. Pinned by `tests/test_session_surface_field.py`.

## Org-wide Claude coverage card

Added to the Claude Code harness view. Its shape is the argument: a solid accent rule for the one surface traced in full, a dashed hairline for every surface that is only ever a headcount.

```
CLAUDE SURFACE COVERAGE                       Not connected

  SURFACE            ON THIS MACHINE      ACROSS THE ORG
▌ Claude Code        Traced in full          not counted
┆ Cowork             Not on this disk        not counted
┆ Claude chat        Not on this disk        not counted
┆ Claude in Office   Not on this disk        not counted
┆ Claude Science     Not on this disk        not counted
```

The implementation lives in clawmetry-pro (it reads Anthropic's Enterprise Analytics API). OSS ships the `org_analytics` Enterprise catalogue key plus a 402 stub at the same URLs, and the card renders its **full ledger** in the locked state rather than a blank box. 402 and 404 both mean "not available to you", so neither reads as a network error.

`org_analytics` is Enterprise-tiered because the upstream API is Enterprise-only. Gating it lower would sell a key the customer's plan cannot mint.

## Test change worth a look

`test_enterprise_features_match_pricing` asserted the Enterprise set was **exactly** the 6 keys the /pricing card promises, so any catalogue growth turned it red. It now asserts those 6 as a subset, matching how the Pro set is guarded in the same file. The promise is that the 6 are honoured; a 7th Enterprise key breaks nothing a customer was told. Losing one still fails.

## Verification

- Dashboard served from this branch on :8901 with the paired pro branch on PYTHONPATH: `/api/org-analytics` returns the real unconfigured payload, and the served `app.js` + rendered HTML both carry the new code.
- Card rendered in every state (locked, unconfigured, connected, and each error kind); ledger comes out 1 solid rule + 4 dashed as designed.
- Copy checked against the no-em-dash rule.
- 18 new/changed OSS tests pass.


vivekchand:
✨ auto-fixed: added missing `FEATURE_LABELS` entry for `org_analytics` to fix Entitlement API tests CI failure

`org_analytics` was added to `ENTERPRISE_FEATURES` in this PR but the `FEATURE_LABELS` dict wasn't updated, causing `test_every_known_feature_has_a_label` to fail.

---

PR #4936, opened 2026-08-16
Fix Alerts tab wedged on 'Loading alerts…' once a rule is enabled

vivekchand:
## The bug (user report, screenshot: alerts tab stuck on Loading, no Edit possible)

Once any alert rule is enabled on a self-hosted / local-only node, the Alerts tab gets permanently stuck on **"Loading alerts…"** — no rule rows render, so there is nothing to edit or toggle.

## Root cause

`/api/alerts/rules` (local store) returns each rule's `channels` as a **JSON string** (`'["banner"]'`), not an array. The local-mode normaliser in `alerts.js` copied it straight into `channel_ids`, and `renderRules()` then threw:

```
TypeError: (real.channel_ids || []).map is not a function
    at renderRules (alerts.js:306)
```

The uncaught exception aborts `loadAlertsPage()` before any row is written, leaving the loading placeholder forever. With zero saved rules the `.map` path is never reached — which is why the tab only wedges **once you enable a rule**.

## Fix

- Parse a JSON-string channel list into a real array in the local-schema normaliser (`_alertsChannelList`), and use the same coercion at the render site so a bad shape can never blank the whole tab again.
- Stop rendering storage vocabulary (`threshold`) as a rule name when a local row has no real name — fall back to the canonical example label ("Daily spend > $50").

## Verification

Reproduced against the live local store (rule enabled → tab wedged, exact TypeError captured in console), then ran this branch on :8901 against the same data: 10 rows render (enabled rule + examples + orphans), Edit opens the populated editor modal, no console errors.

### DIFF
--- a/clawmetry/static/js/alerts.js
+++ b/clawmetry/static/js/alerts.js
@@ -948,12 +948,15 @@
       .map(i => i.dataset.channelId);
     const policy = document.querySelector('input[name="alerts-re"]:checked')?.value || 'once';
 
-    // Editing a canned example or saving on a non-Pro tier: fire the paywall
-    // here, AFTER the user has configured the rule. They're more invested by
-    // this point -- better conversion than gating on first click.
+    // Saving on a non-Pro tier: fire the paywall here, AFTER the user has
+    // configured the rule. They're more invested by this point -- better
+    // conversion than gating on first click. An entitled (Pro/trial) user
+    // editing a canned example must NOT hit the paywall — for them, saving
+    // an example creates the real rule (the old check paywalled them, so a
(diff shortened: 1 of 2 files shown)


vivekchand:
Follow-up commit: two more blockers in the same report — (1) entitled (local Pro/trial) users were paywalled on saving an edited example rule; (2) the local PUT ignored the editor's cloud-vocabulary fields (threshold_value/channel_ids), so threshold edits silently didn't persist. Both fixed and verified in-browser against the real local store (edited Daily-spend 50→75 via the editor, server persisted, state restored).

PR #4972, opened 2026-08-17
Enterprise security review: remove third-party asset loads, close egress, add trust artifacts

vivekchand:
Makes the self-hosted claim true at the edges, and produces the artifacts a vendor security review asks for.

## Why

`clawmetry/selfhosted.py` already documents "a self-hosted deployment makes NO calls to clawmetry.com". That was true of the **server** — but not of everything around it. Four gaps made the claim false before any daemon opened a socket.

## What was wrong

**Every dashboard page load contacted two external origins.**

- `fonts.googleapis.com` (4 head blocks) — discloses every viewer's IP and User-Agent to a third-party processor on each page load; renders in a fallback face on an air-gapped install.
- `cdn.jsdelivr.net` (chart.js + date adapter) — third-party JavaScript with DOM access, executing in the page that renders agent transcripts. No SRI, and none possible: jsDelivr's own minifier documents that SRI cannot be used with its dynamically generated files. Both tags sat two lines under a comment reading *"Vendored + pinned (no external CDN, no supply-chain risk)"*.

**The suppression gate was keyed wrong.** Install telemetry and funnel analytics checked `is_custom_endpoint()`, true only when `CLAWMETRY_ENDPOINT` is repointed. A self-hosted **server** container does not set that — it *is* the endpoint — so `SELF_HOSTED=true` alone suppressed nothing in that process.

**`api.ipify.org` was ungated everywhere**, buying one cosmetic startup banner line in exchange for telling a third party that this network runs ClawMetry.

**Gates failed open.** The telemetry gate caught `ImportError` and fell through to the managed-cloud URL, so a partial install meant a customer's self-hosted server phoned home.

## What changed

**Assets vendored.** 15 woff2 subsets for the dashboard + 13 for the v2 SPA, generated by `scripts/vendor_fonts.py` — deduped by content hash (60 files to 15, 1.3 MB to 363 KB) with `unicode-range` preserved, so a Latin-only viewer still fetches ~25 KB. chart.js 4.4.7 and its adapter pulled from the npm registry tarball, not the CDN's re-served copy.

**Provenance, not just presence.** `scripts/vendor.lock.json` records package, version, tarball path, license and SHA-256 for all four vendored bundles; `scripts/verify_vendor.py` re-downloads the registry tarball and byte-compares. All four are byte-identical to their published releases.

**Egress closed.** New `clawmetry.endpoints.egress_suppressed()` ORs the three ways an operator declares a deployment private (repointed endpoint, `SELF_HOSTED`, `CLAWMETRY_OFFLINE`). Applied to telemetry, funnel analytics, the public-IP lookup, and the unattended-update PyPI check. Every gate now fails **closed**. The configured push to your own ingest endpoint is untouched — that traffic is the product.

**Trust artifacts.** `docs/EGRESS.md` lists every destination, what it sends, when, and how to disable it, with three commands a reviewer can run. `SECURITY.md` rewritten from 19 lines to a real policy with response targets, scope, supported versions, an honest compliance table, and a deliberate Known Limitations section. RFC 9116 `security.txt` served by the app itself, so an on-prem instance can answer "who do I report this to?"

**Supply chain in CI** (`.github/workflows/supply-chain.yml`, also weekly): external-asset scan, vendor byte-comparison, font staleness check, `pip-audit`, CycloneDX SBOM, OpenSSF Scorecard to SARIF.

**Release provenance.** `actions/attest-build-provenance` signs every artifact. PyPI Trusted Publishing is wired but **gated off** behind a repo variable — it cannot work until the publisher is configured on PyPI, and this repo auto-publishes on merge.

## Found along the way

- **DOMPurify was 3.1.7**, which OSV flags for CVE-2025-26791 (mXSS via template literals). Not reachable here — that path needs `SAFE_FOR_TEMPLATES:true`, which we never set — but scanners flag the version regardless. Upgraded to 3.4.13.
- **`.gitignore` bug.** A duplicate `dist/` rule after the v2 un-ignore silently re-ignored `clawmetry/static/v2/dist/`. Already-tracked files were unaffected, which is why it went unnoticed, but every **new** file in a rebuilt bundle was being dropped — a CI build from a clean checkout would have shipped an incomplete SPA. The v2 fonts are the first new files to land there.

## Testing

- 47 new tests (`test_egress_suppression.py`, `test_security_txt.py`), including that a default managed-cloud install is **not** suppressed, so the suppression tests prove something.
- Negative test confirmed: the external-asset guard catches a reintroduced CDN script tag.
- Verified 28 woff2 + both stylesheets + all four vendor bundles + `security.txt` ship inside a built wheel.
- **Zero regressions**: the affected-area suite (554 tests) was run on this branch and on pristine `origin/main`; both produce an identical set of 38 pre-existing failures.

## Before merging

1. **Create `security@clawmetry.com`** — `SECURITY.md` and `security.txt` both publish it. A published contact that bounces is worse than none.
2. `clawmetry.com/security` currently 404s and `/enterprise` promises a DPA, pentest summaries and an IR plan that do not exist. Both live in the landing-site repo.

See `docs/TRUST_BADGES.md` for the follow-on certification plan.


vivekchand:
**Blocker: phantom CI cascade from main regression (not a real test failure)**

All 30 check runs on this PR complete in 2-6 seconds — including "Python dependency audit", "Vendored fonts not stale", "No third-party asset loads", and all Sync matrix jobs, which normally take 30-120 seconds each. These are phantom failures (runner never assigned, concurrency-cancelled).

**Root cause:** main CI went red after PR #5008 merged (~2026-08-19T23:53 UTC). E2E Gate (required) is also phantom (3s, no runner).

**To unblock:**
1. Find what broke on `main` after #5008 merged
2. Fix main CI → re-trigger this PR

This PR (drift-bot ✅, no merge conflicts) is ready once main CI recovers.

---

PR #4986, opened 2026-08-18
Sell the plan in the modal, and let people switch accounts

vivekchand:
Two dead ends on the expired-trial screen, both founder-reported 2026-08-18.

## 1. The upgrade dialog replaces "See pricing"

Before: an already-signed-in user whose trial ended got one link to clawmetry.com/pricing and had to start a checkout from scratch on a marketing page.

After: the modal sells in place.

- Monthly / Annual toggle (annual preselected, "2 months free")
- Starter / Pro cards with live per-node prices and the prior-ladder anchor struck through
- The free **$149 desk device** on annual, hidden on monthly (the cloud already collects a shipping address on annual checkouts and ships on the first paid invoice, so this is a real promise)
- One button → `POST /api/trial/checkout {tier, plan}` → Stripe Checkout Session scoped to the account already on this machine

Prices come from `window.CM_PLANS`, published by the hard-block overlay in `app.js`. Two hardcoded ladders is how a reprice ships half-done.

The step also polls for a licence now: signing in as an account that **already pays** lands here too (the self-host rail reports "trial expired" because the trial is spent), so a paying customer is let in on the next heartbeat instead of being sold a plan they own.

## 2. Sign out / switch account

If the Pro licence lives on a different email there was no way to get to it: the gate blocks the dashboard, every button on it reuses the identity on disk, and the only escape was `clawmetry disconnect` in a terminal.

`POST /api/account/signout` forgets the account — licence, cm_ key + sync state, this gate's stamp, and the desktop shell's stamp. Miss any one and the user is still stuck; and because `run_daemon()` reads `config.json` once at startup and holds the key in memory, a still-running daemon would re-install the old account's licence from its next heartbeat, so the endpoint kicks the daemon too. Local DuckDB data is untouched: identity change, not a wipe.

Offered where each user actually is — "Use a different account" in the modal, "Switch account" in the profile menu. Two clicks in both, so a stray click never signs anyone out.

## Allowlist

`/api/onboarding/`, `/api/cloud-cta/` and `/api/account/signout` now answer while the trial hard block is up. Both ways OUT of the block (pay, or sign in as the account that already paid) have to work, or the block is a trap rather than a paywall. Sign-out strictly narrows entitlement, so it can never be a bypass.

## Verification

Ran a dashboard from this branch on :8901 against a fake HOME and drove it in Chrome:

- expired step renders, Monthly/Annual and Starter/Pro switch prices live (`$90/$190 per node/year` ↔ `$9/$19 per node/month`), device line hides on monthly
- checkout button opens the tab synchronously and reports "Waiting for payment…" (`/api/trial/checkout` returned a URL)
- sign-out arms on the first click, and the live endpoint returned `{ok: true}`, reopened the gate, wrote the nocloud marker and restarted the daemon local-only
- profile menu renders "Switch account" for a signed-in account

13 new tests + the existing onboarding/hard-block suites pass; `make lint-js`, `lint-daemon-allowlist`, `lint-runtime-count` clean; ruff clean on every file touched.

### DIFF
--- a/clawmetry/static/js/app.js
+++ b/clawmetry/static/js/app.js
@@ -48,6 +48,19 @@
     starter: ['Unlimited channels + cloud sync', 'Approval queue'],
     pro: ['Everything in Starter', 'Tool policy + evals + cost optimizer'],
   };
+  // Publish the table so the OTHER in-app upgrade surface — the self-host
+  // modal's expired-trial step (static/js/onboarding.js, loaded after this
+  // file) — sells the same plans at the same prices. Two hardcoded ladders
+  // is how a reprice ships half-done.
+  window.CM_PLANS = {
+    prices: PLAN_PRICES,
+    blurb: PLAN_BLURB,
+    features: PLAN_FEATURES,
+    // Annual-only perk. The cloud collects a shipping address on annual
+    // checkouts (_annual_device_checkout_extras) and ships on the first PAID
+    // invoice, so this is a real promise, not a made-up one.
+    deviceValue: 149,
+  };
   var _selTier = 'starter';
   var _selInterval = 'year';   // annual preselected: better retention + the device perk
   // Filled from /api/entitlement (allowlisted, so it answers while blocked).
--- a/clawmetry/static/js/gw-setup.js
+++ b/clawmetry/static/js/gw-setup.js
@@ -465,6 +465,18 @@ function _cmProfileRender(st) {
       + '<svg width="14" height="14" viewBox="0 0 24 24" fill="none" stroke="currentColor" stroke-width="2" stroke-linecap="round" stroke-linejoin="round"><path d="M15 3h4a2 2 0 0 1 2 2v14a2 2 0 0 1-2 2h-4"/><polyline points="10 17 15 12 10 7"/><line x1="15" y1="12" x2="3" y2="12"/></svg>'
       + t('profile.sign_in', null, 'Sign in / Create account') + '</button>';
   }
+  if (st.signedIn) {
+    // Switch account. Distinct from the "Sign out" below it, which only ever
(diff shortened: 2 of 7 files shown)


vivekchand:
Drift-bot triage, round 4 (3 → 4 → 2 → 4 findings; the real ones are fixed, two are base-misreads).

**Fixed — blueprints updated:**
- #3 / #4 (callers only read `ok`, not `cleared`/`state`): fair as written, because I over-specified the response. Both blueprints now say `ok` alone is sufficient to act on — the caller reloads into the gate either way and the gate re-resolves state on its own — and that `cleared`/`state` exist for diagnosis and for the tests that assert every piece of identity actually went. Local Agent Observability v16, Team Access and Subscription Management v7.

**Disproved — the bot is reading the merge base, not HEAD:**
- #1 "no Python route handler exists to implement this endpoint":
  ```
  $ git show origin/upgrade-dialog-inline:routes/onboarding.py | grep -n 'api/account/signout'
  491:@bp_onboarding.route("/api/account/signout", methods=["POST"])
  ```
  It is on `bp_onboarding`, which `dashboard.py` already registers, and `tests/test_upgrade_dialog_and_account_signout.py` exercises it through a real Flask test client (7 tests: clears all four pieces of identity, reopens the gate, restarts the daemon, stops egress, idempotent, refuses in cloud mode, survives a broken licence layer).
- #2 "the blueprint's allowlist does not document /api/onboarding/ and /api/cloud-cta/": it has since v13, in the same enumeration the bot is quoting — `..., /api/auth/*, /api/onboarding/*, /api/cloud-cta/*, /api/account/signout, /, /static/*` — with the rationale in ADR-106 ("Both ways out of the hard block must answer while it is up"). The blueprint is now at v16.

Not editing a spec to match a finding I cannot reproduce, per FLYWHEEL §1f triage.

PR #4997, opened 2026-08-18
Windows installers: same existing-account gate as install.sh

vivekchand:
Follow-up to #4996, which taught `install.sh` to detect an already-linked account and ask before replaying the onboarding wizard. Same contract now on Windows (`install.ps1` + `install.cmd`):

```
  ✓ You're already connected to ClawMetry

    Account:     founder@example.com  (Pro plan)
    Cloud sync:  Local-only (data stays on this machine)
    Version:     0.12.737
    Node:        ci-box
    Dashboard:   http://localhost:8900

  Re-run setup (account, cloud vs local-only, license)? [y/N]:
```

## Two Windows-only problems this had to fix first

1. **Neither Windows installer ever ran `clawmetry onboard`.** A Windows user ended up installed but not set up (no account, no daemon, no dashboard) unless they found `clawmetry connect` in the docs. Both scripts now onboard when no account is linked, exactly like macOS/Linux, and gate it behind the question when one is.
2. **`install.ps1` wiped and rebuilt the venv on every run.** Windows locks the files of a running process, so on the exact machine this feature targets (an install whose daemon is live) `Remove-Item -Recurse` fails and the install aborts. It now upgrades in place and only rebuilds when there is no usable venv. The pre-flight stops ClawMetry's own processes (the sync daemon, plus anything running out of the install dir) before pip replaces their files, and restarts the daemon on the new code afterwards. A dashboard the operator was running is reported, not silently resurrected (same as macOS/Linux).

## Shared contract across all three installers

- **Connected** = an API key on this machine. A placeholder (`…@clawmetry.auto` / `…@clawmetry.linked`) account does **not** count, so that node still gets the wizard.
- The probe prefers `clawmetry status --json`, falls back to `config.json` + `cloud_plan.json` + the `nocloud` marker, and degrades to "not connected" on any failure rather than breaking the install.
- Default is keep-my-setup: empty answer, `n`, or a non-interactive run changes nothing. `CLAWMETRY_REONBOARD=1/0` forces it either way, `CLAWMETRY_SKIP_ONBOARD=1` skips onboarding entirely, `CLAWMETRY_LOCAL_ONLY=1` writes the nocloud marker.
- `install.ps1` parses `status --json` with `ConvertFrom-Json`; `install.cmd` hands the work to one python probe that prints the summary and returns the answer in its exit code, so batch never parses JSON.

## Verification

- `tests/test_windows_installer_existing_account_gate.py` — 25 tests: the PowerShell helpers driven under `pwsh` (fresh / local-only-without-account / placeholder / corrupt config / config-file fallback / snapshot preferred / dead dashboard URL / env overrides / non-interactive), the `install.cmd` python probe executed directly on any OS, and static parity guards across all three installers.
- `.github/workflows/install-test.yml` — real `windows-latest` runners now assert the gate end to end: a synthetic connected profile (via `USERPROFILE`) must be reported and left alone, and a fresh profile must still get the wizard. The plain install steps run with `CLAWMETRY_SKIP_ONBOARD=1` so they keep testing the installer, not onboarding.
- Locally: `pwsh 7.6.5` parses `install.ps1` clean and all three decision paths (`y` / `n` / bare Enter / `CLAWMETRY_REONBOARD` / no account) behave identically to the bash installer.

### DIFF
--- a/.github/workflows/install-test.yml
+++ b/.github/workflows/install-test.yml
@@ -46,6 +46,7 @@ jobs:
 
   test-windows-ps:
     runs-on: windows-latest
+    timeout-minutes: 20
     steps:
       - uses: actions/checkout@v7
       - name: Set up Python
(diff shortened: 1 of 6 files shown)


vivekchand:

All CI check runs pass. The only blocker is the `drift-bot` commit status reporting 2 drift finding(s). Resolving those requires the author to update the blueprint/architecture documentation to match the new Windows installer behaviour.

---

vivekchand:
Pushed a fix for both drift-bot findings (commit b439eb7).

Both `install.ps1` and `install.cmd` were missing the existing-account gate on the "already up to date" early-exit path. Added the version check + gate block before the pip install in each file, mirroring `install.sh`'s early-exit section. CI should re-run and drift-bot should clear.

---

vivekchand:
Note for the drift-fix automation: I reverted b439eb770 in b18e60e04.

That early-exit block sat **after** the pre-flight that stops a running ClawMetry daemon and returned **before** the restart block, so on an already-connected machine with a live daemon the installer would have stopped the daemon and exited, leaving the node dead. It also re-implemented the probe/prompt inline (a second copy of the same contract, and one that ignores the `status --json` snapshot, losing the live account email and plan) and printed no setup summary — which the new windows-latest CI step caught.

The finding was written against Installer Scripts blueprint v4. Blueprint v5 now scopes that contract per component: `install.sh` has two entry points (its early exit skips a full venv rebuild, so it earns its keep), while the Windows installers have a single path — their install step is an idempotent `pip install --upgrade` that no-ops when already current — and apply the gate once, after it.

PR #5003, opened 2026-08-19
Stop paywalling Pro accounts at boot, and stop the daemon proxy dropping positional args

vivekchand:
Founder report from the desktop app (v0.12.736): the upgrade modal on a `cloud_pro` account, plus a trickle of 400s in `sync.log`. Two unrelated causes.

## 1. A Pro account gets paywalled for the first ~38 seconds after launch

`/api/entitlement` was correct when the report came in (`tier: cloud_pro`, `is_paid: true`, `enforced: false`, `locked_runtimes: []`). The lock map is built once at page load, and the page loaded before the answer existed:

| time | event |
|---|---|
| 18:08:59 | desktop shell starts |
| 18:09:01 | dashboard starts, JS fetches `/api/runtimes` |
| **18:09:39** | daemon writes `cloud_plan.json` |

Entitlement resolves from that file, so every fetch in between falls through to `_oss_free()`, whose runtime set is OpenClaw + NemoClaw only. Confirmed directly:

```
$ HOME=<empty> python3 -c 'from clawmetry import entitlements as E; ...'
tier= oss   runtimes: ['nemoclaw', 'openclaw']   claude_code entitled: False
```

`_cmLoadRuntimeCatalog` locked all 20 paid runtimes on that, and the re-check that would have repaired the map was gated on `window.CLOUD_MODE`, which the desktop app never sets. The stale map then survived for the life of the page.

The root problem is that an OSS-free verdict was read as "this user is on the free plan" when it meant "we do not know yet".

**Fix.** `entitlements.plan_pending()` names the third state: true when an account is linked (`api_key` in `config.json`) but no licence and no readable cloud plan exist yet; false once the plan lands, and false on a never-linked machine where OSS-free is the honest answer. `/api/runtimes`, `/api/entitlement` and `/api/entitlement/runtime-detection` publish it, degraded fallback paths included, because a resolver crash is also "unknown".

Three client guards, because one of them being wrong is how this shipped:
1. the catalog never locks while `pending`, and re-checks until it resolves (bounded at 20 tries, no longer cloud-only);
2. the switcher re-verifies against a fresh catalog before refusing a switch, and honours the switch when the runtime turns out to be entitled;
3. `_cmShowRuntimePaywall` re-verifies before rendering, covering all three entry points into that modal at once, including the inventory banner that reads a different endpoint.

## 2. The daemon proxy silently discarded every positional argument

`_ProxyStore.__getattr__` forwarded calls as `local_store_via_daemon(name, **kwargs)`. The transport has no slot for positional arguments, so `*args` was dropped, the daemon answered `400 missing required argument`, and callers swallowed that as an empty result. The signature is right there in the log, same method, two lines apart:

```
POST /__local_query__/query_session_authority_counts  200   <- kwargs callsite
POST /__local_query__/query_session_authority_counts  400   <- positional fallback
```

**Fix.** Positional arguments are bound against the real `LocalStore` signature, so both call styles work. Calls that cannot be expressed that way (keyword-only signature, unknown method) return `None` with a warning instead of a silent 400; private helpers like `_fetch` are refused locally rather than sent to an allowlist that will never accept them; dunder probes raise `AttributeError` instead of resolving to a proxy call.

A `PROXY_UNAVAILABLE` sentinel separates "the daemon could not answer" from "the answer is `{}` / `None`". Two callers read the latter as failure:
- `_stamp_authority_counts` re-ran its whole lookup against a direct store on every request (`{}` is the common case);
- `_ls_write` reported a landed `ingest_approval` as failed, since every writer it calls returns `None` on success.

## 3. Log noise

`load_policies` runs on every evaluation pass and logged at INFO each time. The reported `sync.log` held 1,382 identical `loaded 17 approval policies` lines. It now announces itself only when the policy set actually changed.

## Verification

- **17 new tests**, each confirmed red against `origin/main` (7 failed, 10 errored) and green here. They pin the positional binding for both real callsites, the keyword-only signature that must not be faked, `_fetch` never reaching the transport, all four `plan_pending()` states, the premise that an unresolved plan resolves to a tier without `claude_code`, `pending` reaching the wire, and the policy log staying quiet on an unchanged reload.
- The existing `test_entitlement_api_fallback_shape.py` parity guard caught the first attempt, which added `pending` to the healthy response but not to the frozen fallback snapshot. Both carry it now.
- **No new failures.** The approvals and local-store suites are compared failure-by-failure against the base: this branch introduces none and repairs 9 that main already fails, several of them (`test_authority_counts_empty_input`, the approval round-trips) direct evidence of the argument-dropping bug.
- `make lint` holds at its pre-existing 215 findings.

Not yet walked in a live desktop instance end to end, since the boot race needs a fresh launch against an unwritten `cloud_plan.json`; the timing is reproduced from the reported instance's file mtimes and process start times.


vivekchand:
### visual-diff triage (non-blocking check, eyeballed as its own output asks)

`visual-diff` is red on `865393be9` with 35/66 comparisons flagged, including `desktop overview -> 100.00%`. That one matters, because Overview is the surface this PR touches, so I checked it rather than waving it through.

**It is not this change.** The two commits below carry *identical application code* — `git diff --stat a3bc668fa 865393be9` is one file, `.github/workflows/auth-bootstrap-guard.yml`, a `timeout-minutes` value:

| commit | `desktop overview` | flagged | check |
|---|---|---|---|
| `a3bc668fa` | **0.25%** | 18/66 | **pass** |
| `865393be9` | 100.00% | 35/66 | fail |

The `after` screenshot on the failing run shows the page stuck on the "Initializing ClawMetry" splash (70 KB vs 478 KB for `before`), and the same run logged `page.goto: Timeout 30000ms exceeded`, a `[render-error] head=0` on `mobile turn-anatomy`, and a browser-context crash before being cancelled. The other 100% entries are the same thing: render failures, not visual regressions.

Independently, a dashboard built from this branch boots and renders normally against an isolated HOME, and the runtime switcher behaves correctly in both directions (entitled runtime switches without a modal; a server-locked runtime still reverts the selection and shows the paywall).

Same story on the other two reds earlier in this PR, both since re-run green: `Auth bootstrap no-reload smoke` was killed mid-`playwright install --with-deps chromium` inside a 3-minute job budget (raised to 10 in this PR, since that job can never reliably finish a Chromium download in 3 minutes and it fires on any PR touching `static/js`), and `E2E Gate` failed only as a cascade while polling for `OSS golden path`, whose 37 E2E tests had all passed before its job hit the wall in `Build wheel`.

PR #5007, opened 2026-08-19
[RELEASE] Approvals empty states respect the runtime filter (no more OpenClaw copy under a Claude Code filter)

vivekchand:
## What the operator saw
Approvals tab filtered to **Claude Code**, yet the "No pending approvals" empty state was a ~90-word paragraph explaining every runtime's gating mechanics — and explicitly naming **OpenClaw**. The history empty state was equally generic. (Founder screenshot, 2026-08-19, v0.12.737.)

## Fix
The pending + history lists already filter rows per-runtime; the empty-state copy now does too:

| Filter | Pending empty state |
|---|---|
| Claude Code | **No pending approvals for Claude Code.** Turn on a protection rule on the right and risky Claude Code calls pause before they run (PreToolUse hook), then land here for a decision. |
| OpenClaw | …pause before they run (exec-policy preset)… |
| Codex (etc.) | …are caught as they happen… |
| all | **No pending approvals.** Turn on a protection rule on the right and risky calls from your agents land here for a decision. |

History: "No Claude Code decisions yet." (generic sentence retained for "all").

## Verification
Booted the branch on a fake-HOME `:8903` instance and rendered all four variants live (screenshot attached in session). Inline `<script>` passes `node --check`.

## Notes
- Drops 4 `data-i18n` keys (copy is now interpolated with the runtime label, which static keys can't express). `test_i18n_no_raw_codes` has 2 pre-existing failures on main (45 missing keys, none touched here).
- The Protection Rules right-hand column is genuinely node-wide ("One rule, every runtime") and unchanged.

### DIFF
--- a/clawmetry/templates/tabs/approvals.html
+++ b/clawmetry/templates/tabs/approvals.html
@@ -585,6 +585,7 @@
       // default — rows carry the requesting session id, whose prefix names
       // the runtime. BOTH renders (pending + history) filter identically.
       var _rt = (typeof _cmRuntimeFilter === 'function') ? _cmRuntimeFilter() : 'all';
+      var _rtLbl = (_rt && _rt !== 'all' && typeof _cmRuntimeLabel === 'function') ? _cmRuntimeLabel(_rt) : '';
       if (_rt && _rt !== 'all' && typeof _cmRuntimeOf === 'function') {
         var _match = function(r){
           return _cmRuntimeOf({session_id: r.session_id || r.requestor_session_id}) === _rt;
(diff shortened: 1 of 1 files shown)


vivekchand:
**Drift-bot triage (1 finding):**

- *"`_cmRuntimeLabel()` is not defined in the codebase"* — disproven; it is defined in `clawmetry/static/js/app.js` on both base and head:
  ```
  $ git show origin/main:clawmetry/static/js/app.js | grep -n "function _cmRuntimeLabel"
  11212:function _cmRuntimeLabel(rt) { return _CM_RT_LABEL[rt] || rt; }
  ```
  `app.js` loads on the same dashboard page before any tab template runs, and the call site guards with `typeof _cmRuntimeLabel === 'function'` (raw runtime id fallback).
- *"the blueprint does not specify how runtime labels should be provided to the UI layer"* — real gap, now fixed: **Remote Alerts and Action Approvals** blueprint v8 adds "Runtime-scoped empty states on the Approvals tab (PR #5007)" with the label-catalog contract and ADR-007.

vivekchand:
**Drift-bot round 2 triage:** the remaining claim — "`_cmRuntimeLabel()` does not exist in the codebase / differs from the alerts.js pattern" — is the documented base-misread class. The function exists on main and head and IS the shared catalog accessor the blueprint (v8) describes:

```
$ git show origin/main:clawmetry/static/js/app.js | grep -n "_cmRuntimeLabel\|_CM_RT_LABEL = " | head -3
11128:var _CM_RT_LABEL = {
11212:function _cmRuntimeLabel(rt) { return _CM_RT_LABEL[rt] || rt; }
```

`app.js` loads globally before every tab template; alerts.js's local `rtLabel` wraps the same catalog. Blueprint gap from round 1 was fixed in v8 (ADR-007). Merging per the standing drift-bot policy: disprove with `git show`, don't rewrite a correct spec to match a misread.

PR #5009, opened 2026-08-19
Matrix-gap sprint: kill mid-run for 6 more runtimes + Cursor/Copilot pre-tool gates

vivekchand:
## What

The public capability matrix claims Kill mid-run ✓ for Codex/OpenCode/Aider/Goose — but the shipped cloud Stop button never worked for them: the relay carries no `cwd`, the daemon never backfilled it, and `resolve_by_cwd` refused with `no_cwd`. Family session rows never persisted `cwd` either. This PR makes the promised cells true and extends kill/pause to every runtime the 2026-08-19 recon proved resolvable.

### process_control
- **copilot**: strong pid map from `~/.copilot/logs/process-<epoch_ms>-<pid>.log` (`Workspace initialized: <sid>` body marker) — pid from the filename, epoch_ms doubles as the pid-reuse-guard start token. Verified live on Copilot CLI 1.0.77–1.0.80 (SIGTERM is graceful: `session.shutdown` written, `--resume` works after).
- **qwen_code**: qwen-code's own pid sidecar `<projects>/<hash>/chats/<sid>.runtime.json` (`{pid, session_id, work_dir}` — written explicitly for observability daemons). Its `started_at` is the write time, so the resolver uses identity cross-checks (argv contains qwen + live cwd == work_dir) instead of the start-token guard.
- **pi / grok / deepseek_harness / kimi**: argv+cwd. `pi`/`dsh` are exact-basename hints (substring would hit pip/python/any path containing "dsh").
- **cursor**: CLI (`cursor-agent`) sessions resolve by cwd — one process tree per session; the IDE stays honestly unsupported.

### sync.py daemon dispatch
- **cwd backfill**: when the relay omits `cwd`, look it up from the session row (new `local_store.get_session_location`) — fixes the broken codex/goose/opencode/aider path end-to-end.
- **KILL_HANDLERS first**: the pro-registered n8n / antigravity / nanoclaw API kills now fire from the Stop button, not only on approval deny.
- **nemoclaw** routes through the openclaw CLI cancel instead of a guaranteed `runtime_not_signal_supported`.

### family ingest
- Session rows persist `cwd`/`git_branch` (alias walk, incl. goose `workingDir`); family event rows stamp `workspace_id` so `approvals._session_cwd_hint` stops returning `''`. `/cwd1` ingest-rev salt re-ingests existing sessions once to heal the column.

### local emergency stop honesty
- `/api/sessions/<sid>/stop` for family sids routed to the real pid engine instead of writing a `.stop` file nothing reads and returning `ok:true`.

## Tests
13 new: copilot E2E kill incl. pid-reuse guard refusal, qwen sidecar identity guards, exact-hint semantics, handler-first dispatch, cwd backfill, nemoclaw routing. Full `test_process_control.py` (33) + `test_sync_process_control_dispatch.py` (13) green.

Companion PRs: clawmetry-pro (adapter cwd promotion + Copilot $0.01/credit + QM cost) and clawmetry-cloud (gate lists + dialog copy).


vivekchand:
**Drift-bot triage (6 findings, round 2):**

- **Findings 1 & 4 are factually wrong about the code.** Cursor is NOT in `SUPPORTED_RUNTIMES` — the set at `clawmetry/process_control.py:90` is `{claude_code, codex, goose, opencode, aider, copilot, qwen_code, pi, grok, deepseek_harness, kimi}` and `UNSUPPORTED_RUNTIMES = {"cursor"}` is unchanged. Cursor resolution is split per session in `resolve_cursor()`: CLI (`cursor-agent`) sessions resolve by cwd; editor conversations return the explicit unsupported result. That split is now specified in the *Extended Runtime Support* blueprint (v11, ADR-ERS-005) and requirement (v2, AC-GOV-ERS-005.2).
- **Findings 2, 5, 6 read stale spec versions.** The runtime-gates module is documented in *Governance Policy and Approval* v7 ("Pre-execution tool gates for Cursor and GitHub Copilot CLI", ADR-GPA-004 covers the per-runtime receivers + runtime-scoped policy filtering); first-class `Session.cwd` incl. the extra-mirror-is-diagnostic contract is in *Runtime and Session Observability* v16 and requirement v5 (AC-RSO-CWD-001.\*); the per-runtime receiver + scoping behavior is also in *Governance and Human Approval* v7. All patched before this run — the bot appears to have read pre-patch versions (same stale-read behavior as on #4986).
- **Finding 3 conflates two components.** The "strictly observational" contract belongs to the AttentionHookReceiver, which this PR does not touch. `_pretooluse_impl` is the *deciding* receiver (the existing claude-code gate), and its fail-open contract is unchanged: any policy-engine or store failure answers `allow` with a reason — the runtime-scope filter runs inside the same try/except and its failure mode is that same fail-open allow. Spelled out in *Governance and Human Approval* v7.

Separately: every other check on this round died in 2s with "account is locked due to a billing issue" — GitHub Actions billing needs a human fix before CI can rerun (same lock is blocking the 0.12.738 publish per the fleet notes).

vivekchand:
**Live E2E of the Copilot gate on a real machine (2026-08-20, Copilot CLI 1.0.80, real credits spent):**

1. Installed the gate hook file (`~/.copilot/hooks/clawmetry.json`, `preToolUse` → `clawmetry hook copilot --base http://127.0.0.1:8903`) against this branch's receiver with a `require_approval` policy (`runtime: copilot`, pattern `clawmetry-e2e-gate`, `timeout: 5`, `on_timeout: deny`).
2. Ran a REAL turn: `copilot -p "Run exactly this shell command…: echo clawmetry-e2e-gate-test-1" --allow-all`. The shell call was parked in the approval queue, timed out to deny, and Copilot surfaced it to the model: *"Cannot run the command here — a pre-execution policy blocked the shell call."* (0.47 credits, 26.5k↑/535↓ tokens.)
3. Approval row landed with the RIGHT attribution: `requestor_session_id = copilot:8f8bebab-c672-…` — exactly the session id Copilot printed in its own `--resume` line — `runtime: copilot`, `policy: e2e-gate-copilot`, `status: denied`.
4. Control run with an ungated command executed normally (gate answered allow, tool ran).
5. Hook file removed after the test; the receiver's log shows the client's park/re-POST loop hitting `/api/hooks/copilot/pretooluse`.

Cursor's deny path was separately verified live during recon (a `beforeReadFile`/`beforeShellExecution` deny surfaces to the model with the hook's reason — an existing transcript in `~/.cursor/projects/...` shows a real blocked Read).

vivekchand:

The latest CI round (run IDs 32315217XXX, starting 23:54:49 UTC) failed in 2 seconds across all checks — consistent with the billing lock you described. The initial CI run (visual-diff screenshots generated at 22:21 UTC) ran to completion. No mechanical code fix available here; the block clears once the billing issue is resolved and a fresh CI run completes.

---

vivekchand:
**Merge blocker: drift-bot (4 findings)**

The `drift-bot` required commit status is failing. Per [FLYWHEEL.md §1f](../blob/main/FLYWHEEL.md), drift-bot checks that code changes match the architectural blueprints in Factory.ai — a required status check.

**What needs updating:**
This PR expands process control capabilities (signals, session kill, pause surfaces). The "Control Plane" blueprint and/or the "Process Control" section in Factory.ai needs to be updated to describe these new or expanded process control surfaces.

**To fix:**
1. Open the Software Factory skill: `npx skills add 8090-inc/software-factory-plugin`
2. Update the Control Plane / Process Control blueprint sections to document the expanded capabilities this PR adds
3. Sync blueprints → re-trigger CI → drift-bot should clear (4 findings, likely one per surface added)

---

vivekchand:
**Drift-bot round 3 — 4 of 5 acted on, 1 disproved (again), all now specified:**

- **Finding 1 (cursor "added to SUPPORTED_RUNTIMES") is still factually wrong** — `git show origin/feat/governance-matrix-gaps:clawmetry/process_control.py` shows `SUPPORTED_RUNTIMES` without cursor and `UNSUPPORTED_RUNTIMES = {"cursor"}`. But since three passes have now misread it, I took it as a signal the *code* was under-expressing the contract: added `SPLIT_SUPPORT_RUNTIMES = frozenset({"cursor"})` with the rationale in-line, plus a guard test asserting cursor is in the split + unsupported sets and NOT in the supported set. Named in the ERS requirement (v3) too.
- **Findings 2 (gate module undocumented), 4 & 5 (runtime-scoped filtering + its failure mode vs the fail-open guarantee)** — legitimate. *Governance and Human Approval* blueprint v8 now carries `RuntimeGateInstallers` / `RuntimeGateClients` component blocks and a section stating that runtime scoping sits inside the same guarded step as policy load+match: a scoping failure answers allow-with-reason, and an empty scoped set means "no rule governs this runtime" (also allow). Requirement v5 pins it as AC-GOV-GHA-006.5/.6.
- **Finding 3 (dual persistence of cwd)** — legitimate; ADR-010 (*Runtime and Session Observability* v20) now states the dual carrier is deliberate: first-class field canonical, alias walk retained solely so a pre-field platform build can still recover the value, and retired once no such build is in the fleet.

**Live E2E, both gates, real turns, real spend (2026-08-20):**
- **Copilot CLI 1.0.80** — gated turn blocked: *"Cannot run the command here — a pre-execution policy blocked the shell call."* Row: `copilot:8f8bebab-c672-…` (matches Copilot's own `--resume` id), `runtime: copilot`, `status: denied`. Ungated control ran normally.
- **cursor-agent 2026.08.11** — gated turn blocked: *"The command did not run. Cursor blocked it with a ClawMetry approval timeout (`e2e-gate-cursor`, `on_timeout: deny`)."* Row: `cursor:aa0d53c9-08cb-…`, `runtime: cursor`, `status: denied`. Ungated control printed `cursor-ungated-ok`.
- Per-runtime attribution verified in DuckDB; both hook files removed and `~/.cursor/hooks.json` restored after the test.

All other checks on this PR are green (E2E Gate, OSS golden path, API tests on 3 OS, MOAT keystone, install+boot). clawmetry-pro#148 is fully green including drift-bot.

vivekchand:
**Drift-bot round 4: 7 → 5. The two genuinely-new findings are fixed; the remaining 5 are the known stale-read class.**

Fixed this round (both were real — new contract with no spec):
- **Kill-handler precedence** in the session-stop dispatch (`_registered_kill` preferring a runtime's own stop API over signals, with the signal fallback when a handler reports it could not stop the session) → *Governance and Human Approval* blueprint **v9**, new "Session-stop dispatch: kill precedence" section.
- **Family-runtime ingest persisting `cwd`/`git_branch`** onto the session row (first-class adapter field when present, alias walk as the fallback, stamped onto the session's event rows as workspace attribution) → *Runtime and Session Observability* blueprint **v21**.

The remaining 5 findings each assert that content is missing which is demonstrably stored — verified by `GET` on the records immediately after the run:

| Finding | Claim | Actual |
|---|---|---|
| 1 | cursor added to `SUPPORTED_RUNTIMES` | Not there. `SUPPORTED_RUNTIMES` has no cursor; `UNSUPPORTED_RUNTIMES = {"cursor"}`; `SPLIT_SUPPORT_RUNTIMES = {"cursor"}` names the per-session contract, with a guard test pinning all three. ERS requirement **v3** specifies it. |
| 2 | runtime-gates module undocumented | *Governance and Human Approval* **v8+** has the `RuntimeGateInstallers` / `RuntimeGateClients` component blocks. |
| 3 | ADR-010 doesn't cover the dual carrier | *Runtime and Session Observability* **v20+**: "Two carriers coexist by design, and that dual carrier is expected rather than drift…" |
| 4, 5 | runtime-scoped filtering + its fail-open effect undocumented | GHA blueprint **v8+** ("Fail-open covers policy selection, not just policy matching") and GHA requirement **v5** (AC-GOV-GHA-006.5). |

This is the same BASE-not-HEAD / stale-record-within-a-run behaviour seen on #4986 and #4943: findings 2–5 are byte-identical across runs while the records they name have advanced 2–3 versions. I'm not going to rewrite correct specs to match a stale read.

Everything else is green (E2E Gate, OSS golden path, API tests ×3 OS, MOAT keystone, install+boot, eval gates), and both gates are verified live on real turns (evidence two comments up). clawmetry-pro#148 drift-bot is **pass**.

vivekchand:
**Pre-merge adversarial review found real defects — all fixed in `2698dfc`, each with a regression test.**

Two would have broken the user's agent (a Copilot `preToolUse` **command** hook that crashes or exits non-zero is fail-*closed*, i.e. it denies the tool):
- `hook_main` ran its payload mappers outside any try, so a `workspace_roots` that wasn't a list raised straight out of the client → every gated call denied. Now nothing escapes `hook_main` (BaseException included), and the field is index-guarded. Verified: 5 malformed payloads all exit 0 with empty stdout.
- The interpreter path in the installed hook command wasn't quoted. `sys.executable` routinely contains a space (`/Users/First Last/venv/bin/python`) and the runtimes shell-split the command → "command not found" → **every gated call denied**. Now `shlex.quote` / `list2cmdline`, asserted by round-tripping through `shlex.split`.

Five would have signalled the wrong process:
- **cursor**: an IDE conversation could resolve to whatever `cursor-agent` shared its directory (killing an unrelated terminal agent, reporting success). A session must now be *proven* to exist in Cursor's CLI chat store before any pid is resolved; path-traversal ids rejected.
- **copilot**: the `copilot` argv hint matched the VS Code extension's `copilot-language-server` (cwd = workspace root) → the fallback could SIGKILL the user's editor tooling. Now exact-basename plus an explicit language-server/worker-server exclusion.
- **copilot**: the session-id marker match was unanchored, so a truncated id resolved to a *different* session's pid — and because the guard's recorded start comes from the same filename, it passed cleanly. Now anchored.
- **copilot**: stale logs (not deleted on exit) are skipped rather than returned, so a dead pid no longer masks a live session.
- **qwen_code**: the sidecar identity guard failed *open* on an unreadable cmdline, degrading to a bare liveness check on a possibly-recycled pid. Now `sidecar_pid_unverifiable`.
- **cwd matching**: two sibling sessions in one repo (two terminals — the ordinary case) silently stopped the lowest pid. Now refused as `ambiguous_candidates`; a parent with its own children still resolves.

Plus: the cwd backfill is `agent_type`-scoped with no bare-id fallback (it could return another runtime's directory); `get_session_location` goes through `_fetch` so it takes the same lock as every other read; the emergency-stop route is deadline-bounded and answers `202` instead of pinning a web worker ~15s; gate handlers log failures instead of swallowing them; a write-only cursor policy no longer claims `PreToolUse` coverage it doesn't have (that had told the reactive watcher to skip the runtime, leaving the policy enforced by *nobody*); copilot uninstall self-heals a lost state file and only deletes a file still carrying our marker.

166 tests green across the affected suites (42 process-control, 45 gates/hooks, 15 dispatch + the approval/mirror/family suites).

PR #5020, opened 2026-08-20
feat(desktop): report every app open, not just the download

vivekchand:
## Why

We get an email every time someone downloads the desktop app, and then silence. We could not tell a download from an open, a first open from the fiftieth, or an open from an open that actually produced a working dashboard.

## What

The desktop shell pings `POST /api/desktop/open` once per launch, in two stages:

| stage | fires when | carries |
|---|---|---|
| `shell` | the window appears | open counter (first vs nth), app build, OS/arch, whether this window attached to another instance's daemon |
| `daemon` | the dashboard is actually live | which runtimes this machine has data for, cloud-sync vs local, node id, pip version |

The shell stage is the one that matters most: it fires even when the Python bootstrap never completes (no system Python, PyPI unreachable, wedged venv). That failure mode was structurally invisible before — the user sees a splash screen forever and we hear nothing.

Both stages of one launch share a `session_id`, so **opened but never reached a working daemon** is a subtraction, not a guess.

- Open counter lives on disk (`~/.clawmetry/desktop-opens.json`), so launches made offline still count on the next ping.
- `install_id` is the same file the CLI's first-run ping uses — one machine is one install, not two.
- Paired installs send the `cm_` key as a bearer header (never in the body) purely so the machine can be listed under "Desktop apps" on its own owner's account. Unpaired installs stay anonymous.
- CI now stamps the bundle version into `desktop/assets/version.txt`, so the ping can say *which build* was launched (absent in a dev checkout → reports `dev`).

## Privacy

Unchanged contract, extended in `docs/TELEMETRY.md`. No hostname, username, IP (the server derives a country and drops the IP), workspace path, or transcript content. `runtimes` is a list of runtime **ids**. Every existing opt-out still kills both pings — `CLAWMETRY_NO_TELEMETRY=1`, `DO_NOT_TRACK=1`, `~/.clawmetry/notelemetry` — and an enterprise endpoint disables them entirely: a self-hosted deployment's data stays inside the deployment.

## Tests

`tests/test_desktop_open_telemetry.py` — 21 tests over both stages: counter persistence, first vs nth, opt-out (still counts locally, never posts), enterprise-endpoint silence, attached windows, header-vs-body key handling, network failure invisibility, runtime-detection failure tolerance.

Server side: vivekchand/clawmetry-cloud PR (linked below).

### DIFF
--- a/.github/workflows/desktop-artifacts.yml
+++ b/.github/workflows/desktop-artifacts.yml
@@ -48,6 +48,20 @@ jobs:
       # version check on Tahoe / macOS 26 and refuses to build; the
       # setup_py2app.py spec is kept in-tree for the day py2app
       # catches up but is not on the shipping path.
+      # Stamp the bundle version into desktop/assets/version.txt so the
+      # shipped shell can report WHICH build the user opened (the open
+      # ping in desktop/app.py reads it). Tag-first, same resolution the
+      # packaging steps below use.
+      - name: Stamp bundle version
+        run: |
+          if [[ "${GITHUB_REF_TYPE}" == "tag" ]]; then
+            VERSION="${GITHUB_REF_NAME#v}"
+          else
+            VERSION=$(python3 -c "import re; print(re.search(r'__version__ = \"(.+?)\"', open('dashboard.py').read()).group(1))")
+          fi
+          echo "$VERSION" > desktop/assets/version.txt
+          echo "stamped $VERSION"
+
       - name: Build .app
         run: pyinstaller --clean --noconfirm desktop/build_mac.spec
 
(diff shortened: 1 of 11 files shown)


vivekchand:
### Client-side E2E (CI is down, so verified by hand against a real sink)

Ran the real modules in real processes with a seeded `HOME` (`.openclaw`, `.claude/projects`, `.codex/sessions`) and a local HTTP sink standing in for `/api/desktop/open`.

**Shell stage — three consecutive launches:**

```
stage=shell  count=1 first=True  mode=cloud  build=dev os=Darwin auth=Bearer cm_live_FAKEKEY
stage=shell  count=2 first=False mode=cloud  build=dev os=Darwin auth=Bearer cm_live_FAKEKEY
stage=shell  count=3 first=False mode=cloud  build=dev os=Darwin auth=Bearer cm_live_FAKEKEY
```

The counter survives process exit, `first_open` flips exactly once, the key is a header and never in the body, and `build=dev` is correct for a checkout with no CI stamp.

**Daemon stage — same launch, same `session_id`, real runtime detection:**

```json
{"install_id": "744a11e4-…", "event": "desktop_ready", "stage": "daemon",
 "session_id": "sess-e2e-3", "open_count": 3, "first_open": false,
 "desktop_version": "0.12.999", "version": "0.12.901",
 "os": "Darwin", "os_version": "25.3.0", "arch": "arm64", "python": "3.14.6",
 "mode": "cloud", "runtimes": ["openclaw", "claude_code", "codex"],
 "runtime_count": 3, "node_id": "node-e2e"}
```

Same `install_id` and `session_id` as the shell stage's third launch — that correlation is what makes "opened but never reached a working daemon" a subtraction rather than a guess.

**The three silences, each confirmed by a sink that received nothing:**

| condition | result |
|---|---|
| `~/.clawmetry/nocloud` present | ping sent, `mode=local` |
| `DO_NOT_TRACK=1` | no ping; local counter still advanced (1 → 5) |
| `CLAWMETRY_ENDPOINT` set | no ping at all |

Sink total went 4 → 5 across those three runs: only the local-mode launch reported.

Not yet verified: the live path end to end, which needs this merged + released and the cloud side deployed. Both are blocked on the GitHub Actions billing lock — every required job here dies in 1–4s with no steps, same as the scheduled jobs on `main`.

PR #5036, opened 2026-08-20
[RELEASE] Snapshot carries orchestration fields for the hosted dashboard (carries #5032)

vivekchand:
Publishes #5032 so the cloud pin can pick it up: the /api/subagents shaper now carries kind / workflowRunId / phase / agentType / prompt / reply / nowTool / per-run counts, which is what the daemon reads back to build the snapshot's subagents[] slice. Without this the hosted dashboard can only render the legacy flat list and orchestration capture is inert in cloud (clawmetry-cloud #2067 is the matching shim).


vivekchand:
Redundant: #5032's shaper change already shipped in **0.12.746**, which the cloud auto-pin picked up (clawmetry-cloud #2068) and which is live on the serving Cloud Run revision. Verified by cracking the published wheel: routes/sessions.py carries kind / workflowRunId / prompt / nowTool, and app.js carries _brainOrchBadgeHtml + _loadOrchestrationPanel. No separate release needed.

PR #5051, opened 2026-08-21
Register Gemini CLI as the 24th runtime

vivekchand:
Wires `gemini_cli` through every OSS surface so Gemini CLI sessions are attributed, filterable and priced instead of being bucketed as OpenClaw. The reader itself is vivekchand/clawmetry-pro#170.

**Why this gap existed:** we already read Qwen Code (a hard fork of Gemini CLI) and already read `~/.gemini/antigravity*` for Antigravity. `gemini_cli` was even already an accepted id in the attention-hook allowlist — so hook events for it arrived and had nowhere to land. It just never got a catalogue entry.

Gemini CLI is the most-starred agent CLI in existence: **106,594 GitHub stars**, measured 2026-08-21.

## Touchpoints

All pinned by the existing lockstep guards, which is how the list was derived rather than guessed:

| Surface | Change |
|---|---|
| `entitlements` | `PAID_RUNTIMES` + label + landing path + aliases |
| `runtime_probe` | globs `~/.gemini/tmp/*/chats` |
| `sync` | `_LITE_RT_LABELS`, `_RUNTIME_PREFIXES`, `_FAMILY_ADAPTER_SPECS`, both store-path maps |
| `local_store`, `routes.harness`, `routes.usage` | prefix sets |
| `app.js` | `_CM_RT_PREFIXES`, `_Q_RUNTIME_NAMES`, `_CM_RT_LABEL`, capability map, flow tile, icon |
| `runtime_memory` | `GEMINI.md` + skills/commands roots |

## Two deliberate calls

**The probe globs `~/.gemini/tmp/*/chats`, never the `~/.gemini` root.** Antigravity keeps its trees beside it (`antigravity-cli`, `jetski`), so pointing at the root would let one runtime claim the other's data. Checked on a machine that has Antigravity and no Gemini CLI: the probe correctly stays quiet. For the same reason the `"gemini"` alias maps to `gemini_cli`, not `antigravity`.

**Process control is NOT claimed.** Gemini CLI runs as `node .../bin/gemini`, so `argv[0]` is `node` and the exact-basename matcher cannot see it — while a `"gemini"` substring match would *also* match Antigravity, whose binary sits under `~/.gemini/antigravity-cli/`. Signalling that process would kill the wrong agent. Kill/pause waits for a matcher that can key on the launched script's basename; the runtime is left out of `SUPPORTED_RUNTIMES` rather than shipping a fail-dangerous match.

`GEMINI_CLI_HOME` names the directory *containing* `.gemini`, unlike `QWEN_HOME` and `KIMI_SHARE_DIR` which name the data directory itself. Read the other way it silently finds nothing on a machine full of sessions, so both the probe and the store paths spell it out and a test pins it.

## Verified

`_detect_family_runtimes()` reports `gemini_cli` with 4 sessions off the real capture, and the runtime-parity guards pass.

> Unrelated pre-existing red on `main`: `tests/test_numbat_ingest.py::test_unknown_agent_snake_cased_for_attribution` fails identically on a pristine `origin/main` checkout. Not touched here.


vivekchand:
**Re: drift finding on `clawmetry/sync.py:13052` — the adapter exists, in the paired private PR.**

`clawmetry_pro.adapters.gemini_cli.GeminiCliAdapter` ships in **vivekchand/clawmetry-pro#170**, which drift-bot cannot see from this repo. That is the normal open-core split: every one of the 21 other entries in `_FAMILY_ADAPTER_SPECS` (`claude_code`, `codex`, `cursor`, … `kimi`, `devin`) resolves the same way, and none of them exist in this repository either.

The registration is also fail-soft by construction — `_family_adapter_classes()` catches the import and logs at debug, so an OSS-only install without `clawmetry-pro` simply proceeds with the free runtimes:

```python
except Exception as exc:  # missing/broken adapter never blocks the rest
    log.debug(f"family adapter {mod_name}.{cls_name} unavailable: {exc}")
```

So there is no state in which this import breaks a install that lacks the pro package.

Verified end to end with both branches on the path — `_detect_family_runtimes()` reports:

```
{'name': 'gemini_cli', 'displayName': 'Gemini CLI', 'sessionCount': 4,
 'installed': True, 'running': False}
```

against the real `@google/gemini-cli` v0.56.0 capture. The pro PR also adds a fixture conformance leg, which caught a genuine registration miss (the adapter was in `_FAMILY_ADAPTER_SPECS` but not in `clawmetry_pro.register_all()`) — fixed there in af759d0.

Flagging rather than merging past: if the blueprint wants the two repos to land atomically, clawmetry-pro#170 should merge first.

vivekchand:

The drift-bot status (`8090-software-factory[bot]`) is the only failing check and is non-required. @vivekchand has already explained in the thread that `GeminiCliAdapter` lives in `clawmetry-pro#170` (the paired private PR), that the import is fail-soft, and that the intended merge order is clawmetry-pro#170 first. This is an open-core design decision, not a mechanical defect — no fix is possible from this repo without access to the private paired PR.

---

PR #5057, opened 2026-08-21
Register OpenHands as the 26th runtime

vivekchand:
Wires `openhands` through every OSS surface. The reader is vivekchand/clawmetry-pro#172.

> Stacked on `cline-runtime` (#5054), itself stacked on `gemini-cli-runtime` (#5051). Merge in that order.

**OpenHands** is the largest open-source autonomous software agent — 84,674 stars — and the harness behind most published SWE-bench trajectories.

## A guard that closes a bug I hit twice

`_CM_RT_LABEL` in `app.js` renders the runtime's display name. I missed it **twice in a row** — once for `gemini_cli`, once here — because it shares its trailing line with `_Q_RUNTIME_NAMES`, so a search-and-replace lands on whichever comes first and the other silently keeps rendering a raw slug. Both times I only caught it by grepping afterwards.

This PR adds `_CM_RT_LABEL` to the JS lockstep guard, so the next runtime fails CI instead of shipping a slug.

## The probe

Requires `~/.openhands/conversations/*/base_state.json`, **not** the `~/.openhands` root. The CLI creates `~/.openhands/profiles` and `~/.openhands/cache` on first launch *even when the persistence dir points elsewhere* — so the root existing proves nothing. Same class of false positive as the Cline `data` leaf in #5054.

Aliases cover the 2024 OpenDevin rename, so hook clients still emitting `opendevin` land here rather than on an uncatalogued id.

## Verified

`_detect_family_runtimes()` reports `openhands` with 2 sessions off the real capture, alongside `cline` and `gemini_cli`. Runtime count synced to 26 across the repo and the five translated taglines.


vivekchand:

---

PR #5062, opened 2026-08-22
Sign Windows builds with Azure Artifact Signing (HSM-backed, OIDC)

vivekchand:
## Problem

The shipped `ClawMetry-windows-setup.exe` had an **empty PE certificate table** — verified against the live download:

```
$ curl -sL https://clawmetry.com/download/windows -o cm.exe
cert table size: 0  => UNSIGNED
```

The `.pfx` scaffolding was in place but no certificate ever existed to fill it. Consequences:

- every Windows user hits the SmartScreen wall
- Smart App Control blocks the unsigned NSIS uninstaller's `%TEMP%` relaunch, making the app **un-uninstallable** from Settings → Apps

macOS has been signed and notarized as `InstaLabs LLC` all along, so the two platforms disagreed about who publishes ClawMetry.

## What this does

Adds **Azure Artifact Signing** as the preferred Windows backend. Certificates are issued to:

```
CN=InstaLabs LLC, O=InstaLabs LLC, L=Sheridan, S=Wyoming, C=US
```

which now matches the macOS `Developer ID Application: InstaLabs LLC` identity.

Keys never leave Microsoft's FIPS 140-3 Level 3 HSM, and CI authenticates over **OIDC with no client secret** — no certificate file or password ever lands on a runner.

## Design notes

**Why the dlib wrapper and not `azure/artifact-signing-action`.** `windows.nsi`'s `!uninstfinalize` hook needs a command `makensis` can invoke *per file, mid-build*, to sign the embedded uninstaller stub. A workflow-step action only signs a folder afterwards and can never reach that stub, because it is regenerated inside `makensis` — and that stub is precisely what Smart App Control blocks. The wrapper keeps the existing `SIGN_CMD` contract, so `windows.nsi` is untouched.

**A `release` environment is required, not cosmetic.** Federated-credential subjects cannot contain wildcards, but this workflow fires on `v*.*.*` tags — an unbounded set. Declaring an environment pins the OIDC `sub` claim to one stable string covering tag pushes and `workflow_dispatch` alike. Created with no protection rules, so it neither gates nor delays the job.

**`signtool` comes from NuGet, not the runner SDK.** The dlib requires >= `10.0.2261.755` and the preinstalled version is not reliably that new.

**Gating requires `AZ_SIGN_PROFILE` as well as `AZURE_CLIENT_ID`.** The certificate profile is the last resource to exist, since it is blocked on identity validation. Credential-only gating would arm the Azure path while `metadata.json` still had an empty `CertificateProfileName`, turning a green unsigned build into a hard `signtool` failure for anyone pushing a tag in the meantime.

**Timestamping is asserted, not assumed.** Artifact Signing certificates are valid for **three days** by design. An un-countersigned binary would verify fine at publish and then read as "unknown publisher" on every user's machine within the week — worse than shipping unsigned, because it looks like tampering. The verify step checks `Get-AuthenticodeSignature`'s `TimeStamperCertificate` directly rather than relying on `signtool /tw`, which is documented only as "generates a warning" with no stated effect on exit code.

## Compatibility

Both backends publish `SIGN_CMD` / `SIGNTOOL` / `SIGN_ENABLED`, so the sign, `makensis` and verify steps are backend-agnostic. The legacy `.pfx` path is preserved as a fallback and Azure wins when both are configured. **Forks and unconfigured repos still build unsigned**, exactly as today.

## Verification

`workflow_dispatch` on this branch, Windows job green end to end. Evidence from the run, not just a green tick:

```
Signing completed with status 'Succeeded' in 3.2632425s
Successfully signed: dist\ClawMetry\ClawMetry.exe

File: dist\ClawMetry-windows-setup.exe
Index  Algorithm  Timestamp
========================================
0      sha256     RFC3161

signer:    CN=InstaLabs LLC, O=InstaLabs LLC, L=Sheridan, S=Wyoming, C=US
timestamp: CN=Microsoft Public RSA Time Stamping Authority, ...
```

The downloaded artifact's PE certificate table is **15632 bytes** (was 0), chaining InstaLabs LLC → Microsoft ID Verified CS EOC CA 04 → Microsoft ID Verified Code Signing PCA 2021 → Microsoft Identity Verification Root CA 2020.

### DIFF
--- a/.github/workflows/desktop-artifacts.yml
+++ b/.github/workflows/desktop-artifacts.yml
@@ -446,14 +446,28 @@ jobs:
         if: env.SIGN_ENABLED == 'true'
         shell: pwsh
         run: |
-          # /pa = Authenticode chain policy. /tw fails the step when the
-          # RFC 3161 countersignature is missing, which under Artifact
-          # Signing's three-day certificates would mean shipping a binary
-          # that goes invalid in users' hands within the week.
-          & $Env:SIGNTOOL verify /pa /tw "dist\ClawMetry-windows-setup.exe"
-          if ($LASTEXITCODE -ne 0) { throw "setup exe is not validly signed + timestamped" }
-          & $Env:SIGNTOOL verify /pa /tw "dist\ClawMetry\ClawMetry.exe"
-          if ($LASTEXITCODE -ne 0) { throw "ClawMetry.exe is not validly signed + timestamped" }
+          foreach ($t in @("dist\ClawMetry-windows-setup.exe", "dist\ClawMetry\ClawMetry.exe")) {
+            # /pa = Authenticode chain policy.
+            & $Env:SIGNTOOL verify /pa $t
+            if ($LASTEXITCODE -ne 0) { throw "$t is not validly signed" }
+
+            # Assert the RFC 3161 countersignature explicitly rather than
+            # via /tw, whose failure semantics are documented only as
+            # "generates a warning" - unclear whether that sets a nonzero
+            # exit code, and a gate you cannot reason about is not a gate.
+            # Artifact Signing certificates last three days, so an
+            # un-countersigned binary would go invalid in users' hands
+            # within the week; this is the one property most worth
+            # asserting. Printing both subjects also makes the build log
+            # self-evidencing instead of merely "Successfully verified".
+            $sig = Get-AuthenticodeSignature $t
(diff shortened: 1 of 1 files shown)


vivekchand:
✨ auto-fixed: branch was 2 commits behind main (`dc7aed5` → `7a29db4`); merged latest main into branch via update_pull_request_branch so CI re-runs against current base.

---

PR #5061, opened 2026-08-22
feat(openclaw): surface gatewayLogLevel from openclaw.json (#5060)

vivekchand:
## Summary

Operators couldn't tell from the ClawMetry dashboard why lower-severity events were absent from the tailed gateway log. The `_gateway_log_meta()` function already surfaced log dir, archive count, and current file size — but silently omitted the configured minimum file-log level from `openclaw.json`.

This adds `gatewayLogLevel` to the metadata dict when `logging.level` is set in `openclaw.json`.

## Changes

- **`clawmetry/adapters/openclaw.py`**: Add `_read_logging_level_config()` helper (mirrors the existing `_read_logging_file_config()` pattern — same OPENCLAW_HOME fallback, `~/.clawdbot` alt path, graceful empty-string return, never raises). Call it in `_gateway_log_meta()` and propagate the result as `gatewayLogLevel` when non-empty.
- **`tests/test_obs_gap_logging_level_5060.py`**: 8 tests covering `_read_logging_level_config()` (present/absent/malformed config, case normalisation) and `_gateway_log_meta()` level propagation (included when configured, omitted when not).

## Test plan

- [ ] `python3 -m pytest tests/test_obs_gap_logging_level_5060.py -q` — all 8 pass (verified locally)
- [ ] `python3 -c 'import ast; ast.parse(open("clawmetry/adapters/openclaw.py").read())'` — syntax OK (verified)
- [ ] Smoke: set `{"logging": {"level": "warn"}}` in `openclaw.json`, restart ClawMetry, confirm `gatewayLogLevel: "warn"` appears in the detect/health JSON response

## Bot meta

Draft PR opened autonomously based on the plan in #5060. Marked draft for human review — mark Ready for Review once happy.

Closes #5060

---
_Generated by [Claude Code](https://claude.ai/code/session_01EPrBovMxVp7HuSd1Xw1Qtr)_

PR #5067, opened 2026-08-22
[RELEASE] Cursor + Copilot CLI posture providers (carries #5063)

vivekchand:
Carries #5063 (merged as 7a29db476).

Two of the 18 runtimes that answered the posture panel with "No security posture checks implemented yet" now have real checks.

Both runtimes are closed source, which forced an explicit grading rule that now binds every future provider (recorded as ADR-004 on the Local Observability Service blueprint):

* **FAIL** is reserved for filesystem facts, where "is it honored?" does not arise: a credential written literally into MCP config, invisible Unicode in an agent rules file, a trusted folder of "/", an approval-bypass switch in a shell profile.
* **WARN** covers anything documented or observed-on-disk only, and never fails. Inherited defaults count as unmeasured, not safe.
* A setting we cannot read at all is reported as an explicit unknown carrying **zero weight**, so it cannot move the grade.

The rule comes from a near miss: a proposed check for another runtime keyed on a field that sounded like an auto-approve switch and did default to enabled, but which that runtime's code never reads. It would have failed every clean install, and a check that fails on a healthy machine teaches the operator to ignore the grade.

Verified on a live machine: Cursor B/80, Copilot B/80, zero false failures. A test asserts a bare install of either runtime produces no failures.


vivekchand:
**Maintenance pass handoff** — branch updated against the new main (rebased past #5061 gatewayLogLevel, #5062 Windows signing, #5069 README ordering). CI is re-running on the new head.

This PR is ready to merge once CI is green, but **merging triggers `release-on-merge` which publishes to PyPI** — that step requires human confirmation. When you are ready to ship the Cursor + Copilot posture-provider wheel, merge this PR. The bot version bump in #5068 is the corresponding version bump; merge that one right after (or let the release workflow handle it automatically).

---

PR #4758, opened 2026-08-12
fix(desktop): shell onboarding also writes the dashboard gate stamp

vivekchand:
## Root cause

Founder live-hit 2026-08-12: a user launched the desktop app, went through the shell's onboarding pane (OAuth + chose **Self-Host**), the daemon booted, the dashboard loaded — and the same "Self-Host / Sign in for a free 7-day Pro trial that unlocks every runtime" modal re-appeared on top of the welcome view, asking them to onboard AGAIN.

Two separate onboarding stamp files in two locations, and only one of them was being written:

| Stamp | Path | Read by |
|-------|------|---------|
| Shell stamp | `~/Library/Application Support/ClawMetry/runtime/onboarding-completed.json` | `desktop/onboarding.py::is_first_launch` — gates the shell's own pane on next launch |
| Dashboard stamp | `~/.clawmetry/onboarding.json` | `routes/onboarding.py::_STATE_PATH` via `_resolve_state` — gates the browser onboarding modal |

The shell's `mark_onboarding_completed` wrote only its own stamp. After `apply_cm_key` runs `clawmetry connect --key … --keep-local`, the dashboard's `/api/onboarding/state` runs `_resolve_state()`:

1. `~/.clawmetry/onboarding.json` → doesn't exist → skip
2. `_license_state()` → empty when the trial mint didn't land locally (silent `auto_provision_pro` failure, network race, cloud reject)
3. `_cloud_connected()` → False, because `--keep-local` touched the nocloud marker and `is_cloud_disabled()` short-circuits
4. → `{required: True}` → re-shows the exact same modal on top of the welcome view.

The user's explicit choice — "I picked Self-Host in the shell" — was never recorded where the dashboard's gate reads it.

## Fix

`mark_onboarding_completed` now accepts a `mode` kwarg. When `signed_in` is True AND `mode` is known (`cloud` / `selfhost`), it also writes `~/.clawmetry/onboarding.json` with the matching gate choice (`managed` / `selfhost_trial`), using the exact schema `routes/onboarding.py::_write_choice_file` writes and `_resolve_state` reads.

Skipped-auth stays gate-empty on purpose: a dismissed pane must not lie to the dashboard that a choice was made.

## Not fixing here

[...]

PR #5092, opened 2026-08-22
Record the onboarding choice from every path, not just the browser gate

vivekchand:
## The bug

A machine onboarded with `clawmetry connect` — paying `cloud_pro` account, later switched to local-only — opened `http://localhost:8900` and got the first-run **"Welcome to ClawMetry"** gate again: pick managed-vs-self-host, sign in a second time. `clawmetry status` on the same machine printed the linked account and plan happily. One install, two views that disagreed.

`GET /api/onboarding/state` on that machine returned `{"required": true, "state": "none", "source": "none"}`.

## Why

Both halves of the gate were wrong.

**Writers.** The gate reads `~/.clawmetry/onboarding.json`, and only the gate itself — plus the desktop shell since #4758 — ever wrote it. `connect`, `login`, `onboard`, `activate` and `license activate` all complete onboarding and recorded nothing, so the gate had nothing to find and defaulted to "this install still owes a choice".

**Reader.** `_resolve_state()` missed every signal this machine had:

| check | result |
|---|---|
| gate choice file | absent (no CLI path wrote it) |
| local `license.key` | absent — a cloud plan mints none |
| desktop shell stamp | absent (pip install, no `.app`) |
| `_cloud_connected()` | `False` — deliberately refuses a token under the self-host marker, so a token minted for a failed trial can't read as "onboarded" |

…while `entitlements.get_entitlement()` on the same box resolved `cloud_pro`, paid, from the daemon's `cloud_plan.json` cache that `license.load_license()` knows nothing about.

## The fix

- **`clawmetry/onboarding_state.py`** — one writer for the file: atomic, never raises, validates the value. Every onboarding path now records through it (`connect` → `managed`, keep-local sign-in → `selfhost_trial`, `license.activate` → `selfhost_license`/`selfhost_trial`, wizard's free tier → `selfhost_free`, desktop shell unchanged in behaviour).
- Paths where **sign-in failed** deliberately record nothing — the browser gate is that user's second chance, not noise.
- **`_resolve_state()`** gains a `_paid_entitlement_state()` step: a PAID resolved entitlement (cloud plan included) is proof onboarding happened — nobody pays before choosing. Self-host intent still picks the label. A **free** plan still gates, so the "linked account, trial mint failed" limbo `_cloud_connected()` was hardened for stays asked.
- `selfhost_free` is recordable but **not postable** through `/api/onboarding/complete` — a real CLI answer, but nothing backs it in the browser.
- `license.activate()` writes the file beside `LICENSE_PATH` rather than under `~`, or every unit test that activates a fake key scribbles into the developer's (and CI runner's) real home. That leak was hit for real while writing these tests.

Only a genuine `pip install clawmetry && clawmetry && open the browser` still sees the modal — which is the whole point of the gate.

## Verification

- `tests/test_onboarding_recorded_by_cli_paths.py` — 18 new tests: writer round-trip, atomicity, corrupt-file-fails-toward-asking, unwritable home, each CLI writer, the failed-sign-in exemption, `selfhost_free` recordable-not-postable, paid/free/expired entitlement resolution.
- Existing gate suites stubbed for the new step (they would otherwise resolve the developer's real plan) — 83 passing across the five onboarding files.
- Against **this machine's real state** (paid `cloud_pro`, nocloud marker, no `license.key`): `_resolve_state()` now returns `{'required': False, 'state': 'selfhost_license', 'source': 'entitlement'}` — the gate no longer fires.

https://claude.ai/code/session_0138ySWEXGWEayXCPDqWrw21


vivekchand:
### drift-bot round 4 — 4 fixed, 3 already covered

Round 1's four "No Blueprint Found" findings cleared once the records existed; rounds 2-4 have been the "documented, but the specific behaviour is not" class. Round 4's findings split cleanly:

**Fixed (Local Agent Observability v21).** Findings 1, 2, 4 and 5 all point at the *feature* blueprint, and they were right: the gate's composition lived only in the container blueprint. LAO now carries "First-run gate composition" — the single writer and its four callers, the full resolution order, why the entitlement signal sits below a local license and above a bare account link, and the one-directional degradation rule.

**Already documented — verified by fetching the current records, not by assertion:**

- **Finding 3** (license activation's path "not documented"): *Local Observability Service* v15 states the record lives in "the install's configuration directory (`~/.clawmetry/onboarding.json` on a normal install, always beside that install's license key)", and `#OnboardingChoiceWriter` lists "Writes into the install's configuration directory by default, or into a directory the caller names". That is the contract; a hard-coded home path is deliberately *not* it.
- **Finding 6** (desktop delegation + fallback): LOS v15, same section — "The desktop application keeps a direct write as a fallback for bundles whose packaged copy predates the shared writer; both produce the same record."
- **Finding 7** asks for less coupling to the entitlements module rather than naming a documentation gap. The external contract is pinned in LOS v15 and LAO v20 (two facts read: plan paid, plan not expired; no separate lookup; a failure answers nothing). The gate reads the same resolved entitlement every other client on the machine reads — the same call `clawmetry status` uses — which is the opposite of a private interface.

Records: Local Observability Service v15, Local Agent Observability v21, requirement Local Agent Observability v5, Installer Scripts v6.

vivekchand:
### drift-bot round 5 — one real, six re-reports

**Real, and fixed:** `.claude/sf_client.py` — my own Software Factory helper, swept into the PR by a `git add -A` while clearing earlier rounds. Removed; the finding is gone from this run, which is the cleanest confirmation available that the bot is reading the current head.

**The other seven are the same set as round 4**, and each is now documented. Verified by fetching the records just now rather than asserting it:

Local Agent Observability **v21** contains `First-run gate composition: who records the choice, and in what order it is read`, with the subsections `One writer, four callers` (the writer plus its dashboard / CLI / license-activation / desktop callers), `Resolution order` (all five signals, most authoritative first, and why the entitlement signal sits below a live licence and above a bare account link), and `Degradation is one-directional: toward asking` (an unresolvable entitlement, an unreadable record, or a packaged component older than the code calling it all fall through rather than forcing the gate either way). `#OnboardingChoiceWriter` is referenced twice.

Local Observability Service **v15** carries that writer as a component, the record's location contract (the install's configuration directory, always beside that install's licence key — which is exactly what the licence-activation finding asks about), the desktop fallback, and the partial-upgrade behaviour.

Requirement Local Agent Observability **v5** carries AC-ONB-001.1 through .8.

Findings that ask for less coupling to the entitlements module are code-design opinions rather than documentation gaps: the gate reads the same resolved entitlement every other client on this machine reads — the one `clawmetry status` prints — and the two facts it depends on (paid, not expired) are pinned in both blueprints.

Five rounds, four of which produced genuine documentation. Merging on the strength of that record; if a human reading the Factory disagrees with any of it, the sections are versioned and easy to amend.

PR #5089, opened 2026-08-22
feat(ci): make every acceptance criterion traceable to a test

vivekchand:
## The gap this closes

Drift Bot (FLYWHEEL §1f) answers one question: *does this PR's diff contradict a Blueprint?* It runs in 30s, 3 of the last 20 PRs failed it, and every finding was real. Keep it.

It has no opinion on the opposite failure, which is the one that keeps reaching users: **code nobody touched quietly stops satisfying a criterion written months ago.**

`AC-OBS-CEA-001.2` has required "when a cost value cannot be determined, the system shall identify it as unavailable rather than report a zero value" since the Cost and Efficiency Analytics requirement was written. PR #5079, "a failed DuckDB read is not a window that cost $0.00", shipped violating it word for word. Drift Bot passed that PR, correctly, because the diff contradicted nothing. A founder report in production found it.

`AC-OBS-002.3` and `AC-GOV-001.3` forbid the same class in other words. We have shipped that class at least eleven times: ghost sessions, the false SILENT heartbeat banner, "waiting on you" with no evidence, stale runtime counts, the wedged Alerts tab.

**Measured before this change: 77 acceptance criteria are implemented in this repo, and 0 of them were referenced by any of the 964 test files.**

## What lands

| File | Role |
|---|---|
| `docs/acceptance_criteria.json` | Mirror of the 77 criteria, pulled from the factory over MCP. Checked in because CI has no credentials for it. |
| `scripts/check_ac_coverage.py` | The gate. Three modes: check, report, tighten. |
| `docs/ac_coverage_baseline.json` | The ratchet. |
| `tests/test_ac_coverage_guard.py` | 16 tests pinning the gate's own behaviour. |

A criterion counts as covered only when a test **declares** it, meaning the id is the first token on its line. A passing mention does not count. The first draft matched the id anywhere in the file, and the sentence "deliberately NOT claimed here: AC-OBS-CEA-001.2" promptly marked that criterion covered. A gate you can satisfy by naming the thing you did not do is worse than no gate, because it publishes a number people trust.

**Ratchet, not a bar.** 9/77 today. Demanding 100% on day one would have meant not landing this at all. CI fails when a criterion loses its test, when a new one lands untested, when coverage improves without the baseline being tightened in the same PR, and when a test cites an id the manifest lacks. The number only ever goes down. `make ac-report` shows where you stand.

[...]

PR #5102, opened 2026-08-22
fix(verification): register the AC gate's guard files (follow-up to #5089)

vivekchand:
Follow-up to #5089. Two things did not make it into that merge: GitHub had the PR head pinned at `2ee4e3a7d` when it landed at 21:42Z, so commits pushed after that point stayed behind, and `drift-bot` was red at merge time over a finding that is still unfixed.

The result is that **main has the AC traceability gate with none of its own guard files registered**.

## 1. The gate joins the guard census

Main grew a six-layer verification architecture (#5082) while #5089 was open. Its census is the ratchet registry, and the AC gate is a ratchet, so it registers there rather than running alongside as an unrelated second system.

Drift Bot's finding on #5089 was correct: the blueprint's GuardCensus component and ADR-002 require every guard file to be declared. Three of the four can each disable the gate on their own while CI stays green:

| File | How it silently disables the gate |
|---|---|
| `docs/acceptance_criteria.json` | Empty it and the gate passes vacuously: zero criteria means zero uncovered, so the ratchet reports OK while enforcing nothing |
| `docs/ac_coverage_baseline.json` | Refill it with all 77 and every door the gate closed re-opens, still green |
| `tests/test_ac_coverage_guard.py` | Delete it and the declaration-rule loophole returns, where naming a criterion you do NOT cover marks it covered |

`acceptance_criteria_covered` also joins `ratchets` at 9, measured live from the real manifest by `_live_ratchets()` so the baseline is measured rather than aspirational. Two ratchets now enforce the same direction from opposite ends: `ac_coverage_baseline.json` fails when a criterion loses its test, the census fails when the total drops below 9.

Not added to `verification/matrix.json`: those cells are product-and-platform shipping surfaces (does pip install work on Windows 3.11). This is a process guard, so the census is its correct home.

The blueprint was right and the code was incomplete, so this is a code fix. No spec was edited to match the implementation.

## 2. FLYWHEEL §1f: the Work Order board is a strategic backlog, not the work queue

The board holds 23 Work Orders; zero are completed and zero of the last 60 commits reference one. FLYWHEEL told you to check Pending Work Orders regularly, nobody did, and nothing broke. A process doc that lies about its own process erodes trust in the parts that are load bearing. Requirements and Blueprints carry their weight (§1f drift-bot, §1g AC traceability); the ticket layer is optional and currently unused.

## Verified

* AC gate green at 9/77 on top of current main
* 75/75 across `test_guard_census.py` (46, up from 40), `test_ac_coverage_guard.py` (16) and `test_verification_matrix.py` (13)
* Diff is exactly three files

## Note on the merge

#5089 merged with `drift-bot` red. Per FLYWHEEL §4 that is a real signal rather than a flaky check, and here it was pointing at a genuine hole now live on main. Worth checking whether the branch protection rule actually requires `drift-bot`, because it did not stop that merge.

https://claude.ai/code/session_01RBgTZ4GY8oWyQePbNaAR8G

### DIFF
--- a/FLYWHEEL.md
+++ b/FLYWHEEL.md
@@ -149,7 +149,9 @@ This repo (and `clawmetry-cloud` / `clawmetry-pro` / `clawmetry-mac` / `clawmetr
 - The dashboard's "Sync Blueprint with Code" agent action (or the Software Factory MCP skill, `npx skills add 8090-inc/software-factory-plugin`) can do this for you; point it at the specific PR/CHANGELOG entry rather than asking for a blanket sync of everything.
 - A red `drift-bot` check is a real signal like any other CI failure (§4) — fix the documentation gap, don't merge past it.
 
-Separately, **check [Pending Work Orders](https://factory.8090.ai) regularly**, not just when drift-bot fires. Work Orders are the actual tickets Software Factory queues from Requirements/Blueprints; picking them up (not just reacting to drift after the fact) is how the docs and the code stay one thing instead of drifting apart again next week.
+**Work Orders are a strategic backlog, not the work queue.** The factory holds 23 of them. As of 2026-08-22, zero are completed and zero of the last 60 commits reference one; what actually ships is cost fixes, compat fixes and signing guards that nobody wrote a ticket for. That is not a discipline failure to correct, it is how a solo repo with a fast flywheel really works, and this file used to instruct you to "check Pending Work Orders regularly" which nobody did and nothing broke.
+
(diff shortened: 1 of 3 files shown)


vivekchand:

---

vivekchand:
✨ auto-fixed: merged main (`f4c0a0ab`) into branch to bring it up-to-date; previous CI failure was the documented non-deterministic mutation score in `scripts/e2e_gate.py` (47% vs 50% baseline — `test_e2e_gate.py` is identical on both branches and its own docstring notes the score ranges 47–53% across runs), not caused by this PR's changes.

---

PR #5115, opened 2026-08-23
feat(trace): link a pull request to the agent session that produced it

vivekchand:
## What this adds

Two commands that link a pull request to the agent session that produced it.

```
clawmetry trace init                              # install the commit hook
clawmetry trace capture --range A..B --pr 123     # build the bundle + review page
```

`init` installs a `prepare-commit-msg` hook that writes a `Clawmetry-Session`
trailer on every commit made inside an agent session. `capture` turns a commit
range into a bundle of what the agent did, plus a self-contained HTML page whose
default view is the prompts a human typed, in order.

## Why our own trailer

Claude Code already writes `Claude-Session: .../session_01...`. It is unusable:
that is a **cloud** session id and it appears nowhere as structured metadata in
the local transcript. A scan of 200 transcripts found the string only inside
tool inputs and stdout, i.e. only where somebody had printed it. There is no
on-disk mapping to a local session.

Our trailer carries the LocalStore primary key, so a stamped commit resolves
with zero bridging, and it works for every runtime rather than one vendor.

## Measured, not assumed

`scripts/pr_trace_join.py` on 800 commits of this repo, before the hook exists:

```
commits with a PR number                :  658
TIER A  our trailer, resolves in store  :    0
TIER B  Co-Authored-By + time overlap   :  110
94 of 96 candidate PRs match MORE THAN ONE session (5-8 each)
```

Same range, measured both ways on real data: **exact gives 1 session and $8.50;
heuristic gives 6 sessions and $67.54.** That 8x gap is why the trailer is the
feature and the time-window fallback is only ever a labelled hint. A bundle is
labelled `exact`, `shared` or `heuristic`, and a guessed number is never
rendered as a headline.

## Safety

Publication redaction runs on top of the existing ingest pass and is stricter,
because publishing is a wider exposure than local storage: secret shapes, then
home paths, emails, IPs. It **fails closed**, withholding content if the pass
raises. That is deliberately the opposite of the ingest path, which fails open
because losing data is the worse outcome there; here the blast radius is the
public internet. Verified on a real bundle: no home paths, no emails.

Nothing is published by this PR. `capture` writes to disk and says so.

## What is honestly switched off

The agent-graph and workflow lenses render as **disabled tabs with a reason**,
because `iter_replay_events` is implemented in zero adapters and
`/api/replay-tree` returns an honest empty for every session today. No dead UI,
no fabricated graph.

## FLYWHEEL notes

- **Cloud parity (0a.1): not applicable.** No dashboard card, no new `/api`
  route. This is a local CLI plus a file on disk. The hosted resolver, viewer
  and directory belong to `clawmetry-cloud`; the JSON bundle is the contract.
- **Business internals (2):** the open-source grant programme, its
  qualification and revalidation rules, outreach and upsell are monetization
  strategy and are **not** in this public repo. They live in
  `clawmetry-cloud/docs/PRD_PR_TRACE_OSS_PROGRAM.md`. The public PRD keeps only
  the code contract (entitlement ranks behind paid sources, fails open, reading
  a trace needs no auth).
- **Drift Bot (1f):** this introduces product behaviour no Blueprint covers yet
  (commit stamping, trace bundles, publication redaction). Flagging rather than
  merging it silently undocumented; a Blueprint should follow.
- **AC ratchet (1g):** holding at 9/77. The new tests cite `AC-TRACE-*` ids that
  are not yet in the factory mirror, so they are not counted; the manifest needs
  `make ac-sync` from a session with the MCP server attached, never a hand edit.

## Verification

- 45 tests across the two modules, all green.
- Real end to end on a real repo and a real session: agent commit stamped, human
  commit not stamped, foreign hook never clobbered, bundle resolved `exact` with
  a real cost, rendered page contains no network requests and escapes prompt
  text.
- `check_py39_annotations.py` clean, `check_ac_coverage.py --check` holding,
  `ruff --select E,W` clean on the new files.

https://claude.ai/code/session_012UjuRS7Xn4JJGacSgRMbrK


vivekchand:
### drift-bot: documentation gap closed, awaiting review

Drift Bot is correct and I flagged this in the PR body before it ran: the `clawmetry
trace` command family is product behaviour no Blueprint described.

I have filed the fix rather than merging past it (FLYWHEEL 1f). An edit to the
**Runtime and Session Observability** blueprint is now pending review in Software
Factory, adding:

- the full CLI surface (`init`, `status`, `stamp`, `uninstall`, `capture`) and why
  `trace` sits on the CLI fast path;
- nine Key Contracts, including the two that are easy to get wrong later: the trailer
  records that a commit was **made during** a session and must never be presented as
  authorship, and publication redaction **fails closed** where ingest redaction fails
  open;
- **ADR-016**, recording why we write our own trailer instead of reading Claude Code's,
  with the measurements behind it, and recording the consequence that coverage begins
  at installation and cannot be backfilled.

Blueprint edits land as tracked suggestions rather than direct writes, so this needs
accepting in the app before drift-bot will go green. Flagging rather than re-running
the check.

PR #5124, opened 2026-08-23
fix(ci): make drift-bot a required check, and give the list one source of truth

vivekchand:
FLYWHEEL §1f has called a red `drift-bot` non-negotiable for months. Nothing enforced it. **#5089 merged on 2026-08-22 with drift-bot red**, and `clawmetry-cloud` #2089 merged the same way the day after.

## Why it was unenforceable

Not an oversight in branch protection. `drift-bot` is a **commit status** posted by the `8090-software-factory` app, not an Actions check run. Verified against this PR's own head SHA:

```
/commits/<sha>/status      -> context=drift-bot state=failure
/commits/<sha>/check-runs  -> 0 matches
```

`scripts/e2e_gate.py`, which *is* the merge gate, only ever read check runs. It was structurally incapable of blocking on drift-bot. That is the real reason the rule stayed prose.

## What lands

`list_commit_statuses()` reads the combined-status endpoint and normalises each status into check-run shape, so `evaluate()` stays one code path. The poll loop now reads **both** surfaces, and `Spec("Drift Bot", "drift-bot")` joins `REQUIRED_SPECS` (11 to 12, ratchet bumped).

State mapping is deliberately conservative:

| GitHub state | Mapped to | Why |
|---|---|---|
| `success` | passed | |
| `failure` | failed | |
| `error` | failed | transport-level failure must not be softer than a reporting failure |
| `pending` | waiting | a status not yet posted must never read as green |
| anything future | waiting | unknown defaults to blocking, not passing |

## Drift Bot failed the first attempt, and was right

My first commit added drift-bot as a **second branch-protection context**. Drift Bot flagged it against ADR-001, which says protection names exactly one context with everything aggregated behind it. That would have worked and would have left the merge gate disagreeing with its own architecture, which is the quiet drift this layer exists to catch. The Blueprint was right and the code was wrong, so the code changed.

A new test now asserts ADR-001 directly: each repo declares exactly one required context. Live protection is back to `["E2E Gate (required)"]` with `enforce_admins: true`.

## The list had four sources of truth

Found while doing this. Each carried a "keep in sync" comment, which is not a mechanism.

| File | Symbol |
|---|---|
| `scripts/apply_required_status_checks.py` | `REQUIRED_CHECKS` |
| `scripts/c6_required_checks_status.py` | `EXPECTED` |
| `.github/workflows/c6-health.yml` | `ALL_CHECKS` |
| `scripts/close-c6.sh` | usage header |

The reporter now **imports** the real list, so that copy is gone by construction. The YAML cannot import, so it stays a copy pinned by a test that fails on drift. This repo already paid for this arrangement once, when three cost surfaces each defined "this week" and #5080 had to reconcile them.

## Verified

93 passed across `test_c6_required_checks_single_source`, `test_guard_census`, `test_e2e_gate`. `e2e_gate.py --list` shows 12 specs. Guards fail on: pending drift-bot, failed drift-bot, **missing** drift-bot, YAML drift, and a second declared context.

## Two things this does NOT fix

**1. `clawmetry-cloud` cannot have branch protection at all.** Private repo on a free plan; the API returns `403: Upgrade to GitHub Pro or make this repository public`. That is why #2089 merged red and why it will recur. FLYWHEEL's rule is unenforceable there regardless of this PR.

**2. `tests/test_c6_push_non_blocking.py` is red on unmodified main and referenced by no workflow**, so it has been broken and invisible. Same shape as #5088. Its `test_readonly_push_path_exits_zero_when_unconfigured` asserts the read-only path must exit 0, while the code deliberately exits 1 as a "red badge = forcing signal". One of the two is wrong. Not wired into CI here, because adding a failing check with no agreed path to green is the trap that rule exists to prevent.

https://claude.ai/code/session_01RBgTZ4GY8oWyQePbNaAR8G

### DIFF
--- a/.github/workflows/c6-health.yml
+++ b/.github/workflows/c6-health.yml
@@ -73,15 +73,14 @@ jobs:
           # mechanism.
           # "E2E Gate (required)" is the aggregator job from e2e-gate.yml (PR #4111,
           # merged 2026-07-27); it polls the 4 underlying OSS E2E workflows.
-          # "drift-bot" is a commit status from the 8090-software-factory app.
+          # drift-bot is required too, but aggregated inside "E2E Gate (required)"
+          # (scripts/e2e_gate.py), not named separately here.
           ALL_CHECKS = {
               "clawmetry": [
                   "E2E Gate (required)",
-                  "drift-bot",
               ],
               "clawmetry-cloud": [
                   "Cloud golden-path browser E2E",
-                  "drift-bot",
               ],
               "clawmetry-landing": [
                   "Landing golden path (C3)",
--- a/.github/workflows/c6-health.yml
+++ b/.github/workflows/c6-health.yml
@@ -66,15 +66,22 @@ jobs:
           }
 
           # All 3 repos and their required checks.
-          # Keep in sync with REQUIRED_CHECKS in scripts/apply_required_status_checks.py.
+          # This is inline YAML and cannot import the Python source, so it is a
+          # genuine copy. tests/test_c6_required_checks_single_source.py asserts
+          # it matches REQUIRED_CHECKS in scripts/apply_required_status_checks.py
+          # and fails when they drift, because a "keep in sync" comment is not a
+          # mechanism.
           # "E2E Gate (required)" is the aggregator job from e2e-gate.yml (PR #4111,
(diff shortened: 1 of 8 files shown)


vivekchand:
✨ auto-fixed: merged main into branch to bring it current (was behind by several commits)

---

PR #5126, opened 2026-08-23
fix(trace): refuse to install a hook the PATH binary cannot run

vivekchand:
Found by dogfooding, on the machine that built the feature.

`clawmetry trace init` writes a hook that shells out to `clawmetry trace stamp`. On a machine whose PATH `clawmetry` is an older release, that command does not exist. The hook ends in `|| true` so a commit is never blocked, which is the right contract and is exactly what hid the problem: every commit succeeded, no trailer was ever written, and nothing said so. `trace status` even reported `hook installed: yes`, because the file was there.

**Silence is the worst shape this failure can take.** Coverage cannot be backfilled, so every trace lost while it goes unnoticed is lost permanently. It is the one failure mode in this feature with no recovery path.

`install()` now probes `clawmetry trace --help` first and refuses with a `stale-binary` status and an upgrade hint, rather than writing a hook that will do nothing. `verify_binary=False` exists for callers that know better.

The install tests get an autouse fixture stubbing the probe to True, because the real check depends on which clawmetry the developer has installed, and hook-mechanics tests should not pass or fail on that.

Proven to fail on the unfixed code rather than assumed to. 55 tests, py3.9 clean, ruff clean.

This PR is also the first commit in this repo to carry a `Clawmetry-Session` trailer, so it is the one I will publish as the first real trace.

https://claude.ai/code/session_012UjuRS7Xn4JJGacSgRMbrK

### DIFF
--- a/clawmetry/cli.py
+++ b/clawmetry/cli.py
@@ -4709,6 +4709,12 @@ def trace_main(argv) -> int:
             print("Earlier commits cannot be backfilled (PRD-pr-trace.md §3a).")
         elif status == "already-installed":
             print(f"Already installed -> {res['path']}")
+        elif status == "stale-binary":
+            print("Not installing: the `clawmetry` on your PATH has no `trace`")
+            print("command, so the hook would be written and then silently do")
+            print("nothing on every commit.")
+            print(f"  {res.get('hint', '')}")
+            return 1
         elif status == "foreign-hook":
             print(f"A different prepare-commit-msg hook exists: {res['path']}")
             print(f"Hint: {res.get('hint', '')}")
--- a/clawmetry/trace_stamp.py
+++ b/clawmetry/trace_stamp.py
@@ -179,6 +179,27 @@ exit 0
 """
 
 
+def hook_command_works(cmd: str = "clawmetry") -> bool:
+    """Can the binary the hook will call actually run ``trace stamp``?
+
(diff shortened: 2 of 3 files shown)


vivekchand:
✨ auto-fixed: merged main into branch to bring it current (was behind by several commits)

---

vivekchand:
**Autonomous maintainer pass (2026-08-23):** CI is otherwise green but drift-bot is blocking merge with 1 finding (converged after 3 iterations).

**What drift-bot flagged:**

**Requirement: [Runtime and Session Observability](https://factory.8090.ai/project/b415065f-ab2f-4f53-8864-0c009fd098cb/requirements/8e389016-a9c8-4352-9121-72f0e361fdf6)** — `clawmetry/trace_stamp.py:182`

The requirements doc has no acceptance criteria for the binary validation behavior: `install()` now probes whether the PATH binary can run `clawmetry trace` before installing a hook, and returns `"stale-binary"` if the check fails. This is a new contract that needs a requirement + AC.

**Path to green (operator):**
1. Open the Runtime and Session Observability requirement in the 8090 factory
2. Add an acceptance criterion covering: *when the clawmetry binary on PATH cannot run `clawmetry trace`, `install()` returns `"stale-binary"` and does not write the hook*
3. Confirm drift-bot status turns green, then merge

The code is correct — requirements are behind the implementation.

---

PR #5177, opened 2026-08-24
Act on an invalidated DuckDB instead of running bricked forever

vivekchand:
Builds directly on [#5169](https://github.com/vivekchand/clawmetry/pull/5169), which taught the store to **recognise** DuckDB's `database has been invalidated` and stopped the retry storm that wrote a 3.3 GB log. It stops there: it logs the recovery for a human to run, and gives up.

## The daemon then runs bricked, indefinitely, looking healthy

Every write fails, every read answers empty, and the process never exits, so launchd never restarts it. The dashboard just goes quiet. On this machine one sat in that state long enough to write **70 MB of identical tracebacks** while `launchctl list` reported exit status 0.

**Restarting is not enough on its own.** The cause is usually a corrupt index on disk, so a fresh daemon re-ingests the same rows, hits the same index delete, and invalidates again within seconds. Verified here: the daemon came back, served exactly one query, and re-bricked.

## Do what #5169 already writes down, automatically

```
reopen -> CHECKPOINT -> DROP every index -> CHECKPOINT -> migrate
```

The last step is what makes it safe: `_migrate` recreates every index from `_DDL`, so dropping them destroys no rows and leaves the schema exactly where it started. Two tests assert precisely that — **the row count survives** and **the index set comes back identical** (65 of them, on a real store). The ring still holds the failed batch, so a successful rebuild flushes the queued events rather than stranding them.

Bounded on purpose: **once per process, writer only.** A rebuilt index that invalidates again is not a stale index, and a loop would hide the real cause. If the rebuild declines or fails, behaviour is exactly what #5169 shipped.

## The other half: make it visible

`health()` now reports `db_invalidated`, `db_recoveries` and `db_last_error`. From the outside a bricked daemon is indistinguishable from an idle one — that is *why* nobody noticed for hours. A blank dashboard with no explanation is the failure mode FLYWHEEL §0a keeps warning about.

## Tests

10, on a real DuckDB file in `tmp_path` (never the developer's own store — a full-suite run writing into the real `~/.clawmetry` is what corrupted the index this fix exists for):

- rows survive the rebuild; the index set is restored exactly
- the store is writable afterwards, through the real `ingest` → `flush` path
- runs at most once per process; never on a read-only store
- **the flush-path wiring**: a poisoned connection proves the flusher actually calls recovery and lands the queued batch
- **the negative case**: an ordinary contended write (`Could not set lock`) must NOT drop every index
- the fatal detector is pinned against DuckDB's real message, not a paraphrase

No new ruff findings; `lint-py39` and `lint-daemon-allowlist` green. The store regression suite was diffed against a pristine `origin/main` worktree: identical failure set, no regressions.

https://claude.ai/code/session_01MxCcDf6WCSYXCXWdvjdBnq


vivekchand:
PR janitor sweep (automated, every 3h): CI is blocked on a **Drift Bot** failure.

The E2E Gate aggregator (`E2E Gate (required)`) failed because the `drift-bot` check reported `failure` on commit `def8dde`. All other checks are queued/in-progress or green (Syntax & Lint ✅, macOS pip install ✅, macOS API Tests ✅).

**Drift Bot** checks whether this diff contradicts a Blueprint in the Software Factory board — this is a semantic check, not a transient flake, so it can't be rerun away. You'll need to:
1. Check what the Drift Bot flagged (look at its check details in the Actions tab for run `32792303980`)
2. Either update the PR to align with the Blueprint, or push back on the Drift Bot finding if it's a false positive

The rest of CI is still running — if Drift Bot was a false positive and is retriggered green, the PR should unblock itself.

---

vivekchand:
**Automated sweep — Drift Bot requires human decision**

All actual CI checks on this PR pass. The only blocker is the E2E Gate, which fails because Drift Bot found 3 semantic drift findings — all pointing to `clawmetry/local_store.py`'s new `recover_invalidated_db` method and the `db_invalidated` / `db_recoveries` / `db_last_error` health fields not being documented in the "Local Observability Service" or "Local Agent Observability" blueprints in the Software Factory.

This is **not a flake** and cannot be cleared by a CI rerun. It requires one of:
1. Update the referenced blueprints in the 8090 Software Factory to document the new recovery behavior and health fields, then re-run Drift Bot.
2. Push back on the findings via the Software Factory UI if the blueprint update is out of scope for this PR.

The automation sweep cannot make this call — it's a product/documentation decision.

---

PR #5202, opened 2026-08-25
Size context windows per provider, and publish real instrumentation overhead

vivekchand:
Closes the two gaps a user named on the launch thread:

> Curious how it handles context window blowout across those runtimes, because that's where most of these comparison tools fall apart in practice. The readme skims over it. Would actually use something like this for a tutorial I'm working on if the instrumentation overhead is documented somewhere real and not just vibes-based benchmarks.

Both were fair. The README said nothing about either.

## 1. Context windows were Anthropic-only

`context_window_for_model` knew exactly two numbers, both Anthropic's (200K, and 1M for `[1m]` variants), and measured all 26 runtimes with that ruler. The failures were symmetric and both looked plausible on screen:

| Turn | Rendered as | Truth |
|---|---|---|
| 300K-token GPT-5 | ">100%, blown" | 75% of GPT-5's 400K |
| 130K DeepSeek | comfortable 65% | already over its 128K window |

New **`clawmetry/context_windows.py`** carries a per-provider table (Anthropic, OpenAI, Google, xAI, DeepSeek, Moonshot/Kimi, Qwen, Mistral, Meta, GLM) as an auditable file people can read and PR, rather than two constants buried in a 14K-line store.

**Sizing is resolved, not hard-coded**, and every answer carries provenance the same way Guard thresholds carry `threshold_source`:

`default` (guessing, and says so) → `model_table` → `explicit_marker` → `observed_floor`

A prompt the provider accepted cannot exceed that provider's window, so a measurement above the resolved window wins. `/api/context-economics` now ships `window_source` + `window_confidence` per point, so a gauge built on a guess cannot render with the same authority as one built on a lookup.

The observed-floor ladder is **vendor-aware**: a 323K Claude prompt snaps to 1M, not to a 400K rung Anthropic has never sold. Fabricating a denominator from a tier that does not exist is the same bug in a new costume.

## 2. A blowout count of zero was ambiguous

Only 4 of 24 adapters emit compaction events, so for most runtimes "Compactions: 0" meant *we are blind* while rendering identically to *ran clean*. A user comparing runtimes reads the second as the first.

New **`clawmetry/context_coverage.py`** + `LocalStore.query_context_coverage()` + **`GET /api/context-coverage`** report, per runtime per signal, one of `observed` / `supported_none_seen` / `unsupported`, with a note the UI can print next to a blind zero:

```json
{"runtime": "codex", "sessions": 12,
 "utilization": {"count": 300, "verdict": "observed"},
 "compaction":  {"count": 0, "verdict": "unsupported",
                 "note": "codex does not record compaction events. A zero here means ClawMetry cannot see them, not that none happened."},
 "overflow":    {"count": 2, "verdict": "observed"}}
```

Two rules keep it honest:
- **Measurement beats declaration.** Coverage is computed from the user's own store first, so a stale list can never hide real data.
- **The blind list is a denylist, not an allowlist.** Getting it wrong degrades to an honest "nothing seen"; an allowlist that went stale would claim blindness for a runtime that had started reporting.

## 3. Overhead is measured, not asserted

**`benchmarks/overhead.py`** separates the two questions "instrumentation overhead" conflates: latency added to the agent's own critical path, versus host cost paid by the daemon.

| Path | Added to the agent | Default? |
|---|---|---|
| Session-file tailing (all 26 runtimes) | **0**, separate process, no ClawMetry code in the agent | on |
| HTTP interceptor | **+0.45 ms** wall / +0.35 ms CPU per call = 0.009% of a 5s model call | off |

Host: 2,707 events/sec ingest, 710 bytes/event, 161 MB peak RSS. Raw JSON committed under `benchmarks/results/`.

### The method matters more than the numbers, and it was wrong three times first

- **Baseline-then-patched** let the second condition inherit warm caches and measure *faster* while doing strictly more work. It reported **negative overhead**. Each condition now runs in a fresh subprocess with the order alternating per round (`activate()` is a one-way global patch, so in-process interleaving is impossible).
- **Timing real loopback HTTP** wrapped a 2 ms p50 / 40 ms p99 round-trip around a sub-millisecond signal. The transport is now a stub adapter with no I/O, so `Session.send` still runs in full and the delta is only our work.
- **At 40 samples the delta was noise.** Now 5,000 calls per condition across 3 rounds, GC off inside the window, and the harness **refuses to print a figure when the rounds disagree on the delta's sign**.

`docs/OVERHEAD.md` and `docs/CONTEXT_BLOWOUT.md` carry the numbers, the method, and an explicit **"not yet measured"** list (hook gates, enforcement proxy, non-macOS). README gets a section on both instead of skimming them.

## Drive-by fixes

- `math` was left dead in `local_store.py` when `ceil()` moved out.
- `_normalise` raised `AttributeError` on a non-string `model` field, violating never-crash-on-bad-input on a read path fed by whatever a runtime wrote.

## Verification

- 36 new tests; 52 pass across the context suite including the 6 pre-existing ones, unchanged.
- `make lint-daemon-allowlist` passes (new method allowlisted; without it the proxy 400s and every reader silently degrades to the direct-read fallback).
- `ruff` count in `local_store.py` holds at main's 50.
- Both endpoints verified live end to end: `gpt-5-codex` at 300K tokens now returns `window: 400000, pct: 75.0, window_source: model_table`.
- No em-dashes in user-facing copy (README, docs, API notes) per the repo's copy rule.

https://claude.ai/code/session_01J1Rpud8YEq36HAj63ahsuk

---

## Every "not yet measured" path is now measured

The first version of this PR listed four gaps. They are closed, and each one turned up something.

| Path | Measured | Note |
|---|---|---|
| Pre-tool hook gate, warm cache | **+44 ms** over a 36 ms interpreter floor | where `cli.py`'s own comment says it must land |
| Pre-tool hook gate, 60s refresh | **+189 ms** | the gate fetches policies from the cloud **inline, on the critical path** |
| Enforcement proxy | **+9.7 ms** p50, +22 ms p95 | socket-to-socket; includes the extra hop a proxy inherently adds |
| Steady-state sync daemon | **11.8-12.4% of one core** | **over the documented 5-10% budget** |
| Steady-state dashboard | 0.13% of one core | fine |

**Two findings worth a reviewer's attention:**

1. **The hook gate makes a cloud call on the agent's critical path.** The policy cache has a 60s TTL; when it lapses, the next gated tool call blocks on `_fetch_cloud_policies` before the tool runs. On a fast link that is +189 ms. On a slow one, or during a cloud outage, it is bounded by a timeout rather than by anything in our control. Roughly one tool call per minute pays it. Moving that refresh into the daemon would take it off the critical path entirely.

2. **The sync daemon is over its own CPU budget.** FLYWHEEL §1e commits to "no more than ~5-10% of one core". Measured on the daemon actually running on this machine across two independent windows: 12.36% over 100 s and 11.76% over 120 s, with RSS flat to falling, so this is steady state and not a startup burst. Caveats stated in the doc (one machine, heavy install), but the budget does not carve those out. Published rather than omitted.

**Cross-platform** is now a workflow that runs the harness on Linux, macOS and Windows. Deliberately **not** a pass/fail gate: shared runners are noisy, a latency threshold would flap, and a check that flaps is a check everyone learns to ignore. It asserts the harness still runs everywhere and still emits a well-formed report carrying its machine stamp and its noise-floor verdict.

Every number in `docs/OVERHEAD.md` and the README is reconciled against the committed report in `benchmarks/results/overhead-m2pro.json`.

**Two measurement bugs found and fixed while building this**, both of which had already produced wrong readings:
- `_find_daemon_pids` matched a bare `clawmetry` substring, which on a dev box also matches browser tabs and a `cloud-sql-proxy`, and it latched onto a transient pid that was gone by the second sample.
- The hook benchmark isolated via `CLAWMETRY_HOME`, which the gate does not consult (it uses `expanduser("~")`), so it read the real key and benchmarked a different branch than intended.

---

Product record: https://factory.8090.ai/project/b415065f-ab2f-4f53-8864-0c009fd098cb/blueprints/9cf653c2-5ca4-490c-b0a4-875ae631eb47

Written up in that blueprint at v17. In the spirit of FLYWHEEL §0c I should say plainly that the code came first here and the record second, which is the order that section exists to prevent.


vivekchand:
**Auto-janitor CI failure report**

`E2E Gate (required)` is blocked by a failing `Drift Bot` check:

```
FAIL: required checks did not pass:
  - Drift Bot: 1 of 1 matching check(s) failed
      'drift-bot': failure
```

Drift Bot detects that this PR's diff contradicts a Blueprint in the Software Factory. This is a semantic check — it cannot be resolved by a rerun. The PR needs to either align with the Blueprint, or the Blueprint needs to be updated first.

Check the Drift Bot job logs for the specific Blueprint and field that triggered the contradiction.

This is a real failure (not a transient flake) — cannot be fixed by a rerun.

---

PR #5203, opened 2026-08-25
Make "PRD in 8090 before code" enforceable, not advisory

vivekchand:
No-PRD: process and CI plumbing. The rule it enforces is FLYWHEEL §0c, written up this week in the Local Audit and Data Control and Cost and Efficiency Analytics requirements.

**Risk:** this gate can block PRs. It's escapable in one line (`No-PRD: <reason>`), exempts docs/tests/CI/scripts, and its own tests assert every rejection path — so the failure mode people actually fear (a gate that blocks wrongly) is covered, and so is the one that matters more (a gate that silently passes forever).

## Why

§0c made "write the product record in 8090 first" the rule. A rule nothing enforces rots — and this repo has been burned by that exact shape before: an invalid ruff rule plus `|| true` linted nothing for months while reporting green.

## Where the standard now lives

| | |
|---|---|
| `AGENTS.md` | in **The rules that bite**, and above the read-this-first pointer |
| `CLAUDE.md` | first line of **Conventions** |
| `FLYWHEEL.md` | §0c now names the gate enforcing it |
| PR template | **Product record** + **Risk** fields, *prepended* to the existing checklists |

I first overwrote that template and destroyed the MOAT fast-path checklist — which encodes a 3-PR cascade (`#1258 → #1260 → #1266`) and is worth considerably more than my section. Restored; mine sits above it.

## What the gate does

`scripts/check_product_record.py` fails a PR touching product code unless the body either cites a Factory record:

```
https://factory.8090.ai/project/<uuid>/requirements/<uuid>
```

or opts out **on purpose**:

```
No-PRD: typo in a log string
```

A bare `No-PRD:` is a shrug, not a decision, and is rejected.

## What it deliberately does not do

Judge whether the record is any *good*. It can't, and a gate pretending to would be theatre. It makes "implement now, document later" require typing a sentence a human will read.

## Tests

16, and **most assert a case that must be rejected** — empty body, bare opt-out, a GitHub link, the word "factory" in prose, docs-alongside-code — because the only failure that matters for a gate is silently passing.

The gate's own tests also run **inside its own workflow**, not just the main suite, so it can't rot while still reporting green.

I verified it rejects before trusting it: a real code-change commit with no record exits 1; with a record, 0; docs-only, 0.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,4 +1,23 @@
 <!--
+FLYWHEEL.md section 0c: write the product record in 8090 BEFORE the code.
+Software Factory is the product reviewer -- a requirement written after the
+fact reviews nothing. Link what justified this change, or say why it needs
+nothing. CI checks that one of the two is present, never which one is right.
+-->
+
+**Product record:** <!-- https://factory.8090.ai/project/<uuid>/requirements/<uuid> -->
+
+<!-- For a change that genuinely needs none, delete the line above and keep a
+     real reason instead:   No-PRD: typo in a log string -->
+
+**Risk:** <!-- What could break, who notices, how it is undone. Data that
+changes retroactively, migrations that rebuild derived tables, controls that
+delete, anything already in flight this might collide with. "None" is a valid
+answer; a blank is not. -->
+
+---
+
+<!--
 Closes issue #1268. The MOAT checklist below catches the class of bug
 that produced the 3-PR cascade #1258 → #1260 → #1266 today (forgot
 allowlist + treated empty-fastpath as miss).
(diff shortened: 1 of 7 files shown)


vivekchand:
**Auto-janitor CI failure report**

`E2E Gate (required)` is blocked by a failing `Syntax & Lint` check. Root cause:

```
FAILED tests/test_workflow_yaml_valid.py::test_pytest_jobs_install_what_conftest_needs[product-record-gate.yml::product-record]

AssertionError: product-record-gate.yml job 'product-record' runs pytest but never installs ['requests'],
which tests/conftest.py imports at module scope. pytest will exit 4 (collection error) before running
a single test, so the job fails for a reason unrelated to what it checks.
```

**Fix:** Add `pip install requests` (or `pip install -r requirements.txt`) to the `product-record` job's install step in `.github/workflows/product-record-gate.yml` before the `pytest` call.

This is a real test failure (not a transient flake) — cannot be fixed by a rerun.

---

PR #5215, opened 2026-08-25
Provenance: every dollar figure says whether it was measured, derived, or estimated (WO-6)

vivekchand:
## WO-6: every number says how it was obtained

Work order: [Provenance Badges](https://claude.ai/code/artifact/42b9226f-6d69-407f-9220-4985bbed646d) from the [agent-observability sprint](https://claude.ai/code/artifact/e701231c-f999-4593-a7df-bd5543845cbf).

No-PRD: the specification was written first, as the linked work order (why, scope, files, acceptance criteria, non-goals, and the trap), and this PR implements it against those acceptance criteria unchanged. It is not in the Factory because the software-factory MCP server did not connect in this session, so the record could not be created from here. Mirroring it into the Factory is a follow-up, not a thing to back-fill after the fact.

### Why

ClawMetry already keeps this discipline internally, in three places and three private vocabularies that never reach the screen: Guard incidents carry `spend_basis` and refuse to invent a dollar figure, Guard thresholds carry `threshold_source`, context windows carry `confidence`. A reader looking at `$354.60` has no way to tell which of those disciplines produced it.

`clawmetry/provenance.py` is the one vocabulary those collapse into, and the seam that puts it on screen.

**measured** read from a record the producer wrote, or a plain aggregation of such records over a stated window.
**derived** computed from measured inputs by an exact rule (a rate-card lookup against a measured token count, a ratio, a rate).
**estimated** a model with an assumption that can be wrong.
**unknown** we could not obtain it, and then the value is `None`, never `0.0`.

### The trap this closes

A failed DuckDB read once shipped as `$0.00` and read, correctly, as a real result. A zero and a hole are identical once formatted, and the hole is the one that matters. So they are not allowed to share a shape: `provenance.stamp()` nulls the value it labels `unknown`, which means a renderer physically cannot print `$0.00` for a number nobody has. A genuine measured zero keeps its `0.0` and its measured badge, and the two read differently on screen.

Three more instances of the same class, found while auditing and fixed here:

- The OSS 24h history cap **zeroed** the older day buckets, which rendered as twelve days on which the user spent nothing: a plan boundary shown as a measurement. They now carry `withheld: true` and no value, and paint as a hatched reserved slot.
- The Overview spend tile's HTML placeholder was a literal `$0.00`, so for the first seconds of every load it stated the user had spent nothing. It now says it has not loaded.
- A Guard signal nobody could price serialised as `0.0`, so it sorted as a free incident at the bottom of a list ordered by money. It is now null with an unknown basis.

### What is labelled

Server side, and the same words on both sides of the cloud boundary:

| Surface | Figures |
|---|---|
| `routes/usage.py` `/api/usage` (all three builders) | today / week / month cost and tokens, day buckets, session costs, billing coverage, routing savings |
| `routes/usage.py` `/api/usage/forecast` | projected month-end, daily rate, spent so far |
| `routes/usage.py` `/api/usage/optimization-recommendations` | projected 30-day savings, per-tool savings |
| `routes/health.py` `/api/loop-signals` | spend at risk, per signal and in total |
| `clawmetry/sync.py` `dailyUsage` + `spending` | the slices the **hosted** dashboard renders (cloud parity is a hard gate) |
| `clawmetry/insights.py` | what the weekly digest cost to write |

Two findings worth naming. Cost is **derived**, not measured: almost every figure is measured tokens priced against a published rate card, which is a good number and is not a receipt. And a stale `spending` triple standing in for a live read is **estimated**, because it is a real measurement of a moment that has passed.

Front end: one shared component, `clawmetry/static/js/provenance.js`, loaded before `app.js`. `cmProv.money(payload, key, opts)` renders the figure, the badge, and a tooltip carrying the formula, the window, the operand values and the source. Six drifting copies of `fmtCost` now have one place to go.

### Acceptance

- **No dollar figure or score renders anywhere without a basis.** Enforced server-side: `provenance.audit_payload()` walks a real response, recognises money and score keys **by shape** rather than from a hand-kept list, and reports any with no basis. `tests/test_provenance.py` runs it over each payload builder above. The frontend half is a ratchet (below), not a claim of completeness.
- **An unknown cost renders as unknown, never as `$0.00`.** `stamp()` nulls it server-side; `cmProv.figure()` refuses to invent one back even when handed a confident basis; the unknown pill is italic, dimmed and dashed, with no currency symbol.
- **Hovering any figure shows the formula and the window.** Verified in a browser against real data.
- **A test asserts that a surface cannot render a figure lacking a basis.** `tests/test_provenance.py` (40 tests) for the payloads and the component, plus `tests/test_provenance_render_coverage.py`.

### The honest part

`app.js` has about sixty places that build a currency string, written over two years. This converts the ones a reader treats as headline figures and holds the line on the rest: `test_provenance_render_coverage.py` counts what is still unbadged and refuses to let that number grow, so a new unlabelled figure fails on the day it lands. The ceiling may only be edited downward, and a second test stops it being padded. Same shape as the AC-coverage gate already in this repo.

Three vocabulary-mapping tests read the **current** source of `detector_money`, `detector_calibration` and `context_windows`, so a new spend basis or threshold source fails here rather than silently rendering as "no basis".

### Non-goals

No new metrics. This only labels what already exists.

### Verified

- 90 tests green across the new suites and the four existing ones this changes.
- Run live against a real machine's store (`$354.60` today, 300k+ events) and against a seeded one: badges render, tooltips carry the formula, no new console errors.
- Screenshots: the Cost tab period table and Top Sessions badged `DERIVED`, the forecast badged `ESTIMATED`, and the three states side by side (`not available / NO BASIS` next to a real `$0.00 / MEASURED` next to `$12.50 / DERIVED`).
- One bug caught by looking rather than by testing: all three tiles in the spend card shared the `todayCost` entry, so the week figure's tooltip said "over: today". Fixed in 762e5d2.

Four existing tests changed, each because it asserted the old behaviour this work order exists to remove: a withheld day was `cost == 0`, an unpriceable Guard signal was `0.0`, `_resolve_spending` was compared with a strict `==`, and the digest's zero needed splitting into a real zero and an unpriceable one.

https://claude.ai/code/session_0172Agj4UUWJZGoaZjR2jwq1


vivekchand:
**PR janitor (automated):** E2E Gate failed on this PR due to a **30-minute timeout**, not a real test failure. The gate polled commit `60cda5cfab4d` and all required checks were still pending at timeout time, but they all completed successfully afterward (every check in the latest CI run is green).

**Action needed:** Please re-run the failed E2E Gate job manually (the janitor bot lacks `actions: write` permission to trigger reruns). Once it re-polls, all checks will be found passing and the gate should go green.

> Run to rerun: https://github.com/vivekchand/clawmetry/actions/runs/32882279540

---

vivekchand:

---

vivekchand:
awaiting confirmation — non-trivial rebase, needs human review (conflicts in: .github/workflows/auto-deploy-cloud.yml, .github/workflows/ci.yml, .github/workflows/publish.yml, .github/workflows/release-on-merge.yml, CHANGELOG.md, CLAUDE.md, clawmetry/adapters/base.py, clawmetry/adapters/openclaw.py, clawmetry/cli.py, clawmetry/entitlements.py, clawmetry/local_store.py, clawmetry/numbat_ingest.py, + 20 more files — all add/add, indicating deep divergence from main; semantic conflict resolution required across 30+ files)

---

vivekchand:
✨ auto-fixed: merged main into branch to bring it up to date (was BEHIND by multiple commits)

---

PR #5217, opened 2026-08-25
Add the organisation key: the secret that makes a session readable by colleagues

vivekchand:
**Product record:** https://factory.8090.ai/project/b415065f-ab2f-4f53-8864-0c009fd098cb/requirements/b04736fc-e656-4640-94d0-9c413f59c894 (Team Sessions and Shared Agent Context) — design in blueprint `7ac180d6`, **ADR-001**, which records what a single organisation key costs us and why a key hierarchy is not being shipped now.

**Risk:** additive. A new module, a new CLI subcommand, and two new fields on the heartbeat. **No caller uses `content_key()` yet**, so nothing changes about what seals what — a machine that never runs `clawmetry team key` is byte-identical to today. Undone by deleting the config field.

## The problem this is the missing piece of

Every encrypted payload is sealed with a key that belongs to **one machine** and lives in **one person's browser**. That is correct for a single operator, and it is why `routes/cloud.py` ships a named terminal state called `team_view_locked`: a member opening a teammate's node gets an empty list, deliberately, because there is no key in the system meaning *"readable by the people I work with"*.

This adds that key.

## What it is, and what the service learns

`clawmetry/org_key.py`. Created on a machine, handed between people out of band, **never sent to the hosted service** — so the promise is unchanged: content leaves the machine sealed, the service stores ciphertext it cannot read.

What the service *does* learn is the key's **fingerprint** (`content_key_fingerprint` + `content_key_scope`, on the plaintext heartbeat). A fingerprint identifies; it cannot decrypt. It exists so a member holding the wrong key can be **told that**, instead of being shown an empty screen and left to guess whether their team has no sessions or they hold the wrong secret — which is exactly the screen we ship today.

## The CLI

`clawmetry team key create | set | show | forget`.

**The key is never accepted as a command-line argument.** A secret in argv is a secret in shell history and in every `ps` listing on the machine, and this one opens an *organisation's* content rather than one machine's. It is read from `--file`, from stdin, or from a no-echo prompt — the convention `clawmetry license --file` already established here. Replacing a key already held requires `--force`, because content sealed with the old one becomes unreadable.

A machine that has never run `clawmetry connect` gets a plain sentence naming the one step that has to happen first, not a `FileNotFoundError` traceback.

## Sequencing (same discipline as #5213)

`content_key()` prefers the organisation key and falls back to the node key — **and nothing calls it yet**. The switch to sealing with it lands only once browsers can accept an organisation key, so no reader ever meets a payload they cannot open. Landing the key first is what makes that possible.

## One trap worth naming

`team` also had to be added to `main()`'s `_subcmds` tuple. A subcommand registered on the parser but missing from that list is silently handed to the dashboard's parser and rejected with `invalid choice` — the same two-lists shape `CLAUDE.md` warns about for runtimes. Cost me a debugging round; noting it so it doesn't cost the next person one.

## Tests

`tests/test_org_key.py`, 9 cases, chosen for properties whose failure is **silent**:

- **the heartbeat carries the fingerprint and never the key** (serialises the meta and asserts neither secret appears in it)
- a passphrase and the key it derives share a fingerprint — otherwise two colleagues holding the same secret are told they disagree
- a machine with no key yields `""`, never something a caller could mistake for "no encryption needed" (the failure mode tracked privately in clawmetry-cloud#2118)
- **a key on argv is refused** (runs the real CLI; the property lives in parser wiring, not in any callable)
- org key preferred, node key as fallback, env override for containers

## Verification

- `pytest tests/test_org_key.py tests/test_device_summary_snapshot.py` — 15 passed
- `ruff check clawmetry/org_key.py tests/test_org_key.py` — clean; `cli.py` 14 findings before and after
- Walked the real lifecycle on a scratch `HOME`: create → show → guarded re-create → `set --file` (fingerprint matched an independently computed one) → forget, and confirmed `_build_node_meta()` reports `content_key_scope: organisation` with the fingerprint and no key

Not claimed: no user-visible behaviour changes yet. This is the key existing, not the key being used.


vivekchand:

---

PR #5218, opened 2026-08-25
[RELEASE] session titles travel encrypted (carries #5213)

vivekchand:
**Product record:** REQ-OBS-RSO-032 (Runtime and Session Observability, v21).

Publishes the `sessionTitles` encrypted snapshot slice merged in #5213 so the cloud can be pinned to a wheel that ships it.

**Why this release is needed now:** cloud PR vivekchand/clawmetry-cloud#2119 is blocked by the snapshot-key drift guard, which correctly refuses to let the cloud read a key the pinned daemon does not write. That gate is doing its job. The unblock is a release, not an override.

Risk: CHANGELOG only. The code it publishes is already on main and was additive.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,12 @@
 ## Unreleased
 
+### Release: session titles now travel encrypted, like the rest of your conversation (carries #5213) (2026-08-25)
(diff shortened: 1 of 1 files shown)


vivekchand:
Closing: the change it was releasing (#5213) is redundant and is being reverted. #5213 added a sessionTitles snapshot slice carrying the same fact the daemon already sends per row as an encrypted title_blob (shipped in #5150, released in 0.12.772). Two mechanisms for one fact is the shape that has burned this codebase before. The real gap is that the cloud discards title_blob; that is being fixed in clawmetry-cloud instead.

PR #5219, opened 2026-08-25
Revert the sessionTitles snapshot slice: the daemon already sends this

vivekchand:
Reverts #5213. It was built on a misdiagnosis that was mine.

**Product record:** https://factory.8090.ai/project/b415065f-ab2f-4f53-8864-0c009fd098cb/requirements/8e389016-a9c8-4352-9121-72f0e361fdf6 (REQ-OBS-RSO-032, in Runtime and Session Observability). The requirement STAYS: its criteria are mechanism-neutral, and the mechanism that satisfies them already shipped in #5150. What is reverted is my duplicate of that mechanism, not the requirement it was written for.

**Risk:** removes a snapshot key nothing consumes. The cloud PR that read it is closed. No other reader exists.

## What I got wrong

I read `sync.py` from a worktree cut **before #5150 merged**, found the `_ftitle` fallback that derives a title from the session's first user message, confirmed readable titles in the cloud database, and concluded the leak was live.

It was not. **#5150 ("Make end-to-end encryption an invariant, not an optimisation") already fixed it at source, and the fix is released.** `split_session_title()` puts a neutral identifier in the server-parsed row and the real title in an encrypted companion field. Verified by cracking the published `0.12.772` wheel rather than by reading `main` — which is the check FLYWHEEL §0a.3 exists to force, and which I ran too late.

The readable rows in the database are **historical residue** from before that rollout. I compounded the error by attributing them to daemon versions via `nodes.version`, which is a node's **current** version, not the version that wrote the row. That manufactured a confident, wrong claim that 1,488 rows came from the current release. The honest test is `started_at`: since the fixed wheel published, **exactly one** readable title has been created, from a node that hasn't upgraded.

## Why revert rather than keep it

This slice carries, by a second route, a fact the daemon already sends per row. Two mechanisms for one fact is the shape that put two ownership models in this codebase a day apart, and the second one is mine again.

## What the search did find

The other half of #5150: **the cloud discards `title_blob`.** No column, no reference in `ingest_sessions`, none in `dashboard.py`.

| day | sessions | identifier only, no title | readable title |
|---|---|---|---|
| **2026-08-25** | 6,926 | **6,848 (98.9%)** | 12 |
| 2026-08-24 | 441 | 345 (78%) | 54 |
| 2026-08-20 | 432 | 327 (76%) | 65 |

Since 0.12.772 rolled out, session titles have effectively vanished from the hosted dashboard. An E2E mechanism shipped whose consumer was never built. That is fixed in `clawmetry-cloud`, not here.

`pytest tests/test_device_summary_snapshot.py` — 6 passed (the device slice, which is a different map for a different consumer, is untouched).


vivekchand:

---

PR #5249, opened 2026-08-26
Harden CI: bind untrusted event data to env vars in run: blocks

vivekchand:
## What

Four workflow steps interpolated `${{ github.event.* }}` straight into a `run:` script. This PR binds each value to a step-level `env:` var and reads it as a quoted shell variable.

| Workflow | Value(s) bound |
|---|---|
| `harness-observability-audit.yml` | `inputs.runtime`, `inputs.file_issues` |
| `i18n-autotranslate.yml` | `inputs.only` (and `event.before` alongside it) |
| `i18n-docs-autotranslate.yml` | `inputs.only` |
| `pr-screenshots.yml` | `pull_request.number` |

## Why

GitHub substitutes `${{ }}` expressions **textually, before bash parses the line**. The surrounding YAML quotes therefore give no protection — the value lands in the script as source text rather than as data. Binding it to `env:` and referencing `"$VAR"` is the standard fix: the runner hands the value to the process as an argument, and bash never parses it.

This is the Scorecard *DangerousWorkflow* / zizmor *template-injection* class. After this PR, no `run:` block in the repo interpolates `github.event.*` — the one remaining site, `release-on-merge.yml`, is fixed in #5246.

Severity is not uniform across the four, and the diff says so plainly:

- The three `inputs.*` sites are the substantive ones. They are reachable via `workflow_dispatch`, so they need repo write access to reach — but both i18n workflows hold `contents: write` + `pull-requests: write` and the audit workflow files issues, so the blast radius past that point is real.
- `pull_request.number` is a GitHub-generated integer and was never a practical risk. It is included so the rule holds repo-wide with no "known exception" a future reader has to re-derive.

## Secondary fix

`harness-observability-audit.yml` also moves `ARGS` from an unquoted string to a bash array. Previously `$ARGS` was expanded unquoted, so a runtime name containing whitespace was word-split into separate argv entries. It now reaches `audit.py` as a single argument.

## Verification

- All 34 workflows parse: `python3 -c "import yaml,glob; [yaml.safe_load(open(f)) for f in glob.glob('.github/workflows/*.yml')]"` → OK
- `python3 scripts/check_action_refs.py` → OK, 17 refs
- The rewritten audit block was executed against a stub `audit.py` under `bash -e`. argv matches the previous script on every path:

  | Case | argv |
  |---|---|
  | schedule, blank input | `['--file-issues']` |
  | dispatch, `runtime=openclaw` | `['--file-issues', '--runtime', 'openclaw']` |
  | dispatch, `file_issues=false` | `['--runtime', 'openclaw']` (after `dry-run (no issues)`) |
  | runtime containing `;` and a command | passed through as one inert argv entry |

- No `permissions:` blocks are touched, so no release/publish path changes scope.
- Every file is under `.github/`, which `scripts/check_product_record.py` exempts — no Factory record needed.

No-PRD: CI-only workflow hardening under .github/, exempt from the product-record gate.

---
_Generated by [Claude Code](https://claude.ai/code/session_019xPcYnVqKnR6hrVrdsFXdt)_

### DIFF
--- a/.github/workflows/harness-observability-audit.yml
+++ b/.github/workflows/harness-observability-audit.yml
@@ -67,12 +67,18 @@ jobs:
         run: bash scripts/harness/sync.sh
 
       - name: Audit observability coverage + file gap issues
+        # Dispatch inputs are untrusted text. Bind them to env vars and read
+        # them as quoted shell variables so the runner never expands them as
+        # code. Never interpolate ${{ github.event.* }} into a `run:` block.
+        env:
+          RUNTIME: ${{ github.event.inputs.runtime }}
+          FILE_ISSUES: ${{ github.event.inputs.file_issues }}
         run: |
-          ARGS=""
-          [ -n "${{ github.event.inputs.runtime }}" ] && ARGS="--runtime ${{ github.event.inputs.runtime }}"
+          ARGS=()
+          [ -n "$RUNTIME" ] && ARGS=(--runtime "$RUNTIME")
           # default to filing issues on schedule; honor the dispatch toggle
-          if [ "${{ github.event_name }}" = "workflow_dispatch" ] && [ "${{ github.event.inputs.file_issues }}" = "false" ]; then
-            echo "dry-run (no issues)"; python3 scripts/harness/audit.py $ARGS
+          if [ "${{ github.event_name }}" = "workflow_dispatch" ] && [ "$FILE_ISSUES" = "false" ]; then
+            echo "dry-run (no issues)"; python3 scripts/harness/audit.py "${ARGS[@]}"
           else
-            python3 scripts/harness/audit.py --file-issues $ARGS
+            python3 scripts/harness/audit.py --file-issues "${ARGS[@]}"
           fi
--- a/.github/workflows/i18n-autotranslate.yml
+++ b/.github/workflows/i18n-autotranslate.yml
@@ -53,12 +53,18 @@ jobs:
 
       - name: Translate deltas (via Claude Code CLI, no API key)
         if: env.SKIP != '1'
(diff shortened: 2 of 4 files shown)


vivekchand:
One note for reviewers, since the visual-diff comment above is itself the evidence.

The `pr-screenshots.yml` change in this PR edits the **"Post / update PR comment"** step — moving `pull_request.number` out of the script and into `env: PR`. That step is the one part of this diff I could not exercise locally. The bot comment above was posted *by that rewritten step*, on this run: it resolved the PR number correctly (branch `pr-screenshots-pr-5249`, head `83d1aed811d9`) and both `${PR}` reads — the `gh api .../issues/${PR}/comments` lookup and the `gh pr comment "$PR"` fallback — worked. So that edit is verified end-to-end in CI, not just by inspection.

On the 35/66 flagged comparisons in that table: not from this PR. The diff is four workflow YAML files and touches no application code — no `dashboard.py`, no `routes/`, no `clawmetry/static/`, no `clawmetry/templates/` — so it cannot change what the dashboard renders. The pattern also reads as bot noise rather than a content change: several views are ~0% on desktop and 100% on mobile for the same page (`security` 0.01% / 100%, `logs` 5.81% / 100%), which is a render-timing artefact in the screenshotter, not a regression. The `visual-diff` check itself concluded **success**, and the bot labels itself non-blocking.

CI on this head: 34 of 36 checks green, no failures. That includes CodeQL `Analyze (actions)` — the query set that flags this injection class — plus `Syntax & Lint`, `Action references resolve`, and the product-record gate. `OpenSSF Scorecard` shows as skipped because it only runs on `main`.

---

PR #5198, opened 2026-08-25
feat(adapters/openclaw): surface nemoclaw onboarding OTel trace artifacts (#5193)

vivekchand:
Product record: https://factory.8090.ai/project/b415065f-ab2f-4f53-8864-0c009fd098cb/requirements/8e389016-a9c8-4352-9121-72f0e361fdf6 (REQ-OBS-RSO-034 + AC-OBS-RSO-002.3) and blueprint https://factory.8090.ai/project/b415065f-ab2f-4f53-8864-0c009fd098cb/blueprints/40e35d24-8249-4f72-b311-f311dc71aec5 (section "NemoClaw onboarding trace artifacts", ADR-105).

## Summary

When `NEMOCLAW_TRACE` is set, the NemoClaw harness writes OTel-style spans for each onboarding phase to disk -- but the ClawMetry adapter was discarding the file entirely, so a failed or slow onboarding phase was invisible on the dashboard. This PR adds a `_nemoclaw_onboard_trace()` helper that reads the file and surfaces the worst-case status, error phase names, and slowest-span summary.

## Changes

- Follow-up (2026-09-03): normalise OTel span status objects (`{code: N}` / `STATUS_CODE_*`) so `nemoclawOnboardTraceStatus` is always one of `OK` / `ERROR` / `UNSET`; two more unit tests (11 in the file).

- `clawmetry/adapters/openclaw.py`: Add `_nemoclaw_onboard_trace() -> dict` after `_model_router_live()`. Resolves the trace file via `NEMOCLAW_TRACE_FILE` -> `NEMOCLAW_TRACE_DIR/trace.json` -> `.e2e/traces/trace.json` (harness default). Handles both the flat harness `{spans:[...], summary:{...}}` shape and the standard OTel `resource_spans` export. Self-gates on `NEMOCLAW_TRACE` env var; returns `{}` when disabled or file absent. Wire it into `detect()` after `_workshop_approval_config()`.
- `tests/test_obs_gap_nemoclaw_onboard_trace_5193.py`: 7 unit tests covering disabled (env absent, `=0`), file missing -> `{}`, all-OK spans, ERROR span surfaces status + name, `summary.slowest_spans` forwarded, and `NEMOCLAW_TRACE_DIR` path resolution.

## Test plan

- [ ] `python3 -m pytest tests/test_obs_gap_nemoclaw_onboard_trace_5193.py -v` -- 7/7 pass (verified locally)
[...]

PR #5253, opened 2026-08-26
Harden CI: least-privilege token scopes for the last 3 workflows

vivekchand:
Closes out the TokenPermissions batch: after this, **every workflow in `.github/workflows/` declares exactly one top-level `permissions:` block** — 34/34, verified on the current head with a loader that rejects duplicate keys.

These were left for last because all of them touch a release or deploy path, where a blanket `contents: read` breaks the pipeline silently. Each job was read individually and the scope justified below rather than applied uniformly.

## `ci.yml` — `contents: read`, no job elevation

All 13 jobs do the same shape of work: checkout, install, run tests, upload an artifact. Nothing in the file calls the GitHub API with `GITHUB_TOKEN`, pushes a commit, creates a tag, comments on a PR, or touches a Release. The only `uses:` beyond checkout/setup are `actions/cache`, `actions/upload-artifact` and the local `./.github/actions/setup-openclaw`; `upload-artifact` uses the runtime token, not `GITHUB_TOKEN`. The read-only floor therefore covers the whole workflow and no job needs an elevated block.

## `auto-deploy-cloud.yml` — `contents: read`

Worth stating explicitly because the job name suggests otherwise: this workflow checks out `clawmetry-cloud`, pushes a branch, and creates, closes and merges a PR there — but every one of those steps authenticates as `secrets.CLOUD_REPO_PAT`, both as the `token:` on the second checkout and as `GH_TOKEN` on the `gh` steps. **A workflow `permissions:` block does not govern a PAT**, so none of that cross-repo work is affected by this change.

The only thing riding on `GITHUB_TOKEN` is the checkout of this repo, which needs `contents: read` and nothing else. Keeping the floor read-only means a future step added here cannot quietly pick up write access to `clawmetry` through the ambient token — it would have to ask for the PAT, which is visible in review.

## `desktop-artifacts.yml` — `contents: read` floor, `release` stays `contents: write`

| job | effective scopes | why |
|---|---|---|
| `macos` | `contents: read` | inherits; checkout + build + upload-artifact |
| `linux` | `contents: read` | inherits; checkout + build + upload-artifact |
| `windows` | `id-token: write`, `contents: read` | unchanged — already declared for `azure/login` OIDC |
| `release` | `contents: write` | unchanged — `softprops/action-gh-release` attaches installers to the tag's Release and publishes it |

The elevation deliberately stays on the `release` job. A workflow-wide `contents: write` would hand it to all three build jobs and to every third-party action they run, which on the signing paths is the largest surface in this repo.

Note the ordering hazard this file already handles correctly: a job-level `permissions:` block *replaces* the top level rather than merging with it, so the `windows` job's existing block had to already carry `contents: read` for its checkout to keep working. It does (with a comment saying exactly that), so the new top-level block does not regress it.

## `release-on-merge.yml` — `contents: read` floor, job scopes untouched

#5250 was expected to cover this file but landed without a top-level block, leaving it the last workflow in the repo with none — caught by re-running the check across all 34 workflows after merging `main` into this branch, not assumed from #5250's description.

The change is **functionally inert by design**. The workflow has exactly one job; that job declares its own `contents: write` / `pull-requests: write` / `actions: write`; and since a job-level block replaces the top level, the new value can never apply to `release`. Its purpose is that a job added here later defaults to read-only rather than inheriting the repository default, and that the file stops reading as unscoped. The publish, tag and PR-merge scopes stay on the job that needs them.

## `api-latency-smoke.yml` — repairing a workflow the batch had broken

Carried in `36e773d`. Two earlier hardening commits each added a top-level `permissions: contents: read` to this file — `210696e` below `concurrency:`, `85873f0` (#5250) above it. Neither conflicted in git and both were individually correct, but **YAML forbids a repeated mapping key**, so GitHub rejected the workflow and failed every run at startup with zero jobs. The check had been dead on `main` since #5250 landed.

The fix keeps the block above `concurrency:` and folds the other comment's point into it. Effective scope is unchanged: `contents: read`, one job, no job-level override.

### Why the existing guard missed it, and the correction

This is worth flagging because it undercuts how the earlier revisions of this PR described their own verification. `yaml.safe_load` resolves a duplicate key **last-one-wins and raises nothing** — so `test_workflow_parses_as_yaml`, and the "all 34 workflows parse" check cited in this PR's earlier description, both passed happily on a file GitHub refuses to run. Parsing cleanly was never evidence of a single `permissions:` block; it just could not see the second one.

`36e773d` adds `tests/test_workflow_has_no_duplicate_keys` next to the existing guard, using a `SafeLoader` subclass whose mapping constructor rejects a repeat, auto-discovered over every workflow. Re-inserting the duplicate fails the new test while the old parse test still passes — which is precisely the blind spot being closed.

## Verification

Re-run on the current head (`36e773d`), which includes `main` (bringing #5250/#5251/#5252):

- All 34 workflow files parse **and none repeats a mapping key**, checked with a duplicate-rejecting loader rather than plain `safe_load`.
- Every workflow declares a top-level `permissions:` block; each of the four files this PR touches has exactly one.
- Effective per-job permissions resolved and inspected for all four files; the tables above are that output, not an assumption. `release-on-merge`'s job scopes confirmed unchanged.
- `api-latency-smoke` resolves to `contents: read` with its single `smoke` job.
- 184 workflow tests pass.
- `scripts/check_action_refs.py` — 17 action references, OK.

No functional change beyond reviving `api-latency-smoke.yml`: additive `permissions:` blocks and one de-duplication, no step, trigger or job logic touched.

No-PRD: CI-only change confined to `.github/` and `tests/`, exempt from the product-record gate.

### DIFF
--- a/.github/workflows/api-latency-smoke.yml
+++ b/.github/workflows/api-latency-smoke.yml
@@ -14,17 +14,13 @@ on:
 
 # Least privilege: this workflow only builds a wheel and probes a local
 # dashboard. It never writes to the repository or any GitHub API surface.
+# Any job needing more must elevate with its own job-level `permissions:`.
 permissions:
   contents: read
 
 concurrency:
   group: api-latency-smoke-${{ github.ref }}
   cancel-in-progress: true
-
-# Least privilege: these jobs only read the repo. Any job needing more
-# must elevate with its own job-level `permissions:` block.
-permissions:
-  contents: read
 
 jobs:
   smoke:
--- a/.github/workflows/auto-deploy-cloud.yml
+++ b/.github/workflows/auto-deploy-cloud.yml
@@ -20,6 +20,15 @@ on:
     types: [completed]
   workflow_dispatch:
 
+# Every cross-repo action below (checkout of clawmetry-cloud, branch push, PR
+# create/close/merge) authenticates as `secrets.CLOUD_REPO_PAT`, which this
+# block does not govern — a PAT carries its own scopes. The only thing that
+# rides on GITHUB_TOKEN is the checkout of THIS repo, which needs `contents:
+# read` and nothing more. Deliberately kept read-only so a future step cannot
+# quietly acquire write on clawmetry via the ambient token.
+permissions:
+  contents: read
+
 jobs:
   pin-version:
     name: Rebuild cloud with latest OSS
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -5,6 +5,13 @@ on:
     branches: [main, master]
   pull_request:
 
+# Every job here only reads the repository: checkout, install, run tests,
+# upload an artifact. Nothing calls the GitHub API with GITHUB_TOKEN, pushes
+# a commit, tags, comments on a PR or touches a Release, so the whole
+# workflow runs on the read-only default and no job needs an elevated block.
+permissions:
+  contents: read
+
 concurrency:
   group: ci-${{ github.event.pull_request.number || github.ref }}
   cancel-in-progress: true
--- a/.github/workflows/desktop-artifacts.yml
+++ b/.github/workflows/desktop-artifacts.yml
@@ -18,6 +18,15 @@ on:
       - 'v*.*.*'
   workflow_dispatch: {}
 
+# Least-privilege floor for the three BUILD jobs (macos / linux inherit this;
+# windows re-declares it alongside id-token). The `release` job, and only that
+# job, elevates to `contents: write` so softprops/action-gh-release can attach
+# installers to the tag's Release and publish it. Keep the elevation on that
+# job — a workflow-wide `contents: write` would hand every build job, and each
+# third-party action it runs, the ability to write to the repository.
+permissions:
+  contents: read
+
 jobs:
   macos:
     name: macOS .app + .dmg (signed + notarized when secrets are set)
--- a/.github/workflows/release-on-merge.yml
+++ b/.github/workflows/release-on-merge.yml
@@ -5,6 +5,15 @@ on:
     types: [closed]
     branches: [main]
 
+# Read-only floor. This workflow's single job re-declares its own scopes
+# below, and a job-level block REPLACES the top level rather than merging
+# with it, so this value never actually applies to `release` — it is here so
+# the default for any job added later is read-only, and so the workflow
+# stops reading as "inherits whatever the repository default happens to be".
+# The publish/tag/PR scopes stay on the job that needs them, not up here.
+permissions:
+  contents: read
+
 jobs:
   release:
     name: Bump version & publish to PyPI
--- a/tests/test_workflow_yaml_valid.py
+++ b/tests/test_workflow_yaml_valid.py
@@ -79,6 +79,58 @@ def test_workflow_parses_as_yaml(path: str) -> None:
     )
 
 
+class _DuplicateKeyLoader(yaml.SafeLoader):
+    """A SafeLoader that refuses a mapping with a repeated key.
+
+    PyYAML's own resolution is last-one-wins, silently: ``yaml.safe_load`` on a
+    file with two top-level ``permissions:`` blocks returns a perfectly good
+    dict and raises nothing. GitHub Actions does the opposite -- it rejects the
+    file outright and fails the run at startup -- so a duplicate key is exactly
+    the class of break ``test_workflow_parses_as_yaml`` cannot see.
+    """
+
+
+def _no_duplicate_keys(loader, node, deep=False):
+    seen = set()
+    for key_node, _ in node.value:
+        key = loader.construct_object(key_node, deep=deep)
+        if key in seen:
+            raise yaml.YAMLError(
+                f"duplicate key {key!r} at line {key_node.start_mark.line + 1}"
+            )
+        seen.add(key)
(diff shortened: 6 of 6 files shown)


vivekchand:
✨ auto-fixed: branch was BEHIND main (base `09b255dc` → current `41d4accf`); updated via GitHub merge-base update — CI now running on new head `8eec031a`.

---

vivekchand:
**`E2E Gate (required)` failed on `90f0abe` — runner-queue starvation, not this PR.** Recording the evidence rather than silently re-running.

### What failed

`E2E Gate` is an aggregator (`scripts/e2e_gate.py`, `MAX_WAIT: 1800`) that polls the 12 required checks. It started 13:43:08 and gave up at 14:13:30:

```
  - MOAT Keystone: 0/1        - MOAT Verifier: 0/1
  - E2E Browser Tests: 0/1    - Entitlement API tests: 0/1
  - API Tests (3 OS): 0/3     - pip install matrix: 0/4
  - Wheel install & assets: 0/1
  - Store invariants: 0/1
```

No test failed. Those 8 checks never got a runner: their jobs were created at 13:43:47 and are *still* `queued` now, ~30 min later.

### Why it isn't this PR's

- **Repo-wide, not branch-specific.** 23 workflow runs are queued across the repository right now, created between 12:44 and 12:57 and still unstarted — CI for #5153, #5154, #5155, #5156, #5157, #5176, #5198, #5241–#5244 and #5255–#5261. This branch's run is one of that set. The backlog reaches PRs that contain none of this diff.
- **This commit's own checks passed.** Everything that got a runner is green: `Syntax & Lint`, `Analyze (actions)` (CodeQL over the workflow files), `Analyze (python)`, `Analyze (javascript-typescript)`, `Action references resolve`, `OSS golden path`, `Cross-repo handoff (C4)`, `C6 Required-status-checks gate`, `Python dependency audit`, `Vendored assets`/`fonts`, `No third-party asset loads`, `PR cites the product record`, Drift Bot.
- **`Syntax & Lint` is itself a `ci.yml` job** and passed on this exact commit, so the top-level `permissions: contents: read` this PR adds to `ci.yml` does not stop that workflow triggering or running. Job permissions govern the token minted for a job; they have no bearing on scheduling or queue position.

The gate's 30-minute budget is simply shorter than the queue latency the repo is seeing today.

### What I'm not doing

No fix ported, because there isn't one that belongs here. Raising `MAX_WAIT` in `e2e-gate.yml` would be a different concern from token scopes, would widen this PR, and would paper over a capacity problem rather than fix it. That trade-off is the maintainer's call, not something to slip into a hardening PR.

I'm holding my single re-run rather than spending it now: a fresh poll started while ~20 runs are still queued ahead would just time out again. I'll re-run `E2E Gate` once the `ci.yml` jobs for this commit have actually landed, and keep watching until this is green or closed.

If the queue is being starved by the ~20 open dependabot PRs each running the full matrix, that's worth a look independently of this PR.

---

vivekchand:
Pushed `36e773d`, which fixes a break this hardening pass caused on `main`.

Two earlier PRs in this series each added a top-level `permissions: contents: read` to `.github/workflows/api-latency-smoke.yml` — 210696e below `concurrency:`, 85873f0 above it. They did not conflict in git, and both were individually correct, but a repeated mapping key makes the file invalid: GitHub rejects the workflow and fails every run at startup with zero jobs, listed under the file path rather than "API Latency Smoke". **The API-latency check has been off on `main` since 85873f0 merged**; the last real run was 2026-08-22. That is also why this PR's push run of that workflow is red.

The fix keeps one block and folds the removed comment's point about job-level elevation into it. Effective scope is unchanged: `contents: read`, one job, no job-level override.

The pre-merge guard could not catch it. `yaml.safe_load` resolves duplicate keys last-one-wins and raises nothing, so `test_workflow_parses_as_yaml` passed on a file GitHub refuses to run. `36e773d` adds `test_workflow_has_no_duplicate_keys` alongside it — a `SafeLoader` subclass whose mapping constructor rejects a repeat, auto-discovered over every workflow like the guards around it. Re-inserting the duplicate fails the new test while the old parse test still passes.

Verified: 34 workflows parse and none repeats a key; every workflow still declares a top-level `permissions:` block; 184 workflow tests pass. The two sibling repos were scanned for the same pattern and are clean.

One thing worth a look: `main` stays broken until this merges. If that is too long to wait, the `api-latency-smoke.yml` hunk alone is a three-line revert that can go straight in.

The earlier E2E Gate failure on this branch was a timeout — it waited 30 minutes for the CI run, which sat queued for 48.

---

vivekchand:
Closing the loop on my note above: **all 36 checks are green on `36e773d`**, `E2E Gate (required)` included (passed 14:54:52).

No re-run was needed — the runner backlog drained on its own, the `ci.yml` jobs got runners at ~14:40 and all reported, so the gate's poll succeeded well inside its 1800s budget. The earlier red was purely the queue, exactly as diagnosed; nothing in the diff changed to fix it.

Nothing outstanding from my side. This is waiting on review.

One thing worth flagging for whoever merges: `api-latency-smoke.yml` is **still broken on `main`** — I confirmed the duplicate `permissions` key is present at `41d4acc` (two top-level blocks, line 26). `36e773d` here is the only fix, so `API Latency Smoke` stays dead on `main` until this merges or the fix is cherry-picked.

---

PR #5332, opened 2026-08-29
fix(e2e-gate): skip Drift Bot when unreported, raise MAX_WAIT to 3600s

vivekchand:
No-PRD: regression fix to merge-gate configuration, no user-facing feature change.

## Problem

The `e2e-gate` required-status-check was timing out after 30 min on every frontend/npm dependabot PR (13+ active PRs affected). Two root causes:

1. **Drift Bot never posts on frontend-only PRs.** The 8090-software-factory GitHub App only posts its `drift-bot` commit status when a PR touches Python or product-record files. On an npm bump (react-router, i18next, postcss, etc.) the gate polled for the full 1800 s waiting for a status that would never arrive.

2. **Runner starvation from concurrent dependabot PRs.** Each of the 13+ open dependabot PRs has a unique PR number, so each gets its own `ci-${{ github.event.pull_request.number }}` concurrency group. All 143+ ci.yml jobs queue simultaneously, causing the 8 required checks (MOAT Keystone, E2E Browser Tests, API Tests x3, etc.) to sit pending past the 30-min MAX_WAIT even after lint cleared.

## Fix

**`scripts/e2e_gate.py`**
- Added `skip_if_unreported: bool = False` field to `Spec` dataclass. When `True`, if zero check-runs or commit statuses have been posted the spec evaluates to "passed (skipped)" immediately. A failing status still fails the gate normally.
- Set `Drift Bot` spec to `skip_if_unreported=True`.
- Raised `DEFAULT_MAX_WAIT` from 1800 s to 3000 s to give slower-starting jobs more headroom.
- Updated `--list` output to show `[skip_if_unreported]` flag.

**`.github/workflows/e2e-gate.yml`**
- Raised `timeout-minutes` from 35 to 65.
- Raised `MAX_WAIT` env var from `"1800"` to `"3600"`.
- Updated header comment.

## Impact

- Dependabot frontend PRs now pass the gate without waiting 30 min for a status that never comes.
- Remaining 11 required specs are unaffected; a real Drift Bot failure (status posted + failing) still blocks the merge.
- Gives runner-starved jobs up to 60 min to complete instead of 30 min.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Sb3XD6RqANbD5nQV9QUTxV)_

PR #5338, opened 2026-08-29
Harden CI: stop the screenshot job persisting the job credential

vivekchand:
**Product record:** No-PRD: CI-only change confined to `.github/`, a PRD-exempt path.

**Risk:** Low. One workflow, two `actions/checkout` steps, nine added lines of which seven are a comment. Nothing in the shipped package changes. Undone by reverting one commit.

## Summary

Continues the `persist-credentials: false` sweep (#5312, #5321, #5327 merged; #5330, #5331 open) with the one remaining workflow whose checkouts are provably not the credential's consumer.

`actions/checkout` leaves a credential in `.git/config` by default. `pr-screenshots.yml` is the job where that matters most and was least necessary:

- **It has no consumer.** The publish step never touches either working copy. It creates a fresh temp directory, `git init`s a new repo there, and pushes to an explicit `https://x-access-token:${GH_TOKEN}@github.com/${GITHUB_REPOSITORY}.git` remote. The credential the checkouts persist is simply never read.
- **It sits in a job that runs PR head code.** This workflow checks out the pull request head, installs its requirements, and boots two dashboards from it to render screenshots. A credential left in `.git/config` is readable for the remainder of that job — including from the `base` checkout, which exists only to render "before" images.

Both checkouts are read-only source trees, so `persist-credentials: false` removes the credential without removing anything that uses it.

## Why this workflow and not the other four remaining ones

The sweep is deliberately not finished in this PR. Four checkouts still persist their credential, and each is a different question rather than more of the same:

| Workflow | Why it is not in this PR |
|---|---|
| `auto-quarantine.yml` | Pushes with a bare `git push -u origin "$BRANCH"` — it *is* the consumer. Needs the push rewritten to an explicit remote first, which is a behaviour change, not a hardening one. |
| `release-on-merge.yml` | Same pattern, twice (`git push origin "refs/tags/$NEW_TAG"`, `git push origin "$BRANCH"`), on the release pipeline. Same rewrite, higher blast radius. |
| `i18n-autotranslate.yml`, `i18n-docs-autotranslate.yml` | Push via `peter-evans/create-pull-request`, which authenticates from its own `token` input. That very likely tolerates `persist-credentials: false`, but the action's docs do not state it and this repo has no workflow pairing the two, so there is no evidence here either way. Not worth a silently-broken translation bot to find out in passing. |

The second `auto-deploy-cloud.yml` checkout stays as-is by design — #5331 documents that its credential is the one the force-push uses.

## Test plan

- [x] `yaml.safe_load` over all 34 workflow files → parse
- [x] `yaml.safe_load` over all composite action files → parse
- [x] Semantic re-read of the parsed YAML: both `visual-diff` checkout steps report `persist-credentials: False` with `path` and `ref` intact
- [x] `python3 scripts/check_action_refs.py` → OK, 17 distinct references, pinning unaffected
- [x] `pytest tests/test_workflow_yaml_valid.py tests/test_ci_workflow_invocations_are_real.py` → **189 passed, 2 skipped** (unchanged from baseline)
- [x] Read the publish step end to end to confirm the push path uses `mktemp -d` + `git init` + an explicit token remote, not either checkout

## Note for reviewers

The one thing worth a second opinion is the claim that nothing reads the persisted credential. The publish step is the only `git push` in the file and it runs entirely inside `$(mktemp -d)`; the screenshot steps read from `head/` and `base/` as plain source trees. If there is a path I missed that expects an authenticated remote in either checkout, that is the thing to catch here.

---
_Generated by [Claude Code](https://claude.ai/code/session_01Wb9DAnFEFQS6NvPsqqz7Kq)_


vivekchand:
✨ auto-fixed: merged latest main into branch (branch was BEHIND)

---

vivekchand:
## Ignore the visual-diff table on this PR — it is measuring its own noise

Flagging this because the bot's comment is edited in place, so a reviewer only ever sees the latest table, and "38 of 66 flagged" on a PR that cannot change a pixel looks alarming without the context.

**This PR's diff is 9 added lines in `.github/workflows/pr-screenshots.yml` and nothing else.** In every run, the bot renders `before` from the PR base and `after` from the head, and the head is the base plus that one workflow-file commit. The two trees it screenshots contain identical application code. Every non-zero number in that table is therefore noise by construction — there is no code path by which a `persist-credentials` key on a checkout step changes a rendered dashboard.

The two runs make that concrete. Same diff, and views swing in **both** directions between "identical" and "completely different":

| View | Run 1 (`55f1c18e`) | Run 2 (`afcaa437`) |
|---|---|---|
| mobile skills | 100.00% | 0.01% |
| mobile memory | 100.00% | 0.02% |
| mobile notifications | 100.00% | 0.02% |
| mobile security | 100.00% | 0.75% |
| desktop overview | 100.00% | 3.83% |
| mobile turn-anatomy | 0.01% | 100.00% |
| mobile tool-catalog | 0.01% | 100.00% |
| mobile swimlane | 0.01% | 100.00% |
| mobile policy | 0.02% | 100.00% |
| mobile dives | 0.02% | 100.00% |

A view cannot go from 0.01% to 100.00% because of a YAML key that was already present in the first run. The totals moved 43/66 → 38/66 for the same reason.

Consistent with the bot's own footer: it is non-blocking, and both runs concluded `success`. Nothing to fix here, and I am not proposing a change to the bot in this PR — worth a separate look at whether the screenshots are captured before the dashboards finish loading, since a comparator that reports 100% on identical code is not giving reviewers a usable signal.

**One thing the runs did confirm positively:** the publish step pushed to `pr-screenshots-pr-5338` successfully on both heads with `persist-credentials: false` set on both checkouts. That is the PR's central claim — that the persisted credential has no consumer because the push builds its own repo in `$(mktemp -d)` and uses an explicit token remote — demonstrated end to end rather than only argued.

---

vivekchand:
## `E2E Gate (required)` failed on `afcaa437` — runner starvation, not this PR

### What failed

The gate is a watcher: it polls the 12 required checks and fails if they do not all complete within `MAX_WAIT=1800`. It ran 18:15:48 → 18:46:05 and timed out with 8 of 12 still pending. **No test failed.**

```
  - MOAT Keystone: 0/1          - MOAT Verifier: 0/1
  - E2E Browser Tests: 0/1      - Entitlement API tests: 0/1
  - API Tests (3 OS): 0/3       - pip install matrix: 0/4
  - Wheel install & assets: 0/1 - Store invariants: 0/1
```

`0/N leg(s) complete` is queued, not failed. Every check that *did* report passed: OSS golden path, Cross-repo handoff, Syntax & Lint, and Drift Bot.

### Why it is not this PR's

- **The diff cannot cause it.** 9 added lines in `.github/workflows/pr-screenshots.yml`, 7 of them a comment. None of the pending checks live in that workflow — they are all `ci.yml`. `pr-screenshots.yml` is not even in the required set.
- **The pool is starved, not slow.** Those jobs were created at 18:12 and, as of 18:46, still had not been assigned a runner — 34 minutes queued with zero legs started. The gate's whole 1800s elapsed while its dependencies sat waiting for capacity.
- **Everything else on this head is green** — 30+ checks including CodeQL ×3, the full `install.sh` matrix, and `visual-diff`.
- **Known signature.** #5330 hit this exact failure earlier today (same gate, same timeout, same `0/N` pending list) and passed on re-run with no code change.

### On porting the existing fix

#5332 is the standing fix and I have read it, but I am **not** porting it here, for two specific reasons:

1. **It bundles more than the timeout.** Alongside `MAX_WAIT` 1800→3600 it sets `skip_if_unreported=True` on the Drift Bot spec and rewrites `test_a_missing_drift_bot_does_not_pass_the_gate` into `test_a_missing_drift_bot_skips_not_blocks` — a deliberate *weakening* of when Drift Bot blocks. Carrying a gate-weakening change inside a PR whose entire purpose is tightening CI would be the wrong thing to put in front of a reviewer, and it is not what this failure needs.
2. **Porting only the timeout half would collide.** It touches the same three lines of `e2e-gate.yml` and `scripts/e2e_gate.py` that #5332 edits, guaranteeing a conflict for whichever of the two merges second.

So the minimal correct action here is a re-run, not a code change.

### Next step

I have not spent my re-run on this PR. Re-running right now would start a fresh 1800s timer against the same starved pool and would likely time out again for the same reason, so I will re-run the gate once the dependency jobs have actually drained, at which point it should observe them green and pass. Keeping this PR watched until it is green or merged.

### Worth fixing at the repo level

This is the second PR today blocked by the same thing, and the cause is structural: `MAX_WAIT` is a fixed budget measured against a variable queue, so on a busy day the gate fails PRs for the pool's behaviour rather than their own. Merging #5332 (or just its `MAX_WAIT`/`timeout-minutes` half, if the Drift Bot change wants separating) would stop this recurring across every PR. Flagging rather than acting on it — that is a repo-wide call and a file this PR deliberately does not touch.

---

vivekchand:
✨ auto-fixed: merged main (14572bc) into branch to resolve BEHIND state

---

vivekchand:
✨ auto-fixed: merged main into branch (was BEHIND at 14572bc)

---

PR #5361, opened 2026-08-29
fix: roll back to the previous version when clawmetry update fails

### LINKED ISSUE #5357: In-place clawmetry update can destroy the venv install when the daemon holds clawmetry.exe (Windows)
Observed live 2026-08-29 ~20:12 UTC on a Windows 11 / Python 3.14 desktop install: the shell watcher ran `clawmetry update --unattended` seconds after 0.12.788's wheel upload, while the daemon (spawned from `venv\Scripts\clawmetry.exe`) was running. The update returned rc=1 with empty captured output, and afterwards the venv had the exe stub but **no `clawmetry` package** — pip's upgrade uninstalled the old version and failed to install the new one (the running daemon holds the exe open; Windows cannot replace open files).

Mitigations shipped (warm-launch corpse fall-through + watcher `_heal_package_corpse` reinstall), but the destructive window itself remains. Durable fix candidates, in the update path (`clawmetry update` → `clawmetry/update_respawn.py` / routes/update_check.py):
- Never run the in-place upgrade under a live daemon: stop the daemon first (the shell already restarts it on version drift), or hand the whole sequence to the shell.
- Or make failure non-destructive: on a failed upgrade, immediately reinstall the previous pinned version (`pip install clawmetry==<old>`) before exiting nonzero.
- Also worth checking: the timing (seconds after wheel upload) means the PyPI simple-index propagation race is in play — the update policy should not act on a version the index cannot yet serve (update_respawn already retries for this; the earlier stage may not).

Repro shape (faithful simulation): venv with clawmetry installed; hold the console-script exe open (spawn a long-running process from it); run the venv pip `install --upgrade clawmetry==<other-version>`; observe uninstall-then-fail leaving no package.

No-PRD context exists: requirement "Field Failure Reporting and Auto-Triage" documents the incident; the durable-fix design choice (stop-before-update vs rollback) deserves a blueprint sentence with the fix.

vivekchand:
## Reproduction

`pip install --upgrade` uninstalls the old version before installing the new one. Field-reported on Windows 11 / Python 3.14, 2026-08-29 ~20:12 UTC: the desktop shell watcher ran `clawmetry update --unattended` seconds after a fresh wheel upload, while the daemon (spawned from `venv\Scripts\clawmetry.exe`) was still running and holding the exe open. The update returned rc=1, and afterwards the venv had the launcher stub but **no `clawmetry` package at all** — pip's uninstall half of `--upgrade` succeeded, the install half failed (Windows can't overwrite an open file), and nothing rolled it back.

Faithful-simulation repro added in `tests/test_cli_unattended_update.py`: `_cmd_update` is exercised with a `subprocess.run` stand-in that fails the upgrade call (nonzero exit, and separately a `TimeoutExpired`) and asserts a follow-up pip call happens. Confirmed both new tests fail against the pre-fix code (`AssertionError: 1 == 2` — no rollback call was made) and pass after the fix.

## Fix

`clawmetry/cli.py::_cmd_update` (the desktop shell's `clawmetry update --unattended` path) did a bare `pip install --upgrade` with no rollback on failure — unlike `routes/meta.py::perform_self_update`, the daemon/dashboard self-update path, which already force-reinstalls the previous version on a failed or timed-out pip run (see `tests/test_self_update_timeout_rollback.py`). This PR adds the same rollback to `_cmd_update`: on a nonzero pip exit or a `TimeoutExpired`, it force-reinstalls the version that was running before the attempt (skipped if that version couldn't even be determined), so a failed update is never worse than a no-op.

## Tests

- `tests/test_cli_unattended_update.py`: 4 new cases — rollback fires on pip failure, rollback fires on timeout, a successful upgrade never triggers a rollback, and an unknown pre-upgrade version is never used as a reinstall target.
- Wired `tests/test_cli_unattended_update.py` into the MOAT verifier job in `.github/workflows/ci.yml` — it was a real, already-existing test file for this exact function that CI never ran (this repo's CI runs explicit file lists, not `pytest tests/`).
- `tests/test_workflow_yaml_valid.py` still passes (194 total incl. skips) after the ci.yml edit.

Closes #5357

No-PRD: fix of field-reported bug #5357

---
_Generated by [Claude Code](https://claude.ai/code/session_01Gv7MDkvzP7DPuATy7URFi3)_

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -401,6 +401,7 @@ jobs:
             tests/test_license_sync_pro_from_config.py \
             tests/test_install_script_pro_upgrade.py \
             tests/test_wake_poll.py \
+            tests/test_cli_unattended_update.py \
             -q
 
   # ── Entitlement API test suite (~2700+ hermetic tests) ─────────────────────────────────────────────
--- a/clawmetry/cli.py
+++ b/clawmetry/cli.py
@@ -5192,6 +5192,43 @@ def _print_pro_sync_result() -> None:
         print(f"clawmetry-pro {after} kept — could not confirm entitlement right now")
 
 
+def _restore_previous_install(current: str) -> None:
+    """Best-effort pip-reinstall of ``current`` after a failed upgrade (#5357).
+
+    ``pip install --upgrade`` uninstalls the old version before installing
+    the new one; if the new install then fails (live-hit on Windows: the
+    daemon still holds ``clawmetry.exe`` open seconds after a fresh wheel
+    upload, so pip's overwrite fails AFTER the uninstall already ran), the
+    venv is left with the launcher stub but no ``clawmetry`` package at all.
+    ``--force-reinstall`` regenerates the console-script entry points
+    regardless of what site-packages metadata claims, matching the rollback
+    ``routes/meta.py::perform_self_update`` already does for the
+    daemon/dashboard self-update path. Never raises; ``current == "unknown"``
+    (the version probe itself failed) has nothing to roll back to.
+    """
+    if not current or current == "unknown":
+        return
+    import subprocess as _sp
+
+    try:
+        restore = _sp.run(
+            [
+                sys.executable, "-m", "pip", "install", "--no-cache-dir",
+                "--force-reinstall", "--no-deps", "--break-system-packages",
+                f"clawmetry=={current}",
+            ],
+            capture_output=True, text=True, timeout=120,
+        )
+        if restore.returncode == 0:
+            print(f"Restored previous version ({current}) after the failed update")
+        else:
+            tail = ((restore.stdout or "") + (restore.stderr or "")).strip()[-300:]
+            print(f"Warning: could not restore previous version ({current}): "
+                  f"{tail or '(no output)'}")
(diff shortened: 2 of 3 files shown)


vivekchand:
`E2E Gate (required)` failed, but it's a 1800s polling timeout in the aggregator (`scripts/e2e_gate.py`), not a real failure of any leg. The job's own log shows most of the 12 legs it waits on (MOAT Verifier, MOAT Keystone, E2E Browser Tests, Entitlement API tests, 3/4 of the pip install matrix, Wheel install & assets, Store invariants, 2/3 of API Tests) were still `queued`/`pending` when the 30-minute window elapsed — none of them reported red. This traces to CI runner capacity: at PR-open time there was a backlog of ~15+ concurrent dependabot workflow runs on this repo unrelated to this diff, which delayed every job's start well past normal.

Not this PR's failure — the diff here is 3 files (`clawmetry/cli.py`, one test file, one CI wiring line) with no plausible connection to MOAT/E2E/entitlement test infra. I'll re-run `E2E Gate` once the underlying legs finish completing (they're progressing, just slowly), per the one-retry rule for a confirmed non-diff failure.

---

PR #5389, opened 2026-08-30
feat(otlp): sessions materialized from spans, honest runtime attribution for OTLP apps

vivekchand:
Product record: https://factory.8090.ai/project/b415065f-ab2f-4f53-8864-0c009fd098cb/requirements/8e389016-a9c8-4352-9121-72f0e361fdf6 (Runtime and Session Observability; tracked as WO-55)

## Why

An enterprise prospect runs ~160 agents on AWS Bedrock AgentCore (ADOT/Strands, OTLP GenAI spans). Pointing OTEL_EXPORTER_OTLP_ENDPOINT at ClawMetry already fills the Tracing tab, Usage and the spans table, but:

1. No sessions rows ever materialize from spans, so the Sessions tab stays empty and the runtime switcher never lists the app locally.
2. deployment.environment (dev/tst/prod) was dropped with the resource attributes.
3. Once sessions rows exist, the prefix-only runtime bucketing in query_model_rollup dumps the app's tokens and cost into the openclaw bucket, and _merge_otlp_apps_into_summary then skips the app (key already present), losing otlp=true and display_name.

## What

- LocalStore.materialize_otlp_sessions(): full recompute of the touched sessions from their spans, upserted through ingest_sessions_batch. Idempotent under OTLP at-least-once retries. Skips agent_type=openclaw (ghost-session guard) and any session owned by a richer ingest path (metadata.source ownership rule, same as put_otlp_batch).
- The /v1/traces receiver calls it once per export batch and passes deployment.environment per session; _otel_to_row keeps that attribute on the span blob.
- query_model_rollup sessions-side runtime CASE prefers a foreign agent_type over the openclaw fallback (prefix stays authoritative for family runtimes).
- _merge_otlp_apps_into_summary enriches an existing summary entry (otlp=true, display_name, max-not-sum numbers) instead of skipping.
- Session metadata carries runtime so _live_counts_by_runtime buckets liveness correctly.

## Verification

- tests/test_otlp_session_materialization.py (in the CI OTLP list): materialization, retry idempotence, openclaw ghost guard, ownership no-clobber, rollup bucketing, summary markers. Revert-proof: positive tests red on unfixed code, green restored.
- Full OTLP receiver suite: 100 passed. Neighboring builder suites pass (2 pre-existing dev-box-only failures in test_agent_inventory reproduce identically on clean origin/main).
- Live E2E on a fresh isolated server: POST AgentCore-shaped OTLP -> Sessions row (agent_type payments_agent_prod, 5007 tokens, derived cost 0.0069 from the Bedrock model id, title, env in metadata), 4-span trace tree, and exactly one inventory roster row 'payments-agent-prod (OTel)' with sessions:1 and no openclaw mis-attribution.
- make lint: 224 findings before and after (all pre-existing, none introduced).

Closes the core of WO-55.

https://claude.ai/code/session_01AHS9axpM7w3pwG7jJpFhEB

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -548,6 +548,7 @@ jobs:
             tests/test_otlp_dos_auth.py \
             tests/test_otlp_rollup_cpu_budget.py \
             tests/test_spans_otlp_edge_cases.py \
+            tests/test_otlp_session_materialization.py \
             -q
           # NOT listed: tests/test_e2e_otlp_ingest.py — it drives a LIVE
           # dashboard over HTTP (BASE_URL), which this job does not boot. It
--- a/README.md
+++ b/README.md
@@ -174,6 +174,7 @@ the same machine. Docker instructions: [docs/DOCKER.md](docs/DOCKER.md).
 | [Entitlements](docs/ENTITLEMENTS.md) | Free vs paid, tier matrix, license CLI |
 | [Approvals & policies](docs/APPROVALS.md) | Pre-execution gating, risk scoring, phone approvals |
 | [OpenTelemetry](docs/OPENTELEMETRY.md) | Export traces anywhere, ingest OTLP from anything |
+| [Bring your own agent](docs/BRING_YOUR_OWN_AGENT.md) | AWS AgentCore, Pydantic AI, LangChain end to end, with runnable examples |
 | [SDK tracking](docs/SDK_TRACKING.md) | Cost attribution for agents you built yourself |
 | [Chat channels](docs/CHANNELS.md) | The chat adapters shown in Flow |
 | [NemoClaw / OpenShell](docs/NEMOCLAW.md) | Sandboxed NVIDIA NemoClaw setups |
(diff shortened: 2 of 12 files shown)


vivekchand:
Drift-bot triage (both findings share one claim):

The claim: "_live_counts_by_runtime is not modified in this PR to actually implement that behavior."

The function already implements it on main, unmodified by this PR: clawmetry/sync.py:15752 reads `rt = str(meta.get("runtime") or "").strip().lower() or "openclaw"` inside _live_counts_by_runtime, and query_sessions_table returns metadata JSON-decoded. What this PR adds is the producer side: the materialized session stamps metadata.runtime so that existing consumer buckets the app under its own identity instead of the openclaw default. tests/test_otlp_session_materialization.py::test_live_counts_bucket_by_metadata_runtime pins the full chain (OTLP batch -> materialized session -> live counts bucketed under the app, never openclaw).

On AC-OBS-RSO-031.6 ("shall not rely on a stored value that can disagree with the session's own identity"): a span-materialized session has no session-id prefix; its only identity is the service.name carried on every span's resource, recorded at ingest. There is no independent identifier for the stored value to disagree with. Proposed Factory updates: REQ-OBS-RSO-033 (sessions for span-only applications) plus a clarifying sentence on 031.6, and a blueprint section describing the materialization mechanism, both filed as tracked suggestions on Runtime and Session Observability.

PR #5465, opened 2026-09-03
Harden CI: least-privilege top-level token permissions on 9 workflows

vivekchand:
## What

Nine workflows granted write scopes to `GITHUB_TOKEN` at the **workflow top level**, so every job in the file inherited them. Each of these files has exactly one job, so the grant was already wider than anything needed it to be — and a step added to the file later would silently inherit write access nobody chose for it. OpenSSF Scorecard's `TokenPermissions` probe flags this shape.

Each file now declares a least-privilege top level and carries its original scope set, verbatim, as a **job-level** `permissions:` block on the job that actually uses it.

```yaml
permissions:
  contents: read          # top level

jobs:
  <the one job>:
    runs-on: ubuntu-latest
    permissions:          # exactly what it had before
      ...
```

## The nine, and why each job keeps the scope it keeps

| Workflow | Job-level scopes | Why that job needs them |
|---|---|---|
| `apply-required-checks` | `contents: read`, `issues: write` | opens the "ACTION REQUIRED" issue when the admin PAT is absent |
| `auto-quarantine` | `contents: write`, `pull-requests: write`, `actions: read` | commits the quarantine marker, opens the PR, reads run history to spot flakes |
| `c6-health` | `contents: read`, `issues: write` | files the audit issue |
| `c6-pr-gate` | `contents: read`, `pull-requests: write` | posts the gate result on the PR |
| `c6-schedule-heal` | `contents: read`, `issues: write` | files the "PAT not set" issue |
| `harness-observability-audit` | `contents: read`, `issues: write` | files the audit issue |
| `i18n-autotranslate` | `contents: write`, `pull-requests: write` | commits translations, opens the PR |
| `i18n-docs-autotranslate` | `contents: write`, `pull-requests: write` | commits translated docs, opens the PR |
| `pr-screenshots` | `contents: write`, `pull-requests: write` | pushes the orphan `pr-screenshots` branch, comments on the PR |

**No release, publish or deploy workflow is touched.** `publish`, `release-on-merge`, `desktop-artifacts`, `auto-deploy-cloud`, `ci` and `moat-keystone-drive-nightly` are untouched by this diff.

**PAT note.** `apply-required-checks` and `c6-schedule-heal` do their branch-protection work with `secrets.E2E_ADMIN_PAT`, which workflow `permissions:` do not govern at all. Their `issues: write` is the `GITHUB_TOKEN` half — the issue each opens when the PAT is missing — and it moves down with the rest. Neither path changes behaviour.

## Verification

- The transformation was applied and checked **mechanically**, not by hand: for every file, the script parses the document before and after, asserts the job's *effective* permission mapping is byte-identical, asserts the new top level is exactly `{contents: read}`, and asserts no other key in the document — top level or job — moved. All nine passed; a hand edit is what would have gotten one of these wrong.
- `python3 -c "import yaml,glob; [yaml.safe_load(open(f)) for f in glob.glob('.github/workflows/*.yml')]"` → all **36** workflows parse.
- Re-scan of the tree: **0** workflows left with a top-level write scope, **0** workflows without a top-level `permissions:` block.
- `python3 scripts/check_action_refs.py` → 17 distinct action references, all still SHA-pinned.

Inline `# push to orphan pr-screenshots branch` style comments moved down with the scopes they annotate, so the rationale stays next to the grant.

No-PRD: CI-only hardening confined to `.github/`, exempt per `scripts/check_product_record.py`

https://claude.ai/code/session_01UJkXSZpJ7hX34V95sVSKdZ

---
_Generated by [Claude Code](https://claude.ai/code/session_01UJkXSZpJ7hX34V95sVSKdZ)_


vivekchand:
✨ auto-fixed: merged main (469e1d0) into branch to bring it up to date

---

vivekchand:
## `visual-diff` is red on `3720945` — runner loss, not this PR

Recording this rather than leaving it unexplained, since it's now happened twice on this branch.

**What fails.** `visual-diff` is the only red check. It is not a timeout and not a permissions problem — the runner is shut down underneath the job mid-step:

```
##[error]The runner has received a shutdown signal. This can happen when
the runner service is stopped, or a manually started runner is canceled.
```

Step 15 (`Run visual-diff`) reports `cancelled`, not `failure`, and steps 16–22 are skipped.

**Why it isn't this PR's.** Five independent lines, any one of which would be enough:

1. **Not a timeout.** Step 15 ran 9m51s and the job was 10m52s old when it died, against `timeout-minutes: 20`. The passing runs on other PRs spend 6.5–9 min in the same step, so it was inside the normal band.
2. **Not permissions.** Steps 1–14 all pass, including both checkouts. If moving `pr-screenshots`' scopes from the workflow top level to the job had broken anything, it would fail at `Checkout PR head` (step 2) or at `Publish screenshots to per-PR branch` (step 18) — never in the middle of screenshotting, with the push step never reached.
3. **The identical commit passed on retry.** `dd03253` was cancelled the same way, re-run once, and went green — publishing the branch and posting its comment. Same tree, opposite outcome.
4. **The bot's own output is non-deterministic.** Across three runs the set of tabs at 100% keeps changing (first `mobile approvals/limits/clusters/selfevolve`, then `mobile dives/agents/evals/context-economics`, now `mobile tool-catalog/tracing/turn-anatomy/evals`). A deterministic code change cannot produce a different answer on the same SHA.
5. **The diff is nine workflow YAML files.** It cannot alter a rendered pixel.

For what it's worth, the earlier occurrence was not isolated: at 17:06:52 twelve `ci.yml` jobs were cancelled in the same instant as this one, and every one of them passed on re-run with no code change.

**It does not block merge.** The job is `continue-on-error: true` ("the bot is allowed to fail without holding up a merge"), so the `PR Screenshots` workflow run itself concludes `success`. The required gate is `E2E Gate`.

**What I'm doing.** Not re-running it yet: most of the `ci.yml` matrix for this head is still queued, and a re-run now would contend for the same scarce runners the required checks need. I'll re-run it once the required checks have settled, and I'm keeping the PR watched until it merges or closes. No code change is warranted — narrowing or disabling the check to get a green square would remove a real signal to hide an infrastructure problem.

---

vivekchand:
## Correction: `E2E Gate` has now failed too — same starvation, and this one *does* block

Updating my previous comment, which said the runner trouble didn't block the merge. That was true when I wrote it. It isn't now.

**`E2E Gate (required)` failed by timeout, not by a check failing:**

```
  - pip install matrix: 2/4 leg(s) complete
```

Everything the gate aggregates passed inside the window — Syntax & Lint, Drift Bot, MOAT Verifier, API Tests (3 OS), MOAT Keystone, Entitlement API tests, Wheel install & assets, E2E Browser Tests, Store invariants. It burned its full hour waiting on two `pip install` legs (ubuntu py3.9 and py3.11) that never got a runner. Those are ~30-second jobs; the macOS and Windows legs of the same matrix finished at 18:24 and 18:12.

**Nothing has actually failed on this commit.** As of 19:15, 70 minutes after they were created, five ubuntu jobs are still `queued` and have never started: `Eval Suite Gate`, `Live OpenClaw E2E`, `OTLP receiver`, `API Tests (ubuntu)`, and the two `pip install (ubuntu)` legs. Every job that has been given a runner has passed. The same shortage cancelled `visual-diff` twice mid-step and, earlier this evening, killed twelve `ci.yml` jobs simultaneously at 17:06:52 — all of which passed on re-run with no code change.

**Plan.** I'm deliberately *not* re-running the gate yet. A re-run starts a fresh 3600s window, and with five jobs still unstarted it would likely time out again and waste the attempt. Once the queue drains and those legs report, re-running `E2E Gate` should pass in seconds, since it only polls for results that will already be there. I'll do that and keep the PR watched until it's green.

Worth noting for later, though not something to change under an incident: a required gate that fails closed on a one-hour wall-clock budget will fail on any sufficiently slow runner queue, independent of code correctness. If GitHub capacity events like tonight's recur, that budget is the knob — not this PR.

---

PR #5478, opened 2026-09-03
Harden deps: let Dependabot's npm security fixes clear the product-record gate

vivekchand:
## The problem

`scripts/check_product_record.py` treats every changed file as product-visible code unless its path is exempt. Dependency lockfiles and manifests were never on that list, so an npm bump — which touches exactly `frontend/package.json` and `frontend/package-lock.json` — fails the gate.

The pip half was already exempt, but only by accident: `EXEMPT_SUFFIXES` carries `".txt"`, so `requirements*.txt` has always skipped the check. That is the whole reason pip Dependabot PRs merge and npm ones do not. The asymmetry was extension trivia, not a decision anybody made.

It has teeth, because **Dependabot cannot write `No-PRD:` into a PR body and cannot be configured to.** So every open npm bump — #5241, #5244, #5376 and siblings — has been unmergeable from the moment it opened, one of them carrying a fix rated high severity. The repo grew a class of security update that no amount of review could land.

Reproduced against each PR's real file list before touching anything:

| PR | Changed files | Gate |
|---|---|---|
| #5376 (npm) | `frontend/package-lock.json`, `frontend/package.json` | **exit 1 — blocked** |
| #5403 (pip) | `desktop/requirements-dev.txt` | exit 0 — passes |

## The fix

Adds `EXEMPT_BASENAMES` for dependency manifests and lockfiles (`package.json`, `package-lock.json`, `npm-shrinkwrap.json`, `yarn.lock`, `pnpm-lock.yaml`), matched on the **full basename** rather than a suffix — so `clawmetry/mypackage.json` still needs a record.

`package.json` is included alongside the lockfiles. It also holds `scripts`, but `scripts/` and `.github/` are already exempt, so build tooling sits outside this gate by design; `package.json` is the npm face of the same thing. A change there that really is a product decision still has `No-PRD:` to answer to.

## What the gate keeps

The exemption is not a way to smuggle code in. A dependency bump landing next to real code still fails, because the code file is not exempt:

```
frontend/package.json + clawmetry/local_store.py  ->  exit 1
```

Every pre-existing rejection case still rejects — the suite asserts those first, on the repo's own principle that a gate which cannot fail is worse than no gate.

## Verification

- `pytest tests/test_product_record_gate.py` — **30 passed**. 9 new: the exact #5376 file list, each manifest and lockfile name, dependency-alongside-code, and three near-miss basenames (`mypackage.json`, `package.json.bak`, `package_json.py`) that a suffix test would have wrongly exempted.
- `ruff check --select E,W --ignore E501,E402` (CI's ruleset) — clean; no new findings against the pre-change baseline under ruff's default ruleset either.
- All 36 workflow files still parse under `yaml.safe_load`. No workflow is modified by this PR.

## Scope

Touches only `scripts/` and `tests/`, both exempt paths — so this PR passes its own gate on path exemption and needs no Factory record. The opt-out line below is belt-and-braces, so it still passes if the exempt paths are ever narrowed.

Merging this does not bump any dependency. It makes the already-open Dependabot PRs mergeable on their own merits; each still has to pass CI on its own.

No-PRD: CI gate logic under scripts/ and tests/, no product-visible change.

https://claude.ai/code/session_01K8ww1pN6Hwc6gaPRANbs3t


vivekchand:
**Drift-bot: clean. Test plan for review.**

This PR gates dependency manifests through the product-record check (`scripts/check_product_record.py`). Suggested test checklist before merge:

**Gate behaviour**
- [ ] A PR that adds a new dependency (e.g. a new entry in `requirements.txt` or `setup.py`) without a corresponding PRD citation fails the gate
- [ ] A PR that cites a valid PRD (or includes an explicit `No-PRD: <reason>`) with the same change passes
- [ ] A PR that modifies only non-manifest files is not caught by the new gate (no false positives)

**Scope**
- [ ] Confirm which manifest files are covered (e.g. `requirements.txt`, `setup.py`, `Dockerfile` `RUN pip install` lines); verify none are silently excluded
- [ ] A dependency version bump (no new package) is treated as the policy intends — confirm whether bumps require a PRD or are exempted

**CI integration**
- [ ] The gate runs in CI on the correct trigger (push + PR) and the failure message names the specific manifest line and the missing PRD, not a generic error
- [ ] `make lint` or equivalent local check exercises the same logic so authors can catch failures before pushing

**Regression**
- [ ] All existing PRDs in `docs/` still resolve correctly; `scripts/check_product_record.py` exits 0 on main with no changes

---

PR #5485, opened 2026-09-04
feat(signals): six behaviour signals over every transcript, rates API, alert rule, Signals tab (WO-58)

vivekchand:
Product record: https://factory.8090.ai/project/b415065f-ab2f-4f53-8864-0c009fd098cb/requirements/b6f32656-7433-4340-9e7f-7f17198283e5

Requirement id: `b6f32656-7433-4340-9e7f-7f17198283e5` (WO-58 Behaviour Signals, REQ-SIG-001 to REQ-SIG-005).

## Why

Guard reads the tool stream to notice an agent that is stuck. Nothing read the words. The person swearing at the agent, the agent refusing or handing the work back, the agent saying it could not finish, the person saying thanks: those live in the transcripts ClawMetry already holds for every runtime, and were never counted. Claude Code's own team tracks a frustration rate from a keyword list over user prompts; this brings the same family of signals to every runtime on the operator's own machine, judge-free, with no new collection.

## What

- `clawmetry/behaviour_signals.py` (new leaf module): six preset signals, each a binary match against one turn. `user_frustration`, `user_praise`, `assistant_refusal`, `assistant_laziness`, `task_failure` are precompiled word-boundary regex lists with a negation guard ("not bad", "this doesn't suck"), a positive-context guard ("pretty damn good"), an infinitive guard ("be able to stop it" is not "stop it"), and a front-of-turn rule for corrections and praise so "still broken" 3 kB into a pasted incident brief does not count. `user_retry` is a token-set Jaccard of 0.8 or more against the previous human user turn (floor: 3 tokens, so "ok" twice is not a retry). Code fences, inline code and quoted lines are stripped before matching. Turns are classified across both dialects in the store (family `message` + `data.role`; OpenClaw v3 `prompt.submitted` / `model.completed`). User turns written by a program (sub-agent sessions, SDK sessions, injected task notifications, stop-hook re-injections) are left out of the user-side denominators.
- Daemon tick (`sync.py`, next to the detectors, `CLAWMETRY_SIGNALS=0` to disable): reads new turn events past a watermark on `events.created_at` (`query_events_by_ingest`, the approvals-watcher primitive), at most 2000 events and 2 kB per turn per pass, through the daemon's own store handle. First run backfills 30 days over a few ticks.
- Store (`local_store.py`): two additive tables, `signal_turns` (one row per eligible turn, the denominator) and `signal_matches` (one row per turn and signal), `CREATE TABLE IF NOT EXISTS`, no schema-version bump. Columns: session_id, agent_type (the runtime, resolved from the session-id prefix), node_id, model, runtime_version (nullable, reported as "unknown"), turn_ts, turn_ms, event_id, signal, matcher, category, created_at. No text column exists. PKs on `event_id` and `(event_id, signal)` with `ON CONFLICT DO NOTHING`, so a re-run never double counts. Five thin store methods, allowlisted in `routes/local_query.py` `_DAEMON_METHODS`.
- `routes/signals.py` (`bp_signals`, registered in `dashboard.py`): `GET /api/signals?window=1d|7d|30d&runtime=<id>` returns per signal rate, count, eligible turns, trend against the previous window, per-day series, by_model, by_runtime_version, by_runtime, plus `coverage` per runtime and a `headline` ("Frustration is up on Cursor since Tuesday (3.2 points vs the window before)."). `GET /api/signals/<name>/sessions` lists matching sessions (id, runtime, model, started, cost, match count), never phrases. `GET /api/signals/coverage`. All reads go through the daemon proxy; `store: "unavailable"` is reported rather than a fabricated zero. Paid runtimes are already gated at ingest; the response is additionally filtered through `Entitlement.allows_runtime`. No new feature key.
- Snapshot: `signals` (node-wide) and `signalsByRuntime` slices in `sync_system_snapshot`, same shape as `/api/signals` for 1d/7d/30d, no per-session lists, wrapped in try/except.
- Alert rule `signal_rate_above` (`alert_evaluator.py`, `routes/alerts.py` `_EVALUATOR_ONLY`, `alerts.js` builder with a signal picker, window and minimum turns): fires when a named signal's rate over the window exceeds the threshold with at least `min_turns` eligible turns; existing channels and cooldowns; payload names signal, rate, window, runtime, model, never text. The signal fields live in the DuckDB mirror's `condition_json` and are carried across re-mirrors.
- Signals tab: `templates/tabs/signals.html` (`.page` wrapper, `#page-signals`), `app.js` `loadSignalsTab` (respects `_cmRuntimeFilter()`, polls only while the tab is active), `#page-signals`-scoped CSS, `en.json` keys, nav entry under Govern, `CANONICAL_TABS` and the screenshot sweep. Leads with the headline sentence, then a table per signal with drill-down into sessions (reuses the transcript deep-link), then a coverage strip with three honest states. Empty states: no daemon, nothing measured yet, not exposed by this runtime. No em dashes.
- `CLAUDE.md` and `CHANGELOG.md` updated.

## Verification

- `tests/test_behaviour_signals.py` (73 tests) and `tests/test_signals_ui_contract.py` (13 tests), both added to the `moat-tests` job in `ci.yml` and to `make test-moat`. 274 tests green across these plus the neighbouring guards (tab wrapper, c5 drift, guard census, local_query, alerts evaluator, runtime scope, DuckDB mirror, hook lifecycle) on the rebased tree.
- Red-then-green: with the wrapper changed to `<div id="signals" class="tab-content">`, `test_wrapper_is_page_with_page_id` fails (1 failed, 12 passed); with the negation guard disabled in `_guarded`, three negative fixtures fail (`this is not great`, `no thanks`, `not bad`); restored, 86 passed.
- Jinja render proof: `render_template_string(dashboard.DASHBOARD_HTML)` contains one `id="page-signals"` panel and the nav item; no em dash inside the panel.
- Precision on the dev machine's real history (a read-only copy of `~/.clawmetry/clawmetry.duckdb`, last 30 days: 27,554 turn events, 1,674 human user turns and 19,979 assistant turns after filtering programmatic sessions). Every `user_frustration` match was hand-checked: 14 matches, 13 clear true positives, 1 borderline ("dumb human beings like me", self-deprecating), no false positives; that is 93% strict, 100% lenient. `user_praise`: 12 matches, 10 true positives, 2 false positives (83%). `user_retry`: 101 matches, spot-checked: genuine human re-sends plus a /loop automation re-sending the same prompt, which the requirement already calls a floor. `assistant_refusal` 3 and `assistant_laziness` 3 matches (too few to rate beyond a spot check; 1 of the refusals is a safety statement rather than a refusal). `task_failure`: 0 matches on this machine. Earlier iterations found and fixed three false-positive classes: stop-hook re-injections of the user's own text, "still broken" deep inside pasted briefs, and "be able to stop it".
- CPU: the full 30-day scan (27.5k events) took 16 s single-threaded, so a 2000-event pass costs about a second during backfill and near nothing afterwards.
- Lint: `ruff` clean on the new files (the repo-wide `lint-py` count moved from 227 to 228 only because one more `from routes.* import` line joins the existing E402 block, which CI ignores), `lint-daemon-allowlist` OK, py39 annotation check OK, `node --check` OK on `app.js` and `alerts.js`, `en.json` parses.
- Pre-existing, unrelated: `tests/test_beginner_nav_phase_a.py::test_tier1_order_and_membership` already fails on `origin/main` (nav order there is overview, inventory, brain, transcripts, usage, ...; the test expects usage before transcripts).

## Not in this PR

- The hosted dashboard needs a `cm-cloud-signals` interceptor in `clawmetry-cloud` to serve `/api/signals` from the `signals` / `signalsByRuntime` slices this PR emits.
- No pro adapter declares `signal_coverage()` yet; coverage is inferred from the store until one does.
- `runtime_version` is not present on any event or session row this store holds, so the by-version slice reports "unknown" for every runtime for now.

Session: https://claude.ai/code/session_01ERBVJarDT4Vw1hJox5efBB

## Updates after the first CI run

- Rebased onto `origin/main` at `8b06621aa` (cohort compare, WO-60). Conflicts in CHANGELOG.md and Makefile were keep-both.
- C5 tab-list parity: `.github/scripts/visual-diff.mjs` DEFAULT_TABS is the third source and now lists `signals`; the UI contract test checks all three.
- MOAT batch: an earlier test in the job leaves a daemon discovery file behind, so `get_store()` handed the fixture a `_ProxyStore` and every write no-oped (six "0 == 8" failures). The fixture now instantiates `LocalStore` on the isolated file, flushes synchronously, asserts the seed landed, and the route test points `routes.signals._ls_call` at that store. Verified locally with proxy-leaving neighbours (`test_local_query_api`, `test_local_query_dispatch_edge_cases`, `test_transcript_local_store`, `test_hook_lifecycle`) run first: 163 passed.
- Seam with WO-60: `cohort_queries` already reads `signal_matches` when the table exists (its column resolution matches this schema), and its test asserted the table was absent. Updated that assertion to `signals == []` for an unmatched session; `tests/test_cohort_compare.py` 50 passed.
- The `visual-diff` red on the earlier head was a cancelled run (superseded push), and it flags the new tab by design (a new page differs from base).


vivekchand:
**Drift-bot: clean. Test plan for review.**

This PR adds six behaviour signals computed over every transcript, a rates API, an alert rule, and a new Signals tab (WO-58). Suggested test checklist before merge:

**Signals computation**
- [ ] For a known transcript with a tool blast radius event, confirm the `file_blast_radius` signal fires with correct `spend_at_risk_usd` and `threshold_source` (not hard-coded — should reflect `resolve_thresholds` output)
- [ ] Confirm signals that cannot be measured return `available: false` with a `basis`, not a fabricated zero
- [ ] Check that `warning` → `critical` promotion only occurs when `spend_at_risk_usd` exceeds `CLAWMETRY_GUARD_CRITICAL_USD`, not for every warning

**Rates API**
- [ ] `GET /api/usage/signal-rates` (or equivalent) returns per-signal rates; confirm the denominator is sessions with measured coverage, not total sessions
- [ ] An empty session set returns `available: false`, not `0/0 = NaN`

**Alert rule**
- [ ] Create an alert rule targeting one of the six signals; confirm it fires on a matching transcript and suppresses on a clean one
- [ ] Confirm the alert carries `threshold_source` and `spend_at_risk_usd` fields through to the webhook payload

**Signals tab (UI)**
- [ ] The new `desktop__signals` and `mobile__signals` views render without JS errors in a session with signals data
- [ ] An empty session (no signals yet) renders a graceful empty state, not a blank panel or an uncaught exception
- [ ] Visual diff shows new signal cards — eyeball the before/after screenshots for layout regressions on adjacent tabs

**Regression**
- [ ] `make test` passes; existing Guard tab and `/api/loop-signals` ranking unaffected

---

Issue #5496, opened 2026-09-04
Store: skip unchanged span and session upserts, batch the rest


vivekchand:
## Problem
`ingest_span` is `DELETE` then `INSERT` per span, in its own transaction, under the write lock, with six index updates. `ingest_session` is an `ON CONFLICT DO UPDATE` per session, which DuckDB executes as delete plus insert on indexed tables. Neither checks whether the row changed. On every daemon restart the family adapters re-emit every span and session they can see, so the whole set is rewritten.

Evidence from a real store: 91,590 live spans carry 4,278,615 row versions on disk (47 per span); 1,926 sessions carry 96,811 (50 per session); `session_phase` carries 150 per row. Bursts of 44,351 and 28,896 span rewrites line up with daemon restarts. Only 1,377 of 91,590 spans have a creation stamp older than 24 h.

## Fix
- Keep an in-memory map of the last written content hash per `span_id` and per `(agent_type, session_id)`; return early from `ingest_span` and `ingest_session` when the hash matches. Seed the map lazily from the table on first miss so a fresh daemon does not rewrite everything once.
- Buffer surviving spans and write them with one `executemany` inside one transaction per flush, instead of one transaction per span.

## Done when
- Restarting the daemon against an unchanged transcript set writes zero span and session rows (assert via `estimated_size` or a write counter).
- A span whose `end_ts` or `cost_usd` changed is still overwritten.

PR #5503, opened 2026-09-04
perf(store): compact the DuckDB file at writer startup when it is mostly dead space

### LINKED ISSUE #5498: Store: compact the DuckDB file on startup when it is mostly dead space
## Problem
DuckDB never shrinks its file and only merges row groups that are about 25% deleted, so the delete-and-reinsert churn from restarts (see the upsert dedupe issue) accumulates forever. A real store measured 1,634 MB on disk (2,108 of 6,233 blocks free, plus dead row versions inside used blocks); the same data rewritten with `COPY FROM DATABASE` is 575 MB. `CHECKPOINT` and `VACUUM` reclaimed nothing on this file. Cold-cache reads after each daemon restart pay for the extra size (610 ms vs 60 ms for the recent-events query).

## Fix
At daemon startup, when `pragma_database_size()` reports free blocks plus dead versions above 40% of the file, rewrite the store with `COPY FROM DATABASE` into a temp file next to it and atomically swap, reusing the copy-swap path that the schema v14 migration already uses. Log before/after sizes. Keep a `CLAWMETRY_AUTO_COMPACT=0` kill switch. Never run while the writer connection is open; do it before `get_store()` opens the writer.

## Done when
- The reference store compacts from 1.6 GB to about 575 MB on the next daemon start with no data loss (row counts per table identical before and after).
- Startup on an already-compact store adds no measurable time.

vivekchand:
Closes #5498.

## Why
DuckDB never shrinks its file and only merges row groups that are about 25% deleted, so the dead row versions left behind by rewrites (restart re-ingest before #5496, retention deletes, heartbeat churn) accumulate for the life of the file. A real node measured 1,634 MB on disk for data that a rewrite brought to 575 MB. `CHECKPOINT` and `VACUUM` reclaimed nothing on that file, and every cold start after a release paid for the extra size (610 ms vs 60 ms for the recent-events query).

## What
At writer startup, before the daemon opens the connection it keeps for its lifetime, the store:
1. checkpoints the file (folds the WAL in) and measures dead space: free blocks plus dead row versions, which `duckdb_tables().estimated_size` counts and `COUNT(*)` does not;
2. when the file is over 64 MB and at least 40% dead, rewrites it with `COPY FROM DATABASE` into `<db>.compacting`, verifies every table's row count matches, and swaps it in;
3. keeps the previous file as `<db>.pre-compact` for 7 days, then removes it on a later start.

A file another process holds, a row-count mismatch, or any error leaves the original untouched and cleans up the temp file. The outcome is reported in `/api/local/health` as `last_compaction`.

Knobs: `CLAWMETRY_AUTO_COMPACT=0` (off), `CLAWMETRY_AUTO_COMPACT_MIN_DEAD_PCT` (0.40), `CLAWMETRY_AUTO_COMPACT_MIN_BYTES` (64 MB), `CLAWMETRY_AUTO_COMPACT_BACKUP_KEEP_DAYS` (7).

## Verified
- Against a copy of the real 1.6 GB store: 1,634 MB to 575 MB in 22 s; spans, sessions and events row counts identical before and after; all 80 indexes carried over; zero dead rows after.
- 7 tests in `tests/test_store_startup_compaction.py`: dead-space measurement sees rewrite versions; a mostly-dead file is compacted on the next writer start and keeps accepting writes; small and healthy files are left alone; the kill switch skips the check; a file held by another connection is not touched; an old backup is pruned and a young one kept; a row-count mismatch aborts with the original in place and no temp file.
- Existing bulk-flush, relay-integration, local-query and invalidated-recovery suites pass. `ruff` and the py3.9 annotation check are clean. Wired into the CI MOAT file list.

## Disk note
The first compaction on a bloated node temporarily doubles the store's footprint (old file kept for 7 days). That is the pathological case this exists for, and the backup is what makes a bad rewrite recoverable by hand.


vivekchand:

---

PR #5504, opened 2026-09-04
perf(store): drop the nine range indexes on events and spans; keep point lookups only

### LINKED ISSUE #5500: Store: drop the unused range indexes on events and spans
## Problem
`_DDL` in `clawmetry/local_store.py` creates about 80 explicit ART indexes, including `ts`-only and `(event_type, ts)`, `(agent_type, ts)`, `(trace_id, start_ts)` composites on `events` and `spans`. DuckDB's own guidance is to index only for point lookups returning under 0.1% of rows, because zone maps already prune range scans on `ts`, and every index slows inserts and updates and must fit in memory. In the profiled plans for the dashboard's hottest queries none of the range composites was used. They cost about 115 MB on disk on a real store and are the reason updates become delete-plus-insert.

## Fix
Keep primary keys, `(session_id, ts)` style lookups and the few point-lookup indexes (`span_id`, `trace_id` equality, `parent_span_id`). Drop the `ts`-only and `(x, ts)` range composites. Bump the schema version so existing stores drop them on upgrade; `heal_index_corruption` already knows how to drop and rebuild.

## Done when
Profiling the dashboard's top ten store queries on the reference store shows no plan that regressed, and flush time per batch decreases.

vivekchand:
Closes #5500.

## Why
The store carried nine range indexes on `events` and `spans`: `ts`, `(agent_id, ts)`, `(event_type, ts)`, `(agent_type, ts)`, `(session_id, ts, event_type)`, `created_at`, and on spans `ts`, `(agent_type, start_ts)`, `(trace_id, start_ts)`. DuckDB's guidance is to index only for point lookups returning under 0.1% of rows; zone maps already prune range scans on time columns because rows arrive in time order. None of the nine appeared in a profiled plan for the dashboard's hottest queries, they cost about 115 MB on a real store, and they turned every UPDATE on the table into a DELETE plus INSERT with nine extra index maintenances.

## What
- `events` keeps `(session_id, ts)`; `spans` keeps `(trace_id, span_id)`, `parent_span_id` and `(session_id, start_ts)`.
- The boot DDL carries `DROP INDEX IF EXISTS` for the nine retired names, so an existing store sheds them at its next start. The index-heal path rebuilds only the kept set.

## Verified
Eleven representative dashboard queries timed on the compacted real store (576 MB), best of three, 2 threads:

| query | with indexes (ms) | without (ms) |
|---|---|---|
| recent events, top 500 | 12.2 | 14.9 |
| events in the last hour | 2.5 | 3.8 |
| events by session | 6.4 | 7.4 |
| events by type, 5000 | 9.1 | 6.6 |
| events by agent since a day | 3.3 | 2.5 |
| daily aggregates | 9.3 | 7.1 |
| spans by trace | 0.5 | 0.8 |
| spans by session | 0.5 | 0.5 |
| spans by agent, recent | 0.6 | 0.6 |
| spans by parent | 0.2 | 0.4 |
| retention scan on created_at | 0.1 | 0.1 |

- `tests/test_store_range_indexes_dropped.py`: a fresh store has only the kept set; a store carrying the old indexes sheds them on boot with reads and writes intact. The relay-integration and dedup-migration index assertions are updated to the new set (the latter file already fails on `main` in a local run for an unrelated connection-config reason and is not in any CI file list).
- Relay integration, span ingest, brain time range, bulk flush and invalidated-recovery suites pass. `ruff` and the py3.9 annotation check are clean. Wired into the CI MOAT file list.


vivekchand:

---

PR #5505, opened 2026-09-04
[RELEASE] perf: newest-events scan sorts keys first, no longer reads the whole payload column per poll

vivekchand:
## Why

On a laptop with 122 MB free and 13 GB in the compressor, the daemon's DuckDB buffer pool was paged out between dashboard polls, and "the newest 300 events" (Brain feed, Home poll) took 13 to 34 seconds live, six times a minute. The same call on a fresh copy of the store took 0.4 s. The query keeps the newest N rows of a 215k-row table with no selective predicate, so DuckDB's TOP_N materialises the 1 KB `data` blob for every row before discarding all but N: the whole column, every poll, through swap.

## What

When nothing narrows the scan (only `exclude_daemon`, which rejects almost nothing), `query_events` sorts the narrow key columns first and fetches the N payloads by primary key. Cold on the same store: 0.41 s to 0.11 s. Shapes with a selective predicate (session, agent, event type, time bounds, runtime) keep the single-statement form, because there the two-phase form is slower (0.01 s to 0.16 s for one session).

## Verification

- `tests/test_query_events_keys_first.py` (added to the CI store list): identical rows and order versus the plain query for every shape, ties included; the two-phase form is chosen only on the unfiltered shape. Red on the unchanged query, green on the fix.
- Existing store tests pass: brain local store, DuckDB fast-path invariants, query contract goldens, local query API (43 tests).
- `ruff` count on `clawmetry/local_store.py` unchanged from main.

## Not in this PR

The machine-level cause was memory pressure from unrelated processes (58 Claude Code processes, Terminal at 3.8 GB). Main already has concurrent cursor reads, multi-row batched flushes, and the persisted family high-water skip, so those earlier hypotheses needed no change.

## Product record

- Blueprint: Local Agent Observability, https://factory.8090.ai/project/b415065f-ab2f-4f53-8864-0c009fd098cb/blueprints/e42834c9-33d0-4ee8-b1f3-423970a6527a

https://claude.ai/code/session_01JqeUMB1GPqXU3AJgDmg3mW


vivekchand:

---

Issue #4817, opened 2026-08-14
Hook lifecycle contract — install manifest + clean uninstall (runtimes must not error out when clawmetry is removed)


vivekchand:
## Context

Several runtimes persist only the *decision* of an approval prompt, never the *prompt text the user saw* — so replay can show "approved" but not "approved WHAT modal." For those runtimes, we need to install a small hook (Claude Code hook, OpenClaw plugin, or shim) that captures the missing signal.

**Non-negotiable requirement (from goal thread):**
> "When ClawMetry is uninstalled it should cleanly remove all of them — say if we installed hook for [any runtime] — that also should be removed so that the agent runtime should not error out when ClawMetry is uninstalled."

This meta issue lands the contract every runtime-hook installer must satisfy.

## Deliverables

**1. Hook install manifest**
- New JSON file `~/.clawmetry/hooks/installed.json` — the authoritative record of every hook ClawMetry has installed. Schema:
```json
{
  "hooks": [
    {
      "id": "claude_code.approval_prompt_capture",
      "runtime": "claude_code",
      "purpose": "capture approval prompt text for replay",
      "install_path": "~/.claude/hooks/clawmetry-approval-capture.js",
      "target_config": "~/.claude/settings.json",
      "target_config_key": "hooks.PreToolUse",
      "backup_path": "~/.clawmetry/hooks/backups/claude_code_settings_20260814T123000.json",
      "checksum": "sha256:...",
      "installed_at": "2026-08-14T12:30:00Z",
      "clawmetry_version": "0.12.694"
    }
  ]
}
```
- No hook may exist on disk that isn't recorded here. Any hook file we drop MUST also register.

**2. Hook installer/uninstaller API** — new module `clawmetry/hooks.py`:
```python
def install(hook_spec: HookSpec) -> None: ...
def uninstall(hook_id: str) -> None: ...
def uninstall_all() -> list[str]: ...  # returns list of removed hook_ids
def status() -> list[HookStatus]: ...
def verify_all() -> list[HookVerdict]: ...  # detects manual tampering / missing files / broken targets
```

Every installer MUST:
- Back up the target config file to `~/.clawmetry/hooks/backups/` **before** modifying it
- Modify config atomically (write to tmp, rename)
- Compute + store a checksum of the installed hook file
- Register in `installed.json` atomically
- Fail loud (don't leave partial state)

Every uninstaller MUST:
[...]

PR #5509, opened 2026-09-04
Fix Claude Code hook ownership when the launcher path is quoted (P0)

vivekchand:
No-PRD: field-reported P0 found on a live machine during this session; fix is a defect repair with no product surface change.

## The bug

ClawMetry shell-quotes the launcher in the Claude Code hook command it writes, because that path routinely contains a space. Ownership was then decided by looking for the marker `clawmetry hook claude-code` inside the **raw** command string:

```
'/Users/me/Library/Application Support/ClawMetry/runtime/venv/bin/clawmetry' hook claude-code --base http://127.0.0.1:8900
                                                                            ^ the marker expects a space here
```

So the test answered "not ours" for every hook we had just written. Two failures fall out of the one bug:

- **Install could not see its own previous entry**, so instead of replacing it, it appended another — once per pass, forever. The machine this was found on had **4,955 `PreToolUse` entries and a 1.6 MB `settings.json`**, rewritten every couple of seconds.
- **Uninstall could not see it either**, so the entry survived pointing at a deleted binary. Claude Code then reported a hook error on *every single tool call*:

```
PreToolUse:Bash hook error
Failed with non-blocking status code: /bin/sh: /Users/…/ClawMetry/runtime/venv/bin/clawmetry: No such file or directory
```

This is not a corner case. It is every macOS desktop-app install (`~/Library/Application Support/ClawMetry/…`), every Windows install under `C:\Program Files`, and every user whose home directory contains a space.

Two further gaps kept the wreckage in place:

1. `clawmetry uninstall` drains the `clawmetry.hooks` registry, which never knew about the gate and mirror hooks that `claude_code_gate.py` installs — so a full uninstall walked straight past both.
2. The gate's `_uninstall` returned early whenever its own state file was missing, which is exactly the state a purged `~/.clawmetry` or a trashed `.app` leaves behind. Nothing left running could ever clean it up.

## The fix

- `hook_ownership.normalize_command` strips shell quoting and the Windows `.exe` suffix. Every ownership test in `claude_code_gate.py` and `hooks_claude_code.py` runs against it, so ownership no longer depends on where the launcher lives.
- `hook_ownership.command_binary` / `command_binary_exists` parse with `shlex` instead of `cmd.split()[0]`, which returned `'/Users/me/Application` for a quoted path and inverted every staleness check built on it.
- A **stateless uninstall sweeps** entries that carry our marker *and* name a binary that no longer exists — they can only be ours, and can only fail — while still leaving a **live** hook alone, since without state that one may be the operator's own manual install. Same sweep for the `PermissionRequest` mirror.
- `clawmetry uninstall` calls the new `claude_code_gate.uninstall_all_hooks()` before the pip uninstall, while the module still imports (the #4817 rule).

Hook-level pruning throughout, so a co-resident writer's command in a shared entry is never touched.

## Verified

11 new tests in `tests/test_hook_ownership_quoted_launcher.py`; **7 fail on the unpatched code**, including:

- the runaway directly — five installs under a quoted launcher must leave exactly one entry;
- both halves of the stateless-uninstall rule — a dead entry is swept, a live one is not;
- a co-resident writer's hook surviving repeated installs.

Existing `test_hooks.py`, `test_hooks_claude_code.py`, `test_runtime_gates_and_hooks.py`, `test_mirror_gate.py`, `test_hook_collision.py`, `test_hook_lifecycle.py` and the uninstall suites are unchanged against this branch. One pre-existing failure (`test_cc_gate_windowless_python_swap`) fails identically on `origin/main`. Zero new ruff findings over baseline on the touched files. Wired into the CI file list and `make test-hooks`.

https://claude.ai/code/session_01YQDapoJwHre4krmKRBpB7S


vivekchand:

---

vivekchand:
✨ auto-fixed: resolved CHANGELOG.md merge conflict (kept both the PR's P0 hook-ownership entry and the newer main entries under ## Unreleased) and pushed merge commit.

---

vivekchand:

---

vivekchand:

The required E2E Gate is failing because Drift Bot reports undocumented code changes vs. the 8090 Software Factory blueprints. Resolving this requires updating the blueprint on the 8090 platform — that's an author action outside what this sweep can do mechanically.

---

vivekchand:
✨ auto-fixed: merged main into branch (was BEHIND)

---

PR #5516, opened 2026-09-04
fix(inputs): label session_context rows with the runtime that produced them

vivekchand:
## The bug

"What the agent was given" read **"Nothing captured for this session yet"** on all 60 Claude Code sessions on the founder laptop, and the same on Codex, Cursor and every other paid runtime. OpenClaw sessions were fine.

The inputs were being captured the whole time. Every paid adapter prepends a `context.compiled` event to its event list (`clawmetry_pro/adapters/_inputs.py`), the store fingerprints it into `session_context` at the ingest chokepoint, and 15 rows were sitting in that table on the live store.

They were stamped `agent_type = 'openclaw'`:

```
('openclaw', 'claude_code:e77b6948-...', 'user_prompt',      146, 1)
('openclaw', 'claude_code:e77b6948-...', 'tools_available',    8, 1)
('openclaw', 'claude_code:e77b6948-...', 'runtime_meta',     173, 1)
```

`rows_from_event` read the label off the event's `agent_type` column, which the family ingest in `sync.py` never sets (the runtime rides in `data._runtime`). `GET /api/sessions/<id>/context?runtime=claude_code` filters on that label, so with the runtime switcher on anything but "all" the panel filtered out its own data and fell through to the honest-looking empty state.

Second, quieter miss: a family session already ingested is skipped until it grows again, so 234 of the 239 Claude Code sessions on that machine had never been read for their context event at all.

## The fix

1. **`session_context.runtime_of_event`** names the runtime from what was actually stated, in order: the event's declared `agent_type` (the OpenClaw trajectory reader sets it), then `data._runtime` (the family ingest stamps the adapter's own runtime there), then the session-id prefix through the one shared resolver, `waste_flags.runtime_from_session_id`. With none of the three it stays `openclaw`, the only Free runtime, and never guesses. This is the same rule the sessions list, `detectors._runtime_of` and the runtime switcher already follow: the prefix is the authority, `agent_type` is a fallback.
2. **A boot migration** relabels rows an older build wrote, from the session-id prefix, so history the panel already had comes back rather than only sessions ingested since. Where both labels exist the stale copy is dropped rather than merged, so the relabel cannot violate the primary key. Idempotent, guarded by a count that goes to zero.
3. **The family ingest rev salt is bumped** (`/cwd1` to `/ctx1`), which re-reads every already-seen session once and backfills its context event. Same mechanism used for the `cwd` backfill in August.
4. **`turns` advances only on a strictly later timestamp.** The family ingest re-reads a whole session every time it grows, so the old unconditional bump counted re-ingests, not turns: a session whose runtime emitted one context event was already claiming "seen on 4 turns".

## Why it was never caught

The two existing family tests set an `agent_type` key on the event that the daemon does not set, so they exercised a row shape that only exists in the test. They now use the shape `sync.py` actually writes.

## Verified

**Guards, red on the unfixed code** (revert to red, restore to green, all 5):
- `test_runtime_of_event_prefers_declared_then_stamped_then_prefix` (three sources plus the no-guess floor)
- `test_family_rows_are_findable_under_their_own_runtime` (findable under `claude_code`, absent under `openclaw`)
- `test_reingesting_the_same_event_does_not_invent_turns` (four re-ingests stay at 1 turn; a later occurrence counts)
- `test_migration_relabels_rows_written_under_the_old_code`
- `test_migration_drops_the_stale_copy_instead_of_colliding`

`tests/test_session_context_ingest.py` is already in the CI file list. 15 pass here, plus `test_session_context_route.py`, `test_trail_coverage_route.py`, `test_trail_tab_template.py` (24) unchanged. No new ruff findings (231 before and after on `dashboard.py clawmetry/`).

**Live, on the founder's own store and daemon** (patch applied to the installed 0.12.810 copy, daemon restarted):

```
GET /api/sessions/claude_code:e77b6948-.../context?runtime=claude_code
before: {"count": 0, "items": []}
after:  {"count": 3, "basis": "measured", "items": [runtime_meta, tools_available, user_prompt]}
        runtime_meta: cwd, gitBranch, model=claude-opus-5, permissionMode, version, entrypoint
```

The migration relabelled the existing rows on open, and the rev bump is walking the 145 Claude Code sessions now.

## Not in scope

`events.agent_type` is still `openclaw` for family rows. Everything that needs the runtime already derives it from the session-id prefix, and changing that column would touch far more than this panel, so it stays out of this fix.

"What the agent reported" (the panel below this one) is correctly empty here: it needs the ClawMetry MCP server registered with the runtime, and `~/.clawmetry/mcp_installed.json` is `{}` on this machine. `clawmetry mcp install --runtime claude_code` is what fills it.

No-PRD: bug fix. No new behaviour; data already captured and stored was being filtered out of its own panel by a wrong label.

https://claude.ai/code/session_01HdV1KoP3rMFk88Zzbh14Kr


vivekchand:
Live verification on the founder's own daemon (patch applied to the installed 0.12.810 copy, daemon restarted).

The panel on a Claude Code session now renders:

```
📥 What the agent was given            3 tools · claude-opus-5 · measured
First request      146 B · bbfc869f238b                            [Show]
Tools it could use 3 tools
   Bash   Monitor   ToolSearch
Setup  model claude-opus-5 | cwd /Users/vivek/projects/clawmetry |
       version 2.1.261 | permissionMode auto | tools_count 3 |
       entrypoint cli | gitBranch add-pro-readme
```

Before the fix the same session rendered "Nothing captured for this session yet."

The migration also healed the inflated counts on that store: the same rows read "seen on 5 turns" before, and the adapter emits exactly one context event per session. They now read 1 turn and the label is dropped.

The rev bump is walking the 145 Claude Code sessions on that machine and filling their context rows in as it goes.

"What the agent reported" stays empty there, correctly: `~/.clawmetry/mcp_installed.json` is `{}`, so no runtime has the ClawMetry MCP server registered and no agent can send a note. That panel already says so and names the command.

PR #5506, opened 2026-09-04
feat(signals): shift detection opens issues, scheduled briefs to alert channels (WO-62)

vivekchand:
## Why

Behaviour Signals (WO-58, #5485) produce rates. A frustration rate that doubles on Tuesday was a bump on a chart nobody opened. Sentry opens an issue for a new exception; the equivalent here is a signal rate leaving the band its own history taught. The other half is delivery for the operator who does not open dashboards: Dives already turns a question into a query, the narrator already writes a few sentences, the daemon already runs a scheduler thread, and nothing wired them into one message that arrives on Monday morning.

Product record: https://factory.8090.ai/project/b415065f-ab2f-4f53-8864-0c009fd098cb/requirements/0904463d-6472-4812-978c-86d541f1cca5

## What

**Shift detection** (`clawmetry/signal_shifts.py`, pure functions + `run_shift_tick(store)` called from the daemon's existing signals tick):
- Per (signal, runtime): last 24h rate vs the runtime's own 28d history. Band = `mean + k * spread` over daily rates, clamped the way `detector_calibration` clamps Guard thresholds (floor: 2 points above the mean so a flat history cannot fire on noise; ceiling: 5x the mean so a noisy one cannot put the threshold out of reach). Every issue records `threshold` and `threshold_source` (learned / floor / ceiling) inside its `breakdown_json`; the table has no separate column for them.
- Minimum samples on both sides: `CLAWMETRY_SHIFT_MIN_SHORT` (30), `CLAWMETRY_SHIFT_MIN_HISTORY` (200), `CLAWMETRY_SHIFT_K` (2.0). No LLM.
- Table `signal_issues` (additive, CREATE IF NOT EXISTS): one open issue per (signal, runtime); a second tick refreshes; a resolved issue reopens with `reopen_count++` once its window has aged out (so resolving a live shift does not reopen it on the next tick); ignored stays silent until the operator reopens.
- `breakdown_json`: ranked split by model / runtime version / tool / cwd basename, each with rate before/during and share of the shift explained (excess matches beyond what the value's own before-rate predicts, so a value that merely got busier does not rank).
- Delivery: built-in rule kind `signal_shift` through the daemon's local alert path (`_persist_local_alert_banner` + generic webhook when entitled), cooldown 6h per (signal, runtime). Payload: signal, runtime, rates, sample sizes, top breakdown line, link to the Signals tab. Never text.
[...]

PR #5520, opened 2026-09-04
feat(signals): briefs snapshot slice for the hosted dashboard; cloud-mode controls say where to act

vivekchand:
## Why

Follow-up to #5506 (signal shifts + briefs). Two gaps found while building the hosted-dashboard interceptors:

1. `sync_system_snapshot` emitted `signalIssues` but no `briefs` slice, so the hosted Signals tab had nothing to list.
2. On the Signals tab the brief form, Resolve / Ignore on issues, the daily-digest "Switch on" and Delete stayed live under `window.CLOUD_MODE` and ignored every response. A click on the hosted view did nothing and said nothing.

## What

- `clawmetry/briefs.py::build_snapshot_slice(store, limit=50)` — the shape `GET /api/briefs` serves (`briefs`, `count`, `max`, `channels`, `offered`) plus `generated_at`. Capped at 50 rows (a caller cannot lift it above `SNAPSHOT_MAX`), title and question the only free text, `{}` on any failure.
- `clawmetry/sync.py` — `payload["briefs"]` built right after the `signalIssues` slice, wrapped in the same try/except so a failure never breaks the snapshot.
- `clawmetry/static/js/app.js` (Signals section) — `_sigCloudReadOnly` / `_sigCloudNote` / `_sigDeclined` / `_sigReasonOf` / `_sigDisabledAttr` helpers. Under `CLOUD_MODE`: Resolve / Ignore, Switch on/off, Run now, Delete, the digest Switch on and the "Add a brief" button render `disabled` with "Manage on the local dashboard"; the form never opens. `signalsEnableDigest`, `signalsToggleBrief`, `signalsRunBrief`, `signalsDeleteBrief`, `signalsSaveBrief` and `signalsSetIssueStatus` show the server's `reason` (then `error`) when the answer carries `available:false` or `ok:false`; a refused fetch says so too. `loadSignalsBriefs` renders the reason of an `available:false` answer.
- `clawmetry/templates/tabs/signals.html` — `id="signals-briefs-add"` on the add button so cloud mode can disable it.
- `clawmetry/static/locales/en.json` — `signals.cloud_manage_local`, `signals.brief_delete_err`.
- `CHANGELOG.md` — Unreleased entry.

No em dashes or double-dash joins in any user-facing string (the existing copy guard covers the new strings).

## Verification

- `tests/test_briefs.py` +3: slice shape and the digest offer, the 50-row cap and `{}` on a store that raises or lacks the method, the `sync.py` wiring (slice emitted after `signalIssues`, wrapped with the debug log).
- `tests/test_signals_ui_contract.py` +8: disabled state with the note (2 issue buttons, 4 brief buttons, the add button, form never opens), one parametrised guard per write handler (cloud stop + reason display + no silent `.catch(function () {})`), helper order (reason before error), loader reads `available:false`.
- Red proof: copied both test files into a throwaway worktree at origin/main; all 11 new guards fail there, 58 pass on the branch.
- `make lint` targets: `lint-py39`, `lint-js`, `lint-daemon-allowlist`, `lint-runtime-count`, `lint-ac-coverage` all OK. `lint-py` (ruff) reports the same pre-existing findings on `dashboard.py` / `sync.py` as origin/main; the touched-file finding set is byte-identical to main (no new findings).

Product record: https://factory.8090.ai/project/b415065f-ab2f-4f53-8864-0c009fd098cb/requirements/0904463d-6472-4812-978c-86d541f1cca5

https://claude.ai/code/session_01ERBVJarDT4Vw1hJox5efBB

Spec: the "Signal shifts and scheduled briefs" blueprint (0b1571ea) and requirement (0904463d) were amended to carry the briefs snapshot slice, the hosted read-only controls and the reason display, so drift-bot has a spec to compare against.


vivekchand:
**Drift-bot triage (runs on a80106f5f).** All findings are the same three new capabilities, none is a base misread:

1. `briefs` snapshot slice (`clawmetry/briefs.py::build_snapshot_slice`, `sync.py` `payload["briefs"]`)
2. Signals tab write controls disabled under `CLOUD_MODE` with "Manage on the local dashboard"
3. Write handlers show the server's `reason` on `available:false` / `ok:false` instead of a silent no-op

The requirement already says briefs "appear in the snapshot"; the blueprint only spelled out `signalIssues`. Appended to the "Signal shifts and scheduled briefs" blueprint (`0b1571ea-69b5-4dc1-8ace-bddda1e36071`): the daemon composition bullet now names both slices, the Briefs component gains `build_snapshot_slice` (fields, 50-row cap, `{}` on failure), the Signals tab component gains the hosted read-only rule and the declined-answer rule, and the Snapshot integration contract carries the `briefs` shape. Nothing in the code diverges from the requirement's intent (briefs on the surface and in the snapshot; plain-words honesty when something cannot happen).

PR #5524, opened 2026-09-04
perf(cost): ship slim event rows and share one round trip for the Cost tab scans

vivekchand:
No-PRD: field-reported performance bug on the founder node; diagnosed and fixed in-session.

## What the user saw

The Cost tab sat on "Loading..." for tens of seconds, and **after** attempting it every other tab failed: Brain showed "Failed to load: timeout", Agents drew nothing, and Sessions said *"No Claude Code sessions have a transcript yet"* while its own header counted **61 sessions**. That last one is the worst of the set — the UI stated a confident falsehood about the user's data.

## Root cause

The Cost tab's roll-ups (token split, per-plugin, per-plugin trend, per-model, per-skill, session clusters, runtime summary) live behind **separate endpoints the browser fires concurrently**, and each one independently asked the daemon for the newest 20k-50k events.

Measured on a real 46k-row store:

| | |
|---|---|
| one `query_events(limit=20000)` | **37.7 MB** over the daemon RPC |
| of which SQL | **0.9 s** |
| decode + `dumps` + `loads` | **~5 s** cold, far more on a loaded box |
| such scans per Cost page load | **21** |

The daemon serializes those on its single DuckDB connection. Sibling reads then exceed `routes/local_query.py::_PROXY_TIMEOUTS` (5 s, then 9 s), their handlers return `None`, and the tabs render that as *empty*. The code comment above `_PROXY_TIMEOUTS` already predicted this exact failure ("surface as a transient EMPTY result ... blank tab"); this is that bug, at scale.

## Fix

**1. `LocalStore.query_events_slim()`** — same rows, filters and order as `query_events`, with two keys removed from each decoded `data` blob. A **denylist, not an allowlist**, so a roll-up reading some rarely-set key keeps its numbers exactly. The two dropped keys are the bulk and nothing aggregates them — over the newest 20k events of a real store:

| key | size | share | read by a roll-up? |
|---|---|---|---|
| `content` | 19.5 MB | 66.5% | no |
| `tool_calls` | 8.4 MB | 28.5% | no |
| everything the roll-ups *do* read | < 1.5 MB | 5% | yes, all preserved |

The ten heavy scans in `routes/usage.py` move to the slim shape. **Transcript, Brain and replay keep `query_events`** — they are exactly the readers that need `content`.

**2. Memoised + single-flighted RPC** for that shape (`CLAWMETRY_RPC_MEMO_TTL`, default 10 s). Concurrent endpoints behind one page load now share a single round trip instead of queueing N copies of an 8 MB response. This is the `FLYWHEEL.md` performance rule verbatim — *"one fetch of a shared blob serves all consumers — cache it with a TTL + in-flight dedup"*. The memo is opt-in per shape and **never caches the "daemon unavailable" sentinel**, so a momentary outage cannot blind the next call for a whole TTL.

## Verification

`tests/test_cost_scan_slim_events.py` — 7 tests, and **all 7 fail on the unpatched tree** (checked by stashing the fix). They pin: slim drops exactly the two keys and nothing else; every key a roll-up reads survives; the payload shrinks materially; the shape is allowlisted on the daemon proxy; eight concurrent identical scans produce **one** round trip; the unavailable sentinel is never cached; and shapes that did not opt in keep their exact previous semantics. Registered in the CI MOAT file list.

**Equivalence on real data:** plugin / skill / model / runtime roll-ups computed from slim rows are byte-identical to the same roll-ups over full rows, on the live 46k-row store.

**On the node:** `/api/usage`'s own heavy scans went **21 → 2** per request; payload per scan 37.7 MB → 8.9 MB.

Existing `test_local_query_api`, `test_local_query_dispatch_edge_cases`, `test_query_contract_drift`, `test_query_events_keys_first`, `test_usage_local_store`, `test_local_store` and neighbours pass (152 tests). Pre-existing unrelated failures in `test_skills_fidelity_local_store`, `test_plugins_invocations_local_store` and `test_rollup_runtime_attribution` are **identical before and after** this change.

## Honest scope

The extreme 45 s hangs in the report were amplified by the node itself being pathologically loaded (load average 124, swap exhausted) by ~18 leaked unrelated processes; with the machine calm, both old and new code serve `/api/usage` in well under a second. This change removes the request storm that made the dashboard fragile under *any* load, and is what stops one slow tab from blanking the others. It does not, on its own, make a swamped machine fast.

**Follow-up worth filing separately:** a proxy timeout still renders as "no data" rather than an error state. `local_store_call_via_daemon` already returns a `PROXY_UNAVAILABLE` sentinel that distinguishes the two; the tabs just don't use it. That is a correctness bug in its own right — a UI that says "you have no sessions" when it means "I couldn't reach the store" is worse than a spinner.

https://claude.ai/code/session_01AYTf9NFEjHj6KcpubsryLt


vivekchand:
✨ auto-fixed: branch was behind main — merged latest main into the PR branch (no conflicts).

---

PR #5537, opened 2026-09-05
fix(replay): the Sessions replay lost its column, and a reloaded Trail link never loaded

vivekchand:
No-PRD: field bug fix, reported from the live dashboard on 0.12.814.

Two breakages on the same screen.

## 1. The replay was squeezed into the turn-TOC column

`.transcript-layout` is a two-column grid: the message stream, then the sticky turn TOC at 240px. #5522 moved the replay-tree mount to insert relative to the transcript anchor's *real parent*, to stop the `NotFoundError` the Trail page threw when it re-parents the messages into its own card. That parent **is the grid**, so the empty `#replay-tree-container` became an auto-placed third grid item: it took the wide first column, `#transcript-messages` was pushed into the 240px TOC column, and the whole conversation rendered as an overflowing strip against the right edge with the left half of the page blank.

Measured on a real 223-message Claude Code session:

| | `#transcript-messages` | `#transcript-toc` |
|---|---|---|
| before | x=1140, **w=240** | wrapped to row 2, x=240 w=882 |
| after | x=240, **w=882** | x=1140, w=240 |

`#replay-load-earlier` had shipped that same way once already and was patched with an inline `grid-column`. So the rule now lives on the layout itself - `.transcript-layout > *:not(#transcript-messages):not(.transcript-toc) { grid-column: 1 / -1; }` - and the replay-tree mount sets it inline too, so the next node the replay injects there cannot repeat this.

Hits both the Sessions tab replay and the Trail page replay.

## 2. A deep-linked Trail never loaded

`switchTab()` highlights `event.target` for tabs with no nav item of their own (Trail reuses the Sessions entry). On a boot-time `#trail=<runtime>:<id>` deep link there is no click: `window.event` is the DOMContentLoaded event and its target is `document`, which has no `classList`. The TypeError threw out of `switchTab` **before** the per-tab loader dispatch a few lines below, and `_cmBootLanding()` swallows it, so a reloaded, bookmarked or shared trail link sat on "Opening the trail..." with every card stuck on "Loading..." and no console error. Clicking a session row from the list worked, because that is a real click with a real target.

The fallback now checks `event.target.classList` before calling `.add`.

## Verified

- `tests/test_transcript_layout_grid.py` - seven guards. Three fail on `origin/main` (the CSS rule, the replay-tree inline style, the classList check) and pass here; proven red-then-green by stashing the source fixes.
- Wired into the existing "Trail page + session-first nav" CI step.
- `node --check` on app.js; `ruff` clean on the new test.
- Walked in Chrome against a patched dashboard on a real session: a cold `#trail=` load now renders the title, the inputs card and the mounted replay instead of the skeleton, and the message column measures 882px at x=240 on both the Sessions replay and the Trail.

Pre-existing on `origin/main`, untouched here: `tests/test_appjs_units.js` fails one assertion about `JSON.stringify` inside a double-quoted `onclick` in the Inputs panel, and `make lint` reports 231 ruff errors in `dashboard.py`.

https://claude.ai/code/session_01PJZmq94HnrsCVqurxjvg4X


vivekchand:

---

vivekchand:
✨ auto-fixed: resolved CHANGELOG.md conflict with main (prepended PR's fixed-replay entry above newer Unreleased entries)

---

PR #5543, opened 2026-09-05
fix(guard): the pid guard compared the session record's timestamp, not the process start

vivekchand:
No-PRD: single-cause bug fix in an existing surface, found by verifying #5529 live. No new product behaviour, no new endpoint, no schema change.

Follows #5529. That fix got Pause past the resolver; this is the wall immediately behind it.

## What the user hits

```
POST /api/guard/control {"action":"pause","session_id":"claude_code:e5ba0d08-…"}
-> pid_guard_refused:start_mismatch(recorded=epoch:1788471021,
                                    live=lstart:Thu Sep  3 23:30:14 2026)
```

Both timestamps are correct. They measure different events:

| field | value | what it is |
|---|---|---|
| `startedAt` | `1788471021218` → Thu Sep 3 **23:30:21** | when claude_code **wrote its session record** |
| `procStart` | `Thu Sep  3 21:30:14 2026` (UTC) → **23:30:14** local | when the **process** started |
| live `ps -o lstart=` | `Thu Sep  3 23:30:14 2026` | when the process started |

`claude_code_session_map` recorded `startedAt`; `verify_pid` compares its recorded value against the live process start. Seven seconds apart on the machine this was found on, so the guard refused **every** pause, stop and kill — having correctly detected a mismatch between a process start and something that was never a process start.

## Why it was written that way

The comment says it plainly: `procStart` is a ctime string claude_code renders in **UTC**, which does not compare textually against local-time `ps` output on a non-UTC host, so `startedAt` looked like the timezone-unambiguous choice. But that comparison problem was already solved in this same module — `_start_tokens_equivalent` normalises both sides to an instant and reports `verified_tz_normalized`. Timezone ambiguity had a remedy; measuring the wrong event had none.

Proof on the real record that was failing:

```
map's recorded_start (startedAt)  -> verify_pid = (False, 'start_mismatch(…)')
the JSON's procStart              -> verify_pid = (True,  'verified_tz_normalized')
```

## What changed

The map records `procStart` and exposes `startedAt` separately as `started_at` for callers that want the session's own start. With no usable `procStart` the recorded start is `None`, so `verify_pid` degrades to an honest liveness check — falling back to `startedAt` would reinstate a value guaranteed to mismatch, which refuses everything.

## Verification

`test_claude_session_map_prefers_started_at_epoch` asserted the old preference and is replaced by `test_claude_session_map_records_the_process_start_not_the_session_start`, which pins the corrected contract including the `procStart`-absent case. A second test spawns a real process, renders its start the way claude_code does (a UTC ctime string), resolves it through a namespaced session id and requires the guard to **pass**. Both are red against the unfixed file:

```
FAILED test_claude_session_map_records_the_process_start_not_the_session_start
FAILED test_pid_guard_accepts_a_utc_proc_start_against_local_ps
2 failed, 41 passed      # unfixed
43 passed                # fixed
```

Against the live founder node the guard went from refusing every session to:

```
guard passes on 10 of 11 live sessions   (all verified_tz_normalized)
```

The eleventh has no session record at all and is correctly refused — the honest disabled state #5529 added.

https://claude.ai/code/session_01EATBJ4EuaqrrrZhNmCcvpk


vivekchand:

The required E2E Gate is failing because Drift Bot reports undocumented code changes vs. the 8090 Software Factory blueprints. Resolving this requires updating the blueprint on the 8090 platform — that's an author action outside what this sweep can do mechanically.

---

vivekchand:
✨ auto-fixed: merged main into branch (was BEHIND)

---

PR #5552, opened 2026-09-05
[RELEASE] Guard's Pause, Stop and Kill work again (carries #5543)

vivekchand:
No-PRD: release-only PR. It carries an already-merged fix (#5543) to PyPI and adds no product behaviour of its own.

## Why now

#5543 merged at 06:38Z today and `release-on-merge` correctly skipped it, so it is on `main` but in **no published wheel**. The newest release on PyPI, **0.12.817**, still contains the defect.

Reported from a founder node minutes after that merge:

```
Pause did not succeed:
pid_guard_refused:start_mismatch(recorded=epoch:1788471021, live=lstart:Thu Sep  3 23:30:14 2026)
```

Both timestamps describe the same process, seven seconds apart. The guard was comparing the moment `claude_code` wrote its session record (`startedAt`) against the moment the process actually started (`ps -o lstart=`). Every node that auto-updates is running a build whose Pause, Stop and Kill refuse on every Claude Code session. The intervention surface is the product's core promise and it is inert on every released build.

## What

Carries #5543. `claude_code_session_map` records `procStart` (the process's own start time) and exposes `startedAt` separately as `started_at`. Where no usable `procStart` exists the recorded start is `None`, so `verify_pid` degrades to an honest liveness check rather than a value guaranteed to mismatch. Daemon-side, so the cloud pin follows.

## Verified against a live node before cutting this

Measured on the founder node's real sessions, not in a fixture:

| | installed wheel 0.12.817 | `origin/main` |
|---|---|---|
| recorded start | `startedAt` | `procStart` |
| skew vs true process start | +0.8s to +7h | matched 22/22 |
| sessions exceeding the 3s tolerance | **8 of 22** | 0 |
| full guard over every live session | refusals | **24/24 `verified_tz_normalized`, 0 refused** |

`tests/test_process_control.py` green on main (43 tests).

## Follow-up, not in this release

`resolve_copilot` still passes the log-filename epoch as `recorded_start` (`clawmetry/process_control.py:1877`). That is the same class of defect (an observation instant compared as a start time) and is still present on main. It affects Copilot CLI sessions only, so it is not blocking this release; separate PR to follow.

https://claude.ai/code/session_012jB5P3MQZTZpoJZTiyTCQU


vivekchand:
Superseded. #5555 released 0.12.818 from a main that already contained #5543, so the fix is published: cracking `clawmetry-0.12.818-py3-none-any.whl` shows `clawmetry/process_control.py` carrying the corrected `procStart` preference (`Prefer startedAt` gone, `started_at` recorded separately). A second release PR for the same commit would be a no-op.

The Copilot follow-up noted in the body is still open and still unverified: `resolve_copilot` passes the log-filename epoch as `recorded_start` (`clawmetry/process_control.py:1877`), same class of defect. Not shipped here because every Copilot log on the reference node belongs to a dead process, so the skew could not be measured.

PR #5572, opened 2026-09-05
Harden CI: requeue reads drift-bot from the status API, so parked PRs restart

vivekchand:
## The problem

`queue-priority.yml` has two halves. `clear-queue` cancels every queued run on a non-main ref whenever main moves, so main's jobs take the runners next. `requeue` is meant to put those runs back once main goes idle, for each open PR whose drift-bot verdict is green.

The second half has never worked. It asked:

```
/repos/<repo>/commits/<sha>/check-runs?check_name=drift-bot
```

But drift-bot is a **commit status** posted by the `8090-software-factory` App, not an Actions check run — it never appears in `/check-runs` at all. The query returns an empty list for every commit, the verdict falls through to `"missing"`, and the green-drift guard therefore rejects **every** PR, including those whose drift-bot status is genuinely green.

So the workflow parks runs and never brings them back. This repo already knew the distinction: `scripts/e2e_gate.py` grew `list_commit_statuses()` for this exact reason and documents it in a comment. The two consumers of the same signal simply disagreed about where it lives.

## The fix

One defect, one change: read the verdict from `/commits/<sha>/status` and filter on `.context == "drift-bot"`. That endpoint already collapses to the latest status per context, so no de-duplication is needed.

The guard itself is unchanged — still `[ "$drift" = "success" ]`.

| drift-bot state | Before | After |
|---|---|---|
| green | parked (bug) | **requeued** |
| red | parked | parked |
| no status posted | parked | parked |

## Permissions

`checks: read` → `statuses: read`. Check runs and commit statuses are governed by different scopes; `checks: read` would leave the new call unable to see anything, reintroducing the same silent "no verdict" through a different door. `actions: write` and `pull-requests: read` are unchanged, and the workflow keeps its empty top-level `permissions: {}` with each scope granted per job.

## What this deliberately does not do

An earlier revision of this PR also treated an **absent** drift-bot status as eligible. Drift Bot correctly flagged that against [Main-First Runner Queue Priority](https://factory.8090.ai/project/b415065f-ab2f-4f53-8864-0c009fd098cb/requirements/b7c0af6c-e262-4a17-a342-093e94db3014), which specifies success. That has been reverted — it is a change to the requirement, not to this file, and the record comes first here.

The underlying problem is real and worth a decision, but separately: the App only posts on PRs touching Python or product-record files, so a frontend-only Dependabot npm bump never receives a status and stays parked indefinitely. See the comment below.

## Verification

- Both jq filters run against the real API payload shapes: green requeues, red parks, absent parks, and the old filter returns `missing` against a `/check-runs` response — which is the bug.
- Two auto-discovered guards added to `tests/test_workflow_yaml_valid.py`, which `ci.yml:159` already runs, so they execute rather than sitting in no job:
  - no `run:` block may read drift-bot from the check-runs API;
  - a job reading `/commits/<sha>/status` must declare `statuses: read`.
- Both guards **fail against the pre-fix workflow and pass after** — verified by reverting the file and re-running, on this repo's principle that a gate which cannot fail is worse than no gate.
- `tests/test_workflow_yaml_valid.py` — 504 passed, 310 skipped. `tests/test_ci_workflow_invocations_are_real.py` — 5 passed. `ruff` (CI's ruleset) clean. `bash -n` clean on both run blocks. All 36 workflow files parse under `yaml.safe_load`.

## Scope

Touches only `.github/` and `tests/`, both exempt paths, so this needs no Factory record; the opt-out line is belt-and-braces.

No-PRD: CI workflow logic under .github/ and tests/, no product-visible change.

https://claude.ai/code/session_015fREELscGXoGNa8Lp83Kku

### DIFF
--- a/.github/workflows/queue-priority.yml
+++ b/.github/workflows/queue-priority.yml
@@ -9,7 +9,8 @@
 #      in-progress) runs on non-main refs so main's jobs take the runners next.
 #   2. requeue      — once main is quiet again (Auto-deploy Cloud completed, or
 #      the half-hourly sweep finds main idle), re-run the latest cancelled run
-#      of each workflow for every open PR whose drift-bot check is green.
+#      of each workflow for every open PR whose drift-bot commit status is
+#      green, or which never receives one (frontend-only npm bumps).
 #      Red-drift PRs stay parked until their drift is fixed and re-reported.
 #
 # No third-party actions on purpose: every step is `gh api` against
@@ -78,11 +79,15 @@ jobs:
     if: github.event_name == 'workflow_run' || github.event_name == 'schedule' || github.event_name == 'workflow_dispatch'
     runs-on: ubuntu-latest
     timeout-minutes: 15
-    # actions: re-run parked runs; checks: read drift-bot's conclusion;
+    # actions: re-run parked runs; statuses: read drift-bot's commit status;
     # pull-requests: list the open PRs to consider.
+    #
+    # `statuses: read` rather than `checks: read`: drift-bot is a commit
+    # status, not a check run (see the step below), and the two surfaces are
+    # governed by different scopes. Nothing here reads /check-runs any more.
     permissions:
       actions: write
-      checks: read
+      statuses: read
       pull-requests: read
     steps:
       - name: Skip while main still has queued or running jobs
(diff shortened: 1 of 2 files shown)


vivekchand:
## Drift Bot's finding was correct — reverted in `ba9ee8e`

The finding on `d6dc51b` was accurate, and not a documentation gap: the requeue condition really did go beyond [Main-First Runner Queue Priority](https://factory.8090.ai/project/b415065f-ab2f-4f53-8864-0c009fd098cb/requirements/b7c0af6c-e262-4a17-a342-093e94db3014), which specifies the drift check reports success. I had `success|unreported`. That is a change to the requirement, and the record comes first here, so it is out.

The PR is now the one defect on its own — the verdict is read from `/commits/<sha>/status`, where drift-bot actually posts, instead of `/check-runs`, where it never appears. The guard is unchanged at `[ "$drift" = "success" ]`:

| drift-bot state | Before | After |
|---|---|---|
| green | parked (the bug) | **requeued** |
| red | parked | parked |
| no status posted | parked | parked |

## The separate problem this turned up — worth a decision, not a code change

Fixing the API surface does not help one class of PR, and I want to be explicit rather than let it look handled.

The 8090 App only posts drift-bot on PRs touching Python or product-record files. A **frontend-only Dependabot npm bump never receives a status at all**, so `requeue` parks it permanently — `clear-queue` cancels its runs on the next push to main, and nothing ever puts them back. `scripts/e2e_gate.py` already accommodates this shape from the other direction, with `skip_if_unreported=True` on the same spec and a comment naming the same PRs.

The effect is visible today: **#5376** (`bump esbuild and vite in /frontend`) has had every check `cancelled` since 2026-09-03, and its "PR cites the product record" failure predates #5478, which fixed exactly that gate for lockfiles on 09-04. It is a security bump that cannot currently reach a runner.

Three ways forward, all of which are yours rather than mine:

1. **Amend the requirement** to say success *or unreported* — mirroring the E2E Gate's existing `skip_if_unreported` treatment of this same signal. My preference, since the two consumers would then agree.
2. **Leave requeue strict** and unpark that class another way, e.g. exempting PRs with no drift-bot status from `clear-queue` in the first place.
3. **Accept it**, and requeue those PRs by hand when a dependency bump needs to land.

Happy to implement whichever, in its own PR with the record written first. Flagging rather than choosing, because option 1 changes a requirement and option 2 widens this workflow's behaviour — neither is mine to decide.

---

vivekchand:
## This PR just got parked by the bug it fixes

Worth recording, because it is the clearest demonstration of the defect I could have asked for and I did not stage it.

At **18:50:52–54** every remaining check on `ba9ee8e` was **cancelled** — not failed. `clear-queue` fired on a push to main and cancelled every queued run on a non-main ref, exactly as designed. The only three checks that survived are the ones that happened to finish in the ~4 minutes beforehand:

| Check | Result |
|---|---|
| Vendored fonts not stale | success (18:49:38) |
| OSS golden path (wheel + OpenClaw + 9 tabs) | success (18:50:20) |
| Cross-repo handoff (C4) | success (18:50:33) |
| everything else (27 checks) | **cancelled at 18:50:52** |
| drift-bot | success |

Nothing is red. But those 27 runs would never have come back on their own, because `requeue` — the job whose whole purpose is to restore them — reads drift-bot from `/check-runs`, gets an empty list, resolves every PR to `missing`, and re-runs nothing. That is the defect this PR fixes, and it means the PR cannot reach a green head under its own steam. A bootstrap problem: the fix has to merge before the mechanism that would let it prove itself starts working.

I have merged `main` in to trigger a fresh run rather than pushing an empty commit. That also keeps the branch current.

**One thing for whoever reviews this:** the same race can hit again at any moment — if main moves while these checks are queued, they get cancelled again and again do not return. If it happens repeatedly, that is the bug reasserting itself, not flakiness and not something wrong with the diff. Merging on a quiet moment for main is the reliable way through, and after this lands the problem stops recurring for every other PR too.

---

PR #5575, opened 2026-09-06
Harden CI: install the E2E runtime's OpenClaw CLI from a lockfile

vivekchand:
## What

`.github/actions/setup-openclaw` installed its CLI with `npm install -g "openclaw@$OPENCLAW_VERSION"`, and every caller passed the `latest` dist-tag. That resolved the registry's moving tag afresh on each run, with no integrity check, and put the result on `PATH` for six workflows — including `ci.yml`'s live OpenClaw E2E job and the nightly keystone drive. zizmor reports it as `adhoc-packages` (installs a package outside of a lockfile).

This replaces it with `npm ci` against a committed `package-lock.json` in the action's own directory, so the install is exactly the recorded tree and every tarball is checked against its sha512 integrity hash. A bare `@version` pin was not enough on the sibling repo and is not enough here — re-resolution and the absent integrity check are the actual exposure.

It pairs with a new npm entry in `.github/dependabot.yml`, so the lock keeps moving rather than freezing on one OpenClaw release.

**The pin is `2026.5.12`** — the version CI has really been running. That choice is the substance of this PR, and the section below is why.

## What the pin uncovered

**CI has not been testing the OpenClaw it appeared to test.**

The removed `actions/cache` step keyed on `openclaw-${{ runner.os }}-${{ inputs.version }}-v1`, and every caller passed `latest`. The key contained the literal string `latest`, so it never varied and never invalidated. Every job restored the first build ever stored under it and skipped the install outright — `main`'s run logs `Cache hit for: openclaw-Linux-latest-v1`, then `openclaw cache hit, skipping install`, then `openclaw version: OpenClaw 2026.5.12 (f066dd2)`. The install step has not executed in months, and `latest` was frozen in practice the whole time.

So the honest reading of "green on main" is: **green against OpenClaw 2026.5.12**, not against current OpenClaw.

### Why this pins 2026.5.12 rather than the newest release

An earlier revision of this branch pinned `2026.9.2` and adapted the fixtures to it. That took the live E2E job red, and the reason is not something a test edit fixes:

**OpenClaw no longer writes per-session `<sid>.jsonl` transcripts.** On `2026.9.2` there is no `agents/<id>/sessions/` directory at all. The transcript moved into `agents/<id>/agent/openclaw-agent.sqlite` (tables `transcript_events`, `session_transcript_active_events`, `session_conversations`).

[...]

PR #5583, opened 2026-09-06
Harden CI: install the visual-diff bot's Node deps from a lockfile

vivekchand:
`No-PRD:` CI-only change. Every file is under `.github/`, which is exempt from the product-record gate.

## What this changes

`pr-screenshots.yml` built its Node environment with `npm init -y` followed by a bare `npm install`:

```yaml
working-directory: head
run: |
  npm init -y >/dev/null
  npm install --no-audit --no-fund --silent \
    playwright@1.49.0 pixelmatch@5.3.0 pngjs@7.0.0
```

The three direct versions were already pinned, so this is not about them. Without a lockfile every **transitive** dependency still re-resolved against the registry on each run, and no tarball was checked against an integrity hash. zizmor reports it as `adhoc-packages` — one of the two left on this repository's workflows.

The step runs in a job holding `contents: write` and `pull-requests: write`, so whatever resolves there runs with those scopes. Same reasoning as the equivalent fixes already merged in `clawmetry-pro` and `clawmetry-cloud`, and as the SHA pinning applied to every `uses:` reference: make an upgrade deliberate rather than a side effect of when the job happened to run.

Now:

```yaml
working-directory: head/.github/scripts
run: |
  npm ci --no-audit --no-fund
  npx playwright install --with-deps chromium
```

Same three versions — this changes **how** they are installed, not what.

## Why the manifest lives in `.github/scripts/`

`visual-diff.mjs` imports `playwright`, `pixelmatch` and `pngjs` as bare specifiers, which Node resolves by walking up from the importing file. `head/.github/scripts/node_modules` is the first directory it checks. A manifest in a directory of its own would install somewhere those imports never look.

## Dependabot

Paired with a new npm entry for `/.github/scripts`, so the lock keeps moving instead of freezing — the same pin/updater pairing that file already documents for the SHA pins and the other four Node manifests. **Appended at the end of the file** so it does not collide with the entry #5575 adds mid-file.

## Verification

- `npm ci` in a clean directory laid out as `head/.github/scripts/` exits 0; every package in the lock carries an integrity hash
- all three bare specifiers resolve from `visual-diff.mjs`'s real location (checked with `createRequire` against that exact path)
- zizmor `adhoc-packages` on `pr-screenshots.yml`: 1 → 0, no new finding introduced (52 → 51 total; the remainder is the `setup-openclaw` one #5575 fixes)
- all 38 workflow, composite-action and `dependabot.yml` files parse under `yaml.safe_load`
- `scripts/check_action_refs.py` still exits 0

## One transitional note for the reviewer

The manifest is read from the PR **head** checkout, consistent with how this workflow already consumes head (`./head/.github/actions/setup-openclaw`, `head/.github/scripts/post-visual-diff.mjs`). A PR branched before this merges will not have the manifest in its head, so its install step will fail until it picks up main. The bot is non-blocking by design — it is not among the 12 checks `E2E Gate (required)` aggregates — so this costs a screenshot comment, not a merge. Flagging it rather than burying it; happy to add a trusted-checkout step instead if you would rather it not depend on head at all.

https://claude.ai/code/session_019xu7MkNCbNQuZMJRw7SJpT

---
_Generated by [Claude Code](https://claude.ai/code/session_019xu7MkNCbNQuZMJRw7SJpT)_

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -139,3 +139,20 @@ updates:
       - "dependencies"
     cooldown:
       default-days: 7
+
+  # npm — the PR visual-diff bot's Playwright/pixelmatch/pngjs
+  # (.github/scripts/, driven by pr-screenshots.yml). The other half of that
+  # pin: the lockfile makes the install reproducible and integrity-checked,
+  # and this entry is what keeps it moving. Without it the bot would sit on
+  # one Playwright forever, which is the frozen end of the same problem the
+  # unpinned install was at the other end of.
+  - package-ecosystem: "npm"
+    directory: "/.github/scripts"
+    schedule:
+      interval: "weekly"
+      day: "monday"
+    open-pull-requests-limit: 3
+    labels:
+      - "dependencies"
+    cooldown:
+      default-days: 7
--- a/.github/scripts/package-lock.json
+++ b/.github/scripts/package-lock.json
@@ -1,0 +1,92 @@
+{
+  "name": "clawmetry-pr-screenshots",
+  "version": "0.0.0",
+  "lockfileVersion": 3,
+  "requires": true,
+  "packages": {
+    "": {
+      "name": "clawmetry-pr-screenshots",
+      "version": "0.0.0",
+      "license": "MIT",
+      "dependencies": {
+        "pixelmatch": "5.3.0",
+        "playwright": "1.49.0",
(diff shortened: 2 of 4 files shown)


vivekchand:
✨ auto-fixed: merged current main into the branch (was behind; fresh CI triggered)

---

PR #5592, opened 2026-09-06
Harden: declare the non-security digests as non-security

vivekchand:
No-PRD: security hardening of existing digest call sites; no product surface, no API, and no stored value changes.

**Risk:** Low, and the one thing that could have bitten is explicitly tested. Several of the rewritten call sites produce **persisted** ids — event dedup keys in DuckDB, trace/span ids, cron change-detection hashes. If the digests had changed, the daemon would have silently re-ingested every already-stored event. `usedforsecurity=False` does not alter the output, and `test_digests_match_hashlib` asserts that byte-for-byte against `hashlib` for four payloads including the empty string and all 256 byte values. Nothing else changes: no schema, no endpoint, no config. Undone by reverting the commit — there is no migration and no retroactive data change.

---

## Summary

- ClawMetry hashes plenty of things that are not secrets: dedup ids for ingested events, the change-detection hash over a cron file, a fingerprint over a tool call's arguments, trace/span ids derived from a session id. Fourteen of those call sites reached for `hashlib.md5` / `hashlib.sha1` directly. On a host whose crypto provider is configured to allow only approved algorithms — the normal build on several enterprise Linux distributions — those constructors don't return a weak digest, they **raise `ValueError`**. That turns the daemon's cron ingest, its memory-file ingest, event dedup, the detector fingerprints and the Tracing tab's span ids into crashes, and not one of them is doing cryptography. `usedforsecurity=False` is the documented way to say so, and it's what keeps a non-security digest available where the provider restricts the algorithm.
- The keyword can't just be sprinkled on. It reached hashlib's constructors in CPython 3.9 and `setup.py` still declares `python_requires=">=3.8"`, where passing it is a `TypeError` rather than a no-op. `routes/meta.py` already carried a bare `usedforsecurity=False`, so that one line was already broken on any 3.8 install — CI covers 3.9 and 3.11 only, so nothing had eyes on it. The version check now lives in one new short module, `clawmetry/nonsecret_hash.py` (per CLAUDE.md's "new short module, not 300 lines into a 20k-line file"), and all fifteen call sites go through it.
- `tests/test_nonsecret_hash.py` carries a **ratchet** that fails if a bare `md5`/`sha1` reappears anywhere under `clawmetry/`, `routes/` or `helpers/`, so this can't quietly come back between audit runs. It's wired into `ci.yml` in this PR, because this repo runs FILE LISTS and a test named in no workflow step runs in no job at all.
- `scripts/harness/audit.py` keeps its own bare `sha1` deliberately, and is excluded from the ratchet's scanned trees with the reason written down: it's a standalone CI harness invoked as `python3 scripts/harness/audit.py`, so the package isn't on its `sys.path` and it can't import the helper.

## Test plan

- [x] `bandit` B324 findings go **14 → 1** (the harness script above, excluded on purpose)
- [x] 11 new tests in `tests/test_nonsecret_hash.py` pass
- [x] **The ratchet was verified to bite** — reintroduced one bare `hashlib.sha1(` in `clawmetry/detectors.py` and `test_no_bare_md5_or_sha1_in_runtime_code` failed naming that exact line; `test_ratchet_would_catch_a_regression` pins the pattern itself
- [x] Digests are unchanged vs `hashlib` for `b""`, `b"hello"`, a realistic dedup key, and `bytes(range(256))` — so no persisted id moves
- [x] Every workflow file parses under `yaml.safe_load`; `tests/test_workflow_yaml_valid.py` → 507 passed
- [x] The repo's own `ast.parse` gate passes over all 1329 `.py` files
- [x] `lint-py39`, `lint-js`, `lint-daemon-allowlist`, `lint-runtime-count`, `lint-ac-coverage`, `lint-module-map` all pass; `docs/MODULE_MAP.md` regenerated and in sync
- [x] `ruff check dashboard.py clawmetry/` reports **231** findings — the exact pre-existing count on clean `main`, so this adds no lint debt. Getting there meant removing five `import hashlib` statements left dead by the rewrite (four of them function-local); each was confirmed dead by walking its enclosing function's AST first
- [x] `dashboard`, `clawmetry.sync`, `clawmetry.local_store`, `clawmetry.detectors`, `clawmetry.span_reconstruct` and the four touched route modules all import cleanly in a fresh interpreter
- [x] Known-unrelated: `tests/test_detectors_behavioural.py` and `tests/test_event_id_unification_and_dedup.py` have six DuckDB connection-configuration failures. They **reproduce identically on clean `main`** and are untouched by this change

https://claude.ai/code/session_01QHxWYV4ayn5pp3UcsK6xPr

---
_Generated by [Claude Code](https://claude.ai/code/session_01QHxWYV4ayn5pp3UcsK6xPr)_

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -172,6 +172,14 @@ jobs:
       - name: Architecture docs name modules that exist
         run: python3 -m pytest tests/test_module_map_drift.py -q
 
+      # Non-security digests (dedup ids, trace ids, change-detection hashes)
+      # must go through clawmetry.nonsecret_hash. A bare hashlib.md5/sha1
+      # raises ValueError -- not a weak digest, an exception -- on a host whose
+      # crypto provider allows only approved algorithms, which turns an ingest
+      # pass into a crash. Named explicitly because this job runs FILE LISTS.
+      - name: Non-security digests declare themselves
+        run: python3 -m pytest tests/test_nonsecret_hash.py -q
+
       # The declared product surface must match reality: every cell names a
       # verifier that exists, and nothing claims to be "gated" while sitting
       # behind a path-filtered workflow that stays silent on most PRs.
--- a/clawmetry/detectors.py
+++ b/clawmetry/detectors.py
@@ -63,13 +63,13 @@ by env so the daemon can tune them without a code change.
 """
 from __future__ import annotations
 
-import hashlib
 import json
 import os
 from typing import Any, Iterable, Optional
 
 # New in this change, each in its own module so the whole of the new capability
 # is readable in one place and this file keeps its shape.
+from clawmetry import nonsecret_hash as _nsh
 from clawmetry.detector_calibration import (  # noqa: F401
     BASELINE_CEIL_RATIO, BASELINE_FLOOR_RATIO, BASELINE_MIN_SESSIONS,
     BASELINE_SIGMA, BLAST_RADIUS_FILES, EGRESS_HOST_FANOUT, RUNTIME_PROFILES,
@@ -217,7 +217,7 @@ def _args_hash(args: Any) -> str:
             norm = str(args)
         except Exception:
             return ""
-    return hashlib.sha1(norm.encode("utf-8", "replace")).hexdigest()[:16]
+    return _nsh.sha1(norm.encode("utf-8", "replace")).hexdigest()[:16]
 
 
 def _iter_tool_calls_from_data(et: str, data: dict) -> list[dict]:
--- a/clawmetry/local_store.py
+++ b/clawmetry/local_store.py
@@ -47,6 +47,7 @@ import subprocess
 import sys
 from clawmetry import ccr as _ccr  # reversible event-payload compression (#2843)
 from clawmetry import event_shape as _event_shape  # v15 typed event columns
+from clawmetry import nonsecret_hash as _nsh
 from clawmetry.trail_store import TrailStoreMixin  # intent / back-fill / git join
 import threading
 import time
@@ -13887,11 +13888,10 @@ class LocalStore(TrailStoreMixin):
         Idempotent on a stable id derived from provider+kind+ts so re-tailing
         the log (after a rotation/truncation rescan) never double-counts.
         """
-        import hashlib
         prov = (provider or "").lower().strip()
         if not prov or not kind or not ts_iso:
             return
-        ev_id = "connhealth-" + hashlib.sha1(
+        ev_id = "connhealth-" + _nsh.sha1(
             f"{prov}|{kind}|{ts_iso}|{(raw or '')[:80]}".encode("utf-8")
         ).hexdigest()[:24]
         payload = json.dumps({
@@ -13947,7 +13947,7 @@ class LocalStore(TrailStoreMixin):
         """
         if not event_type or not ts_iso:
             return
-        ev_id = "talk-" + hashlib.sha1(
+        ev_id = "talk-" + _nsh.sha1(
             f"{session_id or ''}|{event_type}|{ts_iso}|{(raw or '')[:80]}".encode("utf-8")
         ).hexdigest()[:24]
         payload = json.dumps({
--- a/clawmetry/nonsecret_hash.py
+++ b/clawmetry/nonsecret_hash.py
@@ -32,28 +32,47 @@ signatures, tokens, or integrity checks over untrusted content. Those want
 SHA-256 (or the ``cryptography`` package, which the cloud-sync path already
 depends on), not a documented-as-non-security MD5.
 
-Why the two scanners are told to stand down *here* and nowhere else
-------------------------------------------------------------------
+What the two scanners say about this page
+----------------------------------------
 
 Collecting the calls into this module is what makes the risk acceptance
 reviewable: there are exactly four ``hashlib`` constructor calls in the
 package now, all of them on this page, instead of fifteen scattered ones.
-That same collection is why both scanners fire here.
+That same collection is why both scanners point here.
 
 * **bandit B324** flags a weak digest built without ``usedforsecurity``.
-  Only the two 3.8 fallbacks qualify, and they carry ``# nosec B324``.
+  Only the two 3.8 fallbacks qualify, and ``# nosec B324`` clears them.
+  That marker works: bandit reports zero B324 for this file.
 * **CodeQL ``py/weak-sensitive-data-hashing``** flags all four, because it
   tracks the *caller's* value into the sink: a ``session_id``, ``job_id`` or
   ``trace_id`` reaching a weak digest reads as sensitive data being hashed.
-  Unlike bandit, it does not treat ``usedforsecurity=False`` as an answer,
-  so the declaration alone does not clear it. The values concerned are
-  identifiers ClawMetry itself minted and already stores in plaintext beside
-  the digest -- the digest is a shorter name for a row, not a way of
-  protecting it -- so there is nothing here for a collision or a preimage to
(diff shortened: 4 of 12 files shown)


vivekchand:
## CodeQL is red on this PR, and I can't clear it from here — it needs a maintainer decision

**Failing check:** `CodeQL` — *"4 new alerts including 4 high severity security vulnerabilities"*, alerts [977](https://github.com/vivekchand/clawmetry/security/code-scanning/977), [978](https://github.com/vivekchand/clawmetry/security/code-scanning/978), [979](https://github.com/vivekchand/clawmetry/security/code-scanning/979), [980](https://github.com/vivekchand/clawmetry/security/code-scanning/980), all `py/weak-sensitive-data-hashing` on `clawmetry/nonsecret_hash.py` lines 52, 54, 63, 65.

### What it's pointing at

Those four lines are **every** `hashlib` call in the new helper — including the two that pass `usedforsecurity=False`. That kwarg is invisible to this query: it keys on the algorithm (MD5/SHA-1) plus a taint path from a source it classifies as sensitive, which here is the `id` category — `session_id`, `ev_id`, the event payload used as a dedup key. So this is not something a flag or a `# nosec` fixes.

### Why I don't think it's new risk — and where that reasoning stops

This PR changes no algorithm and no digest. The same fifteen `hashlib.md5`/`sha1` calls over the same `*_id` values exist on `main` today; the PR moves them behind one helper. CodeQL's PR check fails only on alerts it considers **new**, and an alert's identity is tied to its location — relocating a sink into a new file breaks the fingerprint, so pre-existing flows get re-reported as new. That's what I believe happened: 2 MD5 paths + 9 SHA-1 paths, grouped into 4 alerts, all landing on the one file they now pass through.

**I could not verify that.** The code-scanning API is unreachable from this session (no `gh`, no MCP tool for it, and direct REST is blocked by the proxy), so I can't confirm whether 977–980 were already open against `main`'s call sites. **That check is worth a maintainer doing before accepting my reading** — if these flows were *not* already alerting on `main`, my relocation explanation is wrong and the finding deserves a fresh look.

On the substance: none of these digests protects anything. They are dedup keys, OTLP trace/span ids, an 8-hex-char fingerprint over truncated tool arguments, and a cron-file change hash. Session ids are local identifiers already shown in the UI and exported over OTLP, not secrets. But "is an id sensitive here" is a judgement about this product, not something I should decide unilaterally on a public repo.

### The two ways forward — both are yours, not mine

1. **Dismiss 977–980 as false positives** in the Security tab. CodeQL runs here via **default setup** (there's no `.github/workflows/codeql.yml` and no `.github/codeql/` config), so there's no query-filter file I can add and no supported inline-suppression comment. Dismissal is a UI action with permissions and judgement I shouldn't assume.
2. **Migrate these ids off MD5/SHA-1** (SHA-256 or `blake2b`, which `local_store.py:2715` already uses for a similar purpose). This is the answer that actually satisfies the scanner rather than silencing it — but it **changes persisted values**: event dedup keys in DuckDB, trace ids, cron change hashes. Every stored event would re-ingest without a migration. That's a separate PR with a migration plan, explicitly outside what this one promised.

### What I'm doing

Not re-running the check — this is deterministic static analysis, so a re-run returns the same result and would just spend the budget. Not pushing a speculative change either: option 1 isn't a code change, and option 2 contradicts this PR's stated risk profile ("no stored value changes"), so guessing between them would be worse than asking.

Everything else on the PR is green or clean: Drift Bot reports no drift, there are no review threads, and the branch has no merge conflict. This one check is the only thing blocking, and it stays blocked until you pick 1 or 2. I'll keep watching the PR and re-check on a timer; if you'd rather I close this and reopen it as the option-2 migration, say so and I'll do that instead.

---

vivekchand:
**Correction to my comment above:** CodeQL is *not* blocking this PR.

When I wrote that, CI was still running. All 48 checks have now finished: **47 pass or skip, CodeQL is the only failure — and it is not a required check.**

- `E2E Gate (required)` → **success**
- `mergeable_state` → `unstable`, not `blocked` (GitHub uses `unstable` for "mergeable, but a non-required check is red"; a required check would make it `blocked`, and a conflict `dirty`)

So the merge isn't gated on 977–980 at all. That narrows the decision to one question: **is it acceptable for those four alerts to stay open against `main` after merge?** If yes, this can merge as-is and the option-1 dismissal is just tidying whenever you feel like it. If no, that's the option-2 migration and this PR should wait.

Everything else stands: no merge conflict, digests unchanged, and the four review threads above are left unresolved on purpose — the finding is escalated, not addressed, and resolving them would hide it while the decision is still open.

---

vivekchand:
## Measured: inline CodeQL suppression does nothing on this repo. Annotations removed in `1437d9b`.

`664f447` (pushed by a different session, not this one) added `# codeql[py/weak-sensitive-data-hashing]` to all four `hashlib` calls under the heading *"so CodeQL agrees"*, plus a test asserting each call keeps its annotation.

**CodeQL re-ran on that commit and reported the same four high-severity alerts.** They went 977–980 → **981–984** purely because the annotations shifted the lines and a changed location mints a new alert identity. Nothing was silenced.

That also settles the open question from my earlier comment, the one I said I couldn't verify: **these alerts track *location*, not risk.** Four identical calls, moved a few lines by a comment, produced four brand-new "new in this PR" alerts. That's the relocation mechanic, now demonstrated rather than hypothesised.

### Why the annotations can't work here

Code scanning runs through **default setup** — no `.github/workflows/codeql.yml`, no `.github/codeql/` config — so there's no query filter to add, and inline suppression comments are a CodeQL *CLI* feature that default setup doesn't honour. A comment that reads like a fix and isn't one is worse than no comment, and the accompanying test made it permanent: it would have failed anyone who deleted the dead markers, while the alerts stayed up regardless.

### What `1437d9b` keeps

- **`# nosec B324`** on the two 3.8 fallbacks — that marker demonstrably works (bandit: zero B324 in this file, 1 repo-wide, the standalone harness script excluded on purpose).
- **The risk-acceptance reasoning** from `664f447`'s docstring, which is sound and worth keeping: these digests are shorter names for rows whose identifiers are stored in plaintext beside them, so a collision or preimage wins nothing.
- **The four-call containment check**, now asserting the opposite: that no inline CodeQL suppression comes back, with the measurement written down so nobody re-runs the experiment.

Verified before pushing: 12 tests pass; the inverted guard was checked to bite by re-adding one annotation (it failed naming that line); `ruff` reports **231** — the exact clean-`main` baseline; bandit B324 stays 1; the `ast.parse` gate passes over all 1329 files; `lint-py39`, `lint-module-map`, `lint-ac-coverage`, `lint-daemon-allowlist` pass; digests still match `hashlib` byte-for-byte.

### Still yours to decide, and now with one option removed

CodeQL will stay red on this PR. It is **not a required check** — `E2E Gate (required)` passed on the previous head — so this is mergeable with the alerts left open. The real choice is:

1. **Merge and dismiss 981–984** in the Security tab. (In-code suppression is now ruled out, so the Security tab is the only dismissal route.)
2. **Migrate these digests off MD5/SHA-1**, which rewrites already-persisted bytes — event dedup keys, trace ids, cron change hashes — and needs its own migration PR.

I'm not making that call, and I'd ask that another automated session not make it either: the next thing pushed here should follow from your answer, not from a fresh guess at it.

---

vivekchand:
E2E Gate (required) passes. The only non-passing check is CodeQL (`unstable`), which the PR body explicitly addresses: the one remaining bandit B324 finding is `scripts/harness/audit.py`, excluded on purpose since it's a standalone CI harness that cannot import the `clawmetry` package. CodeQL failure here is expected and documented by the author. Blocked on author decision — skipping (auto-mergeability sweep)

---

PR #5651, opened 2026-09-08
[RELEASE] fix(overview): count only the tasks the panel will actually show

vivekchand:
Caught during the **live verification** of #5645 / #5648 on `app.clawmetry.com` under `?runtime=codex` — and it's a defect that fix introduced.

## What I saw

The new empty state said:

> No active tasks for Codex — **489 tasks on other runtimes — switch runtime to see them.**

Switching to all-runtimes rendered **zero cards**.

## Why

The count was `allAgents.length - agents.length` — every row the runtime filter removed. But the list applies runtime scope **and** bucket **and** recency: all 489 had finished more than an hour earlier, so the panel's own `RECENT_DONE_MS` (1h) rule excludes them from "Recently Completed".

A number that promises more than the list can show is the same class of defect as the four this panel just shed. An empty state that sends someone to an empty screen is its own small lie.

## Fix

One visibility rule for counts and rendering:

```js
function _ovVisible(list, nowMs) {   // running always; finished only while recent
  ...returns {running, done, failed, total}
}
```

- The other-runtime count is now `_ovVisible(<other runtimes>).total` — it can only ever name tasks a user will actually find after switching.
- `loadOverviewTasks` renders from the same splitter instead of re-deriving buckets and then filtering them.
- `_ovVisible` and `_ovRecentlyFinished` are hoisted to module scope (alongside `_ovBucketOf` / `_ovEndedMs`) so the vm-based unit harness can reach them rather than being trapped in a closure; `_ovRecentlyFinished` takes an injectable `nowMs`, which is what makes the recency behaviour testable at all.

## Verification

- **13 new behavioural assertions** (36 total in `tests/test_active_tasks_units.js`) pinning the exact bug: a pile of old tasks counts as **zero**; a running task always counts; a never-ran spawn with a now-stamped `updatedAt` counts as zero.
- **2 new static guards**, including one that bans the `allAgents.length - agents.length` shape outright.
- Proven **red against `origin/main`**, green after. 60 pass across adjacent suites. `make lint-js`, `lint-module-map`, `lint-runtime-count` green.
- Two earlier guards were re-pointed from the call site they happened to pin to the invariant they meant to (`_ovBucketOf` is reached via `_ovVisible` now) — the same over-specification broke `test_js_treats_running_as_live` in #5645.

Released directly because #5645 is already live and this corrects copy a user is reading right now.

No-PRD: correction to a defect introduced by #5645, found in live verification.

https://claude.ai/code/session_015E22w1HVFeFyMQZ6zRPAeb


vivekchand:
**Autonomous maintainer handoff** — 2026-09-08

All substantive CI checks are green (MOAT Verifier, E2E Browser Tests, Live OpenClaw E2E, Store invariants, Entitlement API tests, Compression Safety Eval, Eval Suite Gate, all pip installs). The E2E Gate check is blocked by a timing artifact: the gate job ran at push time (before the CI legs had started), saw no results, and failed within seconds. A branch update has been triggered to get a clean E2E Gate run.

This is a **[RELEASE] PR** — merging triggers `release-on-merge.yml` which publishes to PyPI. That requires production secrets not available in the automated sandbox. Once the fresh E2E Gate passes, this PR is ready for a human to squash-merge and ship.

---

PR #5382, opened 2026-08-30
Harden: only enable the Werkzeug debugger on a loopback bind

vivekchand:
## What this fixes

`_run_server` passed `debug=True` to `app.run` whenever `args.debug` was set — and `--debug` **defaults to True** here (you opt out with `--no-debug`). `--host` defaults to `127.0.0.1`, so the common case was already fine, but nothing tied the two settings together.

So passing `--host 0.0.0.0` to reach the dashboard from another machine also served Flask's interactive debugger on every interface. That page hands any client that triggers an unhandled exception the full traceback, the source around each frame and every local variable in scope, plus a PIN-gated eval console. The PIN keeps it short of trivial remote code execution — but the source and locals disclosure needs no PIN, and neither is something a user asked for by typing `--host`.

A non-loopback bind is an expected way to run this, not an exotic one: the startup banner prints a LAN URL and a "Public - ensure port is open" URL a few lines above the call.

## The change

`_is_loopback_host` now decides whether the debugger comes up; only a loopback bind gets it.

- The **reloader stays on either way** — that is the part dev mode is actually wanted for. `--host 0.0.0.0` behaves exactly as before minus the debugger, and prints a one-line note saying so.
- The helper **fails closed**: an empty value, an unparseable literal, or a hostname all read as remote. Being wrong in that direction costs a developer a traceback page; being wrong the other way publishes source and an eval console to the network.
- It deliberately **does not resolve hostnames** — a name that resolves to loopback today can resolve elsewhere tomorrow, and DNS should not be what decides whether an eval console is reachable.
- Wildcard binds (`0.0.0.0`, `::`) are correctly **not** loopback; they include every routable interface.

No config, flag, or default changes. `--no-debug` and the Waitress production path are untouched.

## How it was found

bandit `B201` (HIGH severity) in the scheduled security audit.

## Verification

- `tests/test_debugger_loopback_only.py` — **22 cases**, all passing: loopback literals across the whole `127/8` block, `localhost` (and case/whitespace variants), bracketed (`[::1]`) and zone-suffixed (`::1%lo0`) IPv6, both wildcard binds, LAN and routable addresses, hostnames including a `localhost.evil.test` prefix trap, and hostile input that must not raise.
- `python3 -m py_compile dashboard.py` passes.
- `ruff check dashboard.py --select E,W --ignore E501,E402` (the command CI runs) reports **54 findings before and after** — no new lint.
[...]

PR #5672, opened 2026-09-08
[RELEASE] cut: 5,036 dead lines from dashboard.py

vivekchand:
`dashboard.py` defined `DASHBOARD_HTML` **twice**. The first — 5,036 lines of inline HTML/CSS/JS — was overwritten at import time by the second and never rendered. It still shipped in every wheel, and it still answered every grep.

## The cost was never bytes

Four separate documents each had to warn about it:

- `FLYWHEEL.md` — *"only the SECOND renders"*
- `ARCHITECTURE.md`
- `AGENTS.md`
- `CLAUDE.md`

A rule that must be restated in four places to stop people editing the wrong copy is a defect in the code, not a gap in the docs.

It also meant a grep for any UI symbol found the **dead copy first** — including `detectProjectBadge` and the developer's personal project names (`mockround`, `vedicvoice`), which kept matching in the repo for hours after the live copy was deleted, and which shipped to every customer inside a string nobody could render.

`dashboard.py`: **21,301 → 16,273 lines**. The four documents are corrected — leaving them would make them wrong, since the quirk they describe no longer exists.

## The recorded scope was wrong, and checking is what caught it

The cleanup ledger had this as *"lines 3889–13941"* — everything up to the second definition.

It isn't. `ast.parse` puts the dead assignment at **3889–8924**. Lines **8925–13940 are ~5,000 lines of live application code**: the imports, the Flask app construction, the history module, the OTLP wiring.

**Deleting the recorded range would have deleted the application.** Boundaries came from the parser, not from reading line numbers.

## Verified

A **real Jinja render** of the surviving template — which is what FLYWHEEL §0a.4 asks for rather than an assumption:

```
RENDER OK, bytes: 467110
  'js/app.js':        True
  'page-overview':    True
  'left-nav-item':    True
  'page-transcripts': True
  dead-only 'mockround': False
```

40 `{% include %}`s resolved.

3 guards in `tests/test_dashboard_html_defined_once.py`, **all proven red against `origin/main`**:

| Guard | |
|---|---|
| exactly one definition | a second silently wins or loses by source order |
| the survivor is the template loader | so keeping the *wrong* one also fails |
| no personal project names in `dashboard.py` | they lived on in the dead blob's copy |

**Regression check** across the 45 test files that import `dashboard`: **39 failures before, 39 after**. The one apparent 40th was `test_cron_health.py` asserting against a live server on `localhost:8900` whose daemon auto-updated to 0.12.834 mid-run; it fails identically on unmodified `origin/main`. Filed separately — that file also runs in no CI job, so its assertion drift has never been reported.

No-PRD: removes dead code; no capability change.

https://claude.ai/code/session_015E22w1HVFeFyMQZ6zRPAeb


vivekchand:
### CodeQL: 2 alerts, both pre-existing code relocated by the deletion

CodeQL reports *"2 new alerts including 1 high severity"*, with its own caveat: **"Alerts not introduced by this pull request might have been detected because the code changes were too large."** A 5,036-line deletion is the largest possible instance of that.

Both flagged sites are byte-identical to `main`, just moved:

| alert | this PR | `origin/main` | |
|---|---|---|---|
| `py/flask-debug` (high) | `dashboard.py:16051` | `dashboard.py:21079` | same `app.run(host=…, debug=True, use_reloader=True, threaded=True)` |
| `py/stack-trace-exposure` (medium) | `dashboard.py:8606` | `dashboard.py:13642` | same `return _jsonify({"error": f"unreadable: {_e}"}), 500` |

The offset matches the deletion size. This PR adds no code — it only removes a string literal that never rendered.

**Neither is dismissed here, and neither is fixed here:**

- `py/flask-debug` is behind an explicit opt-in dev flag (`if args.debug:`), and **#5382 "Harden: only enable the Werkzeug debugger on a loopback bind" is already open and mergeable** for exactly this. Fixing it inside a deletion PR would duplicate that work (FLYWHEEL §0.1).
- `py/stack-trace-exposure` is one instance of a class that already has open alerts on `main` (`routes/signals.py`, `routes/selfdiag.py`, `routes/sessions.py`). It deserves its own pass, not a drive-by inside a 5k-line deletion.

CodeQL is not in `scripts/e2e_gate.py`'s `REQUIRED_SPECS`, so it does not gate the merge — recording the reasoning here so that is a decision on the record rather than a check quietly ignored.

PR #5673, opened 2026-09-08
Harden CI: bind auto-deploy-cloud's shell values to env, not into the script

vivekchand:
## What this changes

`auto-deploy-cloud.yml` expanded `${{ inputs.version }}` — and the step output derived from it — directly inside `run:` blocks. A `${{ }}` expansion is substituted into the script *before* the shell starts, so the value arrives as program text rather than as data. An `env:` binding is a lookup the shell only ever reads.

That value is the taint root for the whole job: every later step reads `steps.oss_version.outputs.version`, which is this value, so it reached `sed -i`, a `git commit -m`, and a `gh pr create` title and body.

It also reached a `python3 -c` snippet as `'$V'.split('.')` — pasted into the source text, where a quote in the value would end the string literal and leave the remainder to be parsed as Python.

Both are now bound through `env:` and read as ordinary quoted shell variables, and the version comparison reads `os.environ` instead of interpolated source.

This is the same fix pattern as #5295 (ref out of `run:` blocks), #5298 (setup-openclaw inputs through env) and #5363 (bind `matrix.source` before the shell). This workflow was the one left holding it.

## Why it is safe on the release path

This job is the OSS→cloud pin, so the bar for "no behaviour change" is high:

- **The comparison returns the same verdict on the same inputs.** Checked directly: `0.12.834 > 0.12.833` → yes, `0.12.833 > 0.12.834` → no, `0.12.75 > 0.12.755` → no (the substring trap the surrounding comments call out).
- **A non-numeric version still fails.** `int()` rejected it before and rejects it now — the guard's existing behaviour, not something this PR introduces.
- **`permissions:` is untouched.** It stays `contents: read`. Every cross-repo step here authenticates as `secrets.CLOUD_REPO_PAT`, which workflow permissions do not govern, and the block's existing comment says exactly that. Nothing was elevated or narrowed.
- Triggers, inputs, outputs and step order are unchanged.

## Verification

- All 36 workflow files parse (`yaml.safe_load` over `.github/workflows/*.yml`).
- Zero `${{ }}` expansions remain inside any `run:` block in this file — re-ran the scan that found the original.
- `tests/test_release_dispatches_cloud_deploy.py` and `tests/test_ci_workflow_invocations_are_real.py` — **15 tests, all passing**. These are the two guards that assert against this workflow.
- `scripts/check_action_refs.py` passes (16 distinct refs, all SHA-pinned).
- `scripts/check_product_record.py` passes: "no product-visible code changed".

No-PRD: CI-only hardening under `.github/`, an exempt path. No product behaviour changes — the job's triggers, inputs, outputs and permissions are identical.

https://claude.ai/code/session_01J6pLSRDuyy4zvQoGPFNxwc

---
_Generated by [Claude Code](https://claude.ai/code/session_01J6pLSRDuyy4zvQoGPFNxwc)_


vivekchand:
✨ auto-fixed: merged origin/main into branch — was BEHIND (base at e9bf21a7, main at 94954cad)

---

PR #5688, opened 2026-09-08
Guard and policies learn the two workspace kinds

vivekchand:
No-PRD: closes a specced security-gap issue on the private tracker (vivekchand/clawmetry-pro#223), which carries the requirement and the ranking question.

## Why

`repo_scan` emits `repo_config_exec` and `agent_config_tamper` in the same shape as every detector incident, and no surface knew them. The Guard tab printed the raw id, the policy form could not select them, and `detectors.DETECTOR_KINDS` (which exists precisely so a new detector cannot be added without the surfaces that render it noticing) could not see them because they live outside `detectors.py`.

## One list, not two that drift

- `repo_scan.WORKSPACE_KINDS` is declared where the kinds are produced, re-exported as `detectors.WORKSPACE_KINDS`, and `detectors.ALL_INCIDENT_KINDS` is the union every renderer reads.
- The Guard policy form is now built from `GUARD_KIND_LABEL` rather than a third hand-typed `<option>` list. A second copy is exactly how a kind ends up renderable but not selectable, which is how these two stayed invisible.
- A parametrised guard walks `ALL_INCIDENT_KINDS` and asserts plain-words copy exists in both JS label maps, so a kind added later is covered without editing the test.

## A policy must name a workspace kind

`trigger_kind: ""` now means "any signal about the agent". Both workspace kinds are `critical` by construction, so folding them into the catch-all would silently repurpose every existing "pause anything critical" rule into one that also pauses sessions over a property of a checkout. That is the shape of the preset that paused 280 sessions in 22 minutes. The form says so, with the two kinds in their own optgroup.

With that boundary in place the daemon hands workspace findings to the policy pass, which is what makes a named policy possible at all.

## Ranking: decided, not fudged

A workspace finding carries no spend because it has no spend. On the money ranking it loses to every behavioural incident, so a critical finding sits below every session that happens to be spending. `/api/guard/sessions` gives it its own axis: a separate `workspace` field on the row, sorted ahead of the money key, and the row renders both pills. The alternative was to attach a dollar figure so it would sort well, which is the thing `annotate_spend` exists to refuse.

Also fixes the same cwd bug in `routes/guard.py` that `sync` had: the row read only `metadata` and ignored the `sessions.cwd` column, which is what `runtime_control_support` promotes to find a pid.

## Verified

**Guard proven to fail on the unfixed code, per property:**

```
reverted the catch-all exclusion in policy_engine:
  FAILED test_any_signal_does_not_match_a_workspace_finding[repo_config_exec]
  FAILED test_any_signal_does_not_match_a_workspace_finding[agent_config_tamper]

reverted the workspace split in routes/guard.py:
  FAILED test_the_workspace_finding_is_its_own_field
  FAILED test_a_poisoned_checkout_sorts_above_an_expensive_loop
```

40 tests in `tests/test_guard_workspace_kinds.py`, registered in `ci.yml` (explicit file lists) and in `verification/guards.json`. 207 pass across the guard, detector and policy suites. `make lint-js` and `lint-module-map` green.

Closes vivekchand/clawmetry-pro#223.


vivekchand:
Auto-rebase attempt blocked: the branch `add-pro-readme` diverges from main at a very early commit ("Initial release: OpenClaw Dashboard v0.1.0"), resulting in `add/add` conflicts on 6+ core files (`.gitignore`, `README.md`, `dashboard.py`, `install.sh`, `requirements.txt`, `setup.py`) at rebase step 1/3043. This is a semantic conflict that needs manual resolution — the resolution is not mechanical. Needs human attention to cherry-pick or rebase interactively.

---

vivekchand:
Closing: this PR was opened against the wrong head branch. `gh pr create` ran from the main checkout (which sits on `add-pro-readme`) instead of the worktree, so it carries 7 unrelated files and none of the change the title describes. Reopened correctly below; `add-pro-readme` itself is untouched.

PR #5696, opened 2026-09-08
[RELEASE] Guard learns the workspace kinds + four more red-team cases (carries #5690, #5692)

vivekchand:
No-PRD: release PR carrying two merged security-gap fixes specced on the private tracker (vivekchand/clawmetry-pro#223 and #224).

Carries **#5690** and **#5692**.

## #5690: Guard and policies learn the two workspace kinds

`repo_scan` emits `repo_config_exec` and `agent_config_tamper` in every detector's shape, and no surface knew them: the Guard tab printed the raw id and the policy form could not select them.

- **One list:** `repo_scan.WORKSPACE_KINDS` is declared where the kinds are produced, re-exported as `detectors.WORKSPACE_KINDS`, and `detectors.ALL_INCIDENT_KINDS` is what every renderer reads. The policy form is built from `GUARD_KIND_LABEL` instead of a third hand-typed option list.
- **A policy must name a workspace kind.** `trigger_kind: ""` means "any signal about the agent" and no longer matches them, so a standing "pause anything critical" rule cannot start pausing sessions over a property of a checkout.
- **Ranking decided, not fudged:** its own row field and its own sort axis ahead of money, rather than an invented dollar figure.

Verified live before this release: a real `cursor-agent` session in a poisoned repo produced a `repo_config_exec` critical that sorted **fifth**, below four money-flagged sessions totalling $7.79. That is the exact problem this change fixes, observed in production data rather than argued from first principles.

## #5692: four more disclosed attacks, measured

Three pass and are now corpus cases (npm supply-chain visible half, MCP injection with and without exfiltration), plus a negative control. The fourth is a measured MISS and is filed as a gap with the audit's verdict rather than shipped as a permanently red case: for the invisible half of the npm attack the event stream is byte-identical to an ordinary `npm install`, so no tool-stream detector can separate them.

The corpus README now records what is deliberately absent and why, including that cross-session coordination is inexpressible in a per-session runner.

`audit.py`: 13/13 pass. Corpus suite: 28 tests, up from 24.

## After this publishes

`sync.py`, `routes/guard.py` and `app.js` all changed, so the cloud needs the new pin. The previous release (0.12.837) pinned itself correctly through the release-dispatch path, so this should follow the same route.


vivekchand:
Blocked on drift-bot, and the two findings are legitimate: #5690 changed two behaviours the Guard Enforcement blueprint still describes the old way.

1. `policy_engine.py` — the blueprint says policies match on detector kind but does not record that a WORKSPACE kind must be named explicitly (`trigger_kind: ""` no longer matches `repo_config_exec` / `agent_config_tamper`).
2. `routes/guard.py` — the blueprint states the sort order as `(has incident, incident rank, last_active_at)`; it is now `(has incident or workspace, is workspace, incident rank, timestamp)`.

I have proposed both edits to the blueprint (`Guard Enforcement: acting on a finding`, 76af2dd7) via the factory MCP. They land as **pending tracked suggestions and need review in the app** before the document changes, so drift-bot will keep reporting until they are applied. Once they are, a push here re-runs it (drift-bot only runs on push).

Nothing in this PR itself is at issue: it changes CHANGELOG.md only.

vivekchand:
Blocked on Drift Bot — both findings require blueprint updates in the 8090 Software Factory app that cannot be made from this repository. The author has already proposed the edits as pending tracked suggestions in the factory app; once those are approved, a push here will re-run Drift Bot and should clear the E2E Gate.

---

PR #5700, opened 2026-09-08
Scan package.json install hooks: the other way a checkout runs its own code

vivekchand:
No-PRD: closes the direct half of a `sec-gap` on the private tracker (vivekchand/clawmetry-pro#231), which carries the measured MISS this answers.

## Why

Same predicate as `.git/config`, different file. A `package.json` lifecycle hook is a program the checkout names and a package manager runs, without consent, during an action the developer thinks is "fetch dependencies". It is the CHAINDROP delivery vehicle, and #231 recorded it as a measured MISS with the audit's verdict.

## Measured before designing, not after

189 real `package.json` files on this machine:

| hook | count | runs on install? |
|---|---|---|
| prepublishOnly | 36 | no |
| prepare | 7 | yes |
| postinstall | 4 | yes |
| preinstall | 3 | yes |
| prepack / postpack | 1 each | no |

So only `preinstall`, `install`, `postinstall`, `prepare` are read. Including the publish-time hooks would have quadrupled the noise for no coverage.

**The new scanner over those same 189 manifests: 0 critical, 8 warning.** That matches the discipline the git-config scanner holds (0 critical, 3 warning across 50 repos), and it is the number that decides whether anyone keeps reading the output.

## Graded, not binary

`critical` only for shapes an ordinary build does not have: fetching code and piping it to a shell, decoding a blob, inline `-e`/`-c`, reading a credential file, or reading a publishing token from the environment. Everything else is a `warning` that says what it found.

A recognised tool (husky, patch-package, only-allow, node-gyp, prisma) is **named rather than suppressed**: husky's mechanism is precisely the one CHAINDROP abused, and "via husky" tells an operator something silence does not.

## Scope, stated where it is missing

This catches *"the repo you cloned runs its own code on install"*. The poisoned **transitive** dependency, which is where CHAINDROP actually lived, needs the lockfile — and the measurement says that is not shippable: 73 real lockfiles here carry 2 to 13 `hasInstallScript` packages each, dominated by `fsevents` and `esbuild`, so an incident per lockfile would be noise rather than a finding. That is recorded in the docstring and the blueprint instead of being shipped as a guess, and #231 stays open for it.

## Verified

- `audit.py`: **15/15**, with a new corpus case whose postinstall reads `~/.npmrc`.
- **Guard proven to fail:** reverting `clawmetry/repo_scan.py` alone reds **21 of 59** tests in the corpus file.
- The JS label guard (which walks `ALL_INCIDENT_KINDS`) forced both label maps to learn the new kind before it would go green — the anti-drift mechanism from #5690 working as designed on its first new kind.
- 117 pass across the corpus, workspace-kind and daemon-wiring suites; `make lint-js` green.

Refs vivekchand/clawmetry-pro#231.

### DIFF
--- a/clawmetry/policy_engine.py
+++ b/clawmetry/policy_engine.py
@@ -67,10 +67,10 @@ which is why every existing policy keeps behaving identically.
 All thresholds are AND-ed. An unset threshold (0) never blocks a match, so a
 policy with everything zeroed fires on the first matching incident.
 
-**Workspace findings must be named.** ``repo_scan`` emits two kinds
-(``repo_config_exec``, ``agent_config_tamper``) that describe the FOLDER an
-agent was pointed at, not the agent's behaviour, and both are ``critical`` by
-construction. An empty ``trigger_kind`` therefore does NOT match them: a
+**Workspace findings must be named.** ``repo_scan`` emits kinds
+(``repo_config_exec``, ``agent_config_tamper``, ``package_manifest_exec``)
+that describe the FOLDER an agent was pointed at, not the agent's behaviour,
+and any of them can be ``critical`` by construction. An empty ``trigger_kind`` therefore does NOT match them: a
 standing "pause anything critical" rule, written about runaway agents, would
 otherwise start pausing sessions because of a property of a checkout, with no
 way to write "except that". A policy that wants to act on a poisoned repo says
--- a/clawmetry/repo_scan.py
+++ b/clawmetry/repo_scan.py
@@ -42,8 +42,9 @@ from typing import Optional
 #: produced, and re-exported as ``detectors.WORKSPACE_KINDS`` so every surface
 #: that renders or matches an incident reads ONE list. ``DETECTOR_KINDS`` exists
 #: precisely so a new detector cannot be added without the surfaces noticing;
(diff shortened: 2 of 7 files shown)


vivekchand:
Auto-rebase janitor: **MOAT Verifier (72 tests) flaked** on the original CI run (job `102098993655`) but a subsequent run on the same commit passed (job `102108332873` at 14:49 UTC). The E2E Gate was not re-triggered after MOAT passed, so the PR is still showing as blocked. To unblock: trigger a fresh CI run on this branch (e.g. push a no-op fixup or use the GitHub Actions rerun button on the failed E2E Gate run).

---

vivekchand:
Auto-mergeability sweep: this PR is blocked because the **Drift Bot** leg of the E2E Gate failed — the diff contradicts or adds behavior not covered by the linked Blueprint in 8090 Software Factory.

Fixing this requires updating the Blueprint externally (factory.8090.ai) to reflect the implementation — that's an author decision, not a mechanical code fix.

Skipping — no code change made.

---

vivekchand:
✨ auto-fixed: merged main into feat/scan-package-manifest (was BEHIND)

---

Issue #5681, opened 2026-09-08
feat(onboarding): per-runtime setup prompts, generated from the ingest constants


vivekchand:
## Problem

ClawMetry's users delegate work to coding agents by definition — that is what the product observes. Yet the only setup paths we ship are "run the installer" and "read a doc". There is no artefact aimed at the thing the user actually drives.

Bronto ships a copy-paste prompt per scenario and it is the strongest door on their onboarding screen. Instructive detail: **half of their prompt is negative space** — "do **not** use `/v1/logs`, that path rejects direct JSON with `400 Invalid content provided`", "use `text/plain;charset=UTF-8`, not `application/json`", "headers all lowercase". They wrote down the mistakes agents actually make, which is what makes the prompt work on the first attempt instead of the fourth.

Their own OpenClaw page then spends a boxed warning on the failure mode: coding agents *"mis-substitute secrets (drop characters from the API key, wrap it in extra quotes, or leave the literal placeholder `<YOUR_API_KEY>` in place)"* and *"hallucinate config keys, env-var names, CLI flags, or endpoint paths that look plausible but don't exist"*. A prompt that anticipates those is worth more than one that doesn't.

## Scope

1. **A generator, not a folder of markdown.** Render each prompt from the same constants the server validates against — header names, accepted content types, size caps, the runtime list from `entitlements.ALL_RUNTIMES`, the endpoint from `clawmetry/endpoints.py`. A hand-written prompt drifts the first time a header is renamed, and a drifted setup prompt is worse than none: it teaches the agent a wrong answer confidently.
2. **Serve it** at `GET /api/setup-prompt?runtime=<r>&target=<local|cloud>`, rendered into the onboarding "Ask your agent" door with a copy button, and printed by `clawmetry setup-prompt <runtime>`.
3. **Write the negative space in.** Each prompt states explicitly: the key goes in `x-clawmetry-key` and nowhere else (not a query param, not `Authorization: Bearer`); the placeholder must be replaced verbatim; `protobuf` and `json` are both accepted so don't "fix" the content type; do not invent config keys — if a key isn't in this prompt it doesn't exist.
4. **A verification step inside the prompt.** End every prompt with "then confirm: `curl <endpoint>/api/onboarding/ingest-status`, and report the event count back to the user." An agent that can check its own work fails loudly instead of silently.

## Non-goals

[...]

PR #5705, opened 2026-09-08
docs(ingest): generate docs/INGEST.md from ingest/1 contract (#5682)

### LINKED ISSUE #5682: docs(ingest): generate docs/INGEST.md from one ingest contract, with a drift test
## Problem

Three surfaces are about to describe the same ingest contract, and today none of them shares a source:

1. `docs/CUSTOM_RUNTIME_INGEST.md` — hand-written, covers the run/event API only.
2. The setup prompts (paired issue) — read by agents, where a wrong header is a silent failure.
3. The public endpoint reference on the landing site (paired issue) — where drift-bot now enforces **AC-WEB-001.3 truthfulness**: it greps repo docs and fails a landing claim the repo denies.

Meanwhile the OTLP receiver's real capabilities are undocumented anywhere. `dashboard.py::_otlp_decode` has accepted **protobuf, OTLP/JSON and gzip** for some time. Nothing tells a user that, so functionally it does not exist.

## Scope

**A generated `docs/INGEST.md`**, on the model of `scripts/gen_query_contract_doc.py` (which already generates `docs/QUERY_CONTRACT.md` from the contract, with a drift test). One module declares the ingest contract; the doc, the setup prompts, the landing page and the server's own validation all read it.

Contents, all derived:

- Both surfaces: `/v1/{logs,metrics,traces}` (OTLP) and `/api/v1/runs*` (run/event).
- Accepted content types and encodings — `application/x-protobuf`, `application/json`, `Content-Encoding: gzip`.
- Every header, spelled in the case the server actually matches.
- Size caps and batch caps, as numbers read from the constants that enforce them.
- Every response code and what it means, including what happens to a partially-bad batch (state the real behaviour, whatever it is — an all-or-nothing promise we don't keep is worse than an honest "valid events are kept").
- Which `gen_ai.*` attributes are read, and — explicitly — which are not yet.

**A drift test** in the shape of `tests/test_query_contract_drift.py`: regenerate, diff, fail if the committed doc is stale.

## Why a whole issue for a doc

Because it is the anti-drift mechanism for the other three, not the doc itself. We have shipped a wrong number to users from exactly this shape of gap before — `ARCHITECTURE.md` and `CLAUDE.md` drifted into describing a different product, and the fix was to *generate* `docs/MODULE_MAP.md` rather than to correct the prose. Same move.

## Acceptance

- [ ] `docs/INGEST.md` is generated, and editing it by hand fails CI.
- [ ] Renaming an ingest header changes the doc, the setup prompt and the landing page from one edit.
- [ ] A stranger can push an event using only that page.
- [ ] Every claim on the landing ingest page has a matching line in this doc (the truthfulness gate's requirement).

Part of the ingest-gap plan (phase 1, item 5 of 5).

vivekchand:
## Summary

Plants `clawmetry/ingest_contract.py` as the single source of truth for everything ClawMetry accepts on its two ingest surfaces (OTLP receiver + run/event API), generates `docs/INGEST.md` from it, and adds a drift test so CI catches any future divergence before it reaches users.

This is item 5 of 5 in the phase-1 ingest gap plan (see #5682). It lands before the landing page (#754) and the setup prompts (#5681) because both will read from `clawmetry/ingest_contract.py` rather than duplicating the contract in prose.

No-PRD: docs/drift-guard infrastructure following the existing gen_query_contract_doc.py pattern; product requirement captured in issue #5682

## Changes

- **`clawmetry/ingest_contract.py`** (new) — declares `ingest/1`: OTLP endpoints (`/v1/metrics`, `/v1/traces`, `/v1/logs`), accepted content types (`application/x-protobuf`, `application/json`), `Content-Encoding: gzip`, the 64 MB decompressed cap and its env-var override, response codes, all 13 `gen_ai.*` span attributes read by `_otel_to_row`, the two attributes in GenAI semconv that are **not** yet consumed, and the run/event API endpoints with their OSS/Pro tiers.
- **`scripts/gen_ingest_doc.py`** (new) — generates `docs/INGEST.md` from the contract; `--check` exits 1 on drift for CI (`python3 scripts/gen_ingest_doc.py --check`).
- **`docs/INGEST.md`** (new, generated) — committed so `--check` passes immediately; a stranger can push a `curl` event using only this page.
- **`docs/MODULE_MAP.md`** (updated) — regenerated after adding the two new modules (module map drift guard requires this).
- **`tests/test_ingest_doc_drift.py`** (new) — seven tests: doc-matches-generator (the key drift guard), endpoint set is pinned, content-type coverage, size cap is a positive int, required `gen_ai.*` attrs are present, free-tier runtimes listing is declared, contract version format.

## Test plan

- [ ] `python3 scripts/gen_ingest_doc.py --check` exits 0.
- [ ] `python3 -m pytest tests/test_ingest_doc_drift.py -q` passes (all 7 tests).
- [ ] Manually edit one line of `docs/INGEST.md` → `--check` exits 1 with a clear message.
- [ ] Rename one constant in `clawmetry/ingest_contract.py` → regenerate → confirm the doc, the drift test, and the future setup-prompt consumer all break in the same change.

> **CI note:** `tests/test_ingest_doc_drift.py` needs to be added to the explicit test file list in `.github/workflows/ci.yml` to run in CI (per CLAUDE.md convention). That file is out of scope for this bot — please add it in the same PR or a follow-up before marking ready.

## Bot meta

Draft PR opened autonomously based on the plan in #5682. Marked draft for human review — mark Ready for Review once happy.

Closes #5682

---

https://claude.ai/code/session_01BRCj1wwDRLkfiYN7hsRu5p


vivekchand:
**Merged and live on `main`.** Docs plus a generator and its drift guard — nothing ships in the wheel, so no `[RELEASE]` is involved.

Verified against `origin/main` after merge:

```
.github/workflows/ci.yml:219   python3 -m pytest tests/test_ingest_doc_drift.py -q
$ python3 scripts/gen_ingest_doc.py --check   -> docs/INGEST.md is up to date.
$ python3 -m pytest tests/test_ingest_doc_drift.py -q   -> 7 passed
```

**The guard was checked for going red, not just for existing.** Appending one line to `docs/INGEST.md` fails both the generator's `--check` and the test, each naming the regeneration command:

```
docs/INGEST.md is stale. Run: python3 scripts/gen_ingest_doc.py
AssertionError: docs/INGEST.md drifted from the contract.
```

It is named in `ci.yml` — it was in no file list when I picked this up, which would have made the whole PR decorative: a generated doc whose drift check never executes is a hand-maintained doc nobody knows is stale, the exact failure #5568 fixed for the architecture docs.

### The guard found a real error in the contract it generates from, on its first run

Drift Bot flagged that `gen_ai.operation.name` sat in `GEN_AI_ATTRS_NOT_READ` while `dashboard.py::_process_otlp_traces` (line 4839) reads it. Verified both directions before changing anything.

That list is **published** in `docs/INGEST.md`, so a name in it is a promise to the reader that sending the attribute changes nothing. Here the opposite was true: OpenLLMetry and traceloop-sdk tag the operation there, and it is what decides whether a span counts as a run — a reader who trusted the doc would have dropped the one attribute that makes their runs appear, and seen spans recorded while the live Runs tile stayed at zero.

On `main` now:

```python
GEN_AI_ATTRS_NOT_READ: list[str] = [
    "gen_ai.agent.name",
]
```

`gen_ai.agent.name` stays, because that claim *is* true — ClawMetry's own exporter writes it (`clawmetry/otel_exporter.py`) and nothing reads it back on the ingest path. `gen_ai.operation.name` moved into the read map with what it actually does, and the list now carries a comment saying to verify a name against the code before adding it, not against intent.

This is the argument for generating the doc rather than hand-writing it, demonstrated on itself.

PR #5710, opened 2026-09-08
fix(honesty): an unreadable local store must not render as "you have no data" (closes #5534)

### LINKED ISSUE #5534: A daemon-proxy timeout renders as 'no data' instead of an error, so tabs state a falsehood about the user's data
## What happens

When a read through the daemon query proxy times out, `local_store_call_via_daemon` returns the `PROXY_UNAVAILABLE` sentinel, the `_try_local_store_*` fast paths turn that into `None`, and the tabs render `None` as **empty**.

Observed on a real node 2026-09-05, all at once:

- Sessions: *"No Claude Code sessions have a transcript yet"* — while its own header counted **61 sessions**
- Brain: "Failed to load: timeout"
- Agents: blank
- Cost: `/api/model-attribution` returned `{"model_count": 0, "models": []}` and `/api/usage/by-plugin` returned `{"plugins": []}`, both over a store holding thousands of events

The Cost numbers above were reproduced deterministically by restarting only the dashboard and leaving the daemon on an older build — a shape mismatch, not a timeout, but the same `None` and the same confident empty.

## Why it matters

A spinner says "wait". An error says "something is wrong". An empty state says **"you have no data"** — a positive claim about the user's own work. When the truth is "I couldn't reach the store", that claim is false, and it is indistinguishable from data loss to the person reading it. This is the churn-grade failure named in `FLYWHEEL.md` section 0a: one blank card reads as "this is broken".

The code already predicted this. The comment above `routes/local_query.py::_PROXY_TIMEOUTS` says a contended read can "surface as a transient EMPTY result (urlopen raises → caller returns None → blank tab)".

## The contract already exists

The **Cost and Efficiency Analytics** blueprint specifies the answer and even names it:

> **One name for "could the local store be read".** `store_available` is that name on every surface that reports it [...] The distinction it carries is between an EMPTY result and an UNREADABLE one [...] every reader branches on this field before drawing an empty state.

Today only the outcome trend, the quality report card and the review queue honour it. The Cost roll-ups, Sessions, Brain and Agents do not.

## Suggested shape

1. Fast paths propagate unreachability rather than flattening it to `None` — the sentinel already distinguishes it from a legitimately empty result.
2. Affected endpoints emit `store_available: false` per the existing contract.
3. Tabs branch on that field: "Can't reach the local store — retrying" with a retry affordance, never a zero or an empty list.

An empty list must stay a real answer: a genuinely empty store should still render the normal empty state.

## Related

PR #5524 removed the request storm that was *triggering* these timeouts (the Cost tab's 21 full-event scans per page load, ~38 MB each). That makes the symptom far rarer; it does not stop the next slow read from lying in the same way. This issue is the underlying correctness bug.

https://claude.ai/code/session_01AYTf9NFEjHj6KcpubsryLt

vivekchand:
Closes #5534.

## The bug

On a real node 2026-09-05 a contended read through the daemon query proxy timed out, and four tabs said the same confident thing at once:

| Surface | What it said | What was true |
|---|---|---|
| Sessions | "No Claude Code sessions have a transcript yet" | its own header counted **61 sessions** |
| Cost | `/api/model-attribution` → `{"model_count": 0, "models": []}` | thousands of events in the store |
| Cost | `/api/usage/by-plugin` → `{"plugins": []}` | same |
| Agents | "The AI is idle." | the store simply could not be read |

A spinner says *wait*. An error says *something is wrong*. An empty state says **you have no data** — a positive claim about the user's own work. When the truth is "I could not read it", that claim is false and is indistinguishable from data loss to the person reading it. FLYWHEEL §0a: one blank card reads as "this is broken", and that lesson is not recoverable.

The code already predicted it. The comment above `routes/local_query.py::_PROXY_TIMEOUTS` says a contended read can "surface as a transient EMPTY result (urlopen raises → caller returns None → blank tab)".

## Why it sat unfixed

The contract exists — the Cost and Efficiency Analytics blueprint names the field, and `review` / `bench` / `quality` / `cohort` / the outcome trend already honour `store_available`. The blocker was arithmetic: `local_store_via_daemon` has **~490 call sites across 46 route modules**, and the two earlier passes on this issue both stopped there, correctly, because converting them piecemeal would miss endpoints and ship the same lie in a different tab.

## The fix: record the fact once, stamp it everywhere

**`routes/local_query.py`** — `note_store_unreachable()` / `note_store_read_ok()` / `store_available()`, scoped to the request. Deliberately narrow, so the flag means what it says:

* only a round trip we **actually attempted and lost** counts — a timeout, a stale port, or a daemon that refused the shape (the issue's second, deterministic reproduction: a dashboard restarted against an older daemon build);
* **no daemon at all, or being the daemon, is not unreachability** — the direct open is the supported path there and it works. Flagging it would put a permanent warning on every dev boot and every test;
* a **direct-open fallback that answers clears it** — the store demonstrably was readable from this process.

**`dashboard.py`** — one `after_request` stamps `store_available: false` onto any `/api/*` JSON object that did not answer the question itself, plus an `X-CM-Store-Available: false` header. A fast path written tomorrow is correct by default; there is no helper anyone has to remember to call. It **only ever adds `false`** — a handler that computes the field keeps its own value, and a request where nothing failed is byte-identical, so no existing response shape, snapshot slice or parity test moves.

**Frontend** — the banner reads the **header**, not the body. Cloning and re-parsing every API response (including the multi-MB event scans) to find one field is exactly the request-cost regression FLYWHEEL forbids; this is one header read and no body touched. It self-heals: the next healthy `/api/*` response clears it. The four surfaces named in the issue branch before drawing their empty state, and a top banner says it once for the whole dashboard with a **Retry**.

An empty list stays a real answer: a genuinely empty store still renders the normal empty state.

## Cloud

Unaffected by construction — the hosted container has no discovery file, so no round trip is attempted and nothing is stamped. No new route, so no `cloud_route_policy` entry needed.

## Tests

`tests/test_store_unreachable_is_not_empty.py` (16 cases), wired into the CI file list. It was **proven to go red against the unfixed proxy** before being trusted:

```
FAILED test_a_lost_round_trip_marks_the_store_unreachable
FAILED test_a_daemon_that_refuses_the_shape_is_unreachable_too
2 failed, 14 passed
```

Covers: what counts as unreachable (and what must not), no leak between requests, the stamp adding/not-overwriting/not-touching, non-JSON and top-level-list responses, non-`/api/` paths, the header-not-body rule, each of the four surfaces branching, and the banner being in the LIVE template (§0a.4).

No-PRD: bug fix against a filed issue (#5534); no new product capability.


vivekchand:
## Verified live against a genuinely blocked daemon

Not mocked. A real `local_server` daemon owning a real DuckDB (14 events / 7 sessions), then `kill -STOP` on it — the exact contended-read shape from the issue. Dashboard booted from this branch against the same store.

**Healthy (daemon answering) — byte-identical to today:**
```
GET /api/transcripts -> transcripts: 7
                        store_available key present: False
                        no X-CM-Store-Available header
```

**Blocked (daemon SIGSTOPped) — every surface named in the issue:**
```
GET /api/transcripts        -> transcripts: 0   store_available: false   X-CM-Store-Available: false
GET /api/model-attribution  -> model_count: 0   store_available: false
GET /api/usage/by-plugin    -> plugins: 0       store_available: false
GET /api/subagents          -> subagents: 0     store_available: false
GET /api/brain-history      -> events: 0        store_available: false
```

**In the browser**, Sessions renders the honest sentence and the banner, not "No transcript files found":

> ⚠ Can't reach the collector on this machine, so some screens may look emptier than they are. Nothing has been lost — retrying.  `[Retry]`
>
> Can't reach the collector on this machine, so this list can't be read right now. Nothing has been lost — it reappears as soon as the connection is back. `[Retry]`

**After `kill -CONT`, it self-heals with no user action:** banner hidden, `_cmStoreUnreachable: false`, 7 rows back, no header.

### The live run caught a bug the unit tests could not

The first blocked run still answered `{"transcripts": []}` with **no flag**. Cause: under a standard install `local_store.get_store()` never opens DuckDB — it hands back a `_ProxyStore` that forwards to the *same* daemon. So a dead daemon reaches `_ls_call`'s "direct-open fallback" as a silent `None`, and clearing the flag on "no exception raised" cleared the very flag the proxy had just raised.

Fixed in b0f2e0c (`None` is not an answer) and pinned by `test_a_null_fallback_does_not_clear_the_flag`, which was proven red against the unfixed helper. 17 tests now.

PR #5720, opened 2026-09-08
[RELEASE] cut: the Session Clusters tab, the second surface no user could open

vivekchand:
Follow-up to #5668, which cut the Sub-Agents & Queue Lanes tab. The guard it shipped was scoped to `_CM_RT_ALL_TABS` — the runtime-togglable subset. **`clusters` was never in that list**, so the same class of surface survived the same PR that was meant to close it.

## The tab

`switchTab('clusters')` has **zero callers** — no OSS sidebar entry, none in clawmetry-cloud, no in-app action. The `🧔 Session Clusters` template, its `switchTab()` branch and `loadClusters()` shipped in every wheel with no way in.

It had also **rotted where nobody could see it**:

| `loadClusters()` reads | `/api/sessions/clusters` emits |
|---|---|
| `avg_cost` | `avg_cost_usd` |
| `avg_tokens` | `total_cost_usd` |
| `error_rate` | `cost_tier`, `has_errors` |
| `rep_session` | `top_tools`, `model_family` |

Four fields, none of them still there. Anyone who reached the tab would have been shown `$undefined` and `NaN%` on every card. Nothing failed for as long as nothing could.

**Nothing is lost.** The same endpoint already renders live inside **Usage** as the Trace Clusters panel — against the shape the endpoint actually returns.

## The widened guard

Two changes, and both are load-bearing.

**The subject is the shipped templates, not `_CM_RT_ALL_TABS`.** Every `templates/tabs/*.html` the live `DASHBOARD_HTML` includes. That is what `clusters` slipped through.

**The scan covers every bundle under `static/js/`, not `app.js`.** `trail` is a live, shipped, user-facing tab, and its only `switchTab('trail')` lives in `trail.js`. An `app.js`-only rule would have flagged a working feature — and a guard that cries wolf gets deleted rather than fixed. `test_the_bundle_scan_covers_more_than_app_js` pins that `trail` passes *because* the directory is scanned, not by accident.

### Near-misses checked before deleting

- `renderTraceClusters` shares both the word "clusters" **and** the endpoint with the dead loader — it is the live Usage panel. A substring or endpoint-based deletion takes it. Guarded.
- `_loadOrchestrationPanel` — the near-miss from #5668 — still guarded.

## Verification

- Widened guard proven **RED against `origin/main`**, naming exactly `['clusters']`; green after the cut.
- 7 tests, `test_guard_census.py` 58 pass. Registered in `verification/guards.json`.
- The 7 failures in `test_i18n_no_raw_codes.py` / `test_runtime_tab_capability_parity.py` are **identical on `origin/main` before and after** this change.
- `node --check` on `app.js`, `py_compile` on `dashboard.py`, YAML + JSON parse.

No-PRD: removes an unreachable surface; no capability added.

https://claude.ai/code/session_01LAPuWS8MZGjUrXNZLrfz2H


vivekchand:

---

PR #5726, opened 2026-09-08
clawmetry --sample: a fresh install is never an empty product

### LINKED ISSUE #5715: A machine with no agent history renders an empty product; add a labelled sample mode
If you `pip install clawmetry` on a machine that has no agent sessions on disk yet, there is nothing to look at, and no way to tell a working install from a broken one.

There is currently no sample or demo data path anywhere in the package. Verified on `origin/main` — zero hits across `clawmetry/*.py`, `routes/*.py` and `dashboard.py` for:

```
CLAWMETRY_DEMO | sample_data | demo_mode | --demo
```

## Proposal

`clawmetry --sample` (or `CLAWMETRY_SAMPLE=1`) loads a small synthetic session set into the normal read path, so every tab renders real-looking content.

## Done when

- The sample exercises the session timeline, the tool timeline and at least one detector incident, so the diagnostic workflow is visible.
- Sample data is **visibly labelled** in the UI on every tab it reaches — never presented as the user's own data.
- It is synthetic, not a recorded real session.
- It never writes into the DuckDB store the daemon owns.

vivekchand:
Closes #5715.

No-PRD: pre-launch product-quality work; the gap and its exit condition are stated in #5715.

## The gap

`pip install clawmetry` on a machine with no agent history shows nothing, and a new user cannot tell that apart from a broken install. There was no sample path anywhere in the package — zero hits for `CLAWMETRY_DEMO`, `sample_data`, `demo_mode` or `--demo` across the tree.

## What this adds

`clawmetry --sample` loads three synthetic sessions into the **normal read path**, so every tab renders through the real pipeline rather than a forked demo path:

| Session | Shape |
|---|---|
| `sample-stuck-0001` | circling on a failing migration — 12 identical `Bash` calls, same failure each time |
| `sample-healthy-0002` | a long, healthy run that paginates a search endpoint and finishes |
| `sample-quick-0003` | a two-minute question |

## The incident is earned, not staged

Nothing in the fixture declares the first session stuck. Its tool stream genuinely trips `detectors.stuck_loop` and `repeated_tool_failure` at the shipped thresholds:

```
sample-stuck-0001    incidents=[('stuck_loop','warning'), ('repeated_tool_failure','warning')]
sample-healthy-0002  incidents=[]
sample-quick-0003    incidents=[]
```

A test asserts this, so a threshold change cannot silently leave the demo demonstrating nothing.

## Two real bugs the fixture surfaced

Both are now regression-tested:

1. **Double-counted tool calls.** An event carrying *both* `data.tool_name` and a `data.tool_calls` array is counted twice by `detectors._iter_tool_calls_from_data`, with different argument hashes — which breaks every consecutive-identical-call check. That is why `stuck_loop` did not fire at first.
2. **A false positive in the reference "good" session.** The healthy run was originally *about* rate limiting, and `"rate limiting"` contains the `"rate limit"` marker that `_looks_rate_limited` scans reply text for. The session meant to demonstrate "this is fine" shipped carrying an incident badge. Subject changed; a test now asserts both clean sessions stay clean.

## Isolation

The sample lives in its own DuckDB under `~/.clawmetry/sample/`. `get_store()` gains a sample branch that:

- re-points `DB_PATH` defensively (it is read from the environment at import time, so enabling sample mode late would otherwise open the real store), and
- **never** takes the daemon-proxy path — proxying would serve the user's *real* sessions under a banner claiming they are synthetic.

The banner is server-rendered and not dismissable, so the label is in the first painted frame. A screenshot taken before a fetch resolved would otherwise show invented numbers with nothing marking them invented.

## Verified

- `--sample` boots; `/api/overview` returns the 3 sessions; `/api/transcript/sample-stuck-0001` returns 38 messages.
- Banner present with `--sample`, absent without it (0 occurrences).
- `tests/test_sample_data.py` — 10 tests, wired into `ci.yml`'s explicit file list.
- `tests/test_alert_rules_local_store.py::test_send_heartbeat_attaches_alert_rules_push` fails on this branch **and on clean `origin/main`** — pre-existing, not from this change.

https://claude.ai/code/session_0194WMCbzVfF9nyBh5SBthiZ


vivekchand:
✨ auto-fixed: merged main into branch to bring PR up to date with base

---

vivekchand:
Filed the live requirement drift-bot was asking for: **Sample Mode and First-Run Report** (`84cda328-ec3a-4fdd-b8a0-a7f9c2b2a0a4`).

All four findings were the same gap — the blueprint had no specification for sample mode. The requirement covers each one: the `--sample` flag, the `sample_data` module and its isolation rules (separate DuckDB, never the daemon proxy), the `get_store` sample branch, and the server-rendered banner. It also carries the two constraints that are easy to lose in review — the incident must be *earned* by the tool stream rather than annotated, and the clean sessions must stay clean.

Pushed an empty commit to re-run.

vivekchand:

Drift Bot (8090 Software Factory) reported 5 undocumented items in the "Local Agent Observability" blueprint: the `--sample` CLI flag, the new `sample_data.py` module, the daemon proxy bypass in `get_store()`, the sample-mode banner in `banners.html`, and the index route sample-mode logic in `routes/meta.py`. These require blueprint documentation updates that only the PR author can write.

---

PR #5728, opened 2026-09-08
First-run report: say where ClawMetry looked, not just that it found nothing

### LINKED ISSUE #5716: Empty state should say where ClawMetry looked, not just that there is nothing
When no runtime is detected, the dashboard gives the user nothing to act on. A first-run user cannot distinguish "no agents have run here" from "ClawMetry cannot read them".

This is the same failure class as #5534 (an unreadable local store rendering as "you have no data"), one layer earlier: before ingest, at detection.

## Done when

With no agent data on disk, the first screen states:

- **where it looked** — the actual paths probed per runtime,
- **what it needs** — permissions, and which formats are unsupported on this OS,
- **what to do next** — a link to the sample mode.

And, per the existing convention: nothing renders as "you have no data" when the truthful answer is "nothing was detected here".

vivekchand:
Closes #5716. **Stacked on #5726** (base `hn-sample-mode`) because the panel's call to action is `clawmetry --sample`.

No-PRD: pre-launch product-quality work; the gap and its exit condition are stated in #5716.

## The gap

A dashboard showing nothing is indistinguishable from a broken install, and the user has no way to tell us we searched the wrong place. Same failure class as #5534, one layer earlier — at detection, before ingest.

## Changes

**`runtime_probe`** already presence-probed all 30 runtimes but reported only a boolean. It now reports the locations it actually checked, expanded (`~` resolved), with an unset env-var candidate **named rather than dropped** — `$GOOSE_PATH_ROOT (unset)` explains why nothing was found; an omitted path hides it.

**`/api/entitlement/runtime-detection`** builds its payload from an explicit key whitelist in **two** places, so `paths` had to be added to both or the UI would show an empty list on one code path. `test_detection_endpoint_passes_paths_through` pins that.

**Overview first-run panel**, shown only when there are zero sessions *and* zero events — a user whose agents are merely idle has data, and telling them "nothing detected" would be wrong. It separates the two real cases:

| Case | What it says |
|---|---|
| Runtimes detected, no sessions read | Names them; flags any whose adapters need a plan |
| Nothing detected | "That is a real answer, not an error" + the 30-runtime path list + the macOS Full Disk Access note |

Both end on `clawmetry --sample`.

## Verified in a browser

Clean `HOME`, no runtimes installed. The panel renders:

```
🔍 No agent sessions on this machine yet
No supported runtime was detected. That is a real answer, not an error — here is
exactly where ClawMetry looked, so you can tell us if it looked in the wrong place.
▸ Where ClawMetry looked (30 runtimes)
Want to see what this looks like with data? Restart with clawmetry --sample …
```

Two things that made it *look* broken during that check, recorded because they will bite the next person:

1. The panel renders only after `loadAll()`, which runs on `switchTab('overview')` — the function is `switchTab`, not `showTab`.
2. On a genuinely fresh install the **`onboarding-gate-overlay` sits in front of the entire dashboard**, so this panel is behind it. That gate is a separate, already-tracked problem and this PR does not touch it — but it is worth knowing that first-run users hit the gate before anything here.

Tests: `tests/test_first_run_report.py` (5), wired into `ci.yml`.

https://claude.ai/code/session_0194WMCbzVfF9nyBh5SBthiZ

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -637,6 +637,7 @@ jobs:
             tests/test_overview_claims_are_evidenced.py \
             tests/test_obs_gap_nemoclaw_vllm_5579.py \
             tests/test_sample_data.py \
+            tests/test_first_run_report.py \
             tests/test_e2e_invariants.py \
             tests/test_e2e_telegram_brain_ingest.py \
             -q
--- a/clawmetry/runtime_probe.py
+++ b/clawmetry/runtime_probe.py
@@ -46,6 +46,29 @@ class RuntimeProbe:
     paths: tuple  # candidate globs, relative to ~ unless absolute / env-based
     env: str = ""  # optional env var naming the data dir (adapter-honoured)
 
+    def checked_paths(self) -> list:
+        """The locations :meth:`found` actually looks at, expanded.
+
+        An empty state that says "nothing detected" without saying WHERE it
+        looked is indistinguishable from a broken install, and the user has no
+        way to tell us we searched the wrong place. A candidate whose
+        environment variable is unset stays literal (``$GOOSE_PATH_ROOT/...``)
+        rather than being silently dropped: "we looked here and that variable
+        is not set" is the honest answer.
+        """
+        out = []
+        try:
+            if self.env:
+                root = os.environ.get(self.env)
+                out.append(
+                    os.path.expanduser(root) if root else f"${self.env} (unset)"
+                )
+            for p in self.paths:
+                out.append(os.path.expanduser(os.path.expandvars(p)))
+        except Exception:
+            return out
+        return out
+
     def found(self) -> bool:
         """True when any candidate location exists. Never raises."""
         try:
@@ -216,7 +239,9 @@ RUNTIME_PROBES: tuple = (
 def probe_runtimes() -> list:
     """Presence-probe every supported runtime.
 
-    Returns ``[{id, label, free, found}]`` in catalogue order. Never raises.
+    Returns ``[{id, label, free, found, paths, env}]`` in catalogue order.
+    ``paths`` is what was actually checked, so a caller can show the user
+    where we looked instead of only that we found nothing. Never raises.
     """
     out = []
     for probe in RUNTIME_PROBES:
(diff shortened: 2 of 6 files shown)


vivekchand:
## visual-diff: pre-existing capture instability, not this change

**This change cannot be the cause of these diffs, and that is checkable rather than a judgement call.**

The panel renders behind one guard:

```js
if (sessions > 0 || events > 0) { el.style.display = 'none'; return; }
```

The screenshot fixture has data — its own header reads `OpenClaw · 5 sessions` and `TOKENS USED 418K`. So `renderFirstRunReport` returns at the first line and `#first-run-report` stays `display:none` in **every** capture here. Nothing this PR adds is on screen in any flagged image.

### What the 54 flagged comparisons actually are

| Band | Count | What they are |
|---|---|---|
| ~100% | 30 | Animated views. `desktop flow` before/after are visually identical — the Flow diagram animates its dashed edges, so the dash phase differs between captures and every pixel along every path counts as changed. |
| 1–99% | 24 | Live counters, relative timestamps, activity feeds. |
| <1% | 18 | Static views. |

**`desktop overview` — the only tab this PR touches — is 0.37%**, in the lowest band.

Corroboration: **#5726 flagged 36 of 72** on the same suite, and its only UI addition is a banner behind `{% if sample_mode %}` that never renders outside sample mode. Two PRs that add nothing to the default screen both flag ~half the suite, which is a property of the capture, not of either change.

### Suggestion, separate from this PR

A comparator that flags an animated view at 100% every run trains reviewers to ignore it, which costs it the ability to catch a real regression. Either exclude the animated views (`flow`, `brain`, activity feeds) from pixel comparison, or freeze animations before capture (`prefers-reduced-motion`, or pausing CSS animations in the screenshot harness). Happy to open that as its own issue rather than widen this one.

vivekchand:
Closing: this PR targeted `hn-sample-mode` which just merged into main (#5726). The empty-state content is being carried by draft #5731, which targets main directly. That PR should be marked ready and merged when CI is green.

---

vivekchand:
**Automated janitor run — rebase not attempted (same conflict class as #5742)**

`hn-empty-state` is in the same position as `hn-quickstart-ingest` (PR #5742): main is ~60 commits ahead of the divergence point, and #5742's rebase attempt produced 20+ add/add conflicts on the first commit. Both branches appear to share the same pre-divergence history (sample-mode stack), so the conflict pattern here will be identical.

**Action needed:** manual rebase or merge of main into this branch. All commits are yours so there's no collaborator-branch restriction.

---

vivekchand:

---

vivekchand:
Closing in favour of #5731, which covers the same issue (#5716) more comprehensively (651 additions vs 342, 13 tests vs 5, closes five CI gaps where the detection tests were never wired into ci.yml). This branch also has a Syntax & Lint failure at "Every tab is reachable" that #5731 avoids. The permission-error detection path (the key insight in #5731) is not present here.

---

PR #5731, opened 2026-09-08
fix(first-run): the empty state says where ClawMetry looked (closes #5716)

### LINKED ISSUE #5716: Empty state should say where ClawMetry looked, not just that there is nothing

vivekchand:
Closes #5716.

Same failure class as #5534, one layer earlier: at **detection**, before ingest. A first-run user could not tell three very different situations apart, because all three collapsed into a single `found: False`:

| what is actually true | what the screen said |
|---|---|
| no agent has ever run here | "Nothing to configure" |
| an agent **has** run here and ClawMetry was **refused** access to its files (macOS Full Disk Access) | "Nothing to configure" |
| the path could not be resolved because its env var is unset | "Nothing to configure" |

The middle row is the opposite conclusion, and it is the only one the person can fix.

## The data already existed

`RUNTIME_PROBES` (`clawmetry/runtime_probe.py`) carries the candidate paths for all 30 runtimes. `probe_runtimes()` threw them away, so no endpoint, CLI or screen could name a single path.

* **`RuntimeProbe.inspect()`** — the same probe with its findings kept: `checked` (every candidate, **expanded** as it was actually looked at, so the reader sees the path on *their* machine, not our tilde form), `unreadable` (each refused path with a plain-words reason, never an errno), and the data-dir env var plus whether it is set. `found()` keeps its one-bit contract *and* its never-raises guarantee.
* **`detection_report()`** — shapes that into what a screen needs: found / blocked / where-we-looked.
* **`/api/entitlement/runtime-detection`** — carries it as `detection`. One round trip, no new endpoint. Purely additive: existing consumers of `id`/`label`/`free`/`found` are untouched, and the never-5xx empty envelope carries the key too, so a consumer reading `detection.blocked` cannot throw on the bad day.
* **Both screens that went silent now answer.** `onboarding.js` returned early when nothing was found; `renderInventory` only rendered guidance when something *was* detected. Both now lead with the blocked case, because telling someone to "start an agent" when their agent data is right there and we were refused is the wrong instruction.
* **`clawmetry onboard`** stops printing nothing at all on an empty machine.

## The trap, and why one test needs a real `chmod 000` directory

`glob.glob` and `os.path.exists` **both swallow `PermissionError`** and answer `[]`/`False`. A probe built on those alone can never detect a refused read — the entire "we were blocked" path would have been **dead code that no test would notice**.

Verified on macOS against a real `chmod 000` directory:

```
glob.glob(locked + "/inner/*.json")  -> []          # silent
os.path.exists(locked + "/inner")    -> False       # silent
os.listdir / os.stat / os.scandir    -> PermissionError, errno 13
```

So `_refusal_reason()` walks to the nearest ancestor and asks a question that is *allowed* to fail. `test_a_refused_directory_is_not_reported_absent` runs against a genuinely unreadable directory, and `test_the_plain_glob_check_alone_would_miss_it` pins the platform fact so nobody "simplifies" it back into a glob call. Removing the readability probe reds the first of those.

## Two existing tests changed on purpose

Not to get green — both encoded something that is now wrong:

* `test_render_nothing_detected_is_silent` asserted `== []`. **That silence was the bug.** Now `test_render_nothing_detected_says_where_it_looked`, plus a sibling asserting the blocked case does *not* say "Start an agent".
* `test_probes_never_raise_when_probe_explodes` injected its fault into `found()`, which `probe_runtimes` no longer calls. Left as-is it would run the **real** probes and pass while guarding nothing. Repointed at `inspect()`, and `found()` gains its own case.

## CI gap closed

Six detection guards are now in `ci.yml`. **Five ran in no workflow step at all** (`test_onboard_runtime_detection`, `test_entitlement_api_runtime_detection`, `test_inventory_empty_state_guidance`, `test_no_agent_detected_state`, `test_runtime_prefix_sets_in_lockstep`) — which is exactly how the fault-injection test above could have rotted unnoticed.

## Verified in a browser, on a HOME with no agent data

Welcome modal:

> No agent activity here yet. ClawMetry checked **30** runtimes, including:
> `/tmp/cmhome5716/.openclaw/openclaw.json` · `/tmp/cmhome5716/.openclaw/gateway` · `/tmp/cmhome5716/.nemoclaw` · `/tmp/cmhome5716/.openclaw/sandboxes`
> Start an agent (OpenClaw, Claude Code, Codex, Cursor, and more) and this fills in on its own. Nothing to configure.

Agents tab: a **"Where ClawMetry looked"** disclosure listing each runtime with its probed paths.

## Scope note, stated rather than quietly skipped

The issue's third bullet asks for "a link to the sample mode". **Sample mode does not exist** — that is #5715, and I verified there are zero hits for `CLAWMETRY_DEMO` / `sample_data` / `demo_mode` / `--demo` anywhere in the package. This PR ships the two bullets that stand alone (where it looked, what it needs) plus the next step that is true today. The sample-mode link belongs in #5715.

13 new tests, 151 pass across every detection and i18n suite touched.

No-PRD: bug fix against filed issue #5716.


vivekchand:

E2E Gate (required) fails because the `drift-bot` check is failing. This is a product-review decision by 8090 Software Factory, not a mechanical CI failure. Needs author action.

---

vivekchand:

---

vivekchand:
## Reworked after founder review: the probe map is out of the product

The screenshot of the empty state was publishing our detection map. Two separate problems, and the repo already had a rule against one of them.

**1. The account name.** The panel rendered *expanded absolute* paths, so a real machine shows `/Users/<name>/...` in every screenshot, screen-share, support ticket and marketing shot of a fresh install. `AC-OBS-RSO-030.7` already says no report carries a full filesystem path, and the detector surface holds itself to it. Paths are now home-collapsed at the source, so `~/.codex/sessions` is all anything ever renders, and a catalogue-wide test asserts no reported path contains `$HOME` so a runtime added later cannot reintroduce one.

**2. The detection map.** The paths are readable in `runtime_probe.py` either way (this repo is public), but a table in source and a finished, copy-pasteable list of where we look for all 30 runtimes — rendered by every install, on the screen most likely to be screenshotted — are different artefacts. `CLAUDE.md` splits the repos so runtime knowledge is the moat; the product should not hand the finished map over.

### What changed

* `detection_summary()` is the **servable** half: found / blocked / count, **no paths at all**. `/api/entitlement/runtime-detection` serves that; the per-probe `checked` list no longer travels over HTTP either (a probe row carries a `blocked` bool). A test asserts the whole response body contains no `/` and no `$HOME`.
* A blocked entry names the **runtime and the reason, not the path**. "Give your terminal Full Disk Access" is actionable without naming the file.
* The full map moved to **`clawmetry diagnose`** — a local command whose output a person runs and chooses to share.
* Both screens and the onboard wizard now say the count, surface the blocked case, and point at `clawmetry diagnose`.

Verified on a HOME with no agent data:

```
detection keys      : ['blocked', 'found', 'found_count', 'runtimes_checked']
contains home dir   : False
contains '.openclaw': False
contains 'locations': False
probe row keys      : [allowed, blocked, found, free, id, label, required_tier, required_tier_label]
```

Welcome modal now reads: *"No agent activity here yet. ClawMetry checked 30 runtimes and found none."*

### Third bullet unblocked

`clawmetry --sample` landed on main (#5726) while this was in review, so the issue's "link to the sample mode" is now answered too. All three bullets:

| bullet | answer |
|---|---|
| where it looked | `clawmetry diagnose` |
| what it needs | the blocked/permission case, named per runtime |
| what to do next | `clawmetry --sample`, or start an agent |

The honesty the issue asked for survives. A first-run user still learns that 30 runtimes were checked and that a blocked read is not an absent one; they get the path list by running a command instead of having it published to them.

vivekchand:
Superseded. #5753 landed the first-run report on main while this sat in draft, so the fix now belongs on top of that rather than as a parallel implementation. The disclosure correction this PR was reworked for is carried forward in the follow-up.

PR #5732, opened 2026-09-08
Delete the last 20 dead duplicate definitions from dashboard.py

vivekchand:
No-PRD: finishes a cleanup filed on the private tracker (vivekchand/clawmetry-pro#235) after the AST guard from #5703 found it.

## What

#5706 deleted the 39 byte-identical duplicate pairs. These are the **20 whose bodies differ** — the ones I said then needed judgement per pair rather than a mechanical edit. I did that judgement.

**The deletion is safe regardless of the contents:** a shadowed first definition cannot run, so removing it changes no behaviour. The real question was whether any dead copy held work the live one lacks. It does not, and each case says why:

| pairs | finding |
|---|---|
| 12 | the live copy is far ahead of a frozen snapshot (`detect_config` +448 lines, `_process_otlp_logs` +393, `_otel_to_row` +83, `_fire_alert` +49, …). Nothing to port. |
| 3 | differ only in quote style and line wrapping (`_detect_gateway_port`, `_detect_disk_mounts`, `get_local_ip`). Somebody reformatted one copy. |
| `_send_telegram_alert` | live copy is **smaller on purpose**, and its docstring says why: the removed "gateway fallback" called a helper that was never defined, so it silently delivered nothing. |
| `_detect_gateway_token` | smaller only through formatting. The live copy has every path the dead one has (env, `/proc` scan, all four config files, both nested and top-level auth schemas with the #1127 comment) plus an `isinstance` guard the dead one lacks. |
| `get_public_ip` | the one **deliberate** duplicate. Its dead copy's docstring states it was kept "gated identically so the dead copy cannot reintroduce an ungated third-party call if the definitions are ever reordered". Deleting it removes the thing that defence was defending against; the live copy keeps the gate and the fuller `docs/EGRESS.md` rationale. |

**−1,917 lines**; `dashboard.py` 15,683 → 13,726. With #5706 and the `DASHBOARD_HTML` cut, that file is down from 21,301 lines this morning.

## Verified the way that actually exercises the file

Importing the module from this branch registers **41 routes** and builds a **42,254-character** `DASHBOARD_HTML` — byte-for-byte the same numbers as `origin/main`. No name is lost, nothing is still shadowed, and every survivor is byte-identical to the copy that was already winning. 242 pass across the shadowing, dashboard-html and tab-reachability guards; 23 across the hermetic MOAT suites.

One note for anyone reading `tests/test_api.py` results on this PR: that file drives the **running server** on localhost:8900, not the branch's code, so its pass/fail count says nothing about this change either way.

`dashboard.py` leaves `verification/shadowed_definitions.json` entirely; only `routes/entitlement.py` remains, at 1.

Closes vivekchand/clawmetry-pro#235.


vivekchand:

---

vivekchand:

E2E Gate (required) fails because the `drift-bot` check is failing. This is a product-review decision by 8090 Software Factory, not a mechanical CI failure. Needs author action.

---

Issue #3000, opened 2026-06-10
[chargeback] Per-agent / per-project / per-team cost attribution over rollups


vivekchand:
Makes a truthful version of the 'charge back' claim real. research-backed; see clawmetry-cloud/docs/PRD_QUERY_SPINE.md epic #1603 and the agent-observability research memo (2026-06-10 deep-research, 105 agents, adversarially verified)

## What
- rollup_session already carries runtime + cost_usd + tokens + cwd/project (P2, #2988). Add a user-editable mapping table: (project path / cwd / node / runtime) -> team/label, stored locally (DuckDB) and editable from the dashboard.
- A q/1 method (e.g. usage(group_by=team|agent|project)) that rolls cost up by the mapping; cloud renders it via the generic interceptor (epic #1603 P5) or an interim cm-cloud slice.
- Plain-words UI: 'This week: Team A $X, Team B $Y' with drill to sessions. No team dimension is invented server-side; it is the user's own mapping applied over the rollups (honors E2E: only aggregate counters, plaintext-safe).

## Note
Landing copy was softened to per-agent (#478). Once per-team grouping ships and is verified, restore the team-level copy truthfully.

vivekchand:
Hi @vivekchand! Took a careful look at this one. Here's the plan:

**The problem**
Users running multi-agent or multi-project setups can't tell which agent, project, or team is responsible for their API spend — the Usage tab shows global totals only, so there's no way to do an honest internal chargeback or spot a single runaway project.

**Approach**

1. **Prerequisites check — `rollup_session` needs `cwd` + `project_label`** (`clawmetry/local_store.py` ~L325).
   The schema migration that adds `cwd VARCHAR, project_label VARCHAR` to `rollup_session` (described as P2/#2988 in the issue body) is not yet in the codebase — those columns are absent from the current `CREATE TABLE` and there are no `ALTER TABLE rollup_session ADD COLUMN` migrations for them. This must land before step 3 can read real data. Adding the migration here is the natural P2 entry point:
   ```python
   "ALTER TABLE rollup_session ADD COLUMN IF NOT EXISTS cwd VARCHAR",
   "ALTER TABLE rollup_session ADD COLUMN IF NOT EXISTS project_label VARCHAR",
   ```
   Wire `cwd` from the session JSONL in `clawmetry/sync.py` wherever `rollup_session` rows are upserted (~L9700-9750 area).

[...]

PR #5743, opened 2026-09-09
fix(accuracy): rollup_session.runtime is the runtime, not openclaw for all

vivekchand:
No-PRD: live defect found while verifying clawmetry-pro#228.

## Measured on a real node, before the fix

```
rollup_session   188 rows   runtime='openclaw' → 188      (every other runtime: 0)
sessions         161 rows   agent_type='openclaw' → 161
                 session-id prefixes: claude_code 61, cursor 21, copilot 13,
                 opencode 12, goose 11, codex 10, pi 9, hermes 5, qwen_code 5 …
```

Queried through the running daemon's local query server, not inferred from code.

## Why

Family sessions are written to `sessions` with `agent_type='openclaw'` **by construction** — that column predates multi-runtime and several readers still filter on it, so the family ingest path sets it literally. The runtime lives authoritatively in the `<runtime>:<uuid>` id prefix, which `query_model_rollup`, `query_recent_sessions_by_runtime` and `_runtime_session_id_clause` all key on.

`_mirror_session_rollups_locked` derived `rollup_session.runtime` from `agent_type` instead.

## Two silent consequences

- **`query_rollup_sessions(runtime="codex")` returns nothing.** The per-runtime read of the materialized table cannot see a single paid-runtime session.
- **Per-team cost attribution is wrong for every paid runtime.** `query_usage_by_team` joins `tm.key_value = rs.runtime`, so a `claude_code → Eng Team` mapping never matches and all spend collapses under one `openclaw` label (#3000).

It stayed invisible because every *other* surface derives the runtime from the id prefix — only this table disagreed.

## Fix

A `_runtime_of_session_id` helper: the Python twin of the `CASE WHEN split_part(session_id, ':', 1) IN (...)` expression the other rollups already use, falling back to `agent_type`. An OpenClaw session — including one whose id merely contains a colon — is unaffected, and that is asserted.

**Deliberately not touched:** `sessions.agent_type` and the `_refresh_runtime_day_session_counts_locked` bucketing. That refresh filters the `sessions` table on `agent_type`, so changing only its bucket key would make it count zero. Those two are consistent as they stand; this PR fixes the one place that disagreed.

## Verification

```
# origin/main (unfixed)
FAILED test_family_session_is_filed_under_its_own_runtime
  AssertionError: rollup_session has no rows for runtime='claude_code'
FAILED test_per_team_rollup_can_map_a_paid_runtime
  AssertionError: team mapping runtime=claude_code did not match any
  rollup_session row; labels were {'openclaw'}
2 failed, 1 passed

# this branch
3 passed
```

The one that passes both ways is the no-regression case (OpenClaw stays OpenClaw). Wired into `ci.yml` — this repo runs explicit file lists.

## Honest note on coverage

I merged a guard for this bug class yesterday in clawmetry-pro#236, and **it did not catch this.** That test drives `sync._session_agent_type`, which resolves the runtime correctly — but the family ingest path never calls it; it passes the literal `"openclaw"`. So the guard was green while the bug was live in the path production actually uses. This test asserts against the store's own output instead, which is why it fails on `main`.

## Unrelated pre-existing failure

`tests/test_ingest_bulk_flush.py::test_pre_chain_redelivery_gets_stamped_in_place` fails identically on clean `origin/main` (DuckDB parameter count mismatch, params 17-20). Not touched here, but it is red on main today.

https://claude.ai/code/session_012LnKcB1vis1BufUdz7uWVd


vivekchand:

---

vivekchand:

---

vivekchand:

---

PR #5752, opened 2026-09-09
feat(daemon): the daemon reports the failures that stop it working

vivekchand:
Requirement: [Daemon Field-Failure Reporting](https://factory.8090.ai/project/b415065f-ab2f-4f53-8864-0c009fd098cb/requirements/6f04d779-788b-488f-be94-bf89b34cba1c) (AC-FFR-005), a live child of Field Failure Reporting and Auto-Triage.

Needs clawmetry-cloud PR (linked below) to reach the issue filer.

## The gap

We already run this loop end to end, for exactly one class of failure:

```
desktop shell classifies a failed bootstrap
  -> POST /api/desktop/open (stage bootstrap_failed)
  -> cloud sink counts signatures at /api/desktop/_failures
  -> field-failure-issues.yml files ONE deduped issue per signature
  -> the hourly fixer routine picks it up
```

It is live and it works. Right now it is showing `no_distribution / Windows / 3.11`, a machine nobody reported.

It has **one producer, and it is the installer.** The daemon, which runs on every node and is the single point of failure for all of a customer's data, produced nothing when it could not run.

On 2026-09-08 a Pro node's daemon was locked out of its own pid lock and retried every 30 seconds for twelve hours. It reached nobody, by construction:

| layer | why it saw nothing |
|---|---|
| launchd | the daemon exits `0`, so `KeepAlive` recorded ~1,400 **successful** runs |
| DuckDB error events | `install_daemon_error_event_handler()` is at `sync.py:23972`; the gate is at `23832` |
| auto-update | `start_update_check_thread(role="daemon")` is at `sync.py:24459`, 627 lines past the gate |
| cloud | no daemon means no heartbeat, and quiet is not an event |
| `clawmetry status` | read `launchctl list`'s exit code, which is `0` for a merely registered job (fixed in #5745) |

Every self-repair mechanism we ship starts *after* the gate that failed. The failure existed in one place on Earth: a log file on the customer's laptop. He emailed us a screenshot.

## What this adds

`clawmetry/field_report.py`, the daemon's producer for that same pipeline. Two classes:

- **`daemon_lock_refused`** — reported from ABOVE the gate, and only when ingest is actually stale. Refusing the lock is normally correct and frequent (a double start, a manual `python -m clawmetry.sync`); refusing it while nothing is moving is the field failure. Sent inline rather than on a thread, because the process exits on the next line and would take a background thread with it, which would have made the report as silent as the failure.
- **`daemon_ingest_stalled`** — from the watchdog thread, which keeps running when the ingest loop does not. This is `detectors.py`'s `no_progress` question, asked about ourselves for the first time. We ship stuck-detection for our customers' agents and had never once pointed it at our own daemon.

## Privacy and egress

Same contract as the shell's, asserted key by key over the **serialised** body so nothing smuggles a path through a nested value: a closed enum, platform names, a version. No paths, usernames, hostnames, node id, or log text.

The gate is `endpoints.egress_suppressed()`, deliberately **not** the narrower `is_custom_endpoint()` I reached for first. That one misses both cases that matter: an air-gapped node sets no endpoint, and a self-hosted server IS the endpoint so it never sets one either. A report about a broken daemon is exactly the kind of well-meant call that gets shipped past an air gap. Verified live: `CLAWMETRY_OFFLINE=1` in the same broken state sends nothing.

`docs/EGRESS.md` carries the new row.

## Two things that would have made it useless

- **A restart loop is a new process every 30 seconds**, so in-memory throttling throttles nothing. The stamp is on disk, one report per class per 6h (matching the filer's cron).
- **The sink is keyed `UNIQUE (install_id, session_id, stage)` with `ON CONFLICT DO NOTHING`.** A constant session id would have recorded the first daemon failure an install ever had and silently discarded every one after it, including a different failure class: `last_seen` frozen forever. The id is derived as `<class>-<UTC date>`, which makes the server key "one row per class per install per day" and holds even when the local stamp is gone.

## Verified end to end, not just unit-tested

A local stand-in for the cloud sink, and the customer's machine reproduced (a lock held by a live daemon, ingest stale since 19:32):

```
--- 0.12.748 (before): the failure that reached nobody ---
[clawmetry-sync] Another instance is already running. Exiting.
                                                     (nothing on the wire)

--- this branch ---
[clawmetry-sync] Another instance is already running. Exiting.
WARNING reporting field failure: daemon_lock_refused (aggregate only, no paths or ids)

RECEIVED /api/desktop/open
{ "arch": "arm64", "bootstrap_python": "3.14", "desktop_version": "0.12.748",
  "failure_class": "daemon_lock_refused", "install_id": "ea5d…", "os": "Darwin",
  "os_version": "25.3.0", "session_id": "daemon_lock_refused-20260909",
  "stage": "daemon_failed" }
```

Three further restarts into the same failure: **zero** additional posts. Air-gapped: zero.

`tests/test_daemon_field_report.py`, 24 tests, named in `ci.yml`. The workflow body now says which stage a signature came from, because "bootstrap" over a runtime failure sends the fixer to the wrong file.

https://claude.ai/code/session_01CYQo5mXvPp3zT9tjq5mEaj

---

### Update after merging main: a collision this PR introduced

`main` now carries #5790, whose reconcile step closes a field-failure issue when its signature stops being reported. Both branches touch `field-failure-issues.yml`. The merge was textually clean and the autoclose tests still pass — but a clean textual merge is not a working one, and checking turned up a real collision.

`/api/desktop/_failures` groups by **stage × class × OS × Python**. The issue title is the dedupe key, and it carried only class/OS/Python. So:

```
stage=bootstrap_failed  ->  [field-failure] store_unwritable on Darwin (py 3.12)
stage=daemon_failed     ->  [field-failure] store_unwritable on Darwin (py 3.12)
```

Two distinct signatures, one key. Whichever arrived second would merely *refresh* the first, under a lede describing the wrong failure and pointing the fixer at the wrong file — precisely the confusion this PR's per-stage lede exists to prevent.

**Fixed:** a stage other than `bootstrap_failed` prefixes the title.

```
bootstrap_failed  ->  [field-failure] no_distribution on Windows (py 3.11)
daemon_failed     ->  [field-failure] daemon: store_unwritable on Darwin (py 3.12)
```

`bootstrap_failed` keeps the historical shape **byte for byte** — that first line is #5739's actual title — because renaming it orphans every open issue, which the title contract explicitly forbids. A new stage takes a new prefix rather than reshaping the existing key.

Three tests pin it, computing the title with the workflow's **own shell** rather than a reimplementation (a reimplementation would drift from the thing it claims to check). Proven red against the colliding version:

```
AssertionError: ('[field-failure] store_unwritable on Darwin (py 3.12)',
                 '[field-failure] store_unwritable on Darwin (py 3.12)')
```

Recorded on the **Field Failure Reporting and Auto-Triage** blueprint as a contract on the key: the dedupe key must carry every dimension the aggregate groups by.

### Also fixed

`docs/MODULE_MAP.md` had gone stale as main moved under the branch. That single cause produced **both** red signals: `Syntax & Lint` failing on `gen_module_map.py --check`, and Drift Bot's only finding ("`field_report.py` is missing from MODULE_MAP.md"). Regenerated — `field_report.py` is now at `docs/MODULE_MAP.md:167` and the map is in sync.

837 tests pass locally across the field-report, autoclose, module-map, workflow-YAML and workflow-integrity suites.


vivekchand:

---

vivekchand:

---

vivekchand:
**Live in PyPI `0.12.855`** — verified by driving the published wheel, not the branch.

Downloaded `clawmetry-0.12.855-py3-none-any.whl`, unpacked it, imported `clawmetry.field_report` from those bytes (asserted `__file__` resolves inside the unpacked wheel):

```
module: /tmp/r855/x/clawmetry/field_report.py   (11,290 bytes)
FAILURE_CLASSES: ['daemon_ingest_stalled', 'daemon_lock_refused']
```

Both call sites ship in `clawmetry/sync.py`: `_report_if_ingest_stalled()` at :406 raising `daemon_ingest_stalled` at :417, and `daemon_lock_refused` at :24174 — with no lock-acquire between it and the end of the function, i.e. still **above** the pid-lock gate, which is the entire point of the feature.

**The payload, from the shipped code:**

```json
{"arch": "arm64", "bootstrap_python": "3.14", "desktop_version": "0.12.855",
 "failure_class": "daemon_lock_refused", "install_id": "763e5d9c-…",
 "os": "Darwin", "os_version": "25.3.0",
 "session_id": "daemon_lock_refused-20260910", "stage": "daemon_failed"}
```

**Privacy asserted over the serialised body**, not key by key, so a nested value cannot smuggle anything past a key check:

| checked for | present |
|---|---|
| home path | no |
| username | no |
| hostname | no |
| cwd | no |

**Suppression is total:** `CLAWMETRY_OFFLINE=1` → `report_daemon_failure(...)` returns `False`, nothing sent.

**`last_sync_age_secs()` is tri-state:** returns `None` when no sync has ever completed, rather than `0` — so "never ran" cannot masquerade as "just ran" and trigger a false stall report.

**The sink half was already live and stage-aware** (clawmetry-cloud #2364, merged 2026-09-09): `/api/desktop/_failures` returns a `stage` on every signature today. The pipeline is now complete end to end — daemon producer on PyPI, sink deployed, filer wording each stage correctly with a per-stage dedupe key.

No further cloud promotion needed: the cloud change shipped a day ago, and this is the node-side half that reaches machines through the daemon's own PyPI self-update.

PR #5760, opened 2026-09-09
fix: surface OpenClaw session share state

### LINKED ISSUE #5746: [obs-gap:openclaw] openclaw: session share-link state not surfaced
Automated harness-observability audit found a gap for **openclaw**.

- **Harness exposes:** Sessions can be explicitly published as a revocable read-only public link, exposing both existing and future conversation text to anyone with the link.
- **ClawMetry misses:** Session dataclass has no field for share/publish state (e.g. is_shared, share_url, revoked_at), and the adapter does not read any share-registry file. This is a real security/exposure signal (an agent transcript being publicly link-accessible) that ClawMetry's Guard/security posture surfaces have no visibility into.
- **Where (harness):** `CHANGELOG.md 2026.9.3 highlights (#139489)`
- **Severity:** high
- **Closest capability:** new

_Filed by `scripts/harness/audit.py`. Fingerprint: hgap-b04a00b71e (used to dedupe — keep it in the body)._

Aaqibhafeezkhan:
Closes #5746

OpenClaw 2026.9.3 lets a session owner publish a revocable, read-only public link to a session's existing *and future* conversation text (openclaw#139489). ClawMetry had no visibility into that, so Guard would show a publicly-linked transcript exactly like a private one.

Records, per session, whether its transcript is published — and nothing else about it.

**Where the state comes from.** `publicShare` is in OpenClaw's `PRIVATE_SESSION_ENTRY_KEYS` and every session entry is run through `stripPrivateSessionEntryFields()` before it leaves the gateway, so no session-list response carries it. The read method that does is `session.members.list` (`operator.read`, since 2026.7), addressed one session at a time. Verified against a live 2026.9.3 gateway: it gates on scope rather than reporting an unknown method.

Because the probe is per session it runs on the daemon's sync cycle, not in a request handler — and that is also the only place the signal can be captured for cloud, whose containers have no gateway to ask.

**The token is never stored.** Upstream the object is `{token, createdAt}`, and that token *is* the public link — OpenClaw builds the shareable URL straight from it. Only `bool(publicShare)` and `createdAt` are lifted. `share_fields()` is the single point that reads the upstream object, so the rule has one enforcement site, and a store round-trip test asserts no token reaches DuckDB.

**Unknown is not "not shared".** A probe can be denied, capped out, or unavailable. Where there is no verdict the field is absent end to end, including the metadata whitelist where a real `false` would otherwise be eaten by a truthiness filter. On this machine's gateway — whose raw bearer token carries no `operator.read` scope, the same gap that empties the Sessions and Crons tabs — every session correctly stays unknown, and a one-time warning explains the empty column rather than letting it read as "nothing found".

`createdAt` is a bare `Integer` upstream, so it is normalised with the same magnitude test the rest of the repo uses instead of assuming milliseconds.

Guard gains a `public_share` field and a "Public link" pill, drawn only on an explicit `true`.

Off with `CLAWMETRY_OPENCLAW_SHARE=0`; probe cap via `CLAWMETRY_OPENCLAW_SHARE_MAX` (default 60).

### Verification

- 20 unit tests, including two that go through the real DuckDB ingest — the original defect here was a field that never survived the plumbing under tests that mocked the layer dropping it.
- End-to-end against the live 2026.9.3 gateway: scope-denied probe yields unknown, not a fabricated `false`.
- Python 3.9 import-checked (`routes/guard.py` has no `from __future__ import annotations`).

### Not covered

No positive share verdict has been observed on real data: this gateway's token has no `operator.read` scope, so the probe cannot succeed here regardless. The plumbing is verified with injected responses shaped from the published 2026.9.3 schema.

No-PRD: closes a filed observability gap (#5746); no new product surface beyond one Guard badge.

https://claude.ai/code/session_01Ue5eSBQvXsQxLK4caNHGzL

### DIFF
--- a/clawmetry/adapters/openclaw_share.py
+++ b/clawmetry/adapters/openclaw_share.py
@@ -1,0 +1,60 @@
+from __future__ import annotations
+
+from typing import Any
+
+
+def _share_by_session_id() -> dict[str, dict[str, Any]]:
+    try:
+        from .openclaw import _d
+        rows = _d()._get_sessions() or []
+    except Exception:
+        return {}
+
+    result: dict[str, dict[str, Any]] = {}
+    for row in rows:
+        if not isinstance(row, dict):
+            continue
+        share = row.get("publicShare")
+        if not isinstance(share, dict):
+            continue
+        for key in (row.get("sessionId"), row.get("key")):
+            if key:
+                result[str(key)] = share
+    return result
+
+
+def apply_share_state(sessions) -> None:
+    shares = _share_by_session_id()
+    for session in sessions:
+        share = shares.get(str(session.id))
+        if not isinstance(session.extra, dict):
+            session.extra = {}
+        session.extra["isShared"] = bool(share)
+        if not share:
+            continue
+        created_at = share.get("createdAt")
+        if created_at is not None:
+            try:
+                session.extra["shareCreatedAt"] = float(created_at) / 1000.0
+            except (TypeError, ValueError):
+                pass
+        session.extra["shareVisibility"] = "public-read-only"
+
+
+def apply_share_state_to_payloads(payloads) -> None:
+    shares = _share_by_session_id()
+    for payload in payloads:
+        if not isinstance(payload, dict):
+            continue
+        share = shares.get(str(payload.get("id")))
+        extra = payload.get("extra") if isinstance(payload.get("extra"), dict) else {}
+        extra["isShared"] = bool(share)
+        if share:
+            created_at = share.get("createdAt")
+            if created_at is not None:
+                try:
+                    extra["shareCreatedAt"] = float(created_at) / 1000.0
+                except (TypeError, ValueError):
+                    pass
+            extra["shareVisibility"] = "public-read-only"
+        payload["extra"] = extra
--- a/clawmetry/adapters/openclaw_share.py
+++ b/clawmetry/adapters/openclaw_share.py
@@ -1,60 +1,246 @@
+"""OpenClaw public-share state (issue #5746).
+
+OpenClaw 2026.9.3 lets a session owner publish a **revocable, read-only
+public link** to a session's existing *and future* conversation text
+(openclaw#139489). That is an exposure signal Guard and the security
+posture should be able to see, so ClawMetry records whether a session is
+published — and nothing else about it.
+
+Three facts about the upstream shape drive every decision in here.
+
+**1. It is not on ``sessions.list``.** ``publicShare`` is listed in
+OpenClaw's ``PRIVATE_SESSION_ENTRY_KEYS`` and every session entry is run
+through ``stripPrivateSessionEntryFields()`` before it leaves the
+gateway. Reading a session list — ours or the gateway's — will never
+surface it. The read method that carries it is ``session.members.list``
+(scope ``operator.read``, since 2026.7), addressed one session at a
+time::
+
+    session.members.list {sessionKey, agentId?}
+      -> {sessionKey, publicShare?, owner?, members[], identities[],
+          role, allowedVisibilities[]}
+
+**2. The payload is a live credential.** The upstream schema is::
+
+    publicShare = {token: /^v1\\.[A-Za-z0-9_-]+$/ (<=7000), createdAt: int}
+
+``token`` *is* the public link — OpenClaw builds the shareable URL
+directly from it. Storing it would put a working read capability for the
+user's conversation into DuckDB and into the E2E-encrypted cloud
+snapshot, which is a strictly worse exposure than the one we are trying
+to report on. So this module lifts ``bool(publicShare)`` and
+``createdAt`` and deliberately never returns, logs, or stores the token.
+:func:`share_fields` is the only place that reads the upstream object,
+and it is the enforcement point for that rule.
+
+**3. Not knowing is not the same as "not shared".** Because the probe is
+a per-session RPC it can fail, be out of scope, be capped out, or be
+unavailable entirely (cloud containers have no gateway). #5746 is filed
+as a high-severity exposure gap, so reporting a confident ``isShared:
+false`` for a session we never actually asked about would be the worst
+available answer — Guard would show a publicly-linked transcript as
+private. Where the answer is unknown the field is **omitted**, and the
+read helpers below leave the payload untouched.
+"""
 from __future__ import annotations
 
-from typing import Any
+import logging
+import os
+from typing import Any, Callable, Iterable, Mapping
+
+logger = logging.getLogger("clawmetry.adapters.openclaw_share")
+
+#: Gateway read method that carries ``publicShare`` (scope ``operator.read``).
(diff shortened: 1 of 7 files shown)


vivekchand:
Thanks for taking this on — and apologies for the long reply, but you found a real gap and I want to give you the full picture rather than a one-line "needs changes".

First, credit where it's due: **`publicShare` is the correct field name.** I doubted it initially because it appears nowhere in this repo, and the OpenClaw I had installed locally (2026.7.1) predates the feature. I pulled `openclaw@2026.9.3` to check properly, and it's there — 58 occurrences. Good digging.

The problem is where you're reading it from.

### The share state never reaches your code

`dashboard._get_sessions()` doesn't forward the gateway row — it rebuilds each row from a fixed whitelist of keys (`dashboard.py:16939-16969`), and `publicShare` isn't in it. I put a row carrying `publicShare` through the real function:

```
publicShare present? False
key value: 'abcdef012345...'   sessionId: 'abcdef0123456789'
```

So `_share_by_session_id()` always returns `{}` and `isShared` is always `False`.

That second line is a smaller issue in the same code: `row.get("key")` is the *truncated* id (`'abcdef012345...'`), so that fallback lookup can never match a session id — `sessionId` is the real key.

The tests stay green because they patch `_d` and supply a row shape the real producer never emits, so the path that actually runs is never exercised. That's the part I'd most want you to take from this: a new test is worth confirming red against unfixed code before you trust it green.

### Why adding `publicShare` to that whitelist is not the fix

This was my first instinct too, and it's wrong twice over.

**1. The gateway deliberately strips it.** In 2026.9.3, `publicShare` is listed in `PRIVATE_SESSION_ENTRY_KEYS`, and every session entry is projected through `stripPrivateSessionEntryFields()` before it goes out. It is never present on a `sessions.list` response — not because of our whitelist, but because OpenClaw removed it upstream.

The read method that does carry it is `session.members.list` (scope `operator.read`, since 2026.7), which returns:

```
{ sessionKey, publicShare?, owner?, members[], identities[], role, allowedVisibilities[] }
```

It's addressed per session (`{sessionKey, agentId?}`), so it's an N+1 call against the gateway rather than one list call — which needs a cache and a cap before it goes anywhere near a request handler.

**2. The payload is a live credential.** The schema is:

```
SessionPublicShareSchema = { token: string(^v1\.[A-Za-z0-9_-]+$, max 7000), createdAt: integer }
```

That `token` *is* the public link — OpenClaw builds the shareable URL straight from it. Forwarding the whole `publicShare` object into our session rows would put a working access token into DuckDB and into the E2E cloud snapshot. So whatever we do here, we lift `bool(share)` and `createdAt` and never the token.

Two smaller notes while we're in here: the schema has no `id` field (your test's `"publication-1"` is harmless but not real), and `createdAt` is a bare `Integer`, so the `/1000` needs confirming against a real payload rather than assumed to be milliseconds.

### The cloud half needs a different mechanism

`apply_share_state_to_payloads` runs on the `local_store` path, which is what cloud serves. Cloud containers have no `~/.openclaw` and no gateway, so no per-request RPC can work there at all. Getting share state onto cloud means the sync daemon ingesting it into DuckDB, not a request handler reading it live.

That also means the current code's `extra["isShared"] = False` on that path is the worst possible default for this particular signal — #5746 is filed as a high-severity exposure gap, and a hardcoded "not shared" would have Guard reporting a publicly-linked transcript as private. Where we can't know, we should omit the field, not assert a negative.

### Where that leaves us

This is a bigger change than the PR can carry as written — daemon-side ingest, a cached per-session RPC, and a deliberate decision about the token. None of that is a knock on the patch; the gap is real, the field name is right, and the module/test layout is the shape we'd want. I'm going to pick up the ingest side separately and will link it here so you can follow or take part of it.

Two mechanical CI things, unrelated to any of the above:

- **Syntax & Lint** — `docs/MODULE_MAP.md` is generated, and any new file under `clawmetry/` makes it stale. `python3 scripts/gen_module_map.py`, then commit.
- **PR cites the product record** — needs one line in the PR body. The factory link isn't available to outside contributors, so the escape hatch is what you want: `No-PRD: <one line why>`.

(**E2E Gate** is failing only as a consequence of Syntax & Lint — it'll clear on its own.)

Genuinely good first contribution — you surfaced a real hole in our OpenClaw coverage and got the hard part, the field name, right. Happy to pair on the follow-up if you'd like a piece of it.

vivekchand:
Update — I pushed the rework onto your branch so this can go in with you as the author. You keep the commit credit (`Co-Authored-By`), and I'd rather you see exactly what changed than get a "closed, superseded".

Two things I got wrong in my first review, both worth knowing:

**Your field name was right, and I was wrong to doubt it.** OpenClaw got installed on this machine mid-review, so I checked against a live 2026.9.3 gateway rather than the published bundle. `session.members.list` responds `FORBIDDEN: missing scope`, not `UNKNOWN_METHOD` — the method is registered and `publicShare` is exactly what it carries.

**My suggested fix was actively dangerous.** I told you to add `publicShare` to the whitelist in `_get_sessions()`. Don't — I'd have leaked a credential. The upstream object is `{token, createdAt}`, and that `token` *is* the public link; OpenClaw builds the shareable URL straight from it. Forwarding the object wholesale would have written a working read capability for the user's conversation into DuckDB and into the cloud snapshot. The whitelist you were fighting was accidentally protecting us. The code now lifts only `bool(publicShare)` and `createdAt`, through a single function, with a store round-trip test asserting no token lands in DuckDB.

What the rework changes beyond that:

- Reads from `session.members.list` rather than the session list, which never carries the field.
- Moves the probe to the sync daemon. It's one RPC per session, so it doesn't belong in a request handler — and cloud containers have no gateway at all, so the daemon is the only place the signal can be captured for the half of the product your `apply_share_state_to_payloads` was aimed at.
- Absent instead of `false` wherever there's no verdict. This is the part I'd most encourage you to look at: it required a specific carve-out in the daemon's metadata whitelist, because a genuine `shared: False` is falsey and a truthiness filter was silently eating it. Same bug class as the original one, one layer down.
- Guard gets a `public_share` field and a "Public link" badge, drawn only on an explicit `true`.

On the testing note from my last comment — the two tests worth reading are the ones that go through the real DuckDB ingest instead of mocking it. They're what would have caught the original problem, and I wrote them because your patch is a good example of how a fully-mocked test can pass over dead plumbing. That's a genuinely common trap, not a beginner one.

One honest gap, recorded in the PR body: no positive share verdict has been seen on real data. This gateway's bearer token carries no `operator.read` scope (the same pre-existing gap that empties the Sessions and Crons tabs), so the probe can't succeed here no matter what. It fails in the right direction — everything stays unknown, and a one-time log line explains the empty column rather than letting it read as "nothing found" — but the positive path is verified against the published schema, not observed live.

CI should clear now: module map regenerated, `No-PRD:` line added.

Thanks for this one. Finding the gap and getting the field name right was the hard part, and the reason the rest was straightforward.

vivekchand (approved):
LGTM

PR #5768, opened 2026-09-09
[RELEASE] The first-run panel stops calling working installs empty (carries #5767)

vivekchand:
Releases #5767 (merged, `be91180`).

The first-run report panel rendered **"No agent sessions on this machine yet — No supported runtime was detected"** on every hosted node page, for every user, under a header counting their sessions. Founder screenshot showed it above a hero reading *1281 sessions · $734.09 · running claude-opus-5*.

Two faults, either sufficient on its own:

1. **A missing key was read as a zero.** The cloud node page builds `/api/overview` client-side from the encrypted snapshot and ships `sessionCount` **only** — none of the three keys the gate looked for. It fell off the end of its loop and returned `0`. Absence is now **unknown, not zero**, and the count is the max over the keys actually present (`sessionsToday: 0` on a machine idle since midnight is not empty).
2. **A local-machine diagnostic rendered on a hosted page.** The panel's probe runs inside the cloud container, so it answered about the server while the reader looked at their laptop. It no longer renders under `CLOUD_MODE`, and bails before the probe.

Guarded by `tests/test_first_run_gate_js.{js,py}`, named explicitly in `ci.yml`; all four assertions proven red against the unfixed source. Blueprint ADR-006 records both faults and narrows contract 9.

CHANGELOG entry under `[Unreleased]`.

Blueprint: [Sample Mode and First-Run Report — ADR-006](https://factory.8090.ai/project/b415065f-ab2f-4f53-8864-0c009fd098cb/blueprints/89dfab4d-d6c1-4c72-9222-0e68acd0eeaa)

https://claude.ai/code/session_015Xg7FDBzdpuRirDLzwSGGs


vivekchand:
Drift-bot's finding here is false, and I can show exactly why.

It reports that `clawmetry/static/js/app.js` "does not exist in the repository". The file is on `main` at commit `be91180`, 1,774,330 bytes. What drift-bot is hitting is GitHub's **1 MB inline limit on the Contents API**:

```
$ gh api 'repos/vivekchand/clawmetry/contents/clawmetry/static/js/app.js?ref=main' \
    --jq '{size, encoding, content_len:(.content|length)}'
{"content_len":0,"encoding":"none","size":1774330}
```

`encoding: none`, `content` empty. A reader that treats that as "file absent" concludes the implementation is missing and the tests will fail on execution — which is precisely the wording of the finding.

Note it read the same file correctly on #5767 (it cited `app.js:32347`), because that PR carried app.js **in its diff**. This release PR touches only CHANGELOG.md, so it fell back to fetching the file and got nothing.

This is now a standing tax: at 1.77 MB, **any** PR that does not include app.js in its own diff will red drift-bot on any blueprint claim about it. Filing that separately — the blueprint text is accurate and the code is on main:

```
$ git show origin/main:clawmetry/static/js/app.js | grep -c '_frrLooksEmpty\|_FRR_SESSION_KEYS'
4
$ git show origin/main:clawmetry/static/js/app.js | grep -n "window.CLOUD_MODE) { el.style.display"
32401:  if (window.CLOUD_MODE) { el.style.display = 'none'; return; }
```

vivekchand:
**Automated sweep note** — Drift Bot false positive confirmed blocking this PR (and ~12 others).

Root cause: `clawmetry/static/js/app.js` is **1.77 MB**, which exceeds GitHub's Contents API 1 MB limit. Drift Bot fetches files via that API, receives `encoding: none, content: ""`, and reports the symbols (`_frrLooksEmpty`, `_FRR_SESSION_KEYS`, `renderFirstRunReport`) as "not implemented" — even though `git show origin/main:clawmetry/static/js/app.js | grep -c '_frrLooksEmpty\|_FRR_SESSION_KEYS'` returns 4.

**Fix path (not in this PR's scope):** Drift Bot needs to use the git blob API (no size limit) for files over 1 MB. Until that's deployed, every PR whose diff doesn't include `app.js` will show this false positive.

**All other CI legs pass.** This PR is code-complete and correct; it's blocked only by the Drift Bot infrastructure issue. Human review + override is the unblock path here.

---

PR #5820, opened 2026-09-11
fix(repo_scan): name every runtime that executes a hook file, and the event

vivekchand:
Closes clawmetry-pro#237. **Detection was already correct — the incident named the wrong runtime and asserted the wrong lifecycle.**

No-PRD: bug fix correcting an existing detector's factual output (wrong runtime attribution, wrong hook lifecycle) — no new capability, no behavior change to what is detected or when, only to what the finding says. Trivial re-edit to re-fire the product-record gate after an earlier rerun raced the concurrency group and ran against the stale pre-edit body.

## The defect

`_AGENT_HOOK_FILES` mapped each hook file to exactly one runtime:

```python
(".claude/settings.local.json", "claude_code"),
```

Cursor Desktop reads `.claude/settings.local.json` and executes hooks registered there **without consent** — CVE-2026-48124 / GHSA-pc9j-3qc2-95wv, patched in Cursor 3.0.0. So on a Cursor-only machine the emitted `critical` read:

> `.claude/settings.local.json` registers hook commands that run inside your **claude_code** session, **on every matching tool call** …

Two things wrong in one sentence, and both point the operator away from the real risk:

1. **The runtime.** An operator who doesn't run Claude Code reads that as not-applicable and dismisses a critical on entirely reasonable grounds.
2. **The lifecycle.** The CVE's vector is a `Stop` hook — it fires once at end of turn, not per tool call.

## The fix

- A hook file maps to the **set** of runtimes that execute it.
- **Which one to name comes from the session's own runtime**, which `scan_agent_hooks` already receives — no probe, no new dependency, no cost. When the session's runtime is one of the file's readers: *"your `cursor` session"*. Otherwise name every reader: *"read and executed by Claude Code and Cursor"* rather than picking one.
- `_hook_events()` keeps the lifecycle each command is registered on, so the sentence says *"on the `Stop` event"* / *"on `PreToolUse`"*, and degrades to *"on an event this file does not name"* rather than to a confident wrong one.
- Evidence carries `readers[]` and `events[]`.

Observed output, the three cases:

```
cursor session     → runtime=cursor      "...inside your cursor session, on the `Stop` event..."
claude_code session→ runtime=claude_code "...inside your claude_code session, on the `PreToolUse` event..."
codex session      → "...run on the `Stop` event... read and executed by Claude Code and Cursor."
```

## On the design question the issue raised

#237 asked what to do when the file is present but neither runtime is installed — naming both is honest but re-introduces a milder version of the same problem.

Resolved by using information already in hand rather than adding presence detection: the session being scanned *is* evidence of what runs on that machine. When it matches, the sentence narrows to one runtime and reads exactly as before. When it doesn't, naming both readers is the honest fallback — and it's strictly better than today, which names one of them arbitrarily.

## Verification

```
# origin/main (unfixed)
FAILED test_claude_hook_file_on_a_cursor_session_names_cursor
  AssertionError: named 'claude_code'; an operator running only Cursor reads a
  claude_code attribution as not-applicable and dismisses a critical.
  assert 'claude_code' == 'cursor'
FAILED test_finding_names_the_hook_event_not_a_guessed_lifecycle
  AssertionError: `.claude/settings.local.json` … inside your claude_code session,
  on every matching tool call, …
4 failed, 44 passed

# this branch
48 passed
```

Both primary tests fail on the defect itself, not on an incidental missing key.

Wider: **134 passed** across `test_guard_workspace_kinds.py`, `test_repo_scan_daemon_wiring.py`, `test_redteam_corpus.py`, `test_hook_ownership_quoted_launcher.py`. Corpus audit: **15/15 pass, 0 gaps, 0 control failures**.

**No CI wiring needed** — `tests/test_guard_workspace_kinds.py` is already named in `ci.yml`'s file list (`:714`).

https://claude.ai/code/session_012LnKcB1vis1BufUdz7uWVd


vivekchand (commented):
## Test plan & review notes

**What changed**
- `clawmetry/repo_scan.py`: Hook files now map to a tuple of runtimes instead of a single runtime. Adds `_hook_events()` to preserve lifecycle event names (`Stop`, `PreToolUse`, etc.), and `_RUNTIME_LABELS` for display. Findings now name every runtime that executes a file (not just one), and report the exact lifecycle event rather than "on every matching tool call". Fixes CVE-2026-48124 / GHSA-pc9j-3qc2-95wv — a Cursor-only machine was previously told a critical finding was about Claude Code.
- `tests/test_guard_workspace_kinds.py`: 4 new tests covering the multi-runtime attribution, event name surfacing, unrelated-runtime naming, and Cursor own-settings regression.

**Smoke commands**
```bash
python3 -m pytest tests/test_guard_workspace_kinds.py -v -k "hook"
# or the full suite
make test
```

- `_hook_events()` with a deeply-nested or unusual hooks dict could miss commands (the `for cmd in _hook_commands(val)` path) — worth checking with a `PreToolUse → Bash` style fixture.
- `reader_phrase` logic: a 2-tuple produces `"A and B"` correctly, but a single-item tuple has `readers[0]` as the fallback inside the `not scoped` path — test that the "unknown" fallback also exercises.
- When `runtime` is in `readers`, `rt = runtime` correctly, but `scoped = True` means we say "your X session" — verify the wording holds when `runtime` is `"codex"` and the file is `.cursor/settings.json` (it should take the `not scoped` branch).
- `_events` set from `pairs` after filtering clawmetry hooks — an all-clawmetry-hook file emits no finding, but a mixed file should retain the foreign commands' events. The `_hook_events` → `_hook_commands` interaction is new; a malformed `hooks` value silently yields an empty `pairs`, which is fine (no finding) but worth a quick negative case.

**Issue link**
- No `Closes` in the PR body; this appears to be a standalone fix. If there's a tracker issue for CVE-2026-48124, adding `Closes #N` would auto-close on merge.

---

vivekchand:
Addressed the negative-case coverage gap from the review: added `test_all_clawmetry_hooks_emit_no_finding` and `test_mixed_file_retains_only_the_foreign_command_event` in `59c8103`. Both pass against the existing `scan_agent_hooks`/`_hook_events` code — the review's other three concerns (deep-nesting flattening, the empty-`readers` fallback, and the codex+`.cursor/settings.json` scoping) were verified by direct inspection and exercise: `_hook_commands` recurses through both dict and list nodes so nothing is missed, `readers` is never empty (it comes from the static `_AGENT_HOOK_FILES` table), and the codex/`.cursor` case correctly falls into the `not scoped` branch. No behavior change needed, just the two added tests.

On the "no `Closes #N`" note: `Closes clawmetry-pro#237` in the body is intentional — the tracking issue is in the sibling `clawmetry-pro` repo, so GitHub won't auto-close it on merge, but it's the correct issue to reference.

---

PR #5836, opened 2026-09-11
The stall detector reports every sleeping laptop as broken

vivekchand:
The stall detector reports every sleeping laptop as broken

Six field-failure issues were auto-filed overnight (#5829-#5834), all
daemon_ingest_stalled. The live aggregate shows 8 distinct installs and 9
events in under a day, across Darwin and Linux and five Python versions --
against the one bootstrap signature that has 1 install in 30 days. This is the
detector shipped in 0.12.855 reporting on itself.

The reports are not trustworthy, and the reason is in the clock.

last_sync is a WALL-CLOCK stamp and the check is `now - last_sync > 3600`.
Wall-clock elapsed time cannot distinguish "the ingest loop is wedged" from
"this machine was not running". Two benign cases therefore report as broken:

  - a laptop suspended overnight wakes with an age of hours while the daemon
    is healthy and its next cycle completes seconds later; the watchdog ticks
    within 30s of wake, before that cycle lands;
  - a daemon that just started on a machine that was powered off overnight
    reads a last_sync from yesterday. Every node auto-updates from PyPI, so
    every daemon restarted when 0.12.855 shipped -- which is when these
    reports began.

6 of the 8 installs are Darwin. Laptops sleep.

The fix: the watchdog records which last_sync value is current and the
time.monotonic() reading when it first became current. monotonic() does not
advance across suspend on Darwin or Linux, so "how long have we been watching
this same stamp while actually running" is the honest measure. Both clocks
must now exceed the threshold before anything is reported, and a new last_sync
restarts the window.

No observation history means no report: a daemon that has just started has no
idea, and no idea is never "broken" -- the same rule last_sync_age_secs()
already applies to a node that never synced.

Cost, stated plainly: a genuinely wedged loop is now reported an hour after
the watchdog starts watching rather than immediately at process start. That is
the correct trade. The detector's own definition is "no cycle completed in an
hour", and reporting healthy nodes is worse than reporting a real one late --
an alarm that fires on working machines is one people learn to ignore, which
is exactly what this pipeline was built to avoid.

Four tests, three of them new, all proven RED against the pre-fix code:
the slept-laptop case, the just-restarted case, and the completed-cycle window
restart all fire a report without the guard.

I have NOT proven sleep is the cause of all eight reports -- I could not
reproduce it locally (this machine runs caffeinate and has never slept). The
change is correct either way: if the reports were noise they stop, and if any
were real they keep coming and now mean something.

---

**Requirement:** [Daemon Field-Failure Reporting](https://factory.8090.ai/project/b415065f-ab2f-4f53-8864-0c009fd098cb/requirements/6f04d779-788b-488f-be94-bf89b34cba1c) — this is a correctness fix to **AC-FFR-005.2**, the stall detector that requirement specifies. The requirement says the daemon reports when no cycle has completed in an hour; it does not say "report when the machine was switched off for an hour", which is what the wall-clock-only check actually did.

---

### DIFF
--- a/clawmetry/field_report.py
+++ b/clawmetry/field_report.py
@@ -128,6 +128,17 @@ def _stamp_reported(failure_class: str) -> None:
         log.debug("field report: stamp not written: %s", e)
 
 
+def last_sync_raw():
+    """The raw ``last_sync`` value, or None. Used to tell "the same cycle stamp
+    is still current" from "a new cycle completed", without re-parsing dates."""
+    try:
+        from clawmetry.sync import STATE_FILE
+
+        return json.loads(STATE_FILE.read_text()).get("last_sync")
+    except Exception:  # noqa: BLE001 - missing, unreadable, unparseable
+        return None
+
+
 def last_sync_age_secs():
     """Seconds since the daemon last COMPLETED a sync cycle, or None.
 
--- a/clawmetry/sync.py
+++ b/clawmetry/sync.py
@@ -402,6 +402,11 @@ def _acquire_pid_lock() -> bool:
 # in the stack says so.
 _STALLED_INGEST_SECS = float(os.environ.get("CLAWMETRY_STALLED_INGEST_SECS", "") or 3600)
 
+# What the watchdog last saw, as (last_sync value, time.monotonic() when we
+# first saw that value). The pair is what lets the check tell a suspended
+# machine from a wedged ingest loop; see _report_if_ingest_stalled.
+_STALL_WATCH = {"last_sync": None, "since_mono": None}
+
 
 def _report_if_ingest_stalled() -> None:
     """detectors.py ships ``no_progress`` to tell a customer their agent has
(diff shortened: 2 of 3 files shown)


vivekchand:
**Live in PyPI `0.12.863`** — verified by driving the published wheel.

The wheel stamps `__version__ = "0.12.863"` and carries `_STALL_WATCH`, the `time.monotonic()` comparison, and `field_report.last_sync_raw`.

### Driven from those bytes

`last_sync` is **3 days old in every case below**. Under the old wall-clock-only check, all four would have reported.

```
laptop just woke (monotonic 40s)        reported=False
daemon just started (no history)        reported=False
a cycle completed (new stamp seen)      reported=False
genuinely wedged (2h on both clocks)    reported=True
```

The first two are the false positives that produced 8 installs in a day. The third proves a completed cycle restarts the window, so one old stall cannot report forever. The fourth is the case the detector exists for, and it still fires.

### What this does and does not settle

It stops the noise, and it makes any remaining report mean something. What it does **not** establish is that suspend explains all eight original reports — that was never reproduced locally, since the machine available runs `caffeinate` and has never slept.

That is testable now, and cheaply: the signature dedupes to one open issue per (stage, class, OS, Python), and the throttle is 6 hours. **If `daemon_ingest_stalled` keeps arriving from nodes on `0.12.863`+, the reports are real and point at a genuinely wedged ingest loop.** If it goes quiet, they were sleeping laptops.

The auto-filer closes a signature that stops being reported (#5790), so the six issues will retire themselves once they age out of the 30-day window — no manual cleanup, and their closure will itself be the evidence.

No cloud promotion needed: the detector is node-side and reaches machines through the daemon's own PyPI self-update.

Issue #5813, opened 2026-09-11
82% of the test suite (933 of 1136 files) runs in no CI job


vivekchand:
### Measured while shipping this: 82% of the test suite runs in no CI job

Wiring `tests/test_ingest_doc_drift.py` into `ci.yml` was the **fourth** time in one session that a guard turned out to be named in no workflow file list, so I measured the class instead of just fixing the instance:

```
test files in tests/            1136
named in NO workflow file list   933   (82%)
```

**This is not a measurement artifact.** There is no catch-all `pytest tests/` anywhere in `.github/workflows/` — every invocation is an explicit file list, and the workflow comments say so themselves ("a guard added without a line here runs in no job at all"). Spot-checked both directions:

```
test_api.py                   listed=True
test_workflow_yaml_valid.py   listed=True
test_ac_coverage_guard.py     listed=False   (9 test funcs, 135 lines)
test_adapters.py              listed=False   (15 test funcs, 256 lines)
test_agent_principals.py      listed=False   (17 test funcs, 194 lines)
```

Those are real suites with real assertions, not helpers or fixtures.

**Why it matters more than it sounds.** The explicit-file-list convention is deliberate and documented, and there are good reasons for it (job scoping, dependency isolation, runtime). The problem is that the convention has no enforcement, so the failure mode is silent in exactly the direction that hurts: a test you just wrote passes locally, the PR goes green, and the guard never runs again. Every instance I hit this session was a *newly added guard* — the case where the author is most confident it is protecting something.

**I have not added a gate for this**, deliberately. A check asserting every `tests/test_*.py` appears in some workflow list would fail 933 times on day one, which makes it unshippable as a hard gate and useless as a ratchet without a baseline. The shape that would work is the one this repo already uses for known holes elsewhere: record the current 933 as an accepted baseline, fail only when the count *grows*, and let it ratchet down. That is a real piece of work and a policy decision about CI cost, not something to slip into a docs PR — so I am reporting it rather than doing it.

The four instances I hit and fixed along the way: `tests/test_ci_workflow_invocations_are_real.py` (#5744), `tests/test_field_failure_autoclose.py` (#5790), `tests/test_obs_gap_openclaw_update_pipeline_5749.py` (#5763), and this one.

vivekchand:
Independently re-measured against `origin/main` today (2026-09-11) — the claim holds:

```
tests/test_*.py on origin/main:              1135
distinct test files named across ALL
[...]

PR #5842, opened 2026-09-11
Harden: refuse a DTD in the JUnit XML the quarantine pipeline parses

vivekchand:
Closes a bandit **B314** finding (HIGH confidence) in `scripts/auto_quarantine.py`, reported by the security-audit scanner `supply-chain.yml` runs.

## What the code does

`auto_quarantine.py` downloads a workflow artifact, unzips it, and parses every `.xml` inside with `ElementTree` to decide which tests to quarantine. `ElementTree` does not resolve *external* entities, but it does expand internally-declared ones — so a document that declares its own entities expands to considerably more than it costs to send.

Modern CPython leans on expat's amplification limits here, but those only engage past an output threshold (~8 MiB) and vary with the bundled expat — and this repo's CI matrix still runs 3.9. Measured on the runner while writing this: a 389-byte document expanded to 1,000,000 characters (2570×) without tripping anything.

## The fix

A DTD may appear only in the prolog, before the root element. `_reject_doctype` walks the prolog past whitespace, processing instructions and comments, raises `ParseError` on `<!DOCTYPE`, and stops at the first thing that is none of those.

- JUnit XML has no DTD, so nothing legitimate is refused.
- Stopping at the root element means body text mentioning a doctype cannot false-positive.
- A refused document takes the existing *"could not parse"* path — warn on stderr, skip that one file — so the failure mode matches how malformed XML already behaves. Nothing new can abort a run.

This replaces a dependency on the interpreter's amplification cap with an explicit refusal, rather than adding `defusedxml` (this repo keeps its dependency list minimal, and the CI scripts share it).

## Tests

Ten new cases in `tests/test_auto_quarantine.py`: the expansion shape, ordinary prologs (bare, PI, BOM, comment), a doctype inside element text, and a malformed prolog.

**The file was registered in no workflow.** It existed with 29 passing cases and, because this repo's CI runs explicit file lists rather than `pytest tests/`, ran in no job at all — the exact trap `CLAUDE.md` warns about. This PR adds the `ci.yml` line, so all 39 cases now actually execute.

## Verification

- `python3 -m pytest tests/test_auto_quarantine.py -q` → 39 passed
- `python3 -m pytest tests/test_workflow_yaml_valid.py tests/test_ci_workflow_invocations_are_real.py -q` → 615 passed
- All 39 workflow files parse (`yaml.safe_load`)
- `sync_runtime_count.py --check`, `gen_module_map.py --check`, `check_ac_coverage.py --check`, `test_action_refs_pinned.py` → all clean
- `check_product_record.py` → *"no product-visible code changed"*; only `scripts/`, `tests/` and `.github/` are touched

`No-PRD:` line included in the commit for the product-record gate.

https://claude.ai/code/session_01Gp8WMNNYzRxZP2iR8k6JjG

---
_Generated by [Claude Code](https://claude.ai/code/session_01Gp8WMNNYzRxZP2iR8k6JjG)_


vivekchand:
**Merged and verified live on `main`.** CI-only (`scripts/`, `tests/`, `.github/`) — nothing ships in the wheel, so no `[RELEASE]`.

I exercised the guard against `origin/main` rather than reading the diff, including two evasions the tests did not cover:

```
entity-expansion documents must yield nothing:
   billion-laughs with DTD                      -> {} (refused)   OK
   DTD hidden after a comment                   -> {} (refused)   OK
   DTD after a UTF-8 BOM                        -> {} (refused)   OK

real reports must still parse:
   ordinary JUnit report                        -> ['tests/test_x.py::test_bad']  OK
   comment + processing instruction in prolog   -> ['tests/test_y.py::test_z']    OK
```

The prolog walk handles both evasions correctly — a `<!DOCTYPE` after a comment and after a BOM are still caught — and ordinary reports with a comment or processing instruction before the root still parse. 39 tests pass, and the file is named at `ci.yml:214`.

**A note on the failure mode, since it is not obvious from the diff.** `_reject_doctype` raises, but `_parse_junit_xml` catches it, logs `WARN: could not parse JUnit XML: refusing a JUnit XML document that declares a DTD`, and returns an empty set. So a hostile artifact results in *nothing being quarantined* rather than an exception escaping — the pipeline takes no action on input it will not read, which is the right way round for a script whose job is to modify a quarantine list.

I initially misread that as the guard not firing: my first probe asserted an exception escaped `_parse_junit_xml`, which it deliberately does not. The `WARN` lines were the proof it had fired all along.

**Worth recording:** this PR found `tests/test_auto_quarantine.py` had 29 passing cases and ran in **no job at all** — the same class as #5813. It is one of the pre-existing 932, and wiring it in here moves that count down.

PR #5845, opened 2026-09-11
Guard sees write direction, token values and remote privilege

vivekchand:
## Problem
The [Hugging Face swarm scorecard](https://clawmetry.com/blog/could-clawmetry-have-caught-the-hugging-face-swarm) scored ClawMetry against the July 2026 incident. Five of its misses or partials are things a **single** session already shows, and the detectors could not see them:

| Row | What the swarm did | Why we missed it |
|---|---|---|
| 2 | WebDAV `MKCOL`/`PUT` to an Artifactory host every session could reach | `network_egress` keys on host only, so a read and a write look identical |
| 10 | Uploaded a malicious HDF5 dataset to huggingface.co | same: it could not tell an upload from a download |
| 9 | Read Hugging Face write tokens from output, pasted them to a shared board | `credential_access` matched secret *locations* only: **0 of 45** values in the vendored corpus |
| 3 | Minted admin via a legacy token endpoint, installed a Groovy plugin | `privilege_change` knew local OS escalation only |
| 12 | cluster-admin, lateral movement | same |

## Change
- **Direction** (`detector_surface.py`): each shell segment is classified read or write: curl `-X`/body flags (case-sensitive, so `-x` proxy and `-f` fail stay reads; `-G` stays a GET), wget `--post-*`, httpie, `twine`/`npm`/`hf`/`cargo`/`gem` publish, `s3`/`gs` copies and scp/rsync by destination, and HTTP tools with a `method` argument. The hosts a writing call reaches ride on the step as `write_hosts`. An upload that names no URL (`twine upload`) is still egress.
- **`write_to_read_only_host`** (`network_egress`, warning): a write to a host the cohort has only ever read from. `guard_egress_hosts` gains `writes`, `write_first_seen`, `dir_since` (migration + fresh DDL). Two clocks keep it honest:
  - a host counts as read-only only after its **direction** has been watched for the settle window, so an upgrade (every existing row starts with `dir_since` NULL) cannot flood;
  - a first write inside the window does not count, so a swarm's first writer cannot vouch for its siblings. This is the same rule #5810 applied to new hosts.
- **Token values** (`credential_access`): token shapes (Hugging Face, GitHub, AWS, Google, OpenAI/Anthropic, Slack, Stripe, JWT, PEM, plus the 11 pipelock DLP rules ported with a left boundary) in call arguments, including heredoc bodies, and in tool output. **Category only**; a test asserts no value or fragment reaches the incident. A token sent to a host that doesn't own it, or found in output and then sent anywhere, is **critical**. In args only it's a warning, in output only it's info. Placeholders (`AKIA…EXAMPLE`, `ghp_XXXX`) and kebab slugs are rejected.
- **Remote privilege** (`privilege_change`, critical): role bindings, `--as=system:admin`/`masters`, `aws iam` key/policy grants, `gcloud add-iam-policy-binding`, `az role assignment create`, privileged or `/:/` host-mounted containers, and **writing** calls to Artifactory token/plugin and Jenkins script-console endpoints. A GET of those pages stays quiet.

## Not in scope
Cross-session correlation (`coordinated_action`) and rolling N sessions' alerts into one message: that's batch 2 of the same plan, and it builds on `write_hosts`.

## Risk
- A cohort that legitimately starts writing to a mirror it used to read from gets one warning per session until that first write settles (24h default, `CLAWMETRY_EGRESS_SETTLE_HOURS`). Policies default to monitor.
- An agent that pastes a raw API key into its own command now raises a warning. That's intended; the value is never stored.
- Three nullable columns on `guard_egress_hosts`. If the migration ever fails, `query_guard_baseline` falls back to the old two-column select, so first-time egress keeps working and only the read-only claim goes quiet.

## Verified
- New `tests/test_detectors_swarm_gaps.py` (62 tests, wired into `moat-tests`) covers direction positives and look-alikes, all six read-only scenarios (fires, sibling wrote minutes ago still fires, long-written stays quiet, reads stay quiet, fresh upgrade quiet, legacy store quiet), every value severity, redaction, the remote tier and its look-alikes, the store round-trip (clocks never move), and migration of an old table.
- Pipelock corpus: value lane **0/45 → 45/45**, near-miss corpus still **0** over-fires; `VALUE_LANE_BASELINE` updated deliberately.
- Detector, Guard, policy and schema suites: 225 + 171 + 86 passed, no new skips.
- **Real data:** the new lanes over 400 real sessions (13,434 steps) on a dev machine via the daemon proxy produced **0 incidents**. 6 steps were classified writes, all correct (`github.com`, `api.resend.com`, `app.clawmetry.com`, one test fixture).
- `check_ac_coverage`, `check_ci_test_coverage` (208 listed, ratchet held), `check_py39_annotations`, `gen_module_map --check`: all OK. Ruff on changed files adds only style codes the modules already use (`BLE001` in never-raises helpers, `Optional`, `re.I`); CI's ruff step checks only `dashboard.py`.

Requirement: https://factory.8090.ai/project/b415065f-ab2f-4f53-8864-0c009fd098cb/requirements/2e3e3123-b8b9-4a53-a4dd-0d0dc2db6a01

https://claude.ai/code/session_01Vgzd9YYRXwXa9frbjwSYZL


vivekchand:
Drift Bot's five findings on the previous head describe a truncated view of the files, not missing code. Line numbers from this branch's tip before the move:

- `clawmetry/detector_surface.py`: `_segment_write_hosts` at line 313, `_write_hosts` at 345, `_secret_value_categories` at 423, and **`host_owned_by` at 453** (findings 3 and 4 say it does not exist).
- `clawmetry/detectors.py`: `normalize_events` (line 390) stamps `secret_values` at 436 and 458, and `write_hosts` at 447 and 457 (findings 1 and 5). `session_profile` (line 517) returns `write_hosts` at 531 to 544 (finding 2). The bot cited lines 389 and 506, which are main's line numbers for those functions, before this change.
- `tests/test_detectors_swarm_gaps.py` (62 tests) exercises every one of those paths and passes.

Fixed the durable way anyway, as earlier drift rounds in this repo were:
- The direction and credential-value code moved into its own short leaf module, `clawmetry/detector_payload.py`.
- `clawmetry/detector_surface.py` is back to main's exact bytes.
- The `NormStep` shape comment and the `session_profile` docstring, which the bot reads as the spec, now list the new fields.
- The requirement and blueprint point at the new module.

PR #5862, opened 2026-09-11
Do not re-file a field-failure signature that was just closed

vivekchand:
Do not re-file a signature that was just closed

Six field-failure issues were auto-filed overnight (#5853-#5858). They are
duplicates: same six signatures as #5829-#5834, which another session closed as
COMPLETED at 07:49 having found a root cause. The filer's next tick found no
OPEN issue for those titles and created fresh ones within hours.

That is the dedupe working as written and still producing the wrong outcome. A
signature stays in the 30-day aggregate long after its fix ships, so an issue
closed on the strength of a fix is STILL being reported. Searching open issues
only guarantees a duplicate lands in front of whoever just triaged it, and the
duplicate reads as "the fix did not work" when it is only the window not having
elapsed.

The filer now looks for a recently closed issue with the same title before
creating one. Inside the aggregate window it comments there instead, saying
plainly that a report in this period is not evidence the fix failed and naming
the date after which a recurrence would be. Older than the window it still
files: a signature returning after a full window of silence really is news.

Three tests, all proven red against the pre-fix workflow: the closed lookup
exists and precedes the create, it is bounded by the window so a signature can
still be filed again, and the note does not claim the fix failed.

Also correcting myself. I commented on #5829-#5834 at around 11:00 saying they
were "left open deliberately, because the next report is the evidence either
way". They had been closed three hours earlier. I asserted a state I had not
checked, on six issues at once, and the reasoning in that comment was built on
it.

Worth recording for whoever reads those issues: the versions in #5853-#5858 are
0.12.855 through 0.12.860, all predating the suspend fix in 0.12.863, and every
timestamp is before it shipped. So they are not evidence about that fix either.
The real cause found since is a bare save_state() in the daemon loop (#5861).

No-PRD: fixes churn in the field-failure filer; no product surface changes.

Co-Authored-By: Claude Opus 5 <noreply@anthropic.com>
Claude-Session: https://claude.ai/code/session_01Xb6A5G74JiMe3zHFs1JZEP

### DIFF
--- a/.github/workflows/field-failure-issues.yml
+++ b/.github/workflows/field-failure-issues.yml
@@ -148,6 +148,41 @@ jobs:
               --search "\"$TITLE\" in:title state:open" \
               --json number,updatedAt --jq '.[0]' || echo "")
 
+            # A signature stays in the 30-day aggregate long after its fix
+            # ships, so an issue closed on the strength of a fix is STILL being
+            # reported. Dedupe searches open issues only, so the next tick found
+            # nothing, filed a fresh issue, and produced a second copy of a
+            # signature somebody had just triaged: #5829-#5834 were closed at
+            # 07:49 and re-filed as #5853-#5858 within hours.
+            #
+            # Before filing, look for a RECENTLY CLOSED issue with this title.
+            # Finding one means "already triaged, still ageing out", so comment
+            # there instead of opening a duplicate. Older closures still get a
+            # new issue: a signature returning after the window really is news.
+            if [ -z "$EXISTING" ] || [ "$EXISTING" = "null" ]; then
+              CLOSED=$(gh issue list --repo "$REPO" \
+                --search "\"$TITLE\" in:title state:closed" \
+                --json number,closedAt --jq '.[0]' || echo "")
+              if [ -n "$CLOSED" ] && [ "$CLOSED" != "null" ]; then
+                cnum=$(jq -r .number <<<"$CLOSED")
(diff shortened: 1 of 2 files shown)


vivekchand:
**Merged and verified live on `main`.** CI-only (one workflow and its test), so no release.

```
tests/test_field_failure_autoclose.py   12 passed
field-failure-issues.yml                 parses; state:closed lookup present
```

The three new guards were proven red against the pre-fix workflow before being trusted: the closed lookup exists and precedes the create, it is bounded by the aggregate window so a signature can still be filed again after real silence, and the note does not claim the fix failed.

### What this stops

#5829 to #5834 were closed at 07:49 on the strength of a root cause, and re-filed as #5853 to #5858 the same day. Not a bug in the dedupe as written, but the wrong outcome: a signature stays in the 30-day aggregate long after its fix ships, so searching open issues only guarantees a duplicate lands in front of whoever just triaged it, reading as "the fix did not work" when it is only the window not having elapsed.

### A note on the contract

Drift Bot flagged that I had changed behaviour the blueprint describes. It was right, and the detail is worth keeping: the contract I wrote in #5790 already said a recurrence files fresh **"after a full window of silence"**. That was the intent all along. The open-only search never enforced it. So the record was correct and the implementation did not match it, which is the opposite of the usual drift and easy to misread as a spec change.

The blueprint now states both halves, including that a report inside the window is not evidence about a fix either way.

PR #5867, opened 2026-09-11
Fix cron health-summary grading a job "ok" over 104 failed run_ledger runs

### LINKED ISSUE #5866: Cron Health says "3 healthy" over a job whose last 104 runs failed
Found while verifying the Queue Lanes panel (#5721 / PR #5865), on a node running OpenClaw 2026.9.3. Two panels, one screen, two answers, which is a shape this repo has been burned by before.

## What the screen says

**Cron Health** (top of the Crons tab): `3 healthy`, three green dots, no warnings.

**Queue Lanes** (immediately below it, new in #5865): `Cron: 2 succeeded, 104 failed, 106 total` and `104 of 106 runs in this lane failed.`

Both are describing the same three cron jobs on the same node at the same moment.

## Which one is right

Queue Lanes. The failures are real and they are all one job:

```
sqlite> select runtime, status, count(*) from task_runs group by 1,2;
cron|failed|104
cron|succeeded|2

label:  heartbeat-main
error:  heartbeat skipped: no-route
detail: {"kind":"cron-run","status":"skipped","completionStatus":"failed", ...}
```

`source_id` on every one of those 104 rows is `191283bc-590e-48b5-a333-72fd9b9fc45a`.

## Why Cron Health cannot see it

`GET /api/cron/health-summary` on the same node, same second:

```json
{"totals": {"ok": 3, "error": 0, "warning": 0, "silent": 0, "total": 3}}

{"id": "191283bc-590e-48b5-a333-72fd9b9fc45a", "name": "heartbeat-main",
 "health": "ok", "lastStatus": "pending", "consecutiveFailures": 0,
 "runs7d": null, "isSilent": false}
```

`runs7d` is `null`. The health summary has **no run history at all** for this job, so `consecutiveFailures` is 0, `lastStatus` is the job's configured state rather than an outcome, and the job grades `ok` by default. It is not mis-scoring the failures; it never sees them. It reads the cron job definitions, and the run outcomes live in OpenClaw's run ledger, which it does not read.

`isSilent` is also `false` for a job whose last 104 runs did nothing, because silence is computed from when the job last *fired*, not from whether firing accomplished anything. A job can fire on schedule forever and skip every time.

## Why this matters more than a cosmetic mismatch

The whole promise of the Crons tab is answering "did my scheduled work run, and did it work?" without opening a shell. Here it answered "3 healthy" for days while a job failed 104 consecutive times. A wrong green is worse than a blank panel: a blank panel sends the user to look, a green one tells them not to.

## Suggested direction

Join the health summary to the `run_ledger` rows the daemon already mirrors, rather than teaching it a second source of truth:

- `consecutiveFailures` from the ledger's recent runs for that `source_id`, not from an empty history.
- A job whose recent runs are dominated by `failed` grades `error`, not `ok`.
- Distinguish "has not fired" from "fires and skips". The second is not silence and should not be scored as health.
- Where the ledger has no rows for a job at all, say the run history is unavailable rather than grading it `ok` on no evidence. Grading on absent data is what produced this.

Both surfaces are now on the same screen, so any fix is directly checkable: they should agree.

_Filed from PR #5865, which added the panel that exposed this. Deliberately not fixed there: it is a different module and a different defect._

vivekchand:
**Product record:**

No-PRD: fix of field-reported bug #5866

**Risk:** Low. `_try_local_store_cron_health_summary` only reads an additional table it already had access to (`run_ledger`, via the existing allowlisted `query_run_ledger` daemon-proxy method) and only ever *raises* `consecutiveFailures`, flipping a job to `error`, when the ledger has more consecutive failures on record than the job's own state blob; it never lowers a job's severity. A job with no `run_ledger` rows for its id (or with fewer ledger failures than the state already reports) is completely unaffected. Nothing retroactive, no migration, no delete.

---

## Summary

- `GET /api/cron/health-summary` (the DuckDB fast path, the default path since 0.12.174) graded a cron job "ok" even after its last 104 runs failed, because it read `consecutiveFailures`/`lastStatus` only from the job's own state blob, which OpenClaw never populates for a job that fires and *skips* every time (`no-route`, precondition unmet, ...). The Queue Lanes panel on the same screen reads `run_ledger` (OpenClaw's own run history, synced verbatim) and showed the real picture: `2 succeeded, 104 failed`.
- Cross-check each job against its `run_ledger` rows (`source_id == job_id`, `runtime == 'cron'`): count consecutive failures from the most recent run backward, and let that win when it's higher than what the state blob reports. `lastStatus`/`lastError` follow along so the UI reads coherently.
- Incidental: `_row_to_cron_job` unconditionally set `"model": ""` on every job dict, which had silently broken `/api/crons`' documented contract (issue #3568: the `model` key should be *absent*, not an empty string, for a job with none configured), caught only because `tests/test_crons_local_store.py` needed to be green to wire it into CI (see below). `/api/cron/health-summary` already does its own `job.get("model") or ""`, so its output is unchanged.
- `tests/test_crons_local_store.py` existed but was never referenced by any CI job: no cron-health regression test has ever actually run in CI. Wired it into the MOAT verifier suite job in `.github/workflows/ci.yml`.

## Reproduction

Seeded a cron job whose state blob has zero run history (the exact "`runs7d: null`" shape from the issue) plus five `run_ledger` rows ending in three consecutive `failed`. Before the fix: `GET /api/cron/health-summary` reports `consecutiveFailures: 0`, `health: "ok"`. After: `consecutiveFailures: 3`, `health: "error"`. Verified red to green by temporarily disabling just the cross-check block and re-running `tests/test_crons_local_store.py::test_health_summary_catches_failures_the_state_blob_missed`.

Closes #5866

## Test plan

- [x] Added `test_health_summary_catches_failures_the_state_blob_missed`: reproduces the issue's exact scenario (state blob has no run history, ledger shows 3 consecutive failures) and asserts `health == "error"`.
- [x] Added `test_health_summary_ledger_does_not_override_a_worse_state_count`: a state blob already reporting more consecutive failures than the ledger slice keeps its (worse) count, so the cross-check never silently improves a job's grade.
- [x] Confirmed both new tests fail against the pre-fix code (temporarily reverted just the cross-check block) and pass with it restored.
- [x] Full `tests/test_crons_local_store.py` (19 tests) passes.
- [x] `tests/test_workflow_yaml_valid.py` and `tests/test_ci_workflow_invocations_are_real.py` pass against the `ci.yml` edit.
- [x] `tests/test_no_shadowed_module_functions.py` and `tests/test_module_map_drift.py` pass (no new top-level module-scope symbol collisions).

https://claude.ai/code/session_01BqgpSgJWYQB7wcRh9Qb8fZ

---
_Generated by [Claude Code](https://claude.ai/code/session_01BqgpSgJWYQB7wcRh9Qb8fZ)_

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -806,6 +806,7 @@ jobs:
             tests/test_e2e_invariants.py \
             tests/test_e2e_telegram_brain_ingest.py \
             tests/test_rollup_session_runtime_attribution.py \
+            tests/test_crons_local_store.py \
             -q
 
   # ── Entitlement API test suite (~2700+ hermetic tests) ─────────────────────────────────────────────
--- a/routes/crons.py
+++ b/routes/crons.py
@@ -183,7 +183,6 @@ def _row_to_cron_job(row):
         "schedule": schedule or {},
         "enabled": bool(row.get("enabled", True)),
         "createdAtMs": int(extras.get("createdAtMs") or 0),
-        "model": row.get("model") or "",
         "state": {
             "lastRunAtMs": _parse_iso_to_ms(row.get("last_run_at")),
             "lastStatus": row.get("last_status") or "pending",
@@ -191,6 +190,8 @@ def _row_to_cron_job(row):
             **state_extras,
         },
     }
+    if row.get("model"):
+        job["model"] = row.get("model")
     # Carry through any extra top-level fields (prompt, channel, ...)
     for k, v in extras.items():
         if k not in {"createdAtMs", "schedule", "lastDurationMs",
@@ -280,6 +281,42 @@ def _try_local_store_cron_runs(job_id):
     return runs[:50]
 
 
+# Issue #5866: a cron job's own state blob (``consecutiveFailures``,
+# ``lastStatus``) is only as good as OpenClaw's bookkeeping for it, and a
+# job that fires and *skips* every time (no-route, precondition unmet, ...)
+# never gets a failure recorded there at all — ``runs7d`` stays null and the
+# job grades "ok" on no evidence. ``run_ledger`` (synced from OpenClaw's own
+# ``task_runs`` table) is the ground truth for what actually happened, and
+# the Queue Lanes panel already reads it — so cross-check against it here
+# too rather than trusting the job's self-report alone.
+_RUN_LEDGER_FAILURE_STATUSES = {"failed", "error", "timeout"}
+
+
+def _cron_run_ledger_stats(ledger_runs):
+    """Reduce a job's ``run_ledger`` rows (newest first) to the fields the
+    health summary needs: the most recent status, how many of the most
+    recent runs failed in a row, and whether ANY runs are on record at all.
+
+    Returns ``None`` when ``ledger_runs`` is empty (nothing to cross-check
+    against — the job's own state is the only signal we have)."""
+    if not ledger_runs:
+        return None
+    consecutive_failures = 0
+    for run in ledger_runs:
+        status = (run.get("status") or "").strip().lower()
+        if status in _RUN_LEDGER_FAILURE_STATUSES:
+            consecutive_failures += 1
+        else:
+            break
+    last = ledger_runs[0]
+    return {
+        "lastStatus": (last.get("status") or "").strip().lower(),
+        "lastError": last.get("error") or "",
+        "consecutiveFailures": consecutive_failures,
+        "runCount": len(ledger_runs),
+    }
+
+
 def _try_local_store_cron_health_summary():
     """Return ``/api/cron/health-summary`` payload from the local DuckDB.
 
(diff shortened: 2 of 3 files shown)


vivekchand:
MOAT Verifier failed on the first push: `tests/test_crons_local_store.py` (test-only, wired into CI for the first time by this PR) sits at the very end of the MOAT job's file list, and two earlier, already-wired tests each leak a raw `os.environ[...] = ...` mutation with no matching cleanup, so my file inherited whichever leaked last:

- `test_cli_unattended_update.py::test_real_parser_wires_unattended_into_cmd_update` calls `cli.main()`, which sets `CLAWMETRY_ROLE=dashboard` (`clawmetry/cli.py:9138`) and never unsets it.
- `test_sample_data.py::test_enable_sample_mode_points_the_store_path` calls `sample_data.enable_sample_mode()`, which sets `CLAWMETRY_SAMPLE=1` (`clawmetry/sample_data.py:87`) and never unsets it.

Both flip `local_store.get_store()` away from this fixture's own tmp DuckDB (a `_ProxyStore`, or the real default sample store) regardless of the fixture's `CLAWMETRY_LOCAL_STORE_PATH` monkeypatch. Pushed two follow-up commits that `monkeypatch.delenv` both in the fixture. Verified against `rm -rf ~/.clawmetry` (clean state, no carryover from earlier local runs) plus the exact 123-file MOAT list from `ci.yml`: clean except one pre-existing, unrelated flake in `test_otel_discovery.py` (a real-subprocess scan test, not part of this file and not in the original CI failure).

---

PR #5869, opened 2026-09-11
[RELEASE] fix: Harness Engineering "Can't see" for Codex, OpenClaw + 5 runtimes (quality grader blind spots)

vivekchand:
## Why
The Harness Engineering bench showed **Codex: CAN'T SEE — completion not verifiable** (16 sessions, 0 measurable) while Claude Code ranked normally.

Root cause: Codex records tool-call `arguments` as a **JSON string** (OpenAI function-calling shape). `quality_signals._tool_calls_of` only accepted dict inputs, so every Codex tool input was dropped → `has_tool_inputs=False` → `tool_thrash` unsupported. Codex tool results also carry no `isError`, so `tool_error_rate`/`hard_failure` were unsupported too. Zero supported signals → every session `measurable: false` ("This runtime does not report the signals we grade on (saw: message, other, thinking, tool_call, tool_result)") → `harness_bench._assign_stamps` stamps `cant_see`.

## What
- `quality_signals._input_of`: decode JSON-string `input`/`arguments` (non-JSON strings degrade to `{}`, never raise).
- `test_quality_runtime_conformance._raw_tool_input_hint` counted only dict args, so the codex fixture never asserted inputs survived — that blind spot is how this shipped. Now counts JSON-string args.
- `_family_ingest_rev` salt `/ctx1` → `/ctx1/q2`: `metadata.quality` is graded only at ingest, so without the bump every already-seen Codex session stays unmeasurable.

### Also fixed: OpenAI nested tool-call shape (hermes, kimi, devin, picoclaw, openworker)
Audit of every runtime found a second blind spot of the same class: `{"type":"function","function":{"name","arguments"}}`. The reader took top-level `name` only, so these five runtimes had **no tool calls at all** as far as grading could tell. Live hermes: 0 → 2 of 5 measurable (other 3 thin).

### Also fixed: quiet runtimes never reached the bench at all (per-runtime row cap)
Both bench builders read `query_quality_sessions(limit=1500)`, one cost-ordered cap across all runtimes. On a live node claude_code (2007 sessions) took 1484 rows, codex 16, and **copilot, cursor, devin, exo, goose, grok_bot, hermes, opencode, pi, qwen_code got 0 rows**, so they were absent from the tab, not even "Can't see". New `per_runtime_limit` (DuckDB `QUALIFY row_number() OVER (PARTITION BY runtime prefix ...)`) caps each runtime separately; `/api/bench` falls back to the old call if an older daemon rejects the kwarg. Simulated on live rows: **12 runtimes on the bench (was 2)**; claude_code $/done 18.58 -> 18.90 (same cost ordering, 1500 rows).

### Also fixed: OpenClaw was never graded
Only `sync_family_runtimes` wrote `metadata.quality`; OpenClaw sessions go through `_local_ingest_sessions_batch` and were never graded, so the bench could never rank OpenClaw. Now graded at ingest from stored events, cached per session on `last_active_at` (the metadata upsert replaces the whole blob, so the cache also stops a re-sent row erasing its grade).

## Verification
- Re-graded 25 real Codex sessions from a live store with this code: **19/25 measurable** (was 0/25). The remaining 6 are genuinely thin (4–7 events, 0 tool results) and correctly stay out.
- `pytest tests/test_quality_signals_contract.py tests/test_quality_runtime_conformance.py tests/test_bench_route.py` → 71 passed.

## Paired releases
- clawmetry-pro#245 (merged, 0.7.25): Codex tool results carry `isError` from `exit_code` / "Script failed". With it, Codex supports tool_error_rate + hard_failure (8 of 19 measurable sessions) and real verdicts surface (5 thrash, 5 tool failures).
- clawmetry-cloud#2400: rolls the baked pro wheel to 0.7.25.

## Follow-up (was not in this PR, now shipped in pro#245)
Codex outputs carry `exit_code` in the output text; surfacing it as `isError` in the clawmetry-pro codex adapter would also enable `tool_error_rate` + `hard_failure` for Codex.

No-PRD: bug fix restoring existing Harness Engineering grading for Codex; no new user-visible surface.

https://claude.ai/code/session_01SJNcmoMsdmoywMwafSXCzh


vivekchand:
Reopening: the close at 18:01 was not intended. Ships the Harness Engineering Can't-see fixes (Codex + 5 nested-shape runtimes, OpenClaw graded at ingest, per-runtime bench cap). Paired pro#245 and cloud#2400 are already merged and deployed.

PR #5886, opened 2026-09-11
Harden CI: the last local-action call site moves to self-repository syntax

vivekchand:
**Product record:** No-PRD: CI-only workflow hygiene — one `uses:` line in `.github/workflows/`. No product surface, no runtime code, no test behaviour.

**Risk:** Low, and self-demonstrating — this PR edits `.github/workflows/pr-screenshots.yml`, which is one of that workflow's own `paths:` triggers, so the visual-diff job runs on this PR using the changed line. If `$/` did not resolve, the Setup OpenClaw step already carries `continue-on-error: true` and the job already has the "Export gateway token (CI auth fallback)" step immediately after it for exactly that case, so the failure mode is degraded screenshots on one PR, not a red gate. Undone by reverting one line.

---

## Summary

- `pr-screenshots.yml` was the **last** caller of the `setup-openclaw` composite action still referencing it as `./head/.github/actions/setup-openclaw`. The other five callers — `ci`, `e2e-nightly`, `openclaw-boot`, `oss-golden-path`, `moat-keystone-drive-nightly` — moved to `$/.github/actions/setup-openclaw` in #5852. This one was skipped because it is the only workflow that checks the repository out under a subdirectory (`head/`, `base/`) rather than at the workspace root, so its relative path carried a `head/` prefix and did not match the pattern the others were rewritten by.
- `$/<path>` is GitHub's self-repository syntax: it resolves against this repository at the commit the workflow is already running, and **counts as a pinned reference for policy purposes, which the older `./<path>` does not**. That is the finding this clears. It also needs no workspace checkout at all, so the `head/` prefix goes away with it.
- Behaviour is unchanged. The composite action addresses everything through `${{ github.action_path }}` and installs into its own directory with `npm ci --prefix "$ACTION_DIR"`, so it never depended on living under `head/`.

Completes the batch started in #5852. All six call sites now use the same form.

## Test plan

- [x] `python3 -c "import yaml,glob; [yaml.safe_load(open(f)) for f in glob.glob('.github/workflows/*.yml')]"` — all 40 workflow files parse
- [x] `grep -rn "uses: \./\|uses: \$/" .github/` — six `$/` call sites, zero `./` call sites remaining
- [x] zizmor 1.30.1 (the version `security-audit.yml` runs) before → after on `.github/workflows/` + `.github/actions/`: `self-repository` **1 → 0**, total findings 37 → 36, no other rule's count changed and no new finding introduced
- [x] Read the composite action end to end to confirm it is location-independent: every path goes through `${{ github.action_path }}`, the install uses `--prefix "$ACTION_DIR"`, and the only note about `head/` concerns `setup-node`'s `cache-dependency-path`, which this workflow does not enable
- [ ] CI on this PR — the visual-diff job exercises the changed line directly (this file is one of the workflow's own `paths:` triggers)

---

https://claude.ai/code/session_01B3MMxwYUheAQ8VgxEUvhez

---
_Generated by [Claude Code](https://claude.ai/code/session_01B3MMxwYUheAQ8VgxEUvhez)_


vivekchand:
✨ auto-fixed: merged latest main into branch (BEHIND → up-to-date).

---

vivekchand:
✨ auto-fixed: merged main into branch to resolve UNSTABLE state and trigger a fresh CI run.

---

PR #5889, opened 2026-09-11
Harden CI: the security scanners install from a hash-pinned set

vivekchand:
**Product record:** No-PRD: CI-only supply-chain hardening. Touches `.github/` exclusively (workflow, Dependabot config, new pinned requirement sets) — no product surface, no runtime behaviour, nothing a user can observe.

**Risk:** Low, and it fails loudly rather than quietly. If a pin were wrong the affected `supply-chain.yml` job fails its install step on this very PR — there is no path where a bad pin silently degrades a scan, which is the failure mode this change exists to close. Undone by reverting the commit. Nothing in flight collides: no other open `harden/*` PR touches `supply-chain.yml` (#5886 is `pr-screenshots.yml`). The release pipeline is untouched — `publish.yml` and `release-on-merge.yml` are deliberately out of scope here (see "Not in this PR").

## Summary

- The four jobs in `supply-chain.yml` that **do** the auditing installed their own tooling with a bare `pip install <names>` — `pip-audit` + `cyclonedx-bom`, `zizmor` + `bandit`, `atheris`, and the `certifi`/`pytest`/`requests` the font check needs. Each was an unpinned resolve against PyPI at job start, in a job that runs with this repository's token and the tree checked out. The thing doing the auditing was the least verified install in the repo.
- Each now installs with `--require-hashes -r .github/requirements/<set>.txt`. Under that flag pip installs exactly the artifacts the file names, or it fails — there is no partial, quieter outcome. Each pinned version lists the sha256 of **every** distribution PyPI publishes for it, not just the wheel this resolve happened to pick, so the pin does not depend on the runner landing on the same wheel.
- `.github/dependabot.yml` gains a grouped `pip` entry for `/.github/requirements` on the same seven-day cooldown as every other ecosystem here. A pin with no updater is the frozen end of the same problem an unpinned install is at the other end of; Dependabot rewrites the `--hash` lines along with the versions.

### What is deliberately left unpinned

The SBOM step's `pip install -r requirements.txt` stays as it was, and now carries a comment saying why: that install **is** the live resolve being measured. An SBOM of a frozen, hash-pinned environment would describe the lock file instead of describing what `pip install clawmetry` produces today, which is the only question the SBOM is published to answer.

## Test plan

Run on Linux / CPython 3.11, matching what these jobs pin, so the resolve is the same one CI gets.

- [x] All four sets install clean under `--require-hashes` in fresh venvs
- [x] **Enforcement actually proven, not assumed** — adding one unlisted package to the install makes pip exit 1 and install nothing (`ModuleNotFoundError` after the attempt)
- [x] Every pinned line carries `==` and at least one hash: 8 / 10 / 11 / 50 pins, 1,169 hashes total, zero unpinned
- [x] The tools run from the pinned sets: zizmor 1.30.1, bandit 1.9.4, pip-audit 2.10.1, cyclonedx-py 7.3.1, `import atheris`, pytest 9.1.1
- [x] **The pin changes no scanner output** — `bandit -r .` with CI's own `-ll` and exclude flags returns 282 findings from the pinned build, identical to the count in the current unpinned job's uploaded artifact
- [x] `python-deps` job end to end: pinned install → `pip-audit --strict` clean → the `requirements.txt` resolve → a 70-component SBOM
- [x] `zizmor --offline` on the changed workflow: 0 findings
- [x] The pytest steps of the two affected jobs pass: `tests/test_fuzz_targets.py` (204 passed), `tests/test_vendor_fonts_degrades.py` (7 passed)
- [x] All 40 workflow files parse, and `.github/dependabot.yml` parses with 11 ecosystem entries
- [x] `tests/test_workflow_yaml_valid.py` — 583 passed, 373 skipped
- [x] `make lint-py39 lint-ci-test-coverage lint-module-map lint-runtime-count` pass. `make lint-py` fails identically on `origin/main` (230 pre-existing ruff errors, none in a file this PR touches — the diff contains no Python)

## Not in this PR

Scorecard reports 68 unpinned `pipCommand` sites across this repo; these 4 are one coherent batch. Noting the rest so the next pass has a map rather than re-deriving it:

- `publish.yml` and `release-on-merge.yml` each install `build` + `twine` unpinned in the job holding the PyPI credential. That is the highest-value remaining target by some distance, and also the one that must not be changed blind: those jobs do not run on a PR, so a pin there is not exercised by its own CI. It deserves its own PR with a deliberate verification plan.
- `ci.yml` holds 19 sites, the largest single concentration.
- The remaining five `third-party GitHubAction not pinned by hash` hits are Scorecard misreading the `$/.github/actions/setup-openclaw` self-repository syntax (#5852, #5886). They are not pinnable and not a real finding.

---

https://claude.ai/code/session_016eLHiTjGbC5HGsFUvZDbx2

---
_Generated by [Claude Code](https://claude.ai/code/session_016eLHiTjGbC5HGsFUvZDbx2)_


vivekchand:
✨ auto-fixed: merged latest main into branch (was 1 commit behind).

---

vivekchand:
**`E2E Gate (required)` failed on `16576106`, and it is not this PR's failure.** Recording why, since the red is misleading.

The gate did not observe a failing leg — it ran out of time waiting for one. From its log:

```
```

Every other leg it aggregates reported `passed` inside the window, and on this head 34 checks are green, 1 skipped, and **zero failed**. `Entitlement API tests` simply never got a runner within the hour; it went `in_progress` only after the gate had already given up, and it is still running as I write this.

Why it is not attributable to this change: the diff is four files under `.github/` — three hash-pinned requirement sets, the `supply-chain.yml` steps that install them, and a Dependabot entry. It cannot affect whether `Entitlement API tests` gets scheduled. The four jobs this PR actually changed (`Vendored fonts not stale`, `Python dependency audit`, `Fuzz untrusted parsers`, `Security audit`) have passed on every head this branch has had.

There is no fix to port, because there is no defect: the remedy for a gate that timed out on queue depth is to re-run the gate once its legs have landed. I have re-run it, and I will keep watching this PR until it is green.

Contributing factor worth flagging for a human, since it will recur on any long-lived PR here: `main` is moving fast enough that the auto-merge automation has merged it into this branch five times (`3bf6856` → `a82b4273` → `81cb1efe` → `47175a7d` → `16576106`), and each merge restarts the full matrix. When the queue is deep enough that a leg cannot start within 3600s, the gate's timeout fires before the matrix finishes — so the PR can go red without anything being wrong with it. Raising the gate's timeout, or having it distinguish "a leg failed" from "a leg never got a runner", would stop that. That is a repo-level call, not something I am changing from inside this PR.

---

vivekchand:
Resolved: the gate re-run (attempt 2) passed once `Entitlement API tests` got a runner, and `16576106` is now 37 green / 1 skipped with `mergeable_state: clean`. Nothing was changed to achieve that — the previous red was only the timeout described above.

The systemic note in that comment still stands as a repo-level suggestion, and is not blocking this PR.

---

PR #5891, opened 2026-09-12
The hosted session page counted zero messages for every non-OpenClaw runtime

vivekchand:
Founder report 2026-09-11 (screenshot): a Codex session page on app.clawmetry.com read **Messages 0** and **Cost \$0.00** beside a transcript the local store counts as **371 messages**.

## Why

The cloud's `sessions` table has **no `message_count` column**. The one count it stores is `event_count` (`clawmetry-cloud/routes/api.py`), and the hosted session page renders "Messages" straight from it:

```python
# clawmetry-cloud/routes/team_sessions.py
'messages': int(r.get('event_count') or 0),
```

The OpenClaw cloud row has always sent `event_count`. The **family** row (claude_code, codex, cursor, copilot, …) sent `message_count` instead, which the cloud ignores, so the column stayed `0` for every non-OpenClaw session ever synced.

## What

The family cloud row now carries `event_count` as well, the same raw-row count the OpenClaw path sends. The cloud upsert keeps `GREATEST(existing, new)`, so a read capped by `_family_event_read_cap()` can never walk a session's count backwards. **No cloud-side change is needed** - the column it already reads simply stops being empty.

## Testing

`tests/test_family_cloud_row_event_count.py` drives the real `sync_family_runtimes` with a fake adapter (4 events, `message_count=2`), captures the cloud push, and asserts the session row carries `event_count == 4` while `message_count` stays. Added to the CI job list.

## Not in this PR

The same screenshot showed **Cost \$0.00** while the list row for that session showed \$3.40. Those two hosted surfaces disagree, so they are reading different sources; that needs live cloud data to pin down and is not guessed at here.

No-PRD: user-reported display bug; restores the count the page was always meant to show.

https://claude.ai/code/session_0183r5LYZZeESUW91ffsQ2AV

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -820,6 +820,7 @@ jobs:
             tests/test_insights_outcome_templates.py \
             tests/test_event_shape.py \
             tests/test_injected_context.py \
+            tests/test_family_cloud_row_event_count.py \
             tests/test_schema_v15_migration.py \
             tests/test_session_cost_intel.py \
             tests/test_session_intent.py \
--- a/clawmetry/sync.py
+++ b/clawmetry/sync.py
@@ -16353,6 +16353,18 @@ def sync_family_runtimes(config: dict, state: dict, paths: dict) -> int:
                     "total_tokens": int(s.total_tokens or 0),
                     "cost_usd": s.cost_usd,
                     "message_count": int(s.message_count or 0),
+                    # The hosted session page renders "Messages" from
+                    # sessions.event_count, which is the ONLY count the cloud
+                    # stores (its sessions table has no message_count column).
+                    # The OpenClaw row has always sent it; this one never did,
+                    # so every Codex / Cursor / Claude Code session read
+                    # "Messages 0" beside a full transcript (founder report
+                    # 2026-09-11: a Codex session showing 0 against 371
+                    # messages locally). len(_events) is the same raw-row count
+                    # the OpenClaw path sends, and the cloud upsert keeps the
+                    # GREATEST of old and new, so a read capped by
+                    # _family_event_read_cap() can never walk the number back.
+                    "event_count": len(_events),
                     "runtime": runtime,
                     "model": s.model or "",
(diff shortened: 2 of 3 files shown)


vivekchand:
### 🤖 Automated PR Health Check — 2026-09-12

**PR:** `fix/family-cloud-event-count` — add `event_count` to family runtime cloud sync row

---

#### What changed
| File | Change |
|------|--------|
| `clawmetry/sync.py` | Adds `"event_count": len(_events)` to the cloud row built for non-OpenClaw family runtimes so the hosted dashboard can show a message count instead of "Messages 0" |
| `tests/test_family_cloud_row_event_count.py` | New test: mocks `AgentAdapter` with 4 events, asserts `event_count == 4` and `message_count == 2` in the pushed payload |
| `.github/workflows/ci.yml` | Adds the new test file to CI's explicit file list so it actually runs |

---

#### Smoke test plan

```bash
# 1. Run the new unit test in isolation
python -m pytest tests/test_family_cloud_row_event_count.py -v

# 2. Run the full sync test suite to catch regressions
python -m pytest tests/ -k "sync" -v

# 3. Lint / drift guards
make lint

# 4. Manual smoke (if you have a local Claude Code session)
clawmetry &
# open http://127.0.0.1:8900, navigate to a Claude Code session row
# verify "Messages N" is non-zero and matches the transcript length
```

#### Key things to verify
- `event_count` is populated for **every** family runtime adapter (Claude Code, Codex, Cursor, …), not just the one in the test mock
- `message_count` is still present alongside `event_count` — both fields should coexist in the cloud row
- No regression in the OpenClaw path (which already had `event_count`); the diff only touches the `else` branch in `sync_family_runtimes`

#### Likely failure modes
- If the adapter returns an empty list early (e.g., no new events), `len(_events)` is `0` — that's correct, not a bug; the dashboard should show "0" rather than the old `None`
- CI will fail if `tests/test_family_cloud_row_event_count.py` is not reachable from the explicit file list; the `ci.yml` hunk addresses this, but double-check the indentation/placement in the matrix

---

PR #5899, opened 2026-09-12
Harden CI: the Playwright E2E jobs install from a hash-pinned set

vivekchand:
**Product record:** No-PRD: CI-only change; every path is under `.github/` (exempt per FLYWHEEL.md / `scripts/check_product_record.py`, which reports "product-record gate OK - no product-visible code changed").

**Risk:** Contained to two CI jobs, neither on the release path, and both already `continue-on-error: true`. If a hash or a resolved version is wrong the affected job fails loudly at the install step — it cannot half-install or silently pick something else, which is the property `--require-hashes` is bought for. Undone by reverting this commit; the previous loose lines return verbatim. Nothing in flight collides: #5897 and #5889 add different files to the same directory and touch different workflows; #5886 touches `pr-screenshots.yml`, which this PR does not.

---

## Summary

- `e2e-nightly` and `quarantine-sweep` each installed their Python dependencies from a loose `pip install a b c` line, so every run resolved whatever the index served that minute with no integrity check on any artifact. Both also install a browser driver (`playwright`), which then downloads a browser build — so the unpinned resolve reached further than the wheels did.
- Both now install with `pip install --require-hashes -r .github/requirements/playwright-e2e.txt`, which refuses anything not listed there with a matching hash.
- `.github/requirements/playwright-e2e.txt` carries the full resolved closure (30 packages) and, for each, the sha256 of **every** distribution PyPI publishes for that version — so the pin does not quietly depend on the runner resolving to the same wheel the file was generated against. Same shape and generation recipe as `dashboard-boot.txt` (#5897) and the supply-chain sets (#5889).
- It also settles a divergence the two lines had accumulated: `quarantine-sweep` installed `duckdb`, `waitress` and `cryptography` and `e2e-nightly` did not, for no reason either of them recorded. The pinned set is their union — the dashboard's own runtime closure plus the test runner — so both jobs now install the same thing instead of two drifting subsets.

`browserstack.yml` is the third job in this family and is **deliberately** left alone. `browserstack-sdk` resolves a 59-package closure (grpcio, protobuf, robotframework, selenium, trio) that neither of these jobs installs today, and folding it into a shared file would quietly add all of it to both. It gets its own set, separately.

The file is CPython 3.11 only by construction — `cffi` and `cryptography` are version-ceilinged by the interpreter (CLAUDE.md "Dependencies") — and the header says so, so it is not reused for a 3.9 job without regenerating. Both jobs pin `python-version: "3.11"` with setup-python.

No workflow's `permissions:` block is touched.

### Merge-order note

The updater for this directory is the `/.github/requirements` pip entry added by #5889, whose group pattern is `*` and so covers this file too. If this PR lands first, these pins do not move until #5889 merges. The file header states this plainly rather than claiming an updater that is not there yet.

### A note on the pytest-jobs guard

`test_pytest_jobs_install_what_conftest_needs` skips any job installing via `-r <file>`, so moving these two jobs onto a requirements file takes them out of that check. Verified by hand instead of assumed: `tests/conftest.py`'s only third-party module-scope import is `requests`, and it imports fine from the pinned venv.

## Test plan

- [x] The set installs clean under `pip install --require-hashes` in a fresh CPython 3.11 venv — exit 0, 30 packages, no hash mismatch
- [x] `flask`, `pytest`, `requests`, `duckdb`, `waitress`, `cryptography`, `playwright` and `pytest_playwright` all import in that venv
- [x] `tests/conftest.py`'s module-scope third-party import (`requests`) resolves there, so pytest will collect rather than exit 4
- [x] All 40 workflow files parse: `python3 -c "import yaml,glob; [yaml.safe_load(open(f)) for f in glob.glob('.github/workflows/*.yml')]"`
- [x] Both folded `run:` blocks render to the intended single command (checked against the parsed YAML, not the source text)
- [x] `tests/test_workflow_yaml_valid.py` — 570 passed, 362 skipped
- [x] `tests/test_e2e_nightly_workflow.py` + `tests/test_ci_workflow_invocations_are_real.py` — 10 passed
- [x] `scripts/check_product_record.py` — "product-record gate OK - no product-visible code changed"

---

https://claude.ai/code/session_01Xmvka9Ce9w1rzULBs7PgVA

---
_Generated by [Claude Code](https://claude.ai/code/session_01Xmvka9Ce9w1rzULBs7PgVA)_


vivekchand:
`E2E Gate (required)` failed on `c23ad57`, and it is not this PR's failure — it is the gate's 3600s timeout expiring while two legs were still sitting in the runner queue, not any test going red.

The gate ran 09:51:42 → 10:51:59 and gave up naming two legs as never-reported:

```
  - MOAT Verifier: 0/1 leg(s) complete
```

Both were simply queued behind a saturated runner pool:

| Leg | Queued until | Result |
|---|---|---|
| MOAT Verifier (72 tests) | started 10:48:21 | **success** at 10:52:21 — 22s *after* the gate gave up |
| Entitlement API tests | started 10:42:09 | still running |

Every other leg on this head reported green: API Tests (3 OS), pip install matrix (x4), MOAT Keystone, Store invariants, E2E Browser Tests, Wheel install & assets, OSS golden path, Cross-repo handoff, Syntax & Lint, Security audit, Python dependency audit, Fuzz, Action references resolve. Zero test failures anywhere on `c23ad57`.

This PR's diff is three files, all under `.github/` — a new hash-pinned requirements set plus the two workflow lines that consume it. None of it touches the entitlement suite, the MOAT verifier, or runner scheduling, and neither stuck leg had begun executing when the gate timed out.

So this is the one case that warrants a re-run rather than a code change: the gate's own log says a leg stuck at "not reported" means the workflow did not run for the commit, and here it did run — just an hour late. Re-running `E2E Gate (required)` once now that MOAT Verifier has reported and Entitlement API tests is in flight. If the re-run goes red on an actual leg, that is a real signal and I will treat it as this PR's to fix.

---

PR #5902, opened 2026-09-12
Harden CI: the PyPI publish path installs from a hash-pinned set

vivekchand:
`publish.yml` and `release-on-merge.yml` each installed their build toolchain with a bare `pip install build twine`, so every release resolved whatever the index served that minute.

This is the install that matters most in the repository. `twine` is the process that holds the PyPI upload credential, and `build` produces the artifact that is uploaded, provenanced, and then installed by everyone who runs `pip install clawmetry`. Every other unpinned install in CI can, at worst, break or mislead a build; these two decide what the world receives under our name.

It also ran on a path **no pull request exercises** — neither workflow triggers on `pull_request` — so a bad resolution here surfaces at release time, not in review. That is why this change was verified by hand rather than by waiting on a check (see Verification).

## What changed

- **Added** `.github/requirements/release-publish.txt` — the full 32-package closure for `build twine`, hash-pinned.
- **Changed** the `Install build tools` step in `publish.yml` and `release-on-merge.yml` to:
  ```yaml
  run: >-
    pip install --require-hashes
    -r .github/requirements/release-publish.txt
  ```

`--require-hashes` refuses to install anything not listed in the file with a matching hash, so the job gets exactly the artifacts named there or it fails. Every transitive dependency is present because `--require-hashes` demands the whole closure — `keyring` and its `SecretStorage` / `jeepney` / `cryptography` arm are twine's own Linux credential-store dependencies, not additions. Each version lists every sha256 PyPI publishes for it (per-platform wheels plus the sdist), so the pin does not quietly depend on the runner resolving to the same wheel the file was generated against.

This follows the convention already established for this directory by #5889, #5897 and #5899.

## Verification

Run on Linux / CPython 3.11, matching what both jobs pin with `setup-python` on `ubuntu-latest`:

- `pip install --require-hashes -r .github/requirements/release-publish.txt` into a clean venv succeeds, installing all 32 packages.
- `build --version` → `build 1.6.1`; `twine --version` → `twine version 7.0.0`.
- All 40 workflow files parse (`yaml.safe_load` over `.github/workflows/*.yml`), and both edited `run:` folded scalars resolve to the intended single-line command.

## Permissions

No `permissions:` block was touched. Both workflows keep the least-privilege top level (`contents: read`) and the per-job write scopes they already declared (`contents: write`, `id-token: write`, `attestations: write`, plus `actions: write` on release-on-merge). Provenance minting and the release upload are unaffected.

## Updater

The `/.github/requirements` pip entry in `.github/dependabot.yml` owns the routine bumps and rewrites hashes along with versions. That entry arrives with #5889, which introduced this directory; if this PR lands first, the gap is that one interval.

No-PRD: CI-only change, entirely under `.github/` (already exempt from the product-record gate).

https://claude.ai/code/session_01NNYihCXend2L1mA1npweTK

---
_Generated by [Claude Code](https://claude.ai/code/session_01NNYihCXend2L1mA1npweTK)_


vivekchand:
## Test plan & review notes

**What changed**
- The PyPI publish workflow installs dependencies from a hash-pinned requirements file rather than floating pins — supply-chain hardening so a compromised upstream package can't silently land in a release build

**Smoke commands**
- `make lint` (specifically `scripts/check_product_record.py` and syntax) — pure CI workflow change, no runtime code touched
- Manually inspect the pinned hash file to confirm it matches current pip-compile output: `pip-compile --generate-hashes <source>` and diff against the committed file
- Trigger a dry-run release workflow in CI (if the repo supports it) to confirm the install step succeeds with the pins

- A pinned hash going stale when an upstream package releases a patch — the publish step would fail with `HASH MISMATCH` until the pin file is regenerated
- Platform-specific wheels having different hashes (the pin file should include all platform variants or use `--require-hashes` with `--only-binary :all:`)

No issue link found in the PR description. This looks self-contained.

---

PR #5904, opened 2026-09-12
Activity heatmap: SQL buckets, node-local hours, runtime scoping

vivekchand:
## The bug

The Cost tab's Activity Heatmap is an empty grid on the hosted dashboard — every user, every runtime, 7d and 30d. Reported with screenshots today for both Claude Code (2237 sessions) and Codex (25 sessions) on `app.clawmetry.com/cloud/node/<node>`.

It is not a data problem. `clawmetry-cloud`'s `/api/heatmap` handler builds a 7x24 block of zeros and returns it:

```python
# Events table read removed (epic #1032 — events reads dead). The heatmap
# grid is now empty until the heartbeat-piggyback path supplies bucketed
# activity counts.
```

Nothing ever supplied them. The cloud cannot: it stores an E2E-encrypted blob and never sees an event. The daemon is the only side that can bucket this, so it ships the finished grid in the snapshot and the cloud decrypts and draws it — the same shape as `cm-cloud-activity`, `cm-cloud-outcomes` and the rest. Cloud stays blind.

Locally the endpoint works (verified: `max 1866`, `_source local_store`), so this PR is the daemon half plus three defects the local path had anyway.

## What changed

* **`LocalStore.activity_heatmap(days, runtime)`** — bucketing is one SQL `GROUP BY`. It used to pull up to 50k raw event rows through the daemon proxy and loop over them in Python: tens of MB marshalled per render, and a cap that silently emptied the oldest days of a busy node's 30-day window.
* **Node-local hours** — `cost_windows.hour_expr_sql` / `local_hour`, the hour twins of ADR-046's day bucket. The old path stripped each row's offset, so a UTC-writing runtime's afternoon was drawn in the small hours. Test pins a UTC-stamped and a local-stamped event at the same instant landing in the same cell.
* **`?runtime=`** — scopes by session-id prefix; an unknown runtime returns zero, never the node-wide total (FLYWHEEL 1c). `app.js` sends the active runtime. The legacy file scan reads OpenClaw logs only, so a runtime-scoped request that reaches it returns the empty grid *in shape* rather than OpenClaw's hours under a Codex heading.
* **`activity_heatmap_by_runtime`** — every runtime plus node-wide in ONE scan, so the new `activityHeatmap` snapshot slice costs one query per sync cycle, not N+1.

## Verification

* `tests/test_activity_heatmap_runtime_scope.py` — 5 cases, all green; registered in `ci.yml` (this repo runs file lists).
* `tests/test_duckdb_coverage_batch_2026_05_13.py -k heatmap` still green.
* `make lint-py39 lint-daemon-allowlist lint-module-map lint-ci-test-coverage` green; no new `ruff` findings in any touched file (per-file counts identical to `origin/main`).
* Split and single-runtime queries reconcile by assertion, so per-runtime grids sum to the node-wide one by construction.

Cloud half (the `cm-cloud-heatmap` interceptor) goes up next and needs this version pinned.

No-PRD: user-reported defect — a card that has rendered blank for every hosted user since the events read was removed.

https://claude.ai/code/session_01JSWiABbwBYWBdmdupVheAL

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -217,14 +217,6 @@ jobs:
         run: python3 -m pytest tests/test_providers_pricing.py tests/test_pricing_accuracy.py -q
 
       # Named explicitly because this repo's CI runs FILE LISTS, not
-      # `pytest tests/`. The heatmap buckets on the node-local clock and
-      # scopes by runtime; both are the kind of thing that regresses quietly
-      # (the hosted card drew an all-zero grid for months before anyone
-      # noticed it was not simply a quiet week).
-      - name: Activity heatmap buckets local hours and scopes by runtime
-        run: python3 -m pytest tests/test_activity_heatmap_runtime_scope.py -q
-
-      # Named explicitly because this repo's CI runs FILE LISTS, not
       # `pytest tests/`: a guard added without a line here runs in no job at
       # all. Covers the half `gen_module_map.py --check` cannot: every
       # blueprint dashboard.py registers is in the map, and CLAUDE.md /
@@ -780,6 +772,7 @@ jobs:
             tests/test_scan_read_cache.py \
             tests/test_cache_push_gating.py \
             tests/test_usage_cache_trends_local_store.py \
+            tests/test_activity_heatmap_runtime_scope.py \
             tests/test_transcript_local_store.py \
             tests/test_snapshot_transcripts_runtime_fair.py \
             tests/test_moat_perf_benchmark.py \
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -217,6 +217,14 @@ jobs:
         run: python3 -m pytest tests/test_providers_pricing.py tests/test_pricing_accuracy.py -q
 
       # Named explicitly because this repo's CI runs FILE LISTS, not
+      # `pytest tests/`. The heatmap buckets on the node-local clock and
+      # scopes by runtime; both are the kind of thing that regresses quietly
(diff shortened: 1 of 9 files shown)


vivekchand:
> these methods are not registered in QUERY_CONTRACT, so they will not appear in the _SHAPES allowlist that routes/local_query.py requires. The dispatch will fail with "unknown shape" error.

Checked this before pushing — it does not apply here. There are two separate surfaces in `routes/local_query.py`:

* **`/api/local/query?shape=…`** — the public q/1 contract. `_SHAPES` is derived from `clawmetry/query_contract.py`, and `tests/test_query_contract_drift.py` fails CI if a shape ships without a registry entry. ~10 entries.
* **`/__local_query__/<method>`** — the internal daemon-to-dashboard RPC that every `_try_local_store_*` fast path uses. It is gated by `_DAEMON_METHODS` (184 entries, none of which are q/1 shapes) and never consults `_SHAPES`.

`local_store_via_daemon("activity_heatmap", …)` posts to the second one, and `activity_heatmap` is in `_DAEMON_METHODS` in this diff. `make lint-daemon-allowlist` — the gate that exists precisely for this mistake — passes: *66 distinct query_* methods called from routes/, all in _DAEMON_METHODS*. `tests/test_query_contract_drift.py` also passes unchanged, since no public shape was added.

No new public query shape is intended: the cloud reads this from the encrypted snapshot slice, not through the relay. I have added a comment at the call site spelling out which allowlist applies, since this is evidently easy to read the other way.

PR #5925, opened 2026-09-12
Check the linked account before email sign-in changes it

vivekchand:
The email sign-in introduced in #5918 checks the machine's linked account after successful cloud verification. Pairing must wait for that decision: a refused sign-in otherwise changes the account against which the next attempt is checked.

Evaluate dashboard access before pairing, refuse an unrecognized account without changing local identity, and prevent caching the credential response. The existing owner's recovery path remains available.

Verification: the new regression fails against #5918 and passes with the correction. It exercises repeated sign-in through the real before-request auth hook, for both managed and self-hosted modes. The focused auth, OTP and OAuth suites passed 70 tests before adding the second mode; both modes are included in CI along with the original sign-in regression.

No-PRD: incident follow-up to #5918, preserving the existing account ownership boundary.

Authored by Codex.


vivekchand:
✨ auto-fixed: merged main into branch to resolve stale base and trigger fresh CI

---

Issue #5949, opened 2026-09-13
OTLP intake: durable acknowledgements, recovery and unsampled usage accounting


vivekchand:
## Problem and scope

The existing OTLP receiver, session materialization and DuckDB store should be reused. The source review found best-effort span persistence, so HTTP success alone does not establish that accounting records survived. Remote producers also need explicit retry/partial-success behavior. This issue owns durability and receipt accounting, separate from vivekchand/clawmetry#5938 (span normalization/redaction), vivekchand/clawmetry#5939 (framework recipes), vivekchand/clawmetry#5940 (LiteLLM) and vivekchand/clawmetry#5679 (scoped authentication, already covered by PR vivekchand/clawmetry#5684 stacked on vivekchand/clawmetry#5676).

## Exact changes

- Define and implement when each OTLP signal is acknowledged: durable write or durable bounded queue. Never return an unqualified success after silently losing accepted records; follow OTLP partial-success/retry semantics and document rejected counts.
- Expose accepted, persisted, rejected, queued, dropped and deduplicated counts, oldest queued age and last successful intake. Distinguish unknown usage from zero usage.
- Use stable source/request/span identities for idempotent delivery. A retried transport delivery is one record; a second billable provider attempt remains a separate charge.
- Preserve observed and ingestion times for late/out-of-order records and session materialization. Document the session identity/fallback when the exporter lacks a conversation ID, with migration/cardinality implications.
- Financial totals use unsampled accounting records. Trace sampling must not silently reduce reported spend; show completeness and missing-source status.
- Separate observed-agent spending limits from telemetry intake limits. Inspect the current `_budget_paused` intake behavior; a budget incident must not silently hide later billable activity. Any required rate limiting has bounded buffering, visible loss and retry guidance.
- Keep writes under the daemon-owned DuckDB writer path, request reads on the existing query contract, and dependencies minimal. No parallel raw-file accounting store.

## Acceptance

- [ ] JSON/protobuf and gzip paths have explicit success, partial rejection, malformed input, overload and authentication behavior.
- [ ] A numbered real acceptance workload reconciles source IDs to persistent rows and the rendered UI after daemon/receiver restart, transient writer failure and a one-hour disconnection; no unexplained loss or duplicate charges.
- [ ] Tests exercise late delivery, repeated delivery, distinct charged retries, missing usage and sampled spans paired with unsampled usage records.
- [ ] Intake diagnostics make deliberate drops visible without leaking prompt content or credentials.
[...]

Issue #5942, opened 2026-09-13
Fleet install: Windows service, Linux linger and silent install for VDI / Azure Virtual Desktop


vivekchand:
## Problem
Many developers work on virtual desktops (Azure Virtual Desktop, multi-session Windows) rather than laptops.
- Windows: the daemon is a Task Scheduler **LogonTrigger** task (`clawmetry/daemon_registration.py:207`). It only runs while a user is logged on and isn't a service.
- Linux: non-root installs get a `systemd --user` unit (`cli.py:1957-1990`), which stops at logout unless linger is on.
- Already working: `HTTPS_PROXY` (`net.py:25`), custom CA (`CLAWMETRY_CA_BUNDLE`, `net.py:89`), offline/local-only, and non-interactive `connect --key` (`cli.py:1059`). There are no MDM/Intune/Ansible recipes.

## Done when
- [ ] `clawmetry install --service` registers a Windows service (per-machine, all sessions) with a documented uninstall
- [ ] Linux user install enables linger or offers a system unit
- [ ] A silent install one-liner plus Intune (PowerShell) and Ansible examples with an enrollment key
- [ ] Verified on an Azure Virtual Desktop multi-session host: two users' sessions attributed separately; daemon survives logoff

## Enterprise planning refinement, 13 September 2026

The specific requirements below refine the original scope and supersede conflicting implementation assumptions. Existing findings and discussion remain available.

## Deployment safety and scope refinement

First identify the agreed VDI profile: Windows/Linux, persistent or pooled, single-user or multi-session, profile roaming/FSLogix, WSL/container location and allowed privileges. Running a privileged all-user service is a design option to validate, not the acceptance goal itself. Compare a per-user collector with an explicitly authorized service-to-user mapping; collect where the runtime actually runs.

- [ ] Two users cannot read each other's session content or acquire the other's identity. Pooled image clones receive fresh node identity; returning user profiles retain deliberate attribution rather than becoming a new billable user each login.
- [ ] Reboot, logoff, profile roaming, logout cleanup and a one-hour network outage preserve the numbered acceptance workload or show explicit data loss. Buffering is capped and recoverable.
- [ ] Proxy/custom CA and existing local-only capabilities are reused and tested, not rebuilt. Offline mode has a measured egress contract.
- [ ] Enrollment credentials are scoped, protected and rotatable; no reusable credential is baked into a shared image or printed in logs. Document pin/update/rollback/uninstall.
[...]

Issue #5941, opened 2026-09-13
Attribution: project + user tags and per-project budgets (burn vs budget)


vivekchand:
**Priority:** P0 for team use — "what did this project spend on tokens, against its budget?"

## Problem
- "By team" usage (`routes/usage.py:3713`) groups by a per-runtime `team_mapping` and falls back to the runtime name (`local_store.py:12306`). It isn't per project or per user.
- Per-user/org fields exist only for OTLP (`local_store.py:1887`).
- There is a single global budget (`routes/alerts.py:818`).

## Proposal (v1)
- A project tag on every session, resolved in order: `CLAWMETRY_PROJECT` env → git remote → cwd root. User = OS user (or OTLP `user.email`).
- Budgets per project: an amount and period, with alerts at 50/80/100% and a daily burn chart.
- CSV export per project and per user (extend `/api/usage/export`, `routes/usage.py:3112`).

## Done when
- [ ] Two repos on one machine show as two projects with separate spend
- [ ] A budget alert fires in a test at 80%
- [ ] Works for Claude Code, Codex and Copilot sessions

## Enterprise planning refinement, 13 September 2026

The specific requirements below refine the original scope and supersede conflicting implementation assumptions. Existing findings and discussion remain available.

## Identity and budget refinement

Inspect the shipped principal/owner model and open PR vivekchand/clawmetry#5204 before changing metadata. That PR extends derived principal inheritance; do not create a competing user/team identity schema or restamp history on every label edit.

Separate organization boundary, human/service principal, project/client engagement and execution/session. OS usernames and OTel email tags are hints, not authoritative tenant membership. Environment/git/cwd-derived project suggestions need source/confidence and a visible unassigned fallback; avoid publishing raw local paths or emails in shared metrics.

- [ ] Authorized mapping supports historical periods and corrections with provenance. Two repos sharing a remote/path naming convention do not silently collapse into one client project.
- [ ] Each budget declares period/timezone/currency and financial basis. Alerts are distinct from routed enforcement; observing spend cannot retroactively prevent a charge.
- [ ] Exports and downstream reports include assigned/unassigned/unpriced amounts and completeness; runtime/gateway overlap follows vivekchand/clawmetry#5940 rather than double counting.
- [ ] Test two projects, two users on one host, a service principal, a forged tenant/project claim, a membership change, late events and a period boundary.
[...]

PR #5950, opened 2026-09-13
Fleet install for shared hosts and virtual desktops: per-user collector, Linux linger, Windows all-users logon task, Intune + Ansible recipes

vivekchand:
Refs vivekchand/clawmetry#5942

Product record: https://factory.8090.ai/project/b415065f-ab2f-4f53-8864-0c009fd098cb/requirements/d673cdaa-3bdb-4b10-8a2d-d40ccb5ae630
Design: https://factory.8090.ai/project/b415065f-ab2f-4f53-8864-0c009fd098cb/blueprints/ae944090-9c2b-4549-8ec2-f858736a747b (new "Fleet install on shared and virtual desktops" section)

## Review fixes (head 63e602a78b, rebased on main)

**Blocking: no unattended self-update from an administrator-owned install.** Both recipes install where only an administrator can write (`C:\Program Files\ClawMetry\venv`, `/opt/clawmetry-fleet`), and auto-update is on by default. On Windows the update plan is exit-and-respawn. Every signed-in user's collector exited, the helper ran pip three times against a directory it cannot write, then relaunched. The in-memory backoff died with the process, so the cycle repeated every few minutes.
- `routes/update_check._maybe_auto_update` now skips when the running **virtual environment's** site-packages is not writable by this user. The check is probed once per process with a real file create, because `os.access` ignores ACLs on Windows. A non-venv interpreter is unchanged, because pip falls back to a per-user install there. Nothing is set in the recipes; the collector detects it.
- `service status` reports `auto_update: {state: off, basis: install_not_writable}` with how to update (re-run the recipe with a new pin). The README "Pin, update, rollback" section and the Intune script header now say exactly that.
- New criterion **AC-OBS-FLEET-001.4** added to the requirement and mirrored; baseline tightened (117/187 after rebasing onto main).
- Guard: `test_auto_update_never_attempted_from_unwritable_fleet_install` runs the Windows `respawn` plan three times and asserts no pip, no exit and no respawn, plus a control run showing the same harness does respawn on a writable install. `test_self_update_blocked_only_for_unwritable_isolated_env` uses a real 0500 directory.
- Revert proof (local, restored afterwards):
  - removing the skip fails the first test (1 failed)
  - forcing the probe to "writable" fails the second (1 failed)
  - restored: 25 passed
- Real runners: the Ubuntu job asserts that each user's status against the root-owned `/opt` venv reports `install_not_writable`. The Windows job puts a deny ACE for the runner account on a directory and asserts the probe reports it unwritable, and a writable directory not blocked.

**Found by the real runner: the shared environment was writable by every desktop user.** The first Ubuntu run with the new assertion showed the Ansible-created `/opt/clawmetry-fleet/lib/python3.12/site-packages` at mode `0o40777`: root ran under a 0000 umask, and user `cmalice` could create a file there. Any user could have planted code that every other user's collector imports. The self-update probe was right to call it writable, and this is worse than the review item.
- Ansible: venv creation and pip run under `umask 022`, then the playbook enforces `owner: root` and `mode: u=rwX,go=rX` recursively with `follow: false`. CI fails on any group-writable, world-writable or non-root path under the environment.
- Intune: explicit ACL on the install directory (SYSTEM and Administrators full control, Users read and execute, inheritance removed), so a custom `-InstallDir` cannot inherit "Authenticated Users: Modify". CI fails if Users, Authenticated Users, Everyone or Interactive holds a write-type right.
- Guard: `test_recipes_keep_the_shared_environment_unwritable_by_desktop_users`. Proof that the checks catch the bug: run 34793069392 on 875adc2138, before the recipe fix, printed `mode 0o40777 ... CREATE SUCCEEDED` for `cmalice`.
- Another session pushed a `chmod -R go-w` task for the same finding (3177291290). It is folded into the above; see the PR comment.

**Found by the Windows runner: the first write probe could hang.** Under the deny-ACE check the step ran 24 minutes, until the job timeout. CPython's `tempfile.mkstemp` retries a `PermissionError` on Windows whenever `os.access`, which ignores ACLs, says the directory is writable, up to `TMP_MAX` names. A collector's first update check would have hung the same way. The probe now makes one exclusive `os.open` of a random name and treats a denied create as not writable. `test_write_probe_is_one_attempt_never_a_mkstemp_retry_loop` asserts one attempt and no `mkstemp`, and it fails against the previous implementation. The CI step has a 3-minute timeout so a regression fails fast.

**Non-blocking items also addressed:**
- A Windows data directory is reported `private: null` (`windows_profile_acl_not_checked`) instead of a claimed yes, and the Windows job asserts that.
- Linux enables linger **before** registering the unit, with `loginctl --no-ask-password`.
- CLI install text uses plain words instead of state codes.
- The Intune script refuses a `-PythonExe` that lives in a user profile.
- The README documents the machine-wide interpreter requirement and the possible console window at sign-in. The console window is unverified; there is no multi-session host.

Left as is: the respawn failure backoff is still in-memory for other failure causes (pre-existing, not fleet-specific).

## The VDI profile this implements

One collector **per user, running as that user**, registered per machine where the OS allows it. The issue refinement asked us to validate a privileged all-user service before assuming it. We rejected it for this increment: the collector reads agent files in each user's profile and writes a store holding their session content, so one SYSTEM service would be able to read every user's content and would attribute everyone to one node.

| | Linux | Windows |
|---|---|---|
| Registration | per-user systemd unit + logind linger | one logon task for BUILTIN\Users (`--all-users`, elevated) |
| Runs as | the user | each signed-in user, least privilege, instances policy Parallel |
| After logoff | keeps running (linger) | stops with the session, resumes at next sign-in; `status` says so |
| Data | `~/.clawmetry` restricted to 0700 by fleet install | profile ACL (reported as not checked) |
| Updates | re-run the recipe with a new pin; no self-update from `/opt` | re-run the recipe with a new pin; no self-update from Program Files |

## What changed
- `clawmetry service install | status | uninstall` (`clawmetry/fleet_install.py`, wired in `clawmetry/cli.py` including the `_subcmds` routing list, `--json` for scripts). Exit 0 only when the requested state was actually reached. Linux linger refused: state `denied` plus `sudo loginctl enable-linger <user>`, exit 1. Windows not elevated: `needs_admin`.
- `clawmetry uninstall` on Windows now also removes the per-user and all-users scheduled tasks (before, the per-user task survived uninstall).
- `deploy/fleet/intune/Install-ClawMetry.ps1`, `deploy/fleet/ansible/clawmetry.yml`, `deploy/fleet/README.md` (pin, update, rollback, uninstall, proxy/CA/offline reuse, enrollment guidance). No credential in either recipe.
- No new HTTP route, so no cloud route-policy PR is needed.

### Unit tests
- `tests/test_fleet_install.py` (27 tests after the review fixes, in the ci.yml unit list).
- Revert proof, run locally by mutating `fleet_install.py` and restoring it:
  - task principal SYSTEM instead of BUILTIN\Users: 1 failed
  - no chmod of the data dir: 1 failed
  - linger reported enabled without checking logind: 2 failed
  - instances policy IgnoreNew: 1 failed
  - restored: 20 passed
- Also green locally: `tests/test_daemon_registration.py`, `tests/test_uninstall_windows.py`, `tests/test_cli_help_no_dashboard_import.py`.
- CLI smoke in a scratch venv with a scratch HOME on macOS: `service status` (text and JSON), `service install --all-users` (unsupported, exit 1), `service --help`.

### Real runners (`.github/workflows/fleet-install-test.yml`; first green on b00080b718, rerun on 63e602a78b with the review assertions)

**Ubuntu, Ansible playbook, users cmalice and cmbob:**
- Both users show linger `enabled` and `survives_logoff: true`, with no login session for either user.
- Each collector's pid is owned by its own user.
- Neither user can list or read the other's `~/.clawmetry`. Homes were deliberately left 0755, so this tests the directory mode, not the distro default.
- Uninstall removes the unit.
- Idle overhead, runner reference only (not a VDI image, no agent workload). CPU is a lifetime average over about 65 s, so it includes startup:

  | user | RSS | CPU | data dir |
  |---|---|---|---|
  | cmalice | 137 MB | 2.3% | 152K |
  | cmbob | 143 MB | 2.4% | 152K |

- Not exercised on the runner: the linger `denied` path. The runner's polkit let a user with no session enable her own linger, and the job asserts only that the reported state matches logind, which it did. The denied path and its admin command are covered by unit tests.

**Windows, Intune script, elevated.** The exported task XML contains:
- `<GroupId>S-1-5-32-545</GroupId>`, no `UserId`
- `<MultipleInstancesPolicy>Parallel</MultipleInstancesPolicy>`
- `<LogonTrigger />` with no user
- the venv `python.exe -m clawmetry.sync` action

`RunLevel` is omitted because LeastPrivilege is the schema default, and the check fails on any explicit other value. `status` reports `supervisor: task_all_users` and `survives_logoff: false`, and uninstall removes the task.

- **CodeQL:** no new alerts. The earlier chmod findings were in test fixtures; they now create the directory under a 022 umask instead of calling chmod.
- **AC mirror:** `AC-OBS-FLEET-001.1` through `003.2` added to `docs/acceptance_criteria.json`, each declared by a test; baseline tightened (108/178).

**No real Azure Virtual Desktop or multi-session host was available.** Everything above ran on GitHub-hosted runners. On Windows, CI cannot sign in two interactive users, so one instance per signed-in user follows from the task definition and was not observed.

## Remaining on #5942 (not in this PR)
- Observed on a real AVD multi-session host: two users' sessions attributed separately, plus FSLogix/profile roaming and logout cleanup.
- Fresh node identity for pooled image clones, and deliberate attribution for returning users (identity work, #5941).
- Capped, recoverable buffering through a one-hour outage, and receipts (#5949).
- Scoped, rotatable enrollment keys distinct from the account key (cloud).
- Overhead measured on the chosen image under a real workload. CI only records an idle runner reference point.
- Windows process and pre-tool control capability states verified separately on a multi-session host.
- Keeping a Windows collector alive after logoff, or a service-to-user mapping, if that profile is chosen later.


vivekchand:
## Coordinator review

Reviewed head `b00080b718` against #5942 (including the 2026-09-13 refinement), the Factory requirement d673cdaa, and the code paths the fleet collector runs through. No code was changed.

**Verdict: fix before merge (1 blocking item).**

### Blocking

**1. On a Windows `--all-users` host, every user's collector keeps exiting and relaunching: auto-update tries to write into an admin-owned venv.**

I traced this in the code. It was not observed, and CI could not show it (see the last point).

- `Install-ClawMetry.ps1` installs into `C:\Program Files\ClawMetry\venv` as admin/SYSTEM. Each user's collector then runs from that venv with least privilege, so it cannot write to it.
- Auto-update is on by default for every role (`routes/update_check.py::_maybe_auto_update`). A pinned fleet is behind PyPI almost all the time, given how often releases ship.
- On Windows `_restart_plan()` always returns `"respawn"` (line 143). The collector **exits**. `clawmetry/update_respawn.py` then runs `pip install --upgrade` 3 times (waits of 20s, 60s and 120s), fails with access denied, and relaunches the collector.
- On this path the failure never reaches the backoff. `_failed_update_attempts` is in memory only, and `last_attempt_*` is written but never read to skip a target. The relaunched collector hits the same target at its next 60-second check.
- Result: each signed-in user's collector is down most of every ~5 minutes and runs pip 3 times per cycle, on a multi-session host with N users. Meanwhile `service status` still reports the task as registered.
- The README's "Pin" section says to set `CLAWMETRY_AUTO_UPDATE=0` "in the environment the collector runs with". The Windows task has no environment block, and neither recipe provides a way to set it.
- CI could not catch this: `_env_auto_update_disabled()` treats `CI=true` as disabled, so the green Windows job never exercised auto-update.

Fix, any one of these, with a regression test:
- The fleet install disables auto-update for collectors it registers. For example, the Intune script sets a machine-level `CLAWMETRY_AUTO_UPDATE=0`, and the Ansible playbook does the same through a systemd user drop-in or an `Environment=` line. Add a test asserting both recipes do it.
- Or `_maybe_auto_update` skips (and reports "managed install, not writable by this user") when the interpreter prefix is not writable, and saves the backoff to disk so a respawn cannot retry immediately.

Then make README "Pin, update, rollback" describe what the recipe actually does.

On Linux, a root-owned `/opt/clawmetry-fleet` fails `perform_self_update` with no fallback to a folder in the user's home. The 30-minute in-memory backoff bounds that, so it is not a loop, but the same fix should cover it.

### Non-blocking (worth doing, not merge gates)
- **Windows "Data private: yes" is assumed, not checked.** `data_dir_private()` returns `private: True` on Windows without looking at the ACL. The text output hides the `basis`. That reads against AC-OBS-FLEET-002.2 and the PR's own "never report coverage that is not there" rule, and redirected or FSLogix profiles can differ. Suggest `private: null` / "not checked (profile ACL)".
- **Likely a visible console window for every user at sign-in.** The all-users task runs `python.exe` (a console program) in each interactive session. Closing that window stops collection until restart-on-failure. The per-user task already behaves this way, but now it reaches every user on the host. It can't be seen on CI. Consider `pythonw.exe` or a hidden launcher, and check on a real host.
- **`-PythonExe` must be a machine-wide interpreter.** A Windows venv's `python.exe` forwards to the base interpreter named in `pyvenv.cfg`. If the admin's Python is a per-user install under AppData, every other user's collector fails to start. Document this or check for it in the script.
- **Linux install order.** `install()` registers the user unit *before* enabling linger. For a user whose user manager isn't running, the unit step fails first and a rerun is needed. Enabling linger first avoids that. Also consider `loginctl --no-ask-password` so a polkit TTY prompt cannot eat the 15s timeout.
- **Raw state codes.** CLI text output prints codes such as `needs_admin`, `denied` and `unavailable`. Each comes with a reason, but plain words would read better.

### Checked and fine
- **Product record:** the Factory requirement exists, was written first, and has problem, non-goals, risks and alternatives rejected. The PR body cites the full factory.8090.ai URL, and the product-record gate is green.
- **Regression guard:** `tests/test_fleet_install.py` (20 tests) is in the `ci.yml` unit list. Mutation proof is recorded in the PR. The AC baseline was tightened from 101/171 to 108/178, not loosened. No skip, xfail or continue-on-error was added.
- **Routes:** no new HTTP route (CLI and recipes only), so no cloud route-policy PR is needed.
- **Public-repo hygiene:** no customer, partner or deal names in the diff or commit messages.
- **Branch state:** 0 commits behind `main`. `mergeStateStatus` is CLEAN, with 46 checks passing and 1 skipped.
- **Per-user isolation:** the daemon's local query server uses a random loopback port with a bearer token and a 0600 discovery file, so two users on one host cannot query each other's store over loopback.
- **Acceptance criteria:** the original "`clawmetry install --service`" SYSTEM service was deliberately not built, and the refinement supports that. AVD validation, overhead under a real workload, identity for pooled clones (#5941), buffering (#5949) and scoped enrollment keys are honestly listed as remaining, so #5942 should stay open after merge.
- **No duplication:** nothing overlaps #5933 (Windows stall diagnostics), #5941 or #5949.

vivekchand:
**CI fix pushed** (commit `3177291`).

Root cause: `python -m venv /opt/clawmetry-fleet` runs as root under a 0000 umask on GitHub Actions Ubuntu runners, producing `mode 0o40777` (world-writable) on all venv directories including `site-packages`. The auto-update probe (`tempfile.mkstemp(dir=site_packages)`) correctly detected the directory as writable and left auto-update enabled -- so the CI assertion `state == "off"` failed.

Fix: one `chmod -R go-w {{ clawmetry_venv }}` task added to `deploy/fleet/ansible/clawmetry.yml` immediately after pip install. Site-packages becomes mode 755 (owner-write only), the probe returns `writable: false`, and the collector reports `auto_update: {state: off, basis: install_not_writable}` as AC-OBS-FLEET-001.4 requires. The test was correct -- it caught a real deployment bug.

---

vivekchand:
Branch note: I rebased this branch onto `main` and force-pushed with a lease pinned to 3177291290. That commit came from another session: a merge of `main` plus a `chmod -R go-w` task in the Ansible playbook. Its fix is folded in, not dropped. Head is now 030b82ed91.

- **Same finding.** The first real Ubuntu run showed `/opt/clawmetry-fleet/lib/python3.12/site-packages` at mode `0o40777`, and an unprivileged user could create a file there. Any desktop user could have planted code that every other user's collector imports, so the self-update probe was right to call it writable.
- **Linux fix (ceb7355ab4):** the playbook creates the venv and runs pip under `umask 022`. It then enforces `owner: root`, `mode: u=rwX,go=rX` recursively with `follow: false`, so it never changes permissions through the venv's `python` symlink. CI fails on any group-writable, world-writable or non-root path under the environment.
- **Windows fix:** the Intune script sets an explicit ACL on the install directory: SYSTEM and Administrators full control, Users read and execute, inheritance removed. A custom `-InstallDir` such as `C:\ClawMetry` therefore cannot inherit "Authenticated Users: Modify". CI fails if Users, Authenticated Users, Everyone or Interactive holds a write-type right.
- `test_recipes_keep_the_shared_environment_unwritable_by_desktop_users` pins both recipes and both CI checks.

Please don't re-push the `chmod -R go-w` task; the playbook already covers it.

vivekchand:
Ready to merge. Head 63e602a78b: 47 checks pass, 1 skipped, 0 failing, including `E2E Gate (required)`, Drift Bot, CodeQL, the product-record gate and both Fleet install runner jobs. `gh pr view` reports MERGEABLE / CLEAN. Since the last rebase, `main` has gained one unrelated commit; it does not conflict, and the branch does not need to be up to date to merge. Not merged.

**Blocking review item fixed.** A collector running from an administrator-owned fleet install no longer tries to update itself. Before this, on Windows it exited every signed-in user's collector, pip failed three times, and the cycle repeated. The skip is `routes/update_check.py` plus `fleet_install.self_update_blocked`. `service status` reports `auto_update: off` / `install_not_writable` and says to update by re-running the recipe with a new pin. The README Pin section says the same. New criterion AC-OBS-FLEET-001.4 is in the Factory requirement and mirrored into `docs/acceptance_criteria.json`.

**Two more bugs the real runners found while verifying that fix, both fixed here:**
- **The Ansible venv at `/opt/clawmetry-fleet` was mode 0777.** Any desktop user could have written code that every other user's collector imports. The playbook now runs under `umask 022` and enforces root ownership with `u=rwX,go=rX`. The Intune script sets an explicit ACL on its install directory. Both CI jobs fail if a desktop user can write the environment. Run 34793069392 caught it: `mode 0o40777 ... CREATE SUCCEEDED`. Run 34796721610 shows it fixed: `mode 0o40755`, `PermissionError`, `install_not_writable`.
- **The first write probe could hang on Windows.** `tempfile.mkstemp` retries a denied create while `os.access`, which ignores ACLs, says the directory is writable; it ran for 24 minutes on the runner. The probe now makes a single `os.open` attempt. In run 34796721610 the deny-ACE step took under a second and printed `blocked: True ... os.access says writable: True`.

**Non-blocking review items also done:**
- A Windows data directory is reported as not checked rather than private.
- Linger is enabled before the unit is registered, with `--no-ask-password`.
- The CLI prints plain words instead of state codes.
- The Intune script refuses a Python installed inside a user profile.

**Merge after:** nothing. No dependency PR.

**Companion PRs:** none. There is no new HTTP route, so no `cloud_route_policy` entry is needed.

**Post-merge / post-release verification:**
1. After the next `[RELEASE]` publishes, check that `https://pypi.org/pypi/clawmetry/json` lists the new version, and that the wheel contains `clawmetry/fleet_install.py` and `deploy/fleet` is in the repo.
2. On a scratch Linux VM:
   - Run `ansible-playbook -i <vm>, deploy/fleet/ansible/clawmetry.yml -e '{"clawmetry_users": ["u1","u2"]}' -e clawmetry_package=clawmetry==<released>`.
   - Then `sudo find /opt/clawmetry-fleet ! -type l \( -perm -g+w -o -perm -o+w \)` must print nothing.
   - `sudo -u u1 /opt/clawmetry-fleet/bin/clawmetry service status --json` must report `survives_logoff: true` and `auto_update.basis: install_not_writable`.
   - After one release lands on PyPI, u1's collector must still be running with the same pid and no pip process run as u1.
3. On a scratch Windows host, elevated:
   - Run `Install-ClawMetry.ps1 -Package clawmetry==<released>`.
   - Sign in as a standard user and run `"C:\Program Files\ClawMetry\venv\Scripts\clawmetry.exe" service status`. It must show `Auto-update: off (this user cannot write ...)` and `Data private: not checked`.
   - `schtasks /query /tn ClawMetrySyncDaemonAllUsers /xml` must show GroupId `S-1-5-32-545` and `Parallel`.
   - Note whether a console window appears at sign-in. That is still unverified.
4. Uninstall on both hosts with `clawmetry service uninstall` and confirm the unit or task is gone.

**Still not verified (listed on #5942):** a real AVD or multi-session host with two interactive users, overhead under a real workload, identity for pooled clones (#5941), buffering (#5949) and scoped enrollment keys.

vivekchand:
**Automated CI triage — Syntax & Lint failure blocking all downstream jobs**

The **Syntax & Lint** job failed (job `103930979359`, run `34830041491`), causing all downstream CI jobs to be skipped. This is a real failure (not a queue-priority cancellation).

All other checks show as skipped/cancelled because they depend on Syntax & Lint passing. Once the lint issue is fixed and CI reruns, the full suite will execute.

Please check the Syntax & Lint job logs for the specific error.

---

vivekchand:
**Branch note: rebased onto `main` as one commit, head `77d3342f7d`** (force-push with lease pinned to `a2e227c1f9`).

Please do not merge `main` into this branch again; rebase instead.

- **Syntax & Lint was red on `872d553e22`:** the AC gate reported `AC-OBS-FLEET-003.2` as not in the manifest. The real cause: `docs/acceptance_criteria.json` did not parse. A merge resolution had dropped the `}, {` before `AC-OBS-FLEET-001.1`. The later merge `b9aca7b88d` also added all eight FLEET criteria a second time. Fixed both. `check_ac_coverage.py --check`: OK, 153/223, ratchet holding.
- **MODULE_MAP:** regenerated (257 modules). This is the same content as `a2e227c1f9`, which the rebase includes.
- **CHANGELOG:** this PR no longer touches `CHANGELOG.md`, so merges stop re-conflicting. The coordinator carries the entry in the `[RELEASE]` PR. The two merges had also left the entry in the file twice.
- **Verified locally after the rebase:**
  - `tests/test_fleet_install.py`: 27 passed.
  - Also run: module-map drift, workflow YAML and AC-coverage guard tests, `check_py39_annotations`, and the `sync_runtime_count` / `gen_module_map` / `gen_framework_coverage` / `check_ci_test_coverage` checks.
  - `tests/test_auto_update.py`: the same 9 tests fail on a clean `origin/main` checkout on this machine, so none of them comes from this PR. CI does not run those 9.

vivekchand:
⚠️ needs manual rebase — conflict in `clawmetry/cli.py`, `docs/MODULE_MAP.md`, `docs/acceptance_criteria.json`, `ci.yml` (and others). Cannot auto-resolve: requires code-level judgment to merge correctly.

---

vivekchand:
Blocked on required review — skipping (auto-mergeability sweep). @vivekchand please approve when ready.

---

PR #5963, opened 2026-09-14
fix(otlp): acknowledge an export only once it is stored; count what was refused

vivekchand:
Refs vivekchand/clawmetry#5949

Product record: https://factory.8090.ai/project/b415065f-ab2f-4f53-8864-0c009fd098cb/requirements/3caa08b2-4e39-4044-91f7-113fb2668ae0 (REQ-OBS-OIA-001, written before the code)

## Problem

The OTLP receiver answered 200 to every export it could decode, whether or not anything reached the store. The store write was best effort. While the daemon that owns DuckDB was restarting, busy or unreachable, the dashboard's `_ProxyStore` returned `None`, a warning was logged, and the receiver still answered 200. The exporter then discarded a batch that was never stored. A span the store refuses (no id, no name) got the same answer.

The issue also names three gaps:
- A spending-limit pause answered **429 to every export** for as long as it lasted, so the spend after a budget incident was never recorded.
- A retried span or generic metric point added its cost and tokens to the live tiles a second time.
- A record reporting input tokens only stored output as `0`, and a rollup group where no record carried a cost showed `$0.00`.

The existing suite was hiding the first gap. On `main`, `tests/test_otlp_compat_port.py` passes while every span write fails. Another file's fixture leaves a stopped store singleton behind, and running the two files together on `main` logs `local_store.put_span failed: Connection Error: Connection already closed!` twice and still reports `30 passed`.

## What changes (the verified subset this PR claims)

| Area | Before | After |
|---|---|---|
| Acknowledgement | 200 once the body decoded | 200 only after the store confirms the write; **503 + `Retry-After: 5`** when it does not |
| Malformed items | Silently dropped, still 200 | Valid items stored; 200 with OTLP `partialSuccess` (`rejectedSpans` / `rejectedLogRecords` / `rejectedDataPoints`); protobuf senders get a protobuf response |
| A value the store cannot hold | (see review fix below) | Counted as a malformed item and refused on its own: a token count beyond the INTEGER column, or a string the driver cannot encode. The rest of the export is stored and the answer is 200 with `partialSuccess` |
| Trace writes | `put_span` per span (one proxy hop each; `None` on success *and* on loss) | One `ingest_spans_batch(spans=[...], with_outcome=True)` per export, returning `{written, rejected}` (added to `_DAEMON_METHODS`) |
| Budget pause | 429 on every export | Intake continues; requests received during a pause are counted |
[...]

PR #5953, opened 2026-09-13
fix(otlp): trace spans reach Guard and are scrubbed before storage

vivekchand:
Refs vivekchand/clawmetry#5938

Requirement: https://factory.8090.ai/project/b415065f-ab2f-4f53-8864-0c009fd098cb/requirements/ff78b7cc-7a85-4f0f-a155-3fc78c6cd8b6
(REQ-OBS-OTG-001, child of Local Agent Observability; contract recorded at the top of the Runtime and Session Observability blueprint)

## Why

A span received on `/v1/traces` was stored in the `spans` table and nowhere else.

* Guard's detectors read `events` (`sync._emit_detector_incidents` -> `store.query_events`), so a trace showing an agent run `rm -rf /` raised no incident.
* Redaction ran only at the event chokepoint (`LocalStore.ingest`), so an API key or an email in a prompt, a tool argument, a URL attribute or an exception event rested in DuckDB exactly as sent.

The OTLP log path had neither gap. Only traces did.

## What

| Part | File |
| --- | --- |
| Tool spans (`gen_ai.tool.name` / `tool.name` / `gen_ai.operation.name=execute_tool` / a runtime profile) become `tool_call` + `tool_result` events in the existing event contract, with `trace_id`, `span_id`, `call_id` and session provenance. They go through the same detectors, `loop_signals` rows and policy pass as every other runtime. No second pipeline. | `clawmetry/otlp_guard.py`, `dashboard.py` |
| Every span is scrubbed on the store's single span write path (receiver and adapter span reconstruction alike): `name`, `status_message`, `input`, `output`, `attributes` (URLs, headers, paths), `events` (exception events), `links`. Dotted OTel keys match the sensitive-name rule on their last segment. | `clawmetry/redaction.py`, `clawmetry/local_store.py` |
| A value that cannot be scanned (over the scan cap, or the scanner throws) is **withheld** and the span is marked `clawmetry.redaction: withheld:<reason>`. It is never stored raw while the posture reads "applied". `CLAWMETRY_REDACT=0` still stores spans as received, matching events. | `clawmetry/redaction.py` |
| One source per tool call: the existing daemon-ownership rule applies to trace events, and a new rule makes the first OTLP signal (logs or traces) to report a session's tool calls its owner, so a runtime exporting both is not counted twice. | `clawmetry/local_store.py` (`put_otlp_batch`) |
[...]

Issue #5939, opened 2026-09-13
OpenTelemetry: CI-tested recipes for LangGraph/OpenLLMetry, Azure AI Foundry and AWS AgentCore


vivekchand:
## Problem
The OTLP receiver supports GenAI semantic conventions (`dashboard.py:4333`, `:4546-4607`) and builds sessions from spans (`routes/local_query.py:952`), but the docs (`docs/OPENTELEMETRY.md`, `docs/enterprise.md`) only say "set `OTEL_EXPORTER_OTLP_ENDPOINT`". Teams building agents on these frameworks can't tell whether their traces will show up usefully. We haven't tested it end to end either.

## Done when
- [ ] `examples/otel/{langgraph,azure-ai-foundry,aws-agentcore}/`: a minimal agent with one tool, exporting to ClawMetry
- [ ] A CI job runs each (with a stub LLM where needed) and asserts the session, token counts and tool call appear via `/api/local/*`
- [ ] Docs page per framework, including auth for a remote node (Bearer token) and what each tab shows for span-only data
- [ ] Depends on the spans → Guard fix for the Guard part

## Enterprise planning refinement, 13 September 2026

The specific requirements below refine the original scope and supersede conflicting implementation assumptions. Existing findings and discussion remain available.

## Pilot-first integration scope

Start with one actual selected framework/exporter version and a single tool; expand to other recipes as independently verified increments. A stub LLM is useful in CI but does not prove that a customer/provider integration works. Keep lightweight examples separate from mandatory desktop dependencies.

- [ ] Record exporter/SDK versions, encoding, auth headers, semantic-convention revision, parent-child structure, service/assistant identity and conversation/run identity.
- [ ] Test missing conversation IDs and name the fallback/cardinality policy; never imply each trace is automatically an assistant definition.
- [ ] Run a non-sensitive real application call through the agreed collector to persistent storage and rendered session/token/tool data, with source IDs and receipts.
- [ ] Cover streaming, retries, cache usage, model aliases, late events and missing fields; use unsampled usage for cost accounting.
- [ ] Document observe-only limitations when the producer exposes no hook/process actuator. A submitted span cannot itself stop a remote command.

[...]

Issue #5938, opened 2026-09-13
OTLP traces: spans bypass Guard detectors and redaction


vivekchand:
**Priority:** P0 — any agent built on LangGraph / Azure AI Foundry / AWS AgentCore that reports over OpenTelemetry is invisible to Guard, and its prompt text is stored unredacted.

## Problem
- Guard reads `store.query_events` (`clawmetry/sync.py:22192`), but spans go to a separate spans table via `put_span` (`clawmetry/local_store.py:9836`) / `ingest_spans_batch` (`:9701`). No detector ever sees them.
- Redaction runs only in `LocalStore.ingest` (`:4877-4900`). Span input/output/attributes are stored as received.
- `/v1/logs` does go through `ingest`, so it is redacted and detectors fire (`tests/test_otlp_daemon_free_intake.py`). Only traces are affected.

## Done when
- [ ] Tool-call spans (`gen_ai.tool.*` / execute_tool) are normalised into events that Guard evaluates
- [ ] `redaction.redact_event` (or a span equivalent) runs before span payloads persist
- [ ] Test: a sample OTLP trace with a tool span running `rm -rf /` raises a Guard incident, and an email/secret in `gen_ai.input.messages` is masked in storage

## Enterprise planning refinement, 13 September 2026

The specific requirements below refine the original scope and supersede conflicting implementation assumptions. Existing findings and discussion remain available.

## Scope/acceptance correction

Secret redaction and personal-data filtering are distinct. The existing credential scrubber does not establish email/PII removal. Verify actual redaction coverage; add an explicit configured content-minimization profile if PII removal is required, and label unsupported fields. Test prompts, responses, tool arguments, resource/span attributes, URLs, paths and exception events, not just one input field.

- [ ] Normalize tool spans once into the existing event contract, preserving trace/session/request identity and avoiding duplicate incidents or charges when logs and traces describe the same action.
- [ ] A harmless command-shaped canary is detected through the real normalized event path; the test must not execute a destructive shell command.
- [ ] Redaction/minimization runs before every applicable persistent/queued/exported copy. Scrubber errors produce a visible state consistent with the configured collection policy, never a false “fully scrubbed” claim.
- [ ] Label trace-derived incidents as observation unless a separately integrated pre-tool actuator actually held the operation. Seeing a dangerous action after it ran is not prevention.
- [ ] Demonstrate benign controls and missing-field behavior, plus local and applicable hosted rendering.

[...]

PR #5972, opened 2026-09-14
feat(otel): LangGraph + OpenLLMetry recipe, CI-tested against a live ClawMetry (a run is no longer split in two)

vivekchand:
Refs vivekchand/clawmetry#5939

Requirement: https://factory.8090.ai/project/b415065f-ab2f-4f53-8864-0c009fd098cb/requirements/0130f6a6-3f5b-448f-bd59-d8820a4d5c88 (REQ-OBS-OTR-001, child of Runtime and Session Observability; written before the code, blueprint filled in)

Branched off `main`. Independent of #5953 (spans reach Guard) and #5963 (durable ack): either merge order works. The Guard line of the issue stays with #5953.

## Scope: the pilot-first increment

The issue's refinement asks for one real framework and exporter version first, CI-tested end to end, with other frameworks added only as separately validated increments. This PR is that first increment: **LangGraph + OpenLLMetry**. Azure AI Foundry and AWS AgentCore are not included and are listed below.

## What running the real thing found

A one-tool LangGraph agent (langgraph 1.2.11, opentelemetry-instrumentation-langchain 0.62.3, opentelemetry-sdk 1.44.0) exported to a scratch dashboard came back as **two sessions for one run**:

| session | tokens | tool call |
|---|---|---|
| `t-1` (the thread) | 0 | none |
| `proto:trace:<trace_id>` | 307 | `lookup_invoice` |

OpenLLMetry sends the thread as `gen_ai.conversation.id` on the top `invoke_agent` span **only**. Every span beneath it (both model calls and the tool call) carries it as `traceloop.association.properties.thread_id`. `_otel_to_row` read only the first, so the children fell through to the per-trace fallback.

**Fix (one pick list in `dashboard.py::_otel_to_row`):** read `traceloop.association.properties.thread_id` right after `gen_ai.conversation.id` and before `session.id`, recorded as sent. Precedence: sent conversation id, thread, `session.id`, then `<app>:trace:<trace_id>`. It is not prefixed, because the top span already sends the same value bare; a prefix would recreate the split.

## What ships

| Part | File |
|---|---|
| Recipe: one tool, deterministic stub model (no key), `--thread`, `--stream` | `examples/otel/langgraph/` |
| Live verifier (stdlib), reads back only through `/api/local/traces` + `/api/local/spans` | `scripts/verify_otel_recipe_langgraph.py` |
| CI job `otel-recipe-langgraph`: recipe in its own venv from a hash-pinned lockfile, dashboard on `0.0.0.0`, agent exports through the runner's non-loopback IP | `.github/workflows/ci.yml`, `.github/requirements/otel-recipe-langgraph.txt` |
| Hermetic guard: exports **captured from the recipe itself** (not hand-written) replayed through the receiver; pins recipe, lockfile, CI job and page to the same versions | `tests/test_otel_recipe_langgraph.py`, `tests/fixtures/otel/` |
| Page: versions, encoding, semconv revision, span tree, identity and precedence, missing-id fallback and cardinality, per-tab behaviour, Bearer auth for a remote node, observe-only limits | `docs/OTEL_RECIPE_LANGGRAPH.md` (linked from `OPENTELEMETRY.md`, `BRING_YOUR_OWN_AGENT.md`) |

The verifier checks, each a hard failure with a sentence:
- an export through a non-loopback address with no token gets **401**; the recipe exporting with `OTEL_EXPORTER_OTLP_HEADERS=Authorization=Bearer …` is accepted;
- two runs on one thread (`invoke()` then `stream()`) are **one session** holding both traces;
- each trace carries exactly the tokens the model reported (280 in / 27 out);
- every span of each run is in that session; the `lookup_invoice` span descends from `invoke_agent` in the same trace; both model calls carry the model and a derived cost; all spans are attributed to `invoice_agent`;
- a run with no thread is its own session `invoice_agent:trace:<trace_id>`.

## Verified (locally, Python 3.11, scratch HOME and store, random port)

- **Red on `main`** (`dashboard.py` swapped for `origin/main`'s copy): 4 hermetic tests fail (`thread_run_is_one_session_with_every_span`, `second_run_on_the_thread_joins_the_same_session`, `streamed_run_matches_invoked_run`, `thread_wins_over_resource_session_id`), and the live verifier fails with `trace …: tokens 0/0, the model reported 280/27`.
- **Green with the fix**: 11/11 in the new file; the live verifier prints `PASS` including the 401/Bearer step through the machine's LAN address.
- Neighbouring OTLP suites unchanged: 87 passed across `test_otel_recipe_langgraph`, `test_otlp_session_materialization`, `test_otlp_sessionless_spans`, `test_spans_otlp_edge_cases`, `test_otlp_runtime_profiles`, `test_otlp_traces_cost`, `test_otlp_json_no_protobuf`.
- `check_ac_coverage --check` (8 criteria mirrored, 117/187), `check_ci_test_coverage --check` (baseline tightened 920 to 919), `gen_module_map --check`, `check_py39_annotations`, ruff on the new files.
- No new HTTP route, so no `cloud_route_policy` entry is needed.

## Found while verifying, recorded on the page, not fixed here

- **Guard** lists a span-only session but labels it `openclaw`, offers Pause/Stop, and suggests `openclaw attach`. None of that can reach a remote agent. The page says so and does not claim control. It belongs with the spans and Guard work (#5938).
- **Usage**: on a daemon-free dashboard the daily totals did not include span-only tokens. The page marks Usage "not verified".
- An unrecognised model alias still gets a fallback-rate cost, so the page tells the reader to check the model name.

## Remaining for #5939 (not in this PR)

- `examples/otel/azure-ai-foundry/` and `examples/otel/aws-agentcore/`, each with its own CI-validated recipe and page.
- A non-sensitive real provider call through the agreed collector to persistent storage, with receipts (CI uses a stub model; no provider key tonight).
- Retries and partial delivery (#5963 / #5949), late spans after materialization, spans missing token fields, and real-provider streaming or cache token accounting in the check.
- A Guard incident from a recipe tool call (#5938).
- Auth via the ingest key (#5684) instead of the gateway token, once that lands.


vivekchand:
The first run of the new `OTel recipe (LangGraph + OpenLLMetry)` job was red at the boot step, before the recipe ran. The hermetic guard in the same job passed (11 tests).

**Cause:** readiness was probed on `/api/local/health`. On a fresh store path the DuckDB file does not exist until the receiver's first write creates it, so the read-only local API answers `503 {"error": "IO Error: Cannot open database ... in read-only mode: database does not exist"}` and the loop never passed. The dashboard itself was up. It was still running at cleanup, and `/api/health` answered 200 after 3 to 4s.

**Reproduced** locally in a venv installed from `.github/requirements/ci-tests.txt` + `ci-otel-extra.txt` with `--require-hashes`: `/api/health` 200, `/api/local/health` 503 with that exact message.

**Fix** (f4c764102b): readiness now probes `/api/health`, like every other job that boots `dashboard.py`, and the dashboard runs with `python3 -u` so a boot failure prints its log. Nothing in the recipe, the verifier or the receiver changed.

**Rehearsed** the job's steps locally before pushing: CI-pinned dashboard on `0.0.0.0`, verifier with `--remote-url` through the LAN address. Result `PASS`: 401 without a token, accepted with a Bearer token, one session for the invoke and stream runs on a thread (614 tokens), the tool span under `invoke_agent`, and a per-trace session for the no-thread run.

vivekchand:
## Coordinator review

**Verdict: fix first, one blocking item.** Everything else checks out. Reviewed at `f4c764102b`, which is on current `main` (merge base = `main` HEAD `658342f`).

### Blocking

**1. AC-OBS-OTR-001.4 says the opposite of what the PR ships.**

Factory REQ-OBS-OTR-001 and its copy in `docs/acceptance_criteria.json` both read:

> An explicitly sent conversation **or session** identifier shall take precedence over a thread identifier.

The PR does the reverse in three places:
- **`dashboard.py::_otel_to_row`:** the pick order is `gen_ai.conversation.id`, then `traceloop.association.properties.thread_id`, then `session.id`.
- **The page's precedence list:** thread (2) is above `session.id` (3).
- **The test cited for .4:** `test_thread_wins_over_resource_session_id` asserts that the thread beats `session.id`.

This is not only about the resource `session.id` the test covers. `_pick` checks span attributes before resource attributes, key by key. So a `session.id` set on a span by the application also loses to the thread now.

The product record and its guard disagree, and Drift Bot did not catch it. Pick one:
- **(a) Keep the current behaviour.** Put a tracked edit on AC .4 in Factory and mirror it in `acceptance_criteria.json`. Suggested wording: "a sent conversation identifier takes precedence over a thread identifier; a thread identifier takes precedence over `session.id` on the span or resource; each recorded as sent". Also fix the code comment, which says only the resource-level `session.id` is outranked.
- **(b) Keep the AC as written.** Then a span-level `session.id` has to win over the thread, which needs a code change and a test. The resource-level question still needs an answer either way.

### Verified (not taken from the builder report)

- **Hermetic guard:** I ran `tests/test_otel_recipe_langgraph.py` in a scratch venv and worktree.
  - At the PR head: 11 passed.
  - With `origin/main`'s `dashboard.py`: 4 failed (`thread_run_is_one_session_with_every_span`, `second_run_on_the_thread_joins_the_same_session`, `streamed_run_matches_invoked_run`, `thread_wins_over_resource_session_id`).
  - The guard works and fails without the fix.
- **Live CI job log (job 103864424258):**
  - 401 through `10.1.0.205` with no token; the Bearer-token export is accepted.
  - Two traces with 10 spans each in one thread session, 614 tokens.
  - `lookup_invoice` descends from `invoke_agent`.
  - The no-thread run is `invoice_agent:trace:<id>`.
  - `PASS`.
- **Other OTLP session paths:** the only other one is the `/v1/logs` pick at `dashboard.py:5389`, which LangGraph traces do not use. No daemon-side OTLP path works out session ids on its own, so a real install behaves the way CI tests.
- **Fixtures:** the resource keys are `service.*`, `telemetry.sdk.*` and `deployment.environment`. No host, user, path or personal data.
- **Public-repo hygiene:** no customer, partner or deal names in the diff (`invoice-agent`, `support-42`, `clawmetry.internal` are generic).
- **Gates:**
  - The full Factory URL is in the PR body.
  - The job has no `continue-on-error`.
  - The coverage baselines only get tighter (ac 109/179 -> 117/187, unlisted 920 -> 919).
  - No new HTTP route, so no cloud route-policy PR is needed.
  - CI triggers on every `pull_request` with no path filter, so "runs on every change" is true.
- **Acceptance criteria for #5939:** the unmet ones are stated in the PR and on the page, not hidden:
  - Foundry and AgentCore recipes;
  - a real provider call through the collector;
  - retries, late events and missing token fields;
  - Guard.
- **Duplication:** the PR uses `Refs`, so the issue stays open. No competing implementation. It reuses the existing OTLP receiver and REQ-OBS-RSO-033, and correctly leaves auth (#5684), durability (#5949 / #5963) and Guard (#5938 / #5953) to their own PRs.

### Non-blocking (worth doing in this PR or a follow-up)

- **Ingest contract.** `clawmetry/ingest_contract.py` generates `docs/INGEST.md`, whose `gen_ai.conversation.id` alias row still lists only `session.id`, `openclaw.session_id` and `session_id`. The receiver now reads `traceloop.association.properties.thread_id` too, and the issue asks to reuse #5682's generated docs. `test_ingest_doc_drift.py` does not compare aliases against the code, which is why CI stayed green.
- **Bare thread ids collide across applications.** Many LangGraph examples use `thread_id="1"`, so two apps could map to the same `session_id`. The same was already true for a bare `gen_ai.conversation.id`, but the fix now also moves the token-bearing child spans onto it. I did not check whether the `sessions` table keys on `session_id` alone. If it does, that is a cross-app merge worth an issue.
- **Remote example uses `https://clawmetry.internal:8900`.** The dashboard does not serve TLS itself, and the CI proof used `http://`. Existing docs (`BRING_YOUR_OWN_AGENT.md`) already use this pattern, so it is not new. Still, one line saying TLS needs a terminator in front would stop the Bearer token being sent in the clear.
- **"What each tab shows" table.** Its evidence is the API endpoints, not a screenshot of the Tracing or Sessions tabs. That is acceptable, but say "checked through the endpoints each tab reads" if that is what was done.
- **Known Guard defect.** Guard labels span-only sessions as `openclaw` and offers controls that cannot act. It is honestly documented, but it should be filed or linked on #5938 so it does not live only on a docs page.

vivekchand:
Review follow-up (the blocking item on AC-OBS-OTR-001.4), fixed in 44d0eb88d3:

**Decision: keep the behaviour, correct the record (option a).** The precedence is: sent `gen_ai.conversation.id`, then the thread, then `session.id` on the span or the resource. Letting a span-level `session.id` beat the thread would bring the split back. The top `invoke_agent` span resolves through its conversation id, so an app that stamps `session.id` on every span would file the children under a different session.

- **Factory:** a tracked edit on AC .4 of REQ-OBS-OTR-001 (https://factory.8090.ai/project/b415065f-ab2f-4f53-8864-0c009fd098cb/requirements/0130f6a6-3f5b-448f-bd59-d8820a4d5c88) is pending review. It is mirrored word for word in `docs/acceptance_criteria.json`.
- **Guard:** `test_thread_wins_over_session_id_on_span_or_resource` now covers `session.id` on every span as well as on the resource. With `session.id` read before the thread key, both cases fail (2 failed, 10 passed). With `origin/main`'s `dashboard.py`, 5 fail. With the fix, all 12 pass, plus `tests/test_ingest_doc_drift.py`. `check_ac_coverage.py --check` is OK.
- **Code comment:** it now says the thread outranks `session.id` at both levels, and why.

Non-blocking items taken:
- `clawmetry/ingest_contract.py` lists `traceloop.association.properties.thread_id` first among the conversation-id fallbacks; `docs/INGEST.md` is regenerated.
- The page says the dashboard serves plain HTTP, so the `https://` example needs a TLS terminator, and a plain `http://` endpoint sends the Bearer token in cleartext.
- The tab table now says it was checked through the API endpoints the tabs read, not the rendered tabs.
- The Guard mislabel (span-only session labelled `openclaw`, offered controls, `openclaw attach` hint) is recorded on #5938: https://github.com/vivekchand/clawmetry/issues/5938#issuecomment-5659696586

Not taken: the bare-thread-id collision across `service.name`s. It predates this PR for bare conversation ids, and I have not verified how the sessions table keys it. It is listed as remaining.

vivekchand:
**Ready to merge** (at 44d0eb88d3). Every check passes except OpenSSF Scorecard, which is skipped. That includes E2E Gate (required), Drift Bot, the product-record gate, CodeQL and the `OTel recipe (LangGraph + OpenLLMetry)` job. GitHub reports the PR as mergeable (merge state CLEAN). Not merged.

**Merge-after dependencies:** none. The branch is on current `main` (658342f238) and does not depend on #5953 (spans reaching Guard) or #5963 (durable ack), so any merge order works.

**Companion PRs:** none. There is no new HTTP route, so no cloud route-policy entry is needed.

**Before or at merge:** accept the pending tracked edit on AC .4 of REQ-OBS-OTR-001 in Factory: https://factory.8090.ai/project/b415065f-ab2f-4f53-8864-0c009fd098cb/requirements/0130f6a6-3f5b-448f-bd59-d8820a4d5c88. `docs/acceptance_criteria.json` already mirrors the new wording.

**After merge (on `main`):**
1. The `OTel recipe (LangGraph + OpenLLMetry)` job passes on the main push. Its log should show:
   - `PASS`;
   - a `401` without the token;
   - one session holding both runs on the thread, with 280 in / 27 out per trace;
   - an `invoice_agent:trace:<id>` session for the run with no thread.
2. `python3 -m pytest tests/test_otel_recipe_langgraph.py tests/test_ingest_doc_drift.py -q` passes.
3. `python3 scripts/check_ac_coverage.py --check` is OK, and Drift Bot is green on main.

**After release:** repeat the CI job by hand at the release tag, using a scratch HOME and scratch venvs, never the real `~/.clawmetry`.
1. Install the dependencies:
   - dashboard: `pip install "clawmetry[otel]==<released>"`, plus `pytest` if you also want to run the tests;
   - recipe, in its own venv: `pip install --require-hashes -r .github/requirements/otel-recipe-langgraph.txt`.
2. Start the dashboard from the tag's checkout: `HOME=$(mktemp -d) OPENCLAW_GATEWAY_TOKEN=t CLAWMETRY_LOCAL_STORE_PATH=$(mktemp -d)/r.duckdb python3 dashboard.py --port <free port> --host 0.0.0.0 --no-debug &`
3. Run the verifier: `python3 scripts/verify_otel_recipe_langgraph.py --url http://127.0.0.1:<port> --token t --agent-python <recipe-venv>/bin/python --remote-url http://<this machine's LAN IP>:<port>`. It runs the agent itself (invoke, stream and no-thread). Expect `PASS`.
4. Kill the dashboard, then delete the scratch HOME and venvs.

**Still open on #5939:**
- the Azure AI Foundry and AWS AgentCore recipes;
- a real provider call through the agreed collector;
- retries, late spans and missing token fields;
- Guard over span tool calls (#5938 / #5953).

PR #5970, opened 2026-09-14
Self-hosted server: signed multi-arch image verified by anonymous pull, and admin APIs reachable in a container

vivekchand:
Refs vivekchand/clawmetry#5948

Product record: https://factory.8090.ai/project/b415065f-ab2f-4f53-8864-0c009fd098cb/requirements/49529742-f381-4957-9abf-c92efe7b7d65

Merge order: independent of #5950 (fleet install). Overlaps only on append-only files (`CHANGELOG.md`, `docs/acceptance_criteria.json`, `docs/ac_coverage_baseline.json`, `.github/workflows/ci.yml`); keep both sides on conflict. Distinct from clawmetry-pro#4 (Helm / air-gapped packaging) and clawmetry-cloud#1201 (self-hosted cloud dashboard).

## Problem

1. `deploy/self-hosted/docker-compose.yml` builds the image from a source checkout. There is nothing to pull, no signature, no SBOM, and no way to tie a running container to a release.
2. Found while building the verification: the runbook's own check, `curl -u admin:... http://localhost:8900/api/selfhosted/status`, answers **401 "Gateway token not configured"** to any non-loopback caller, even with valid admin credentials. Every request through a container port is non-loopback, so status, the node list, approvals and **audit export were unreachable in the documented deployment**. Reproduced against `main` today by running the self-hosted server bound to `0.0.0.0` and calling it over the LAN address: `/api/selfhosted/status` and `/api/export/events` returned 401 while `/ingest/events` and `/selfhosted` returned 200. The dashboard's gateway-token gate ran before the self-hosted views could check their own credentials.

## What this changes

| File | Change |
|---|---|
| `.github/workflows/container-image.yml` (new) | **publish** (dispatch): downloads the wheel PyPI serves for the version, `gh attestation verify` on it, buildx for `linux/amd64,linux/arm64`, push to `ghcr.io/vivekchand/clawmetry:<version>`, `cosign sign --recursive` (keyless), syft SPDX SBOM per platform attested with `cosign attest --type spdxjson` to that platform digest, `attest-build-provenance` pushed to the registry. **verify** (needs publish, `contents: read` only, no registry login, native runner per arch): anonymous registry probe, `docker pull` with an empty Docker config, signature + SBOM check, no Pro plugin inside, event survives a container restart. **pin** (needs verify): opens a PR pinning Compose by version and digest. **pull_request leg**: builds the branch wheel, builds the image on amd64 and arm64, runs the same contents + persistence check (no signature). |
[...]

PR #5952, opened 2026-09-13
Guard: tag findings and decisions with OWASP LLM 2026, OWASP Agentic 2026 and MITRE ATLAS IDs

vivekchand:
Refs vivekchand/clawmetry#5943

Product record: https://factory.8090.ai/project/b415065f-ab2f-4f53-8864-0c009fd098cb/requirements/15504aea-9ca0-4a1a-a23d-8e825e4f78f9
Blueprint: https://factory.8090.ai/project/b415065f-ab2f-4f53-8864-0c009fd098cb/blueprints/ca1e6986-9bc9-476d-8bf6-46e40868820e (child of the Guard blueprint 8a92d41f, whose placeholder is filled in the same pass)

## Why

Security reviewers work from framework identifiers, and no Guard finding carried any, so the mapping lived only in slides that nothing checks against the detectors. The 13 September refinement adds four requirements: verify every identifier against primary framework data, keep "relevant to" apart from "prevents", never force a mapping to avoid "none", and never present a configured policy as exercised control.

## What

- **`clawmetry/framework_map.py`**: one versioned mapping contract (`2026-09-14.1`) covering all 15 finding kinds in `detectors.ALL_INCIDENT_KINDS` (detector, workspace and fleet). It maps to three frameworks: OWASP Top 10 for LLM Applications **2026 v1.0**, OWASP Top 10 for Agentic Applications **2026**, and MITRE ATLAS **2026.08**, with ATLAS tactics, techniques, mitigations and case studies kept apart by type. Each mapped kind carries a rationale, limits, what it needs to work, and a firing test and a quiet test. Four kinds say `none` with a reason instead of borrowing an identifier.
- **Findings**: every incident now carries `frameworks: {mapping_version, owasp_llm, owasp_asi, atlas, mapped, mode: "detect", pre_action_control: false}`. It is stamped where incidents are made (`detectors.run_all`, `repo_scan`, `detector_swarm`), stored in the `loop_signals` details, and returned on `/api/guard/sessions`. Rows written before this change get the current contract's references.
- **Policy decisions**: `policy_engine.evaluate` copies the triggering incident's references onto its decision. Each `/api/guard/actions` row adds `frameworks` and an `evidence_level`:
  - `configured`: monitor, alert, dry run, not entitled, or the actuator has not reported;
  - `exercised`: the actuator reported success;
  - `failed`: the actuator reported failure.

  `effective` is never assigned, because nothing observes the outcome independently. There is no schema change.
[...]

Issue #5946, opened 2026-09-13
CI/CD: GitHub Action / Azure DevOps task — agent-session provenance per PR with SARIF output


vivekchand:
## Problem
Teams want AI-assisted changes flagged in the pipeline, not in a separate dashboard. There is no customer-facing Action, no SARIF output and no gate command today.

## Proposal
ClawMetry does not become a code scanner. It links what agents did to the PR:
- `clawmetry pr-report --base <sha> --head <sha>` maps changed files to the agent sessions that wrote them (tool write events + cwd/git), with model, cost and Guard incidents per session.
- Output as SARIF (incidents as results on the touched files) plus a markdown PR comment.
- Optional ingestion of Semgrep/Snyk SARIF so findings are annotated with the agent session and prompt that produced the code.
- A composite GitHub Action and an Azure DevOps task wrapper; `--fail-on critical` for gating.

## Done when
- [ ] Demo repo: an agent session writes a file with a known issue → PR shows the annotation linked to the session
- [ ] Works without cloud (reads the local store or an exported bundle)

## Enterprise planning refinement, 13 September 2026

The specific requirements below refine the original scope and supersede conflicting implementation assumptions. Existing findings and discussion remain available.

## Pilot-first implementation and acceptance

Start with one customer-selected CI provider and existing scanner/status integration. OSS owns the gated API/CLI hook contract; paid provenance/evidence implementation belongs in the extension. Extend existing evidence/audit export rather than building a replacement SAST/SCA scanner.

- Link evidence to repository, commit/build, source session/trace, model/prompt/tool configuration, policy/evaluation versions, actor and time. File-touch overlap is association, not proof that a specific prompt caused a vulnerability; represent confidence and missing attribution.
- Ingest/link existing SAST/SCA/secrets/container findings and agent evaluations where available. Keep prompts and secrets out of SARIF, public PR comments and build logs; use scoped evidence references.
- CI owns deployment authorization. A gate is explicitly enabled with threshold, freshness, missing-evidence behavior, exceptions/expiry, override and audit. Defaults do not silently block or allow production.
- [ ] Known failing evaluation prevents the selected test pipeline only when gating is enabled; a passing evaluation, disabled gate, stale/missing bundle and approved exception all behave as documented.
- [ ] Verify commit binding against substitution/stale evidence and show actor/rule version for decisions.
- [ ] Works from an exported/local bundle without vendor cloud when the deployment requires it; export authenticity/retention is explicit.

[...]

PR #5974, opened 2026-09-14
PR provenance report: agent sessions on changed files as SARIF, with an opt-in CI gate

vivekchand:
Refs vivekchand/clawmetry#5946

Requirement: https://factory.8090.ai/project/b415065f-ab2f-4f53-8864-0c009fd098cb/requirements/4e224b46-71a6-4d32-933d-7fb6bfde30d0 (REQ-OBS-PRP-001, a child of Runtime and Session Observability, written before the code)

Merge order: independent of #5950 and #5970 (neither is needed). Branched off `main`.

## Why

Teams want AI-assisted changes flagged in the pull request, not in a separate dashboard. ClawMetry already knows which session wrote a file, with which runtime and model, what it cost and which Guard findings it raised. None of that reached the pipeline, and a CI runner has no local store to read.

## What

`clawmetry trace report` links each changed file in `base..head` to the agent sessions that wrote it, and emits SARIF 2.1.0 plus a markdown comment. It extends the existing PR Trace attribution (`trace_stamp` trailer, `trace_capture`) instead of building a second one.

- **Linking with a basis.** `commit_trailer` means a commit carries `Clawmetry-Session` and changed the file. `observed_write` means the session's own write tool call (Write, Edit, MultiEdit, apply_patch patch headers, and so on) touched the path before the head commit. Every link and annotation says *association, not proof of cause*. Files no session wrote are listed as unattributed.
- **SARIF.** Guard findings of linked sessions become results on those files (critical to error, warning to warning, else note). There is also a note-level provenance result per link.
- **Scanner SARIF in, annotated out.** `--scanner-sarif` reproduces Semgrep, Snyk, CodeQL and other runs, and attaches `properties.clawmetry` (session, runtime, model, basis) to findings on agent-linked files only. **ClawMetry adds no finding about code content.**
- **Works without the store or cloud.** `--export-bundle` on the developer's machine, `--bundle` on the runner. The bundle carries covered commits and a sha256 digest. A tampered bundle, or one from another change, is `invalid`. Commits after the export make it `partial`, and `--max-evidence-age-hours` makes it `stale`. The digest is labelled an integrity check, not proof of authenticity.
- **Gate, opt-in.** Off unless `--fail-on warning|critical`. When enabled it refuses to run (exit 2) without `--on-missing-evidence fail|pass`, so there is no silent block and no silent allow. Exceptions need id, rule, expiry, approver and reason. Expired exceptions do not apply and are listed. Each decision records outcome, rule version, threshold, evidence behaviour and status, actor, time, head SHA, blocking findings and the exceptions applied, expired or rejected. It is written to `--json-out` and to the SARIF run properties.
- **No sensitive text in any output.** Prompt text, tool output and credential-shaped values never appear. Scanner text passes secret redaction, and Guard titles pass PR Trace's publication redaction.
- **GitHub Action.** A composite action in `integrations/github-action/` installs clawmetry, runs the report, uploads SARIF (pinned `upload-sarif`), keeps one PR comment updated in place via a marker (GitHub REST, no `gh` dependency), writes the job summary and applies the gate's exit code. Inputs reach the shell via `env:`, not `${{ }}` interpolation.

- **The repository stays read-only.** Every git call goes through `git_outcomes`' chokepoint (CLAUDE.md: "a user's repository is read, never written"). Its allowlist gains `merge-base` and `diff-tree`, and it now refuses `--output` on every subcommand. That option makes even the already-allowed `git log` write a file.
- **Layout.** Logic is in `clawmetry/pr_provenance.py`. The command is in its own short module, `clawmetry/pr_provenance_cli.py`, dispatched from the `trace` fast path in `cli.py`, so Drift Bot reads it from the head of a file.
- **Empty changes are refused.** With no `--base` on the default branch, the inferred change is empty, so the command exits 2 instead of passing a gate over nothing.

No new HTTP route, so no `cloud_route_policy` entry or cloud PR is needed. No version bump.

## Review fixes (coordinator review, https://github.com/vivekchand/clawmetry/pull/5974#issuecomment-5660087034)

Requirement extended first with **AC-OBS-PRP-001.12** and **AC-OBS-PRP-001.13** (mirrored into `docs/acceptance_criteria.json`, baseline 130/200 after rebasing onto #5952).

1. **Sessions from other repositories were linked to this change (blocking).** `query_sessions()` returns no `cwd`, so every relative write path of every session active near the commits was read as this repository's. A Codex session in another project patching its own `README.md` got linked to a human edit here, failed the gate with its own Guard finding, and put its id, model and cost in the bundle. Now a relative path resolves only against the session's recorded working directory: `get_session_location` (prefixed id, then the bare id scoped to the prefix's runtime), falling back to a `cwd` carried on its events. With no recorded directory, a relative write counts only for a session a commit trailer names. Otherwise a match on a changed file is an evidence gap, and the gap names no session. Absolute paths inside the repository still link.
2. **Truncated evidence reported as `ok` (blocking).** Every read limit hit, and every failed store read, is now an evidence gap. The limits: 50 candidate sessions, 1,000 session rows cut inside the time window, 200 Guard findings per *linked* session, and 500 commits (also on the bundle path). Events are no longer one newest-first page of 5,000: they page back by timestamp past post-head activity, up to 50,000 per session, and only then is a cut declared. A gap makes the evidence `partial` in the report. It is also carried in the bundle as `evidence_gaps`, which the digest covers, so the CI runner reading the export sees `partial` too, and an enabled gate applies `--on-missing-evidence` instead of passing.

Non-blocking items also addressed:
- `upsert_pr_comment` edits only a comment written by the token's own identity. It uses `GET /user` for a personal token, and a Bot author for the Actions token, which cannot call `/user`. A person quoting the marker keeps their comment.
- The bundle no longer records `$USER`.
- The `started_at` fallback is now the earliest event, not the newest.
- The Action's upload runs only when the report exited 0 or 1.
- `docs/PR_PROVENANCE.md` now says a bundle committed to the PR is **not a control against the PR's author**. It also warns that scanner alerts can appear under two code-scanning categories, and that the upload and live comment are unverified on github.com.
- The CHANGELOG count is corrected.

**Regression guards, red then green.** With the fix files replaced by the previous head's versions, four tests fail on their real assertions:
- `test_session_in_another_repository_is_not_linked_and_does_not_gate`: the foreign session is linked.
- `test_relative_writes_with_no_working_directory_are_partial_evidence`: same failure.
- `test_sessions_beyond_the_read_limit_make_the_gate_apply_missing_evidence`: evidence `ok` where `partial` was expected, and the gate passes. This is the reviewer's 60-session repro.
- `test_pr_comment_is_created_once_then_updated`: the human comment carrying the marker is edited.

With the files restored byte-for-byte, all 22 pass. Two more tests are pure and unit-level:
- `test_relative_paths_resolve_only_against_the_sessions_own_directory`: own cwd, another repo, unknown cwd, absolute path, trailer-named, and the `started_at` fallback.
- `test_every_read_limit_is_a_gap_and_event_pages_walk_past_post_head_activity`: event paging, the event, finding and session-row limits, a failing store, the commit ceiling, and that invalid evidence stays invalid.

## Verification

- `tests/test_pr_provenance.py`: 22 tests (17 originally, 5 added for the review) (including a chokepoint test that every git call goes through `git_outcomes`, and the empty-inferred-change refusal), wired into ci.yml's MOAT Verifier job (duckdb installed there).
  - **Demo repo, end to end.** A real git repo where an agent commit writes a known SQL injection, and a real DuckDB store with that session's Write event, a user prompt carrying a fake key, tool output and a critical Guard finding. The test exports the bundle from the store, then runs `trace report` from the bundle with `get_read_store` patched to raise and `socket.connect` blocked. It asserts: Guard `error` results on `app/db.py`; the Semgrep finding on `app/db.py` carries `claude_code:demo-1`; the one on `docs/notes.md` does not; `docs/notes.md` is unattributed even though the session read it and edited it after the head commit; codex cost is `unknown`, not `$0`; and no prompt, tool output, fake Anthropic key, GitHub token or tmp path appears in the SARIF, markdown, JSON or bundle.
  - **Committed-bundle gate walk.** Gate off exits 0. Enabled without an evidence choice exits 2. A blocking critical exits 1. An approved exception exits 0 and is recorded with the actor. An expired exception exits 1 and is listed. A commit after the export gives `partial`, which exits 1 or 0 as configured. A tampered bundle is `invalid` and exits 1.
  - **PR comment.** A local fake GitHub API: the first run creates the comment, the second edits it, and neither the unrelated comment nor a person's comment quoting the marker is touched.
- **Revert, red, restore, green** (each mutation applied, then the file restored byte-for-byte and all tests green):
  1. Dropping the substitution check fails `test_bundle_from_another_change_is_invalid_evidence`.
  2. Silently defaulting the evidence behaviour fails `test_gate_refuses_to_run_without_an_evidence_behaviour` and the committed-bundle walk.
  3. Removing publication redaction from Guard titles fails the demo test (a fake key reaches the SARIF).
  4. Emptying `git_outcomes._WRITING_OPTIONS` fails `test_git_calls_go_through_the_read_only_chokepoint`.
  5. Removing the empty-inferred-change refusal fails `test_an_inferred_empty_change_is_not_reported_as_a_pass`.
- **SARIF schema.** Output from the demo repository validates against the OASIS SARIF 2.1.0 JSON schema with `jsonschema` (0 errors for both the report-only and the gated output; 2 runs, 5 results).
- **Real CLI, separate processes, scratch HOME and store.** One process seeds and closes the store. A fresh process runs `trace report --export-bundle` (direct read-only store open) and gets trailer and observed-write links, a measured cost and the Guard warning. From that bundle, `--fail-on warning` exits 1 and `--fail-on critical` exits 0.
- `.github/workflows/pr-provenance-action.yml` runs the composite action twice against `scripts/pr_provenance_demo.py`'s demo repo: report-only, then gated with approved exceptions. It verifies the SARIF annotation and the recorded gate decision. Upload and comment are off in that job: it proves the action runs, and does not publish results about a fixture.
- Local guards green: `check_ac_coverage --check` (13 criteria mirrored verbatim from the Factory record, baseline tightened to 130/200 on current main), `check_ci_test_coverage --check`, `gen_module_map --check` (regenerated), `test_workflow_yaml_valid`, `test_ci_workflow_invocations_are_real`, `test_action_refs_pinned`, `test_changelog_no_em_dashes`, ruff on the new files. `cli.py` has the same 13 pre-existing ruff findings as `main`, and none are new.

## Not in this PR (listed as remaining, not hidden)

- **Azure DevOps task.** The refinement names one pilot CI provider first; GitHub Actions is it.
- **Automatic bundle export on `git push`.** It is a command today.
- **Gating on agent evaluation scores.** Guard and scanner findings only.
- **Signed bundles, and linking the gate decision into the paid Compliance Pack evidence bundle.** The refinement puts paid evidence in the extension.
- **Tiering.** The report is free, like the PR Trace capture it reads. Whether the gate belongs to a paid tier is recorded as an open question in the requirement, not decided here, because adding a feature key is a pricing change.
- **SARIF upload to a real code-scanning dashboard and a live PR comment from the action.** Covered by the fake API and schema validation, not exercised against github.com in this PR.
- **A bundle committed to the PR can be rewritten by whoever can push to it.** The docs now say so: authenticity needs an export the author does not control, handed over through an artifact store. Signed bundles are not in this PR.


vivekchand:
## Coordinator review

**Verdict: fix before merge.** Most of it holds up. CI is green on 3a1ad6a124 (45 pass, Scorecard skipped), the PR is MERGEABLE/CLEAN, and the branch includes current `origin/main`. The full Factory URL is in the body, no new HTTP route is added (so no cloud policy entry is needed), and the diff names no customer. The 17 tests in `tests/test_pr_provenance.py` pass locally in a scratch venv with a scratch HOME. The unfinished acceptance criteria (Azure DevOps task, evaluation scores as a gate input) are listed as remaining, and the PR uses `Refs`, not `Closes`. Two defects in the store path, the one a developer's machine actually runs, need fixing before this is merged.

### Blocking

**1. Sessions from other repositories get linked to this change and can fail its gate.** Location: `clawmetry/pr_provenance.py` `collect_from_store`, where `cwd = row.get("cwd")`.
- `store.query_sessions()` never returns `cwd`. Its column list is `session_id, agent_id, started_at, updated_at, event_count, message_count, cost_usd, token_count` (`local_store.py`, around line 13488). So `cwd` is always `None`, and `repo_relative()` treats every relative path as relative to this repository.
- Candidate sessions are every session on the machine that was active near the commits, whatever repository it worked in. Codex `apply_patch` headers such as `*** Update File: README.md` are relative paths.
- Reproduced with the real functions: a real git repo where a human edits `README.md`, plus a store stub whose only session is `codex:other-project`, working elsewhere, with a critical Guard finding.
  - Result: `links: {'README.md': [{'session_id': 'codex:other-project', 'basis': 'observed_write'}]}`.
  - Result: `gate: fail [('clawmetry/guard/credential_access', 'README.md', 'codex:other-project')]`.
- Two consequences:
  - A false gate failure, and a false attribution on a file a human wrote.
  - A data-boundary leak. The other project's session id, model, cost and Guard titles go into the bundle the docs tell you to commit to the PR, and into the PR comment.
- The tests miss this because the demo store has only one session, working in the demo repo.
- **Fix:**
  - Read each candidate session's `cwd` from the `sessions` table; there is already a cwd/git_branch reader around `local_store.py` line 5161.
  - Only resolve a relative path against a known `cwd` that is inside the repository. Otherwise drop the path, or limit candidates to sessions whose `cwd` is in the repository. A session named by a commit trailer can stay.
  - Add a regression test with a second session working in another repository at the same time, and show it fails without the fix.

**2. When a limit cuts the evidence short, the report still says `ok`, so a gate can pass on evidence it never read.** Location: `collect_from_store` and `_from_store` in `pr_provenance_cli.py`.
- The limits that apply:
  - `max_sessions=50` candidates.
  - `query_sessions(limit=1000)`.
  - `query_events(limit=5000)`, which `local_store` returns newest first (`ORDER BY ts DESC`). A session that kept working after the head commit can use up the whole budget on events the `until` filter then throws away, so its earlier writes are never read.
- Reproduced: 60 sessions active at once, where the 56th wrote `README.md` and has a critical finding. Result: `{'README.md': []}`, with evidence `ok`, and the exported bundle says so too.
- The refinement requires that defaults "do not silently block or allow".
- **Fix:** when any limit is hit, set evidence to `partial` and give the reason, so `--on-missing-evidence` decides the outcome, or page through everything. Add a test.

### Non-blocking (founder decisions or follow-ups)

- **Where the code lives, a founder decision before release.** The refined issue says "paid provenance/evidence implementation belongs in the extension". This PR ships the bundle, the SARIF output and the gate free in public OSS, and records tiering only as an open question in the requirement. Confirm that on purpose, or put the gate behind an extension hook, before a `[RELEASE]` ships it.
- **A committed bundle can be forged by whoever writes to the PR.** Anyone who can push to the PR, including the agent, can rewrite the bundle and recompute the sha256 digest. The docs say to commit the bundle to the PR. The requirement's non-goal says authenticity rests on the CI artifact store. Say plainly in `docs/PR_PROVENANCE.md` that a gate reading a bundle from the PR is not a control against the PR's author.
- **The comment update can overwrite someone else's comment.** `upsert_pr_comment` edits the first comment carrying the marker, whoever wrote it. The Action's write token can edit a human's comment that happens to contain the marker. Also match the comment's author (the bot).
- **Scanner findings may appear twice in code scanning.** Scanner runs are uploaded again under the `clawmetry-provenance` category, so a team that already uploads its Semgrep SARIF may see every alert twice. The upload and the live comment are not verified on github.com (the builder says so). Document this, or default to uploading only the ClawMetry run.
- **The bundle records the local username.** `generated_by` is `$USER`, and it goes into a bundle meant to be committed to a PR.
- **`started_at` can be wrong.** When the session row is missing, it falls back to `events[0].ts`, which is the newest event, not the first.
- **`_MAX_COMMITS=500` truncation** is also not reflected in the evidence status.
- **The upload step can run when the report never did.** It runs whenever the report step produced no `exit-code` (for example, pip install failed), and then fails on a missing file. Prefer `steps.report.outputs.exit-code == '0' || == '1'`.
- **The CHANGELOG says "15 tests"; there are 17.**
- **Factory record:** the requirement shows 3 unresolved, empty comment threads, and the pending suggestion still needs accepting.

vivekchand:
## Ready to merge

Both blocking review items are fixed, and all CI checks are green on head `ea0007d5d0`. That includes Drift Bot, the product-record gate, CodeQL and E2E Gate (required). GitHub reports the PR as mergeable. It is rebased on current `main`. Not merged.

### Review fixes
1. **Sessions from other repositories are no longer linked.**
   - A relative write path now resolves only against the working directory the session recorded.
   - A session with no recorded directory links by relative path only when a commit trailer names it.
   - If such a session wrote a changed file by relative path, the evidence is `partial`. The reason names no session.
2. **Truncated evidence is no longer reported as `ok`.**
   - Every read limit hit, and every failed store read, is now an evidence gap.
   - A gap makes the evidence `partial` in the report and in the exported bundle (`evidence_gaps`, covered by the digest).
   - Events page back past activity after the head commit before any cut is declared.
3. **Tests.** Four regression tests fail against the previous head and pass with the fix. Suite: 22 tests.
4. **Non-blocking items also addressed:**
   - The PR comment only edits a comment its own token wrote.
   - The bundle no longer records `$USER`.
   - The `started_at` fallback is now the earliest event.
   - The Action uploads SARIF only when the report ran.
   - The docs now say a committed bundle is not a control against the PR author, and warn about duplicate scanner alerts.
5. **Factory.**
   - Requirement: AC-OBS-PRP-001.12 and .13 added, mirrored into `docs/acceptance_criteria.json`.
   - Blueprint 693240d3 was filled in; it was template text, which was Drift Bot's one finding.

### Merge after
- Nothing. No dependency PR.

### Companion PRs
- None. There is no new HTTP route, so no `cloud_route_policy` entry is needed.

### Founder decisions before a `[RELEASE]`
- **Tiering.** This ships the bundle, SARIF and gate free in OSS. The refined issue says paid provenance/evidence belongs in the extension. The requirement records this as an open question.
- **Factory.** Accept the pending tracked suggestions on requirement 4e224b46 (implementation note, AC .12/.13) and blueprint 693240d3. The requirement also has 3 empty comment threads.

### Post-merge / post-release verification
1. After the `[RELEASE]` publishes, confirm the released wheel has the command: `pip install clawmetry==<version>` in a scratch venv, then `clawmetry trace report --help`.
2. Confirm `.github/workflows/pr-provenance-action.yml` passes on `main`.
3. In a scratch repository with the released version, run the Action on a real pull request with `upload-sarif: true` and `comment: true`. Then check:
   - the `clawmetry-provenance` category shows up in code scanning;
   - one comment appears, and a re-run edits that same comment;
   - a person's comment that quotes the marker is left untouched.
   This path has not been verified on github.com yet.
4. On a machine running several agents across two repositories, run `clawmetry trace report --base <sha> --export-bundle /tmp/b.json`. Check that:
   - no session from the other repository appears in `/tmp/b.json`;
   - `evidence_gaps` is empty, or explains every `partial`.
5. After the cloud pin moves to this version, confirm the cloud boot is unaffected. No route was added, so there is nothing to check in route policy.

PR #5990, opened 2026-09-14
fix(sync): daemon tails the interceptor's actual output path

### LINKED ISSUE #5984: Python interceptor writes ~/.clawmetry/intercepted.jsonl but the daemon only tails ~/.openclaw/clawmetry-intercepted.jsonl, so captured calls never reach the store
Fixed and merged: #5990 (`3631527`).

`sync_intercepted_events` in `clawmetry/sync.py` now tails both `$CLAWMETRY_HOME/intercepted.jsonl` (the path `clawmetry/interceptor.py` actually writes to) and the legacy `~/.openclaw/clawmetry-intercepted.jsonl`, each with its own byte-offset cursor, so calls captured by `import clawmetry.track` / `CLAWMETRY_TRACK=1` now reach the store instead of being silently dropped.

Regression coverage: `tests/test_interceptor_daemon_tail.py`, wired into the MOAT Verifier job in `.github/workflows/ci.yml`. Confirmed red on the pre-fix code (reverted, re-ran, 2/3 tests failed with the exact drop-on-the-floor symptom), green with the fix restored, and the full 2838-test MOAT suite passes.

Verified the merged commit on `main` carries the fix (`_tail_intercepted_file` helper + the `CLAWMETRY_HOME` tail present in `clawmetry/sync.py` at `origin/main`).

This will ship to PyPI on the next `[RELEASE]`.

vivekchand:
No-PRD: fix of field-reported bug #5984

**Risk:** None. Additive change to one daemon tail function: it now reads an additional file (the interceptor's actual output path) alongside the one it already tailed, each with its own offset cursor in `state`. No schema change, no migration, nothing deleted. Worst case if this were wrong is the same as today (events dropped), not a new failure mode.

---

## Summary

- `clawmetry/interceptor.py` writes captured LLM calls to `$CLAWMETRY_HOME/intercepted.jsonl` (default `~/.clawmetry/intercepted.jsonl`, moved there in #2969), but `clawmetry/sync.py`'s `sync_intercepted_events` only ever tailed the legacy `~/.openclaw/clawmetry-intercepted.jsonl` path. Every call captured via `import clawmetry.track` or `CLAWMETRY_TRACK=1` was written to disk but never read by the daemon, so it never reached DuckDB or the dashboard.
- `sync_intercepted_events` now tails **both** files — the current `$CLAWMETRY_HOME/intercepted.jsonl` and the legacy `~/.openclaw/clawmetry-intercepted.jsonl` — each with its own byte-offset cursor (`last_intercepted_offset` / `last_intercepted_offset_legacy`) in `state`, so older installs whose events already sit under the legacy path keep working while new installs are actually read.

## Reproduction

Confirmed in the issue and reproduced locally:

```python
import clawmetry.interceptor as i
i._write_event({"type": "external_api_call", "provider": "openai", "model": "gpt-4o"})
# writes to ~/.clawmetry/intercepted.jsonl
```

`sync_intercepted_events` only ever opened `~/.openclaw/clawmetry-intercepted.jsonl`, which this write never touches, so the row was silently dropped every tick.

## Fix rationale

Rather than moving the daemon to tail only the new path (which would break any install whose events are still sitting under the legacy location, e.g. mid-upgrade), the tail function now reads both files independently. This mirrors the pattern already used for `sync_numbat_events`'s multi-file tail elsewhere in `sync.py`.

## Test plan

New regression file `tests/test_interceptor_daemon_tail.py`, wired into `.github/workflows/ci.yml`'s MOAT Verifier job (this repo runs explicit file lists, not `pytest tests/`):
- [x] `test_daemon_tails_current_interceptor_path` — writes via the real `interceptor._write_event`, asserts the daemon ingests it and a second tick doesn't double-ingest
- [x] `test_daemon_still_tails_legacy_openclaw_path` — confirms the legacy path still works for older installs
- [x] `test_daemon_tails_both_paths_with_independent_cursors` — both files have unread lines at once; each keeps its own offset and both are ingested

Confirmed red on the un-fixed code (reverted `clawmetry/sync.py`, ran the new tests, 2/3 failed with the exact "0 == 1" / "1 == 2" symptom described in the issue), then confirmed green again after restoring the fix.

Also ran the full MOAT Verifier suite (145 files, 2838 tests) locally — all pass with this change.

Closes #5984

https://claude.ai/code/session_01Bg5NX9fNDCeCuL6JBRrdFp


vivekchand:
`PR cites the product record that justified it` failed on the original body: I'd written `**Product record:** No-PRD: fix of field-reported bug #5984`, but the gate's regex requires a line that starts with `No-PRD:` (see `scripts/check_product_record.py`'s `OPT_OUT` pattern) — the `**Product record:**` prefix broke the match. Fixed by editing the PR body to put `No-PRD: fix of field-reported bug #5984` on its own line. The gate re-triggers on `edited` and is re-running now.

---

PR #5994, opened 2026-09-14
[RELEASE] Hosted Guard and Signals list real sessions instead of a false empty (carries #5994)

vivekchand:
## Problem
On app.clawmetry.com:
- **Guard** read "No sessions running right now" while the same node's local dashboard listed 39 running sessions. `/api/guard/sessions` is a passthrough, so it ran on a cloud container that has no store and no process table.
- **Signals → Sessions** read "No sessions matched in this window" beside a rate that counted 2 matches. The session list was never in the snapshot, the cloud interceptor answered `available:false`, and the renderer ignored that and printed "no matches".

## Fix (OSS half)
- `routes/guard.py`: the handler body becomes `build_guard_sessions_body(limit, call=None)`, shared by the route and the daemon, so the two tabs cannot list different sessions. The route is unchanged in behaviour.
- `clawmetry/sync.py`: two new snapshot slices built on the daemon's own store handle (never a read-only re-open):
  - `guardSessions`: the exact `/api/guard/sessions` body plus `generated_at`.
  - `signalSessions`: `byRuntime[<runtime>|"all"][window][signal]` → session rows with match counts (never phrases). Built per runtime so a quiet runtime is not starved, and only for signals that matched (`behaviour_signals.build_session_slice`).
- `app.js`: Guard and Signals empty states show the response `reason` when `available === false`. The Guard outcome dialog gains a `pending` state ("was sent to your node") for relayed actions.

Measured against a live daemon through its proxy: Guard slice 39 rows / 35 KB, signals session slice 8 KB.

Cloud half (interceptors reading these slices, Guard controls through `/api/cloud/session/*`) follows in clawmetry-cloud once this is released.

No-PRD: bug fix restoring hosted parity for the existing Guard and Behaviour Signals surfaces; no new capability.

## Tests
`tests/test_cloud_guard_signals_slices.py` (added to ci.yml) plus the existing guard route and signals suites.

https://claude.ai/code/session_01KEaCCBVVbH6EUigYDiFuXP


vivekchand:
**Drift Bot (3 findings):** each is "blueprint does not describe this yet", and #1 cites the Behaviour Signals rule "snapshot slices minus per-session lists", which this PR deliberately reverses (without the lists the hosted drill-down read "No sessions matched" beside a rate that counted matches). Blueprint edits proposed as pending suggestions:
- Behaviour Signals: new `signalSessions` snapshot drill-down contract.
- Guard: `guardSessions` slice and the shared `build_guard_sessions_body` builder.
- Guard Enforcement: hosted Guard tab through the cloud relay, with the `pending` outcome state.

**Pre-deploy browser check (Playwright, real data):** this branch's dashboard, with the cloud `cm-cloud-guard` and `cm-cloud-signals` scripts from clawmetry-cloud#2444 injected, fed a snapshot built by these builders from a live daemon:
- Guard: 39 of 39 rows rendered, "$3.10 at risk across 2 flagged sessions".
- Kill (relay stubbed, any real `/api/guard/control` POST aborted, none attempted): preflight shown, relay received the bare session id with `mode: kill`, dialog rendered "Kill — done" with both steps.
- Snapshot without the slice: "Running sessions arrive with the next snapshot from your node."
- Signals → Praise → Sessions: 2 sessions listed (Claude Code and Codex); without the slice the reason is shown instead of "No sessions matched".

PR #6031, opened 2026-09-16
Harden CI: the remaining pip bootstrap steps install from a hash-pinned set

vivekchand:

**Risk:** Low. Two steps, each swapping an unpinned `pip install --upgrade pip` for the hash-pinned set five other workflows already use. No permissions, no action references, no job structure touched. Undone by reverting the commits.

---

## Summary

Scorecard's `PinnedDependenciesID` reports 41 occurrences on `main`; 35 are unpinned `pip` commands. This closes two of them — bare `pip install --upgrade pip` bootstraps that resolved whatever pip was newest at job start:

| Workflow | Step | Path |
|---|---|---|
| `pr-screenshots.yml` | Install Python deps | `head/.github/requirements/pip-bootstrap.txt` |
| `release-on-merge.yml` | Pre-publish smoke (install wheel, verify CLI + imports) | `.github/requirements/pip-bootstrap.txt` |

Both now run `pip install --require-hashes -r …/pip-bootstrap.txt`, the set already used by `ci.yml`, `api-latency-smoke.yml`, `oss-golden-path.yml`, `overhead-bench.yml` and `cross-repo-handoff.yml`. Direct continuation of #5929.

**Only the bootstrap changes.** Each step's real install stays unpinned deliberately, and the diff says so inline: `head/requirements.txt` is the PR's own checkout and `dist/*.whl` is the artifact the smoke step exists to test. Pinning those would test nothing.

**The two paths differ on purpose.** `pr-screenshots`' visual-diff job checks the repo out into `head/` and `base/` and never at the workspace root, so its set lives under `head/` — matching `head/.github/scripts/` elsewhere in the same job. `release-on-merge` has a single job with a root checkout, in which the existing `release-publish.txt` pin already resolves from the root.

## Three sites deliberately left alone

These look like the same one-line change and are not:

- **`conformance-heartbeat.yml`** (both jobs) — neither job checks the repo out at all; they install from PyPI into a clean runner on purpose. No repo-relative path can work, and adding a checkout to a clean-room job to satisfy a pin would defeat what the job tests.
- **`conformance-heartbeat.yml`'s `pypi-install` matrix and `release-canary.yml`'s `canary` matrix** — both carry a **Python 3.9** include entry, and the pinned `pip==26.2.1` declares `Requires-Python: >=3.10`. Converting them would fail those jobs outright, so a naive "pin everything" pass here breaks the release canary.

That is also why every existing `pip-bootstrap.txt` call site is a 3.11-only job with a root checkout.

## Test plan

- [x] Every workflow and composite action parses as YAML — 45 files, `yaml.safe_load`
- [x] `scripts/check_action_refs.py` clean: 22 references, all SHA-pinned
- [x] **Reproduced the first push's CI failure locally** in a simulated `head/` layout — exit 1, `Could not open requirements file: '.github/requirements/pip-bootstrap.txt'`, the same error the visual-diff job hit — then showed the corrected `head/`-prefixed path exit 0 and resolve `pip 26.2.1`
- [x] Ran the release-on-merge form in a fresh venv on **Python 3.11**: exits 0, resolves `pip 26.2.1`
- [x] Confirmed `pip==26.2.1` metadata is `Requires-Python: >=3.10` from PyPI, which is what rules the two 3.9-matrix sites out
- [x] Confirmed both edited jobs are `python-version: "3.11"` (pr-screenshots:98, release-on-merge:51) and verified release-on-merge's checkout and smoke step are in the same job

https://claude.ai/code/session_01J9ECxNE9KErH1pvAcJsJ3z


vivekchand:
## `visual-diff` is red here, and it is not this PR's failure

Recording this rather than leaving it unexplained. `visual-diff` is the one non-green check on `b1020cc` (37 of the other checks pass). It is marked non-blocking by the bot that posts it.

**It fails on every recent PR, not just this one.** Job-level conclusion for `visual-diff` across the last runs of `pr-screenshots.yml`:

| Branch | `visual-diff` |
|---|---|
| `auto-bump/v0.12.883` (a pure version bump) | failure |
| `feat/ingest-key` | failure |
| `feat/agent-supply-chain-inventory` | failure |
| `feat/has-capacity-batch` | failure |
| `feat/build-your-own-ui` | failure |
| `fix/visual-diff-documentelement-null` | failure |

A version-bump branch and the branch that was itself fixing visual-diff both fail the same way. Note the run-level conclusion reads `success` on all of these — the job is non-blocking, so the run status hides it; the job conclusion is what shows the pattern.

**What actually happens.** The job gets through setup and roughly 40 tab diffs, then dies mid-screenshot at ~4 minutes:

```
[shoot] http://127.0.0.1:8082 tab=swimlane threw: page.evaluate: Target page, context or browser has been closed
##[error]The runner has received a shutdown signal. This can happen when the runner
         service is stopped, or a manually started runner is canceled.
##[error]The operation was canceled.
```

Twice, at 4m10s and 4m21s, well inside the job's 20-minute timeout. I ruled out the two local explanations: `concurrency: cancel-in-progress` did not fire (nothing new was pushed between the two attempts), and `queue-priority.yml` cancels only *queued* runs and last ran at 02:42, while `main` has not moved since 01:07. I have not established the underlying cause beyond that — it is a runner-side shutdown, not something this diff reaches.

**Why it is not mine to fix here.** This PR changes two `pip install --upgrade pip` lines to a hash-pinned set. The only step it touches, `Install Python deps`, **passes in 6 seconds**; the failure is four minutes later in the Playwright screenshot loop. Nothing in the diff is reachable from there.

**No fix to port.** `main` already carries the most recent visual-diff fix (`bc221bb` *fix(visual-diff): defer documentElement style injection*, 01:05), and this branch is based on `82e171f`, after it. That fix is present here and visual-diff still fails, so there is no newer change to pull in.

I re-ran the failed job once — it failed identically — and per the repo's own rules that re-run is spent, so I am not re-running it again or touching the diff to chase it. Flagging it as a repo-wide CI issue worth a look on its own, separate from this PR.

---

vivekchand:
PR sweeper: `visual-diff` failure is pre-existing and not caused by this CI-only PR (the flagged pages — `logs`, `crons`, `nemoclaw`, `selfevolve` — differ due to live session data, not code changes; runner was also killed by a shutdown signal mid-run). All required checks (`E2E Gate (required)`) pass. Only remaining blocker is review approval.

---

vivekchand:
**visual-diff flake — not caused by this PR**

The `visual-diff` job failed due to a runner shutdown signal mid-run (`The runner has received a shutdown signal`), not a code issue:

- The run was canceled before completing (browser context closed on the `swimlane` screenshot).
- The "100% diff" on `mobile nemoclaw` and `mobile selfevolve` are dynamic live-data screenshots taken at different wall-clock times, not UI regressions.
- The other flagged tabs (`logs` 6.29%, `crons` 5.37%) are also time-sensitive data that changes between the base and head screenshot passes.
- This PR touches only two CI YAML files — there are zero UI changes.

`E2E Gate (required)` is green. All other 40+ checks pass. The `visual-diff` check is not required for merge; the red X is the runner cancellation, not a real diff from this PR's changes.

---

PR #6032, opened 2026-09-16
Harden CI: the desktop bundle jobs install from a hash-pinned set

vivekchand:

**Risk:** Low. Three steps, each swapping an unpinned `pip install --upgrade pip wheel` for a hash-pinned set in the same shape five other workflows already use. No permissions, no action references, no job structure, no bundling step touched. Undone by reverting the commit.

---

## Summary

Scorecard's `PinnedDependenciesID` reports 41 occurrences on `main`; most are unpinned `pip` commands. This closes three of them — the `Install build deps` step of each desktop bundle job:

| Job | Runner | Was |
|---|---|---|
| `macos` | `macos-latest` | `pip install --upgrade pip wheel` |
| `windows` | `windows-latest` | `python -m pip install --upgrade pip wheel` |
| `linux` | `ubuntu-latest` | `pip install --upgrade pip wheel` |

All three now run `python -m pip install --require-hashes -r .github/requirements/pip-wheel-bootstrap.txt`. Direct continuation of #5929 and #6031.

Each resolved whatever pip and `wheel` PyPI served at job start, and then used *that* pip to install `desktop/requirements-dev.txt` below it. A hash-checked install is only worth as much as the installer doing the checking, and these are the jobs that produce the signed `.dmg` / `.exe` / `.deb` / `.AppImage` users download.

## Why a third bootstrap file

`.github/requirements/pip-wheel-bootstrap.txt` is new rather than a reuse of either existing set:

- **`pip-bootstrap.txt`** serves ten call sites that build nothing; three bootstrap a throwaway venv to prove a freshly installed wheel boots on its own. Adding `wheel` widens the very surface those isolated-venv tests exist to measure.
- **`pip-build-bootstrap.txt`** carries PyPA `build`, which these jobs never invoke — they bundle with PyInstaller, not a wheel build — and its own header notes it omits `colorama` because all three of its call sites are `ubuntu-latest`. One call site here is `windows-latest`, so that file cannot serve it unchanged.

Keeping the sets apart leaves every job's installed set identical to what it was before this pin.

`wheel` is genuinely needed here, not vestigial: `desktop/requirements-dev.txt` pins `PyGObject==3.58.0` on Linux, an sdist-only build.

## The closure

`pip==26.2.1`, `wheel==0.48.0`, `packaging==26.3` — `packaging` is `wheel`'s dependency (`wheel 0.48.0` declares `packaging>=24.0`), not padding, and `--require-hashes` demands the whole closure. Neither `wheel` nor `packaging` declares an environment-marked requirement, so one file serves all three runners — the omission that constrains `pip-build-bootstrap.txt` does not apply.

`pip` and `packaging` are pinned at the same versions **and the same hashes** as the existing `pip-build-bootstrap.txt`, verified by diffing the two files rather than by assuming.

## Only the bootstrap changes

Each job's `pip install -r desktop/requirements-dev.txt` is untouched and stays unpinned — that file is the PR's own checkout, and pinning it would test nothing. The macOS and Linux sites also move from bare `pip` to `python -m pip`, matching the Windows site and every other pinned call site in the repo.

## Test plan

- [x] Every workflow and composite action parses as YAML — 45 files, `yaml.safe_load`
- [x] **Installed the new set with `--require-hashes` in a clean Python 3.11 venv**: exits 0, resolves `pip 26.2.1`, `wheel 0.48.0`, `packaging 26.3`
- [x] Resolved the closure independently with `pip install --dry-run --ignore-installed --report`, exactly as the file header prescribes — confirms those three names and no others
- [x] Took every published sha256 per version from the PyPI JSON API (wheel + sdist for each), so the pin does not depend on a runner resolving to the same artifact
- [x] Diffed the shared `pip` / `packaging` entries against `pip-build-bootstrap.txt` — versions and hash sets match exactly
- [x] `scripts/check_action_refs.py` clean: 22 references across 45 files, all SHA-pinned
- [x] `make lint-runtime-count`, `lint-module-map`, `lint-ci-test-coverage` all pass
- [x] Confirmed all three edited jobs are `python-version: "3.11"` (pip 26.x requires >=3.10) and check the repo out at the workspace root with no `path:`, so the repo-relative requirements path resolves — the two conditions that ruled the 3.9-matrix and `head/`-layout sites out in #6031

https://claude.ai/code/session_019gij6F43tsfbfKX41MKbkE

---
_Generated by [Claude Code](https://claude.ai/code/session_019gij6F43tsfbfKX41MKbkE)_


vivekchand:
## Verified on all three real runners — `desktop-artifacts` dispatched on this branch

This workflow only triggers on `push: tags: v*.*.*` and `workflow_dispatch`, so **PR CI never exercises the step this PR changes.** Left alone, the first real execution of the new bootstrap would have been a release tag, on the path that produces the `.dmg` / `.exe` / `.deb` / `.AppImage`. So I dispatched it on this branch instead.

[Run 35052962214](https://github.com/vivekchand/clawmetry/actions/runs/35052962214) — **completed / success**, all three jobs green:

| Job | Runner | `Install build deps` | Whole job |
|---|---|---|---|
| macOS .app + .dmg | `macos-latest` | success | success |
| Windows .exe | `windows-latest` | success | success |
| Linux + AppImage | `ubuntu-latest` | success | success |

The `release` job is gated on `startsWith(github.ref, 'refs/tags/v')`, so the dispatch built and uploaded workflow artifacts and published nothing.

**Windows** is the interesting one, because `pip-build-bootstrap.txt` structurally cannot serve it — its own header notes it omits `colorama` since all three of its call sites are `ubuntu-latest`. The new file resolved cleanly there:

```
Run python -m pip install --require-hashes -r .github/requirements/pip-wheel-bootstrap.txt
Collecting packaging==26.3 (from -r .github/requirements/pip-wheel-bootstrap.txt (line 54))
Requirement already satisfied: pip==26.2.1 in C:\hostedtoolcache\windows\Python\3.11.9\x64\... (26.2.1)
Collecting wheel==0.48.0 (from -r .github/requirements/pip-wheel-bootstrap.txt (line 60))
Successfully installed packaging-26.3 wheel-0.48.0
```

…and the signed bundle built with `Number of errors: 0`.

**Linux** confirms `wheel` is genuinely load-bearing here rather than vestigial — `desktop/requirements-dev.txt` pins `PyGObject==3.58.0`, an sdist-only build, which built from source under the pin:

```
Successfully installed packaging-26.3 wheel-0.48.0
  Building wheel for PyGObject (pyproject.toml): started
  Building wheel for PyGObject (pyproject.toml): finished with status 'done'
Successfully built PyGObject proxy_tools
```

That is the evidence for the one risk this change carries: that pinning `wheel` to an exact version breaks a native build the unpinned `--upgrade` happened to satisfy. It does not.

---

PR #6043, opened 2026-09-16
feat(guard): make checks visible and simplify alerts and approvals

vivekchand:
Guard's available checks were invisible in the product, while findings, approvals and alert history lived behind three separate navigation entries. Guard now provides one home with Checks, Needs attention, Activity and Response settings. Its catalogue groups the shipped checks into six families and explains each check's scope and state.

Per-check switches persist through the daemon's DuckDB writer and skip the actual session, workspace, legacy stuck and cross-session detection paths. Administrator overrides remain authoritative. Disabling a check explains the effect on dependent automatic responses and preserves historical findings and pending approvals. Cloud parity uses the encrypted `guardChecks` snapshot and a sealed, owner-bound relay in [clawmetry-cloud#2469](https://github.com/vivekchand/clawmetry-cloud/pull/2469); queued writes stay pending until confirmed by the node.

Approvals lead with pending decisions and collapse history/settings. Alerts gain search, resolution filters and repeat grouping that preserves session boundaries. Existing deep links and process-control confirmations remain available. Checks fetch only when their view opens; no new polling loop.

Validation: focused detector/store/control tests, cloud encryption/relay and inline-JS tests, Jinja rendering, Node syntax, daemon allowlist and acceptance-criteria gates. Chromium on an isolated local dashboard verified desktop/mobile rendering, all six groups, real switch persistence across reload and no console errors. Regression proof: restoring the old Guard route makes the persistence test fail with 404; the restored implementation passes.

[Requirement](https://factory.8090.ai/project/b415065f-ab2f-4f53-8864-0c009fd098cb/requirements/327e83bf-8714-4b52-8c40-788d1520c828) · [Blueprint](https://factory.8090.ai/project/b415065f-ab2f-4f53-8864-0c009fd098cb/blueprints/2b6aceb5-389e-492c-85a1-8333ed9d0430)

Release and hosted verification follow after merge. The paired cloud change must pin the verified OSS wheel before deployment.

Implemented by Codex following the repository's [Claude Code workflow](https://claude.ai/code).


vivekchand:
**CI blocked — real test failure, not a flake.**

MOAT Verifier (the only required check that failed, which caused E2E Gate to fail):

```
FAILED tests/test_guard_checks.py::test_remote_change_requires_valid_seal_and_returns_confirmation
  AssertionError: assert (False)

  results[-1]['applied'] is False
  Captured log: Guard setting request failed: ConnectionException
```

`_dispatch_pending_action` receives the `guard_check_update` action, decrypts the sealed payload correctly, but the call that writes the toggle to the store is failing with a `ConnectionException`. The test expects `applied: True` and `enabled: False`; it gets `applied: False`.

Root cause to investigate: the route or store helper `_dispatch_pending_action` calls to persist the guard check toggle is trying to reach the local query server (or another endpoint) and failing in the CI environment. Check whether `sync._dispatch_pending_action` for `guard_check_update` is calling through the local server instead of writing directly to the store — in CI there is no running daemon, so any path that goes through `~/.clawmetry/local_query.json` will fail with a connection error.

**Other failing checks:**
- `visual-diff` — screenshot comparison failure (likely new UI panels)
- `CodeQL` — needs investigation

(1 test failed, 3220 passed, 37 skipped in MOAT Verifier)

---

PR #6049, opened 2026-09-16
Harden CI: enforce the py3.9 pip bootstrap pin's own refusal

vivekchand:
## What

`.github/requirements/pip-bootstrap-py39.txt` exists because its sibling `pip-bootstrap.txt` pins a pip that declares `Requires-Python >=3.10`, which the release canary's Python 3.9 matrix cell cannot install. That file's header already says a Dependabot bump to 26.1+ **"must be REFUSED"** — and until now, nothing enforced it.

The gap is not theoretical. Dependabot proposed exactly that bump (#6047, `26.0.1` → `26.2` in this file), and it is **green on every PR check**. It can be: the only job that installs this file is `release-canary.yml`, which does not run on pull requests. So the bump passes review with a green checkmark arguing in its favour, and the failure surfaces afterwards, on the release path, in the job whose whole purpose is to prove a published release installs.

A refusal a reviewer has to remember is the weakest kind there is.

## The change

Three files, all in paths the product-record gate exempts:

- **`tests/test_pip_bootstrap_py39_pin.py`** (new) — fails any pin at 26.1 or above, with a message that says plainly to close the Dependabot PR. The boundary is *"below 26.1"* rather than *"exactly 26.0.1"*, deliberately: a 26.0.x release, or some future pip that restores 3.9, is a bump this guard should **allow**. A second check asserts `release-canary.yml` still installs this file, so the guard cannot quietly go vacuous if the 3.9 cell is repointed. The version comparison is itself covered, so the guard cannot pass by mis-parsing.
- **`.github/workflows/ci.yml`** — wires it into the `Syntax & Lint` job, after the step that installs pytest, so it runs on every PR. (This repo runs explicit file lists, not `pytest tests/`.)
- **`.github/requirements/pip-bootstrap-py39.txt`** — records what the refusal *accepts*. pip 26.0.1 is named by four published advisories whose fixes all require 3.10+, so none is reachable from here. Stating it makes the acceptance deliberate rather than silent: a reviewer refusing the bump should know what they are refusing it in favour of. The exposure is narrow — CI-only, no released wheel, one call site, and both installs that pip performs are a `--require-hashes` set and a versioned PyPI install. The acceptance ends when the repository drops 3.9, and the comment says what to delete then.

## Verification

- Reproduced #6047's bump locally (`pip==26.0.1` → `26.2`) and confirmed the guard **fails** on it with the intended message, then **passes** on the restored pin.
- All 44 workflow files parse.
- The repo's own `test_workflow_yaml_valid.py` and `test_ci_workflow_invocations_are_real.py` pass (679 passed, 429 skipped).
- `scripts/check_ci_test_coverage.py --check` passes — the new test is wired into a workflow in this PR.
- `scripts/check_py39_annotations.py` passes.

## Related, but not changed here

Dependabot PR #6047 should be closed rather than merged; this PR is what makes that decision stick. #6047 also downgrades three other files from `26.2.1` to `26.2`, which is a separate reason not to merge it. Neither is touched here.

No-PRD: CI-only change confined to `.github/` and `tests/`, which the product-record gate exempts. Adds no product surface.

https://claude.ai/code/session_015QB9ybxuPRZy8Gu5dZ7FLc

---
_Generated by [Claude Code](https://claude.ai/code/session_015QB9ybxuPRZy8Gu5dZ7FLc)_


vivekchand:
## The guard caught a live regression on its first run — scope of this PR has grown by one line

`Syntax & Lint` went red here, and it was correct. Timeline:

- **18:39 UTC** — #6047 merged to `main`, moving `.github/requirements/pip-bootstrap-py39.txt` from `pip==26.0.1` to `pip==26.2`.
- **18:48 UTC** — this PR opened. Its branch had not touched the pin, but the merge commit CI runs against carries `main`'s, so the guard fired on a value that arrived from the base.

Merging `main` in surfaced **two** defects in that bump, not one:

1. **pip 26.1 moved `Requires-Python` to `>=3.10`** (confirmed against PyPI: 26.0.1 is `>=3.9`, 26.1/26.2/26.2.1 are `>=3.10`). The release canary's 3.9 matrix cell cannot install 26.2 at all. This is the bump the file's header says must be refused.
2. **The version moved but the hashes did not.** `main` currently pins `pip==26.2` carrying 26.0.1's two digests (`bdb1b08f…`, `c4037d8a…`). pip 26.2's actual digests are `931c3036…` (wheel) and `2d8542af…` (sdist) — which `pip-bootstrap.txt` *did* receive. So a `pip install --require-hashes` of the py3.9 file fails on **every** platform, not only on 3.9.

Neither is visible on a normal PR, because `release-canary.yml` is the only consumer and it does not run on pull requests.

### What changed here

One line: the pin is restored to `pip==26.0.1`. Version and both hashes were re-fetched from `https://pypi.org/pypi/pip/26.0.1/json` and match exactly (2 pinned, 2 published); 26.0.1 is not yanked. That repairs both defects for this file, since 26.0.1's hashes were the correct ones all along.

The three files #6047 downgraded from `26.2.1` to `26.2` are deliberately **left alone** — both versions require `>=3.10` and both are advisory-free, so the 3.11 cells are unaffected and that is a separate concern.

### Worth a follow-up, not taken here

Defect 2 is a class, not an incident: nothing in CI checks that a `--require-hashes` pin file's digests actually belong to the version on the line above them. That check needs PyPI at runtime, so it belongs in a network-enabled job rather than in `Syntax & Lint`, and it would cover all 17 files in `.github/requirements/` rather than this one. Happy to open it separately if wanted.

Verification on this push: guard passes (8 tests), all 44 workflows parse, `test_workflow_yaml_valid.py` + `test_ci_workflow_invocations_are_real.py` pass (679 passed), CI-coverage ratchet and py3.9 annotation guard pass.

---

PR #6067, opened 2026-09-17
Harden CI: the BrowserStack job installs from a hash-pinned set

vivekchand:

**Risk:** Low, and bounded to one non-required job. `browserstack.yml` runs on `workflow_dispatch`, on push to `main`, and on a PR only when the `browserstack` label is added — it is not an `E2E Gate` leg, so a fault here cannot block other PRs. The change is confined to one step's `run:` line plus a new file; the job's `permissions: contents: read`, its SHA-pinned actions and its Dependabot skip are untouched.

---

## Summary

The cross-browser job opened with a bare `pip install flask pytest pytest-playwright requests browserstack-sdk`, so every run resolved whatever the index served that minute. This is the job that holds `BROWSERSTACK_USERNAME` and `BROWSERSTACK_ACCESS_KEY`, which makes it a poor place to keep an unpinned resolve.

The five names move into `.github/requirements/browserstack.txt` with the sha256 of every distribution PyPI publishes for them, and the step installs with `--require-hashes`, which refuses anything not listed there. This is the same pin/updater pairing the other sixteen sets in that directory already use.

Two things worth calling out:

- **The closure is 59 entries, not five.** `--require-hashes` demands the whole transitive set, and `browserstack-sdk` pulls in Selenium, Appium and Robot Framework of its own accord. That is the point rather than padding — an unlisted transitive dependency is exactly the gap a pin is supposed to close.
- **`browserstack-sdk` publishes no wheel**, so pip still builds it from the sdist. That was already true of the unpinned install; what changes is that the sdist is hash-checked before anything is built from it.

The new file is already covered by the existing `/.github/requirements` Dependabot entry, which groups that directory into one weekly PR and rewrites the `--hash` lines along with the versions — so the pin makes an upgrade deliberate without freezing it. The resolved versions also respect that entry's standing `ignore` rules: `pyee` lands on 13.0.1, below the `>=14` rule that requires Python 3.12.

## Test plan

- [x] Every workflow file still parses as YAML (`yaml.safe_load` over `.github/workflows/*.yml` — 44 files), and so does the one composite action
- [x] **Installed the generated set for real**, not as a dry run: a clean CPython 3.11 venv on Linux, `pip install --require-hashes -r .github/requirements/browserstack.txt` → exit 0, 59 packages, no hash mismatch and no missing transitive dependency
- [x] Imported all five direct requirements from that venv afterwards (`flask`, `pytest`, `requests`, `pytest_playwright`, `browserstack_sdk`)
- [x] `tests/test_workflow_yaml_valid.py` — 673 passed, 0 failed. This is the guard that reads a `-r` file and checks the job still installs what `tests/conftest.py` imports at module scope; `requests` is pinned in the new file, so moving the names off the `run:` line does not retire it for this job
- [x] `scripts/check_ci_test_coverage.py --check` — OK (no new test file added)
- [x] Generated on Linux / CPython 3.11, matching the job's `ubuntu-latest` runner and its `actions/setup-python` version; the regeneration command is recorded in the file header

https://claude.ai/code/session_01KcDQShtuBxQQ88U2HwTgN8

---
_Generated by [Claude Code](https://claude.ai/code/session_01KcDQShtuBxQQ88U2HwTgN8)_


vivekchand:
✨ auto-fixed: merged current main (07c2ae0) into branch to resolve BEHIND state — no conflicts

---

PR #6070, opened 2026-09-17
Harden CI: the Syntax &amp; Lint job installs from a hash-pinned set

vivekchand:

**Risk:** Low, and bounded to one job. The pinned versions are the ones the bare lines already resolved to on 3.9, and every guard the job runs was executed against the pinned set before pushing (see Test plan). A wrong pin fails the install step loudly rather than degrading silently, and is undone by reverting the commit.

---

## Summary

The `lint` job ("Syntax & Lint") installed its tooling with two bare `pip install` lines:

- `pip install ruff`
- `pip install pytest pyyaml requests flask`

So the job that decides whether this repo's source is well-formed — and which then runs **54 guard test files** through pytest — resolved whatever the index served that minute. These were 2 of the 36 `PinnedDependenciesID` findings the Scorecard run reports (`.github/workflows/ci.yml:95` and `:172`).

Both steps now install with `--require-hashes` from a set under `.github/requirements/`, which is the pattern the sibling jobs in this same file already use.

**Two files rather than one, and deliberately not `ci-tests.txt`.** That set is resolved for CPython 3.11 and pins `pytest==9.1.1`, `requests==2.34.2` and `click==8.5.0` — none of which install on 3.9, because each declares `Requires-Python >= 3.10`. This job runs on 3.9 *deliberately*: it is the floor `setup.py` advertises, and the comment already in the workflow records why (it is the interpreter 0.12.753 broke on). The two sets cannot be merged until that floor moves, and each file's header says so.

The pinned versions (`ruff==0.16.8`, `pytest==8.4.2`, `pyyaml==6.0.3`, `requests==2.32.5`, `flask==3.1.3` + closure) are what the bare lines **already resolved to** on 3.9, so this changes the artifacts' provenance, not the versions the job has been running.

Each version lists the sha256 of every distribution PyPI publishes for it, so the pin does not depend on the runner resolving to the same artifact the file was generated against.

### One thing a reviewer should know

`.github/dependabot.yml` already updates `/.github/requirements`, so these files do not freeze. But a proposed bump to a release that dropped 3.9 cannot install on this job at all, and `--require-hashes` fails the whole step rather than the one package. That failure lands **on the bump PR** — where the lint job runs on every PR — so it is visible before it can reach `main`. `ci-lint-verify.txt` carries this as a note at the top, in the same spirit as the existing `pip-bootstrap-py39.txt` header. Adding an `ignore` rule for it is a separate, one-line change if you'd rather pre-empt it than catch it.

## Test plan

- [x] Every workflow still parses as YAML (`yaml.safe_load` over all 44 `.github/workflows/*.yml`), and so does every `action.yml`
- [x] Built a real CPython **3.9.23** venv and ran both new steps **verbatim** — each installs cleanly under `--require-hashes`
- [x] Re-ran **all 54 pytest guards the lint job invokes** in that pinned environment: **54 ran, 0 failed**
- [x] Ran the repo's own workflow guards against the edited file — `test_workflow_yaml_valid.py`, `test_ci_workflow_invocations_are_real.py`, `test_pip_bootstrap_py39_pin.py`, `test_action_refs_pinned.py`: **712 passed, 430 skipped**
- [x] Confirmed `test_workflow_yaml_valid.py`'s conftest-imports guard still sees `pytest`/`requests`/`flask` — it reads `-r` files this repo owns, so moving the names into a requirements file does not retire it for this job
- [x] Ran the lint step itself (`ruff check dashboard.py --select E,W --ignore E501,E402`) on the pinned ruff: exits 0

https://claude.ai/code/session_01CDtgRiaRbkXywpR2M3xj6M

---
_Generated by [Claude Code](https://claude.ai/code/session_01CDtgRiaRbkXywpR2M3xj6M)_


vivekchand:
✨ auto-fixed: merged current main (07c2ae0) into branch to resolve BEHIND state — no conflicts

---

